PR #104, opened 2022-12-19
Parse for loops

egiurleo:
This PR implements parsing for loops in the structure:

```ruby
for i in 0..10 do
  i # or some other expression here
end
```

This PR handles multiple indexes e.g. `for i, j in 0..10` and makes error handling more robust.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1229,6 +1229,26 @@ nodes:
 
           while foo do bar end
           ^^^^^^^^^^^^^^^^^^^^
+  - name: ForNode
+    child_nodes:
+      - name: keyword
+        type: token
+      - name: index
+        type: node
+      - name: keyword_in
+        type: token
+      - name: collection
+        type: node
+      - name: statement
+        type: node
+      - name: end_keyword
+        type: token?
+    location: keyword->statement
+    comment: |
+      Represents the use of the `for` keyword.
+
+          for i in a end
+          ^^^^^^^^^^^^^^
   - name: YieldNode
     child_nodes:
       - name: keyword
--- a/config.yml
+++ b/config.yml
@@ -1241,11 +1241,11 @@ nodes:
         type: node
       - name: do_keyword
         type: token?
-      - name: statement
+      - name: statements
         type: node
       - name: end_keyword
         type: token
-    location: for_keyword->statement
+    location: for_keyword->statements
     comment: |
       Represents the use of the `for` keyword.
 
--- a/config.yml
+++ b/config.yml
@@ -772,6 +772,16 @@ nodes:
 
           module Foo end
           ^^^^^^^^^^^^^^
+  - name: MultiLeftHandNode
+    child_nodes:
+      - name: targets
+        type: node[]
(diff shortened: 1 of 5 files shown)


config.yml
```diff
@@ -772,6 +772,16 @@ nodes:
 
           module Foo end
           ^^^^^^^^^^^^^^
+  - name: MultiLeftHandNode
```
egiurleo, line 775: @kddnewton, this is my and @dirceu's take on the MLHS node!

src/yarp.c
```diff
@@ -1484,6 +1486,9 @@ missing(yp_token_t *token, const char *location) {
 static yp_node_t *
 parse_expression(yp_parser_t *parser, binding_power_t binding_power, const char *message);
 
+static yp_node_t *
+parse_left_hand_side(yp_parser_t *parser, binding_power_t binding_power, const char *message);
```
egiurleo, line 1490: C was having a hard time finding our `parse_left_hand_side` method, so we thought to follow the pattern of `parse_expression` and include this declaration here. I'm not really sure why we need this!

config.yml
```diff
@@ -772,6 +772,16 @@ nodes:
...
+    child_nodes:
+      - name: targets
```
egiurleo, line 777: We called these "targets" because in the assign nodes, the variables being assigned are also called "target". Happy to change the name to whatever makes more sense.

src/yarp.c
```diff
@@ -2046,6 +2051,37 @@ parse_expression_prefix(yp_parser_t *parser) {
...
+        do_keyword = parser->previous;
+      } else {
+        not_provided(&do_keyword, parser->previous.start);
+      }
+
+      yp_node_t *scope = yp_node_scope_create(parser);
+      yp_node_t *parent_scope = parser->current_scope;
+      parser->current_scope = scope;
+
+      accept(parser, YP_TOKEN_SEMICOLON);
```
egiurleo, line 2074: @dirceu and I added handling for semicolons as well as newlines (and added a test case for it). Let me know if that looks alright.

src/yarp.c
kddnewton: So C is a one-pass compiler, meaning it has to know about all of the symbols in the program when they are referenced. This is called a forward declaration, that tells the compiler "hey I'm going to have a function later that is named like this and has these types". It allows the compiler to keep compiling, even without the function body. When it finds it lower in the file, it'll fill in the body.

In this case though, you should move the function up to this point, because it's not necessary that it lives lower. In that case you will be able to remove the forward declaration since it all can live right here.

config.yml
kddnewton: Targets is good!

kddnewton (changes requested):
One small change to move the function up, and the rest looks good!

src/yarp.c
egiurleo: Ah thanks for the explanation! Moved the method up here and removed the declaration.

PR #107, opened 2022-12-20
Implement ensure statement

vinistock:
Implement ensure statements.

I was a bit in doubt about the usage of `not_provided` for the end token. Is that the correct way of using it? Without it, I kept running into segmentation faults.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -342,9 +342,11 @@ nodes:
         type: token
       - name: statements
         type: node
+      - name: ensure_clause
+        type: node?
       - name: end_keyword
-        type: token
-    location: begin_keyword->end_keyword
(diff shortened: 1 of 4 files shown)


kddnewton:
@vinistock the only other thing I just thought of is that `ensure` blocks need their own scope node as well.

vinistock:
Do ensure and begin have their own scope? I tried this and some variations in Ruby 3.2, but I might be missing something.
```ruby
begin
  a = 123
ensure
  b = 321
end

puts a # 123
puts b # 321
```

kddnewton:
@vinistock ugh I got confused with scopes versus frames again. Nevermind!

PR #123, opened 2022-12-22
Implement return + Update next and break

bitwise-aiden:
I've added two new helpers `parse_arguments_without_parenthesis` and `parse_arguments_list_without_parenthesis` to perform the logic of their counter parts, but using `NEWLINE|SEMICOLON|EOF` as the mark of the end of arg list. 

This captures all of the following cases: 
``` ruby 
return 
return 1
return 1, 2, 3
return 1,
  2, 3
return ()
return (1)
return (1), (2), (3)
return (
  1
  2 
)
```

While refusing to parse: 
```ruby
return(1, 2, 3)
return 1,;
```

The parts that I would like extra review on would be: 
- Is it sensible to include the `EOF` in the termination check for arg list?
- Are the new methods overkill?
- Should this be split into multiple PRs?

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -1,4 +1,6 @@
 #include "yarp.h"
+#include "ast.h"
+#include "node.h"
 
 #define STRINGIZE0(expr) #expr
 #define STRINGIZE(expr) STRINGIZE0(expr)
@@ -1600,7 +1602,29 @@ parse_arguments(yp_parser_t *parser, yp_node_t *arguments) {
       yp_node_list_append(parser, arguments, &arguments->as.arguments_node.arguments, expression);
 
       if (accept(parser, YP_TOKEN_PARENTHESIS_RIGHT)) break;
+      // This needs to allow for optional new line
       expect(parser, YP_TOKEN_COMMA, "Expected an ',' to delimit arguments.");
+    }
+  }
+}
+
+// Parse a list of arguments without parenthesis.
+void
+parse_arguments_without_parenthesis(yp_parser_t *parser, yp_node_t *arguments) {
+  if (!accept_any(parser, 3, YP_TOKEN_NEWLINE, YP_TOKEN_SEMICOLON, YP_TOKEN_EOF)) {
+    while (parser->current.type != YP_TOKEN_EOF) {
+      yp_node_t *expression = parse_expression(parser, BINDING_POWER_NONE, "Expected to be able to parse an argument.");
+
+      yp_node_list_append(parser, arguments, &arguments->as.arguments_node.arguments, expression);
+
+      // If parsing the argument resulted in error recovery, then we can stop
+      // parsing the arguments entirely now.
+      if (expression->type == YP_NODE_MISSING_NODE || parser->recovering) break;
+
+      if (accept_any(parser, 3, YP_TOKEN_NEWLINE, YP_TOKEN_SEMICOLON, YP_TOKEN_EOF)) break;
+
+      expect(parser, YP_TOKEN_COMMA, "Expected an ',' to delimit arguments.");
+      accept(parser, YP_TOKEN_NEWLINE);
     }
   }
 }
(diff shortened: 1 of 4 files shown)


shopify-aiden:
> So with that in mind, we can't quite reuse the same codepath for the return node unfortunately.
I've just been going through these and realizing that. Will update now.

shopify-aiden:
Looking at the other similar keywords, it appears as though `break` and `next` will need to be updated to have the same treatment.

src/yarp.c
```diff
@@ -1530,25 +1554,69 @@ parse_arguments(yp_parser_t *parser, yp_node_t *arguments) {
 // present.
 void
 parse_arguments_list(yp_parser_t *parser, yp_arguments_t *arguments) {
```
shopify-aiden, line 1556: I'm unsure if we want to make the change to this function as we will be allowing `return(1, 2, 3)`. 

Depending on how we want to parse `return(1)`, we could add a max arguments check to it. Though if this parses the argument to `(1)`, then that is a separate issue.

src/yarp.c
```diff
@@ -1884,17 +1952,52 @@ parse_expression_prefix(yp_parser_t *parser) {
...
+
+        yp_node_t *expression = parse_expression(parser, BINDING_POWER_NONE, "Expected to be able to parse an argument.");
+
+        // For arg list we would break here, but should this be a return NULL instead?
+        if (parser->recovering) return NULL;
+
+        arguments.arguments = yp_node_arguments_node_create(parser);
+
+        // I feel like the third argument here is redundant
+        yp_node_list_append(parser, arguments.arguments, &arguments.arguments->as.arguments_node.arguments, expression);
```
shopify-aiden, line 1974: // I feel like the third argument here has some redundancies

kddnewton:
@shopify-aiden Coming back to this now, what's the status? I can't quite tell. Also if you don't mind, can you sync with @dirceu who also started working on `return`?

shopify-aiden:
Current state of things: 
- We think that we have parsing for args without parenthesis implemented correctly
- We suspect that there is a general problem with `parse_expression_prefix` not being able to parse expressions that are surrounded by parenthesis, eg: 
```ruby
return (1), (2), (3)
```
- There is a problem with `yield` consuming the first `(` when it actually belongs to the expression being returned, e.g: 
```ruby
yield (1), (2), (3)
```
Here it is in IRB: 
```
irb(main):001:1* def fn 
irb(main):002:1*   yield (1), (2), (3)
irb(main):003:0> end
=> :fn
irb(main):004:0> fn {|a, b, c| puts "A: #{a}, B: #{b}, C: #{c}"} 
A: 1, B: 2, C: 3
```

shopify-aiden:
Preserving the previous description: 

> Converted this back to a draft because there is a lot to unpack with this work. Before the holidays I want to write down my current context so it isn't lost to time. 
> 
> `return`, `next`, and `break` all have the same requirements of `Args|nil` (Although `next` and `break` don't have that in their syntax tree docs, someone might be able to explain that to me). This means that they will accept the following cases: 
> ``` ruby 
> return
> return 1
> return 1, 2, 3
> ```
> 
> Another fun case it will accept is:
> ``` ruby
> return 1, 
>   2, 3
> ```
> 
> And then the final fun cases it will accept are: 
> ```ruby 
> return()
> return(1)
> ```
> 
> Don't let these cases fool you though, because it won't accept: 
> ```ruby
> return(1, 2, 3)
> ```
> 
> Thankfully it also doesn't support: 
> ```ruby 
> return 1,;
> ```
> 
> My current (incomplete) thinking on how to parse this: 
> - if `LPAREN`: 
>   - true -> if `RPAREN`:
>     - true -> done
>     - false -> parse a single expression, followed by `RPAREN`
>   - false -> if `NEWLINE` | `SEMICOLON`: 
>     - true -> no arguments to parse
>     - false -> parse expressions while trailing `COMMA`, stopping for `SEMICOLON` or.... (This is where things get incomplete *)
> 
> 
> \* Does it stop parsing on `SEMICOLON` | `END` | `RESCUE` | `THEN` | `WHEN` | `ENSURE` | etc? 
> 
> You can technically have the following, so I guess no?
> ```ruby
> return 1
> 1
> ```
> 
> At this point my remaining question is how to detect the end of the statement correctly. 
> 
> My working assumption would be that we parse expression, check for `COMMA`, if found, accept `NEWLINE` then parsing expressions until there is no longer a `COMMA`. 
> 
> This also raises the point that the current `parse_arguments` stops parsing on `NEWLINE`, but this is a valid yield statement: 
> ``` ruby
> yield 1, 
>   2
> ```
> 
> Left some additional comments in code, particularly around the `return ()` and `return(1) cases.

src/yarp.c
```diff
@@ -1948,6 +1988,7 @@ parse_expression_prefix(yp_parser_t *parser) {
       yp_node_t *parentheses;
 
       if (parser->current.type != YP_TOKEN_PARENTHESIS_RIGHT && parser->current.type != YP_TOKEN_EOF) {
+        accept(parser, YP_TOKEN_NEWLINE);
```
shopify-aiden, line 1991: @kddnewton do you want me to handle this in this PR or should we pull it out?

PR #126, opened 2022-12-23
Implement StringConcat

### LINKED ISSUE #115: String concatenation
We need a node to represent string concatenation nodes. This is for circumstances like

```ruby
"foo" "bar"
```

or

```ruby
"foo" \
  "bar"
```

It should have a left node, a right node, and an optional `\` operator token. For documentation, see https://github.com/ruby/ruby/blob/master/doc/syntax/literals.rdoc#string-literals-.

vinistock:
Closes #115

Implement string concatenation node. I based it on the current behaviour in Ruby, where it will ignore as many blank spaces as there are after a backslash, but only ignores a single newline. E.g.:
```ruby
"foo" \             "bar" # => "foobar"

"foo" \





  "bar" # => "bar" (doesn't seem to include the "foo")
```

I extracted the logic to advance the parser during lexing on blank spaces and line breaks and tried reusing it. The backslash part is not working just yet.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -1,4 +1,6 @@
 #include "yarp.h"
+#include "ast.h"
+#include "unescape.h"
 
 #define STRINGIZE0(expr) #expr
 #define STRINGIZE(expr) STRINGIZE0(expr)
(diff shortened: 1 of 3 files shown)


src/yarp.c
```diff
@@ -734,10 +739,9 @@ lex_token_type(yp_parser_t *parser) {
             return YP_TOKEN_TILDE_AT;
           return YP_TOKEN_TILDE;
 
-        // TODO
         case '\\':
-          return YP_TOKEN_INVALID;
-
+          advance_blank_characters(parser);
+          break;
```
kddnewton, line 744: We need to return something here and not just `break`. I think if we `break` then it's going to return `YP_TOKEN_INVALID` by default. If we're truly just skipping past these characters, maybe we should just return `lex_token_type(parser)`?

vinistock: I changed it and it's now a little closer to being correct. However, on parsing tests are failing with
```
<[[2, 1], :on_tstring_beg, "'"]> expected but was
<[[1, 7], :on_nl, "\n"]>
```
I guess we're not matching Ripper exactly.

kddnewton: We probably need to emit a newline token after the `\` is found if there is one. Then in the parser you should `accept(parser, YP_TOKEN_NEWLINE)`

kddnewton:
Closing in favor of #268 - thanks for the work here! This got us to the solution

PR #166, opened 2022-12-24
Strings with %, %q, and %Q

### LINKED ISSUE #155: Strings with %
We currently do not support string literals starting with `%`.

DRBragg:
This PR adds support for string literals starting with `%`. String literals starting with `%q` and `%Q` were already implemented but untested so I added tests for them.

TODO:
- [x] Add tests for `%` and `%Q` with and without interpolation 
- [x] Add test for `%q`
- [x] Add tests for using any token, closing or non-closing
- [x] Update `lex_token_type` to handle `%` + token strings

closes #155


src/yarp.c
```diff
@@ -123,6 +123,14 @@ char_is_whitespace(const char *c) {
   return char_is_non_newline_whitespace(c) || *c == '\n';
 }
 
+static inline bool
+char_is_symbol(const char *c) {
```
DRBragg, line 127: This could also be done with `ispunct` but that would involve pulling in `ctype.h`. Since there was already other functions that did char checking I felt this was the best path forward but let me know if pulling in `ctype.h` is preferred.

test/parse_test.rb
```diff
@@ -894,6 +918,66 @@ class ParseTest < Test::Unit::TestCase
...
+        StringNode(nil, STRING_CONTENT(" ccc"), nil)
+      ],
+      STRING_END("}")
+    )
+
+    assert_parses expected, "%{aaa \#{bbb} ccc}"
+  end
+
+  test "% string with any non-closing token" do
+    %w[! @ # $ % ^ & * _ + - | \\ : ; " ' , .].each do |token|
```
DRBragg, line 958: Let me know if this is overkill

src/yarp.c
```diff
@@ -781,7 +789,12 @@ lex_token_type(yp_parser_t *parser) {
               lex_mode_push(parser, (yp_lex_mode_t) { .mode = YP_LEX_STRING, .term = terminator(*parser->current.end++), .interp = true });
               return YP_TOKEN_PERCENT_LOWER_X;
             default:
-              return YP_TOKEN_PERCENT;
+              if (char_is_symbol(parser->current.end)) {
```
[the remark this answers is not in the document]
DRBragg: Ah! Good catch on the extra symbols!  I'll update the test.

I think I understand what you're looking for. I'll give it a whirl.

kddnewton:
Thank you very much @drbragg! I've pulled your commits into #269 and rebased and fixed up some errors, so I'll close this one in favor of that one.

PR #167, opened 2022-12-25
Lex numerics prefixed with plus sign

FnControlOption:
(I hope this PR isn't too small)


kddnewton:
Not too small at all, thanks!

PR #170, opened 2022-12-31
Implement other symbols

### LINKED ISSUE #159: Symbols for other tokens
Currently we handle symbols for identifiers and dynamic strings, but there are other kinds of symbols in Ruby. Those include:

- [ ] Instance variables (`:@foo`)
- [ ] Class variables (`:@@foo`)
- [ ] Global variables (`:$foo`)
- [ ] Keywords (`:self`)
- [ ] Operators (`:+`)
- [ ] Constants (`:Foo`)

We need to handle all of these cases.

ccocchi:
I'm not 100% convinced of what I did (especially when we encounter an error) but here's a quick summary:

* When lexing, I used the presence of whitespace in the next character after a ":" to return an early `SYMBOL_COLON` (Natalie's parser is doing something similar) and correctly determine the start of a symbol. 

* Next when lexing with the `LEX_SYMBOL` mode, I used the same behavior as the `LEX_DEFAULT` mode to get all the tokens independently of being within a symbol, and let the parsing determine what is a correct symbol. It seems to match what Ripper does. 

* Then in parsing, I made a manual list of allowed tokens for symbols. Upon finding an unallowed one, I ended up returning a `TOKEN_INVALID` to fill the `SymbolNode` but it does not feel right. 

I also found an infinite loop when trying to parse `send(:[invalid])` but not when parsing `:[invalid]` nor `send(:[invalid)` which really makes me think I missed something in the error case 😅 

Closes #159

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -774,6 +774,7 @@ static yp_token_type_t
 lex_token_type(yp_parser_t *parser) {
   switch (parser->lex_modes.current->mode) {
     case YP_LEX_DEFAULT:
+    case YP_LEX_SYMBOL:
     case YP_LEX_EMBEXPR:
     case YP_LEX_EMBVAR: {
       // First, we're going to skip past any whitespace at the front of the next
@@ -802,6 +803,8 @@ lex_token_type(yp_parser_t *parser) {
 
       // Next, we'll set to start of this token to be the current end.
       parser->current.start = parser->current.end;
+
+      if (parser->lex_modes.current->mode == YP_LEX_SYMBOL) lex_mode_pop(parser);
 
       // Finally, we'll check the current character to determine the next token.
       switch (*parser->current.end++) {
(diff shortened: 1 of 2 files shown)


src/yarp.c
```diff
@@ -596,7 +599,7 @@ lex_token_type(yp_parser_t *parser) {
         case '!':
           if (match(parser, '=')) return YP_TOKEN_BANG_EQUAL;
           if (match(parser, '~')) return YP_TOKEN_BANG_TILDE;
-          if ((parser->previous.type == YP_TOKEN_KEYWORD_DEF || parser->previous.type == YP_TOKEN_DOT) && match(parser, '@'))
+          if ((parser->previous.type == YP_TOKEN_KEYWORD_DEF || parser->previous.type == YP_TOKEN_DOT || parser->previous.type == YP_TOKEN_SYMBOL_BEGIN ) && match(parser, '@'))
```
ccocchi, line 602: I was also thinking about making an helper method `previous_was_any` similar to `accept_any`, don't know if that makes sense or not.

test/parse_test.rb
```diff
@@ -968,6 +968,35 @@ class ParseTest < Test::Unit::TestCase
...
 
+  test "keyword symbol" do
+    assert_parses SymbolNode(SYMBOL_BEGIN(":"), KEYWORD_SELF("self"), nil), ":self"
+  end
+
+  test "constant symbol" do
+    assert_parses SymbolNode(SYMBOL_BEGIN(":"), CONSTANT("Foo"), nil), ":Foo"
+  end
+
+  test "intance variable symbol" do
```
paracycle, line 979:
```suggestion
  test "instance variable symbol" do
```

kddnewton:
Closing in favor of #271. Thanks for the help here!

PR #171, opened 2022-12-31
Support **nil in method signature

### LINKED ISSUE #160: `**nil`
When a method declaration uses `**nil`, it specifies that it accepts no more keyword arguments. To support this, we will need to update our `parse_parameters` function.

DmitryTsepelev:
Added support for `**nil`, please let me know if the test is correct (maybe we want to wrap in `NilNode`?)

Closes #160

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -822,6 +822,17 @@ nodes:
 
           nil
           ^^^
+  - name: NoKeywordsParameterNode
+    child_nodes:
+      - name: keyword
+        type: token
+    location: keyword
+    comment: |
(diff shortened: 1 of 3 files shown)


kddnewton (changes requested):
Looking at this now, I think it makes sense to make a `NoKeywordsParameterNode` that indicates the presence of `**nil` instead of trying to reuse `KeywordRestParameterNode`.

DmitryTsepelev:
Makes sense, updated the implementation and spec

PR #172, opened 2023-01-02
Implement safe operator

### LINKED ISSUE #156: `&.` safe navigation operator
We need to handle method calls with `&.`. It should be treated with the same precedence as regular method calls with `.` or `::`.

ccocchi:
I added the new token in lexer and used it in the parser to behave like the `.` operator.

I added a test that does not pass yet, using `omit` but I wasn't sure if there was a better way or if we do not want failing tests at the moment. I did not fix it as I didn't think it was part of this PR's scope.

Closes #156

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -14,6 +14,8 @@ tokens:
     comment: "&&"
   - name: AMPERSAND_AMPERSAND_EQUAL
     comment: "&&="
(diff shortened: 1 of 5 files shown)


kddnewton:
Nice work! This is really good!

PR #174, opened 2023-01-05
Improve binary expression handling

HParker:
binary / without a space after it was interpreted as the start of a regular expression Now expressions without spaces should produce the expected result more often.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -722,7 +722,8 @@ lex_token_type(yp_parser_t *parser) {
         // / /=
         case '/':
           if (match(parser, '=')) return YP_TOKEN_SLASH_EQUAL;
-          if (*parser->current.end == ' ') return YP_TOKEN_SLASH;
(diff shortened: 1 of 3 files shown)


src/yarp.c
```diff
@@ -722,7 +722,8 @@ lex_token_type(yp_parser_t *parser) {
         // / /=
         case '/':
           if (match(parser, '=')) return YP_TOKEN_SLASH_EQUAL;
-          if (*parser->current.end == ' ') return YP_TOKEN_SLASH;
+          if (parser->previous.type == YP_TOKEN_INTEGER || parser->previous.type == YP_TOKEN_IDENTIFIER || parser->previous.type == YP_TOKEN_PARENTHESIS_RIGHT)
```
HParker, line 725: I think this check still isn't 100% correct. Can I check that the previous node is a type of `expression` maybe?

kddnewton (changes requested):
At the moment it seems like this PR is doing 2 things: improving our parsing for `/` and handling parentheses. To make it a little easier to review, could you split this into 2 commits? The other comments are inline with the code.

Issue #144, opened 2022-12-23
Parentheses


kddnewton:
We don't currently handle parentheses, but we should. For example:

```ruby
1 * (2 + 3)
```

This should continue parsing but lower the binding power down to the beginning and then expect a closing parenthesis.

jpcamara:
Is there more you could expand on about the scope of this particular issue?

Is this generally about precedence with any arbitrary parens anywhere in a file? 

I pulled down the repo and got everything running. I don't have much C experience but I'd be willing to give it a go!

kddnewton:
Fixed by #175

PR #175, opened 2023-01-05
Improve parenthesized expression handling

HParker:
Expressions in parentheses get their own node which wraps the expression in it.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -1926,6 +1926,22 @@ parse_expression_prefix(yp_parser_t *parser) {
 
       return array;
     }
+    case YP_TOKEN_PARENTHESIS_LEFT: {
+      yp_token_t opening = parser->previous;
+      yp_node_t *parentheses;
+
+      if (parser->current.type != YP_TOKEN_PARENTHESIS_RIGHT && parser->current.type != YP_TOKEN_EOF) {
+        yp_node_t *statements = parse_statements(parser, YP_CONTEXT_BEGIN);
+        parentheses = yp_node_parentheses_node_create(parser, &opening, statements, &opening);
+      } else {
+        yp_node_t *statements = yp_node_statements_create(parser);
+        parentheses = yp_node_parentheses_node_create(parser, &opening, statements, &opening);
+      }
+
+      expect(parser, YP_TOKEN_PARENTHESIS_RIGHT, "Expected a closing parenthesis.");
+      parentheses->as.parentheses_node.closing = parser->previous;
+      return parentheses;
+    }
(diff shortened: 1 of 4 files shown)


src/yarp.c
```diff
@@ -1926,6 +1926,24 @@ parse_expression_prefix(yp_parser_t *parser) {
 
       return array;
     }
+    case YP_TOKEN_PARENTHESIS_LEFT: {
+      yp_token_t opening = parser->previous;
+      yp_node_t *parentheses;
+
+      if (parser->current.type != YP_TOKEN_PARENTHESIS_RIGHT && parser->current.type != YP_TOKEN_EOF) {
+        yp_node_t *statements = yp_node_statements_create(parser);
+        yp_node_t *statement = parse_expression(parser, BINDING_POWER_DEFINED, "Could not parse content in parenthesis.");
```
kddnewton, line 1935: I think we're really close here! The thing is, it's not like an array because it actually allows multiple statements, so it's the lowest binding power.

You can also have something like:

```ruby
(
  a
  b
)
```

and it'll work. So in this case we should use `parse_statements` instead of `parse_expression`.

src/yarp.c
```diff
@@ -1926,6 +1926,22 @@ parse_expression_prefix(yp_parser_t *parser) {
...
+
+      if (parser->current.type != YP_TOKEN_PARENTHESIS_RIGHT && parser->current.type != YP_TOKEN_EOF) {
+        yp_node_t *statements = parse_statements(parser, YP_CONTEXT_BEGIN);
+        parentheses = yp_node_parentheses_node_create(parser, &opening, statements, &opening);
+      } else {
+        yp_node_t *statements = yp_node_statements_create(parser);
+        parentheses = yp_node_parentheses_node_create(parser, &opening, statements, &opening);
+      }
+
+      parentheses->as.parentheses_node.closing = parser->previous;
```
kddnewton, line 1941: So close! We should flip this statement with the next one so that we set the `closing` token to be the right parenthesis after the `expect` function has moved past it.

kddnewton:
Fixes #144

test/parse_test.rb
```diff
@@ -79,6 +79,31 @@ class ParseTest < Test::Unit::TestCase
...
+           ArgumentsNode([IntegerLiteral(INTEGER("1"))]),
+           nil,
+           "+"
+         )]
+      ),
+      PARENTHESIS_RIGHT(")")
+    )
+
+    assert_parses expected, "(1 + 1)"
+  end
```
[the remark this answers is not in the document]
kddnewton: Whoops merged too fast haha. Go ahead and open another PR for that test, that'd be great

PR #185, opened 2023-01-05
Add support for empty statements

HParker:
`parse_statements` should be able to ignore newlines and semicolons before starting to parse an expression.


kddnewton:
Nice, this is great.

PR #187, opened 2023-01-06
Add not keyword

### LINKED ISSUE #116: `not` keyword
We need to handle the `not` keyword. It has two forms: with parentheses and without. It is very similar to a unary operator. It looks like:

```ruby
not foo
not(foo)
```

DmitryTsepelev:
Adding `NotNode` for `not foo` and `not(foo)`. The code is a copy of the `defined?`, should I somehow unify them?

Closes #116.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -824,22 +824,6 @@ nodes:
 
           nil
           ^^^
-  - name: NotNode
-    child_nodes:
-      - name: keyword
-        type: token
-      - name: lparen
-        type: token?
-      - name: value
-        type: node
-      - name: rparen
-        type: token?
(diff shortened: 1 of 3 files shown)


kddnewton:
Sorry I should have put this on the issue.

Would you be able to make this into a `CallNode`? Fundamentally this is calling `!` on the expression inside the `not`. For example:

```ruby
class Foo; def !; "hi"; end; end

not Foo.new
# => "hi"
```

This way we can reuse the nodes we have. In this case the `message` field would be the `not` keyword and the `name` field would be the `"!"` constant string.

kddnewton:
Nice work!!

PR #188, opened 2023-01-06
Support hashrocket hash literals

HParker:
This adds support for parsing hash literals with hash rocket operators. splats are also supported, but other hash key syntax are not implemented.

Closes: https://github.com/Shopify/yarp/issues/136


kddnewton:
Sorry @hparker do you mind rebasing?

HParker:
No prob! Updated

config.yml
```diff
@@ -338,6 +338,52 @@ nodes:
...
+        type: token
+    location: opening->closing
+    comment: |
+      Represents an hash literal
+
+          { a => b }
+          ^^^^^^^^^^
+  - name: AssocNode
+    child_nodes:
+      - name: opening
```
kddnewton, line 357: Ahh I see how you would make this inference, but we don't need these tokens here. You can drop the `opening` and `closing` tokens here, and just use the `key->value` syntax for the location. That should simplify the tree.

config.yml
```diff
@@ -338,6 +338,52 @@ nodes:
...
+        type: token
+    location: opening->closing
+    comment: |
+      Represents an hash key/value combo
+
+          { a => b }
+            ^^^^^^
+  - name: AssocSplat
+    child_nodes:
+      - name: opening
```
kddnewton, line 375: Same here, I think we should call this `operator` and then you can drop the `closing` token, for location it can be `operator->value`.

kddnewton (changes requested):
Just a couple of changes that should hopefully simplify things, then good to go!

test/parse_test.rb
```diff
@@ -1740,6 +1740,47 @@ class ParseTest < Test::Unit::TestCase
...
+  test "parses empty hash" do
+    assert_parses HashNode(BRACE_LEFT("{"), [], BRACE_RIGHT("}")), "{}"
+  end
+
+  test "parses hash with hashrocket keys" do
+    expected = HashNode(
+      BRACE_LEFT("{"),
+      [AssocNode(
+        CallNode(nil, nil, IDENTIFIER("a"), nil, nil, nil, "a"),
+        EQUAL_GREATER("=>"),
```
HParker, line 1752: Do you think we should track this here?

I realized Ruby maybe doesn't keep track of the type of hash syntax you used in the parse tree,

```
ruby --dump=parsetree -e "{ a: b }"                                               11:28:48
###########################################################
## Do NOT use this node dump for any purpose other than  ##
## debug and research.  Compatibility is not guaranteed. ##
###########################################################

# @ NODE_SCOPE (id: 5, line: 1, location: (1,0)-(1,8))
# +- nd_tbl: (empty)
# +- nd_args:
# |   (null node)
# +- nd_body:
#     @ NODE_HASH (id: 4, line: 1, location: (1,0)-(1,8))*
#     +- nd_brace: 1 (hash literal)
#     +- nd_head:
#         @ NODE_LIST (id: 2, line: 1, location: (1,2)-(1,6))
#         +- nd_alen: 2
#         +- nd_head:
#         |   @ NODE_LIT (id: 1, line: 1, location: (1,2)-(1,4))
#         |   +- nd_lit: :a
#         +- nd_head:
#         |   @ NODE_VCALL (id: 0, line: 1, location: (1,5)-(1,6))
#         |   +- nd_mid: :b
#         +- nd_next:
#             (null node)
```

I am starting to think about how we would do `{ a: b }` and maybe it would be difficult to parse the `COLON` as its own token instead of being part of the `LABEL`/ key.

Maybe we can skip this and just track the key and value?

kddnewton (approved):

PR #190, opened 2023-01-08
Rescue keyword

### LINKED ISSUE #141: `rescue` keyword
Right now we have `begin` keyword handling but don't parse `rescue` clauses. These should be modeled like a linked list like the `elsif` clauses are now. The root node should live as an optional node on the `begin` node.

DmitryTsepelev:
Please let me know if tree should be different 🙂

Closes #141.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -344,6 +344,8 @@ nodes:
         type: token
       - name: statements
         type: node
+      - name: rescue_clause
+        type: node?
       - name: ensure_clause
         type: node?
       - name: end_keyword
@@ -1085,6 +1087,27 @@ nodes:
 
         foo rescue nil
         ^^^^^^^^^^^^^^
+  - name: RescueNode
+    child_nodes:
+      - name: rescue_keyword
+        type: token
+      - name: exception_classes
+        type: node[]
+      - name: equal_greater
+        type: token?
+      - name: exception_variable
+        type: token?
+      - name: statements
+        type: node
+      - name: consequent
+        type: node?
+    location: rescue_keyword->statements|consequent
+    comment: |
+      Represents a rescue statement.
+
+          rescue
+            foo
+          ^^^^^^
   - name: RestParameterNode
     child_nodes:
       - name: operator
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -29,6 +29,7 @@ debug_context(yp_context_t context) {
     case YP_CONTEXT_PARENS: return "PARENS";
     case YP_CONTEXT_POSTEXE: return "POSTEXE";
     case YP_CONTEXT_PREEXE: return "PREEXE";
+    case YP_CONTEXT_RESCUE: return "RESCUE";
     case YP_CONTEXT_SCLASS: return "SCLASS";
     case YP_CONTEXT_UNLESS: return "UNLESS";
     case YP_CONTEXT_UNTIL: return "UNTIL";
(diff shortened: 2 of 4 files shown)


src/yarp.c
```diff
@@ -2241,10 +2244,67 @@ parse_expression_prefix(yp_parser_t *parser) {
...
+      // Parse any number of rescue clauses. This will form a linked list of if
+      // nodes pointing to each other from the top.
+      while (parser->current.type == YP_TOKEN_KEYWORD_RESCUE) {
+        parser_lex(parser);
+        yp_token_t rescue_keyword = parser->previous;
+
+        yp_token_t exception_variable;
+        not_provided(&exception_variable, parser->previous.end);
+
+        yp_node_t *statements = yp_node_statements_create(parser);
```
DmitryTsepelev, line 2264: I have to init statements to avoid exception in the generated code but I'll override them down the line 😞Maybe there's a better way to do that

DmitryTsepelev:
BTW looks like current implementation fails where there's more than one `\n` after statements:

```ruby
begin
  a
  <- it will fail because it expects end here
end
```

Maybe we need a follow up for that

src/yarp.c
```diff
@@ -2241,10 +2244,65 @@ parse_expression_prefix(yp_parser_t *parser) {
...
+      }
+
+      switch (parser->current.type) {
+        case YP_TOKEN_KEYWORD_END: {
+          parser_lex(parser);
+          parent->as.begin_node.end_keyword = parser->previous;
+          break;
+        }
+        default:
+          expect(parser, YP_TOKEN_KEYWORD_END, "Expected `end` to close `begin` statement.");
```
kddnewton, line 2300: You can just use `expect` here, as that will effectively do what you're doing in this switch

src/yarp.c
```diff
@@ -2241,10 +2244,65 @@ parse_expression_prefix(yp_parser_t *parser) {
...
+        not_provided(&exception_variable, parser->previous.end);
+
+        yp_node_t *statements = yp_node_statements_create(parser);
+        yp_node_t *rescue = yp_node_rescue_node_create(parser, &rescue_keyword, &exception_variable, statements, NULL);
+
+        while (parser->current.type == YP_TOKEN_CONSTANT) {
+          yp_node_t *expression = parse_expression(parser, BINDING_POWER_NONE, "Expected to find a class.");
+            yp_node_list_append(parser, rescue, &rescue->as.rescue_node.exception_classes, expression);
+
+          if (accept(parser, YP_TOKEN_NEWLINE) || accept(parser, YP_TOKEN_EQUAL_GREATER)) break;
```
kddnewton, line 2271: accept_any will do here

config.yml
```diff
@@ -1043,6 +1045,25 @@ nodes:
           def a(b)
                 ^
           end
+  - name: RescueNode
+    child_nodes:
+      - name: rescue_keyword
+        type: token
+      - name: exception_classes
+        type: node[]
+      - name: exception_variable
```
kddnewton, line 1054: I'd like to capture the `=>` token as well. In general we try to gather up all of the tokens into _some_ node on the tree.

kddnewton (changes requested):
Looks really good so far! Could you rebase and address a couple of comments?

kddnewton:
@DmitryTsepelev here's the issue: https://github.com/Shopify/yarp/issues/205

PR #203, opened 2023-01-10
Support the alias and undef keyword for global variables

Max-Leopold:
Closes https://github.com/Shopify/yarp/issues/117


kddnewton (changes requested):
Just marking this as "request changes" for easier tracking

Max-Leopold:
I changed this now to first parse the arguments and then check for invalid combinations afterwards. 
One of the disadvantages here is that we have two `default: { return NULL }` clauses to make the compiler happy even though we know it can't hit those cases and the nesting for the global variable path is actually one level deeper. Instead of only checking for the token, we now have to check if the node is a `GLOBAL_VARIABLE_READ` and then check the type of the global variable to be able to distinguish between global variables, backreferences and nth references.

I implemented a couple of different variations with varying combinations of nested `switch` and `if` statements. The one in this PR is the one I think is the most readable and easiest to understand even though it is quite verbose.

I also decided to default to erroring instead of default to accepting token combination we didn't expect. Since we know that what type of nodes the `parse_alias_argument` will return this might be overkill. It was meant to be a safeguard if parse_alias_argument should be changed the future.

Example of what I mean by this: Instead of writing

```c
switch(right->as.global_variable_read.name.type) {
  case YP_TOKEN_BACK_REFERENCE:
  case YP_TOKEN_GLOBAL_VARIABLE: {
    break;
  }
  case YP_TOKEN_NTH_REFERENCE: {
    yp_error_list_append(&parser->error_list, "Can't make alias for number variables.", parser->previous.start - parser->start);
    break;
  }
  default: {
    // Cannot hit here.
    return NULL;
  }
}
```
we could also write

```c
switch(right->as.global_variable_read.name.type) {
  case YP_TOKEN_NTH_REFERENCE: {
    yp_error_list_append(&parser->error_list, "Can't make alias for number variables.", parser->previous.start - parser->start);
    break;
  }
}
```

which defaults to accepting the parameter instead of explicitly stating the accepted parameters.

kddnewton (approved):
This looks good!

PR #208, opened 2023-01-12
Add support for `#[]` and `#[]=` calls

### LINKED ISSUE #147: #[] and #[]= calls
We need to handle syntax that calls `#[]` and `#[]=`. This looks like:

```ruby
foo[bar]
```

and

```ruby
foo[bar] = baz
```

These should both be call nodes with a synthesized method name.

paracycle:
Fixes #147


kddnewton:
Yeah this looks good!

PR #215, opened 2023-01-13
Fix parsing binary `+` and `/` with surrounding spaces

haldun:
The following piece of code is being parsed incorrectly as an integer literal:

```ruby
1+1
```

```
irb(main):004:0> YARP.parse('1+1').node
=> Program(Scope([]), Statements([IntegerLiteral(INTEGER("1"))]))
```

This PR fixes the parser so that the code is parser correctly:

```
irb(main):004:0> YARP.parse('1+134').node
=> Program(Scope([]), Statements([CallNode(IntegerLiteral(INTEGER("1")), nil, PLUS("+"), nil, ArgumentsNode([IntegerLiteral(INTEGER("134"))]), nil, "+")]))
```

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -1037,6 +1037,7 @@ lex_token_type(yp_parser_t *parser) {
         case '/':
           if (match(parser, '=')) return YP_TOKEN_SLASH_EQUAL;
           if (*parser->current.end == ' ') return YP_TOKEN_SLASH;
+          if (token_type_is_numeric(parser->previous.type)) return YP_TOKEN_SLASH;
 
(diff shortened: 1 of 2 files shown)


HParker:
How do you feel about adding a test for: 

* `1+(2+3)`
* `1r+2r`
* `1i+2i`

I worry there are other cases I haven't thought of. I think this change is fine, but we should probably consider a different more general whitespace technique eventually.

HParker:
I think we can update this to test `/` now that we are sure we won't confuse it with the start of a regular expression after https://github.com/Shopify/yarp/issues/176

PR #216, opened 2023-01-13
Implement `case` statement

### LINKED ISSUE #137: `case` keyword
We need to parse `case` statements. This ticket is specifically for `case` and `when`, not for pattern matching, which will come later. Important to remember that the `case`'s predicate is not required.

FnControlOption:
Closes #137

I'm not 100% on the names for some of the node fields, so please suggest any better ones that come to mind...

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -34,6 +34,7 @@ debug_context(yp_context_t context) {
     case YP_CONTEXT_UNLESS: return "UNLESS";
     case YP_CONTEXT_UNTIL: return "UNTIL";
     case YP_CONTEXT_WHILE: return "WHILE";
+    case YP_CONTEXT_WHEN: return "WHEN";
   }
   return NULL;
 }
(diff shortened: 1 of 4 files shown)


src/yarp.c
```diff
@@ -1967,11 +1970,11 @@ parse_targets(yp_parser_t *parser, binding_power_t binding_power, const char *me
   return multi_target;
 }
 
-// Parse a list of statements separated by newlines or semicolons.
-static yp_node_t *
-parse_statements(yp_parser_t *parser, yp_context_t context) {
+// Parse statements separated by newlines or semicolons, and append them to
+// the given list of statements.
+static void
+parse_and_append_statements(yp_parser_t *parser, yp_context_t context, yp_node_t *statements) {
```
kddnewton, line 1976: Not sure I understand the need for this change. Were you forced to create the statements early?

kddnewton:
Thank you very much @FnControlOption! I ended up merging another one recently that solves this in a slightly different way, but I really appreciate the work here!

PR #217, opened 2023-01-14
Start parsing lambda literals

### LINKED ISSUE #145: Lambda literals
Currently we don't handle lambda literals at all, but we need to. This involves both the brace and `do`..`end` forms. It should parse parameters, but also parse any potential block locals, as in `-> (; foo) {}`.

paracycle:
Fixes #145


src/yarp.c
```diff
@@ -850,7 +852,9 @@ lex_token_type(yp_parser_t *parser) {
 
         // {
         case '{':
-          if (parser->previous.type == YP_TOKEN_MINUS_GREATER) return YP_TOKEN_LAMBDA_BEGIN;
+          if (parser->previous.type == YP_TOKEN_MINUS_GREATER ||  parser->previous.type == YP_TOKEN_PARENTHESIS_RIGHT) {
```
kddnewton, line 855: I'm okay with doing this for now, but it's definitely going to be a problem later. For example, this would say

```ruby
foo(bar) {}
```

would have a lambda begin, which is incorrect. There's a lot of state we need to track to get these working though, so if you want to push forward here that's okay.

paracycle: Yeah, I thought about that too, and honestly, I am not even sure why this is a different token from `YP_TOKEN_BRACE_LEFT`. It feels like it shouldn't matter to the lexer what context the `{` appears in.

I had gotten the parsing code to work with matching either `YP_TOKEN_LAMBDA_BEGIN` or `YP_TOKEN_BRACE_LEFT`, but it failed to match the test output against the Ripper output, which we will have to resolve if we will implement this differently.

paracycle: Ok, I reverted this commit, since it wasn't helping me in the case of non-parenthesised lambda arguments anyway. Now I don't know how to resolve the test failures due to the mismatch between YARP and Ripper outputs.

PR #219, opened 2023-01-14
Top level constants

### LINKED ISSUE #146: Top-level constants
Right now we parse constants, but don't handle if they're prefixed with `::`. We should parse top-level constant lookups, like:

```ruby
::Foo
```

This will involve adding a clause to the `parse_expression_prefix` function. When this is done, we will also need to support assigning to top-level constants, as in:

```ruby
::Foo = 1
```

That will involve updating our handling of the `=` operator in the `parse_expression_infix` function.

DmitryTsepelev:
Somehow it worked out even without changing `parse_expression_infix`, please let me know if result tree is wrong 🙂

Closes #146.


ccocchi:
Didn't see you tackled this already 😭 at least we went for different implementations.

src/yarp.c
```diff
@@ -2621,6 +2621,12 @@ parse_expression_prefix(yp_parser_t *parser) {
       return yp_node_class_variable_read_create(parser, &parser->previous);
     case YP_TOKEN_CONSTANT:
       return yp_node_constant_read_create(parser, &parser->previous);
+    case YP_TOKEN_COLON_COLON: {
+      yp_token_t delimiter = parser->previous;
+      yp_node_t *child = parse_expression(parser, BINDING_POWER_CALL, "Expected constant after ::");
```
ccocchi, line 2626: I think Ruby doesn't allow anything other than a constant after `::`, and if I'm not mistaken `parse_expression` will accept something like `::a` here.

DmitryTsepelev: No worries! @kddnewton could you please take a look at both implementations and decide if we want a new node or keep current ones? I'll fix my implementation if we decide to go with this direction because yeah, it allows doing `::A::b` now 😞

DmitryTsepelev:
@kddnewton done, please check 🙂 cc @ccocchi

src/yarp.c
```diff
@@ -2508,6 +2508,22 @@ parse_interpolated_string_parts(yp_parser_t *parser, yp_token_type_t end_type, y
...
 
+static inline yp_node_t *
+parse_top_level_constant(yp_parser_t *parser, yp_node_t *parent) {
+  yp_token_t delimeter = parser->previous;
+
+  expect(parser, YP_TOKEN_CONSTANT, "Expected constant after ::.");
+  yp_token_t constant = parser->previous;
+  yp_node_t *constant_node = yp_node_constant_read_create(parser, &constant);
+
+  if (accept(parser, YP_TOKEN_COLON_COLON)) {
```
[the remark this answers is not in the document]
DmitryTsepelev: done!

kddnewton (approved):

PR #221, opened 2023-01-16
Implement calling with unnamed rest parameter 

haldun:
This PR contains implementation for calling a method with unnamed rest parameter.

```ruby
def foo(*)
  bar(*)
end
```

The behavior is very similar to the forwarding parameter `...`.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -2050,9 +2050,9 @@ parse_argument(yp_parser_t *parser, yp_node_t *arguments) {
   if (accept(parser, YP_TOKEN_STAR)) {
     yp_token_t previous = parser->previous;
     if (*parser->current.start == ',' || *parser->current.start == ')') {
-      // if (!yp_token_list_includes(&parser->current_scope->as.scope.locals, &parser->previous)) {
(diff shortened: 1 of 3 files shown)


src/yarp.c
```diff
@@ -2001,18 +2001,26 @@ parse_statements(yp_parser_t *parser, yp_context_t context) {
...
-    argument = yp_node_forwarding_arguments_node_create(parser, &parser->previous);
-  } else {
-    argument = parse_expression(parser, BINDING_POWER_NONE, "Expected to be able to parse an argument.");
+    return yp_node_forwarding_arguments_node_create(parser, &parser->previous);
   }
 
-  return argument;
+  if (accept(parser, YP_TOKEN_STAR)) {
+    yp_token_t previous = parser->previous;
+    if (*parser->current.start == ',' || *parser->current.start == ')') {
```
DmitryTsepelev, line 2013: Should do the same but more idiomatic:
```suggestion
    if (accept_any(parser, 2, YP_TOKEN_COMMA, YP_TOKEN_PARENTHESIS_RIGHT) {
```

kddnewton (changes requested):
(Just marking this a request changes)

kddnewton (approved):
Looks good!

PR #225, opened 2023-01-18
Add checking errors in `assert_parses`

haldun:
In order to check both the parse result and possible errors, I extended `assert_parses` in a way that it will also check for the expected errors. Not used anywhere yet.


kddnewton:
Ahh, that's what https://github.com/Shopify/yarp/blob/main/test/errors_test.rb is for. The idea is to put syntax errors in there so `parse_test` is just the valid Ruby.

Issue #153, opened 2022-12-23
Method parameter order enforcement


kddnewton:
At the moment we allow any method parameter declarations to be in any position in the list. So for example `def foo(&bar, *baz)` is permitted by our parser but shouldn't be.

To maintain flexibility, we want to continue to allow the parameters to be parsed in any order, but we want to add errors if they are found to be incorrect. This allows the user to type them incorrectly and to have a nice experience, as opposed to just crashing.

HParker:
Related, the compiler needs to know how many params are `pre` or `post` optional arguments for methods like,

```
def a(x, y = 1, z)
end
```

PR #227, opened 2023-01-18
Enforce method parameters order

ccocchi:
I tried to tackle #153. The error message is very generic, I don't know if that's you had in mind either with the message or the implementation.

I also added an `assert_errors_only` helper to make the test more readable but let me know if you want me to remove it and use `expression` in those tests instead (I only found that it checks the parser's result is the same twice and had no impact on errors).


kddnewton:
Thanks @ccocchi Adam is picking this up as part of #669

PR #233, opened 2023-01-20
auto-generate specs based on parser testsuite

DerekStride:
## What

I'm interested in ways to position this project into a standard set of specs for ruby parsers. I'm taking inspiration from [tree-sitter tests](https://tree-sitter.github.io/tree-sitter/creating-parsers#command-test). I'd like to introduce some tooling to auto-generate specs for the AST generated by ruby code.

> The new parser should be able to be built and linked as its own static library.

One of the stated goals of this project is portability, allowing other projects to build and link against YARP easily. Having a set of specs aimed at improving compatibility with other parsers might seem counter-intuitive. However, other parsers have established eco-systems of tools built on top of them already. Having a spec allows other parsers to have confidence that the AST they produce is compatible with native ruby tooling.

## How

Introduces `bin/specs`, a program to hook into the parser testsuite to auto-generate specs for the YARP parser. The specs are saved in a format that is easy for other parser implementations to consume in order to validate compatibility with YARP.

The spec generator hooks into `YARP#parse` to capture all of the source code used in the parser tests.

## Comments

* I've chosen the format that tree-sitter uses for its tests for the specs. I think it's easy to generate & parse. It's also very human friendly.
* The `bin/spec` program is a proof-of-concept and I'm happy to move the tooling into the gem itself (or into a   development specific library).
    * It might also be nice to move it into a rake task instead.
* Having the specs generated as part of CI to ensure they are up-to-date would be great.
    * Alternatively, having the generation run as part of CI (to ensure it doesn't break) and moving the specs into an       artifact published with releases of the gem would work well too.


DerekStride:
Thanks for the feedback, I'm going to close this until I have time to come back to it in the future.

PR #238, opened 2023-01-27
Support block with braces

HParker:
This adds support for block arguments syntax. Focusing on `do/end` and `{/}` blocks with `|` variables.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -297,6 +297,7 @@ debug_context(yp_context_t context) {
     case YP_CONTEXT_ELSE: return "ELSE";
     case YP_CONTEXT_ELSIF: return "ELSIF";
     case YP_CONTEXT_EMBEXPR: return "EMBEXPR";
+    case YP_CONTEXT_BLOCK_BRACES: return "BLOCK_BRACES";
     case YP_CONTEXT_FOR: return "FOR";
     case YP_CONTEXT_IF: return "IF";
     case YP_CONTEXT_MAIN: return "MAIN";
@@ -2259,6 +2260,8 @@ context_terminator(yp_context_t context, yp_token_t *token) {
       return token->type == YP_TOKEN_KEYWORD_ELSE || token->type == YP_TOKEN_KEYWORD_ELSIF || token->type == YP_TOKEN_KEYWORD_END;
     case YP_CONTEXT_EMBEXPR:
       return token->type == YP_TOKEN_EMBEXPR_END;
+    case YP_CONTEXT_BLOCK_BRACES:
+      return token->type == YP_TOKEN_BRACE_RIGHT;
     case YP_CONTEXT_PARENS:
       return token->type == YP_TOKEN_PARENTHESIS_RIGHT;
     case YP_CONTEXT_BEGIN:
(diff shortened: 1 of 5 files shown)


src/yarp.c
```diff
@@ -2154,6 +2157,49 @@ parse_arguments_list(yp_parser_t *parser, yp_arguments_t *arguments) {
...
+    }
+
+    accept(parser, YP_TOKEN_NEWLINE);
+
+    arguments->arguments =  yp_node_arguments_node_create(parser);
+
+    yp_node_t *statements;
+
+    if (accept(parser, YP_TOKEN_BRACE_RIGHT) || accept(parser, YP_TOKEN_KEYWORD_END)) {
+      statements = yp_node_statements_create(parser);
```
kddnewton, line 2188: Instead of creating this here, how about we make this optional?

kddnewton: Additionally, we'll have to do something different for braces versus do/end because with do/end you can use `rescue`/`else`/`ensure`.

src/yarp.c
```diff
@@ -2154,6 +2157,49 @@ parse_arguments_list(yp_parser_t *parser, yp_arguments_t *arguments) {
...
+
+  if (accept(parser, YP_TOKEN_BRACE_LEFT) || accept(parser, YP_TOKEN_KEYWORD_DO)) {
+    not_provided(parser);
+    not_provided(parser);
+
+    yp_token_t opening = parser->previous;
+
+    accept(parser, YP_TOKEN_NEWLINE);
+
+    yp_node_t *block_arguments = yp_node_arguments_node_create(parser);
```
kddnewton, line 2169: I think we want to parse parameters here, not arguments. Because you can do things like

```ruby
foo do |bar:, baz:|
end
```

src/yarp.c
```diff
@@ -1714,6 +1715,8 @@ context_terminator(yp_context_t context, yp_token_t *token) {
       return token->type == YP_TOKEN_KEYWORD_ELSE || token->type == YP_TOKEN_KEYWORD_ELSIF || token->type == YP_TOKEN_KEYWORD_END;
     case YP_CONTEXT_EMBEXPR:
       return token->type == YP_TOKEN_EMBEXPR_END;
+    case YP_CONTEXT_BLOCK:
```
kddnewton, line 1718: We'll want different contexts here because this is going to accept `end` for `{`

kddnewton (changes requested):
A good start!

src/yarp.c
```diff
@@ -2283,6 +2225,106 @@ parse_parameters(yp_parser_t *parser) {
...
+    not_provided(parser);
+
+    yp_token_t opening = parser->previous;
+
+    accept(parser, YP_TOKEN_NEWLINE);
+
+    yp_node_t *block_arguments = NULL;
+    if (accept(parser, YP_TOKEN_PIPE)) {
+      block_arguments = parse_parameters(parser);
+      expect(parser, YP_TOKEN_PIPE, "expected block arguements to end with '|'.");
```
HParker, line 2303: I didn't handle `accept(parser, YP_TOKEN_SEMICOLON)` here like @paracycle 's change did since I think I would end up duplicating almost exactly what he did there and we should probably share that code somewhere.

paracycle: I'd extracted lambda/block parameter parsing code to its own function in my branch, which you can cherry-pick and start using if you want: https://github.com/Shopify/yarp/pull/217/commits/4a03263cb3f23e2dd55ae3057b8f51adc8f754c4

paracycle: Though I guess you would need this too: https://github.com/Shopify/yarp/pull/217/commits/e9c086dad959d98d694fa30cc4f3bca2967a9fb5#diff-5a7db8dbadaef1b1b5a8738ba70b5ffac82b8e243732154165911284e08aad4bR402-R413

HParker: Maybe I should wait for lambda support to merge? I think this PR would be ok waiting.

kddnewton (approved):

PR #239, opened 2023-01-28
Parse optional keyword parameters

orhantoy:
This PR adds parsing of optional keyword parameters.
Resolves https://github.com/Shopify/yarp/issues/231

### Concerns

Not sure if

```
if (parser->current.type != YP_TOKEN_COMMA && parser->current.type != YP_TOKEN_PARENTHESIS_RIGHT) {
```

is sufficient to handle all kinds of optional keyword parameters.


kddnewton (changes requested):
This looks really good! One architectural note: to keep it consistent with other trees, could you make it so that `KeywordParameterNode` has an optional value as opposed to making a new node? That should hopefully make it easier for consumers. Other than that, if you can sign the CLA this looks about ready to merge!

src/yarp.c
```diff
@@ -2224,8 +2224,22 @@ parse_parameters(yp_parser_t *parser) {
...
+          // If parsing the value of the keyword parameter resulted in error recovery,
+          // then we can put a missing node in its place and stop parsing the
+          // parameters entirely now.
+          if (parser->recovering) return params;
+
+          parsing = false;
+        } else {
+          yp_node_t *param = yp_node_keyword_parameter_node_create(parser, &name, NULL);
+          yp_node_list_append(parser, params, &params->as.parameters_node.keywords, param);
+        }
```
orhantoy, line 2241: There is opportunity to refactor the duplicate `yp_node_keyword_parameter_node_create` calls but for now I just wanted to confirm if the new direction made sense.

Max-Leopold:
Hey, I also had a look at this issue and one headache I had was that optional keyword parameters behave differently depending
on if the method is defined with or without parenthesis.
The specific edge case is that if you define method parameters with parenthesis, you are allowed to have newline breaks between
the keyword parameter name and the expression.
E.g.
```ruby
def a(b:
1
)
 puts b
end
```
is valid ruby code and is interpreted very differently to
```ruby
def a b:
 1
 puts b
end
```
For default optional parameters this isn't a problem because they have the `=` specifically indicates that this parameter is optional.
This is only one of the edge cases I found, maybe there are more.

I did go down the route of passing a parameter `uses_parenthesis` into `parse_parameters` to allow for parenthesis aware parsing of the parameters.
I wasn't particularly happy with it though and wanted to search for a better alternative which is why I never opened a PR for it.

I just pushed my code to a draft PR [here](https://github.com/Shopify/yarp/pull/240). Maybe we can come up with a better solution together :)

orhantoy:
@Max-Leopold Thanks for highlighting your findings ✨ 

My first thought was to look into how the argument list is parsed but turns out YARP currently doesn't support keyword parameters:

```ruby
  test "call with parentheses and keyword arguments" do
    expected = nil # TODO

    assert_parses expected, "a(b, c: 2)"
  end
```

<details>
<summary>Parse error</summary>

```
  <[#<YARP::ParseError:0x0000000100d1c058
    @location=(4..4),
    @message="Expected to be able to parse an argument.">,
   #<YARP::ParseError:0x0000000100d17fd0
    @location=(4..4),
    @message="Expected a ')' to close the argument list.">]> was expected to be empty.
```
</details>

Newlines in the argument list are also not supported at the moment:

```ruby
  test "call with parentheses and arguments separated by newlines" do
    expected = nil # TODO

    assert_parses expected, "a(b,\nc)"
  end
```

<details>
<summary>Parse error</summary>

```
  <[#<YARP::ParseError:0x0000000100c8d678
    @location=(4..4),
    @message="Expected to be able to parse an argument.">,
   #<YARP::ParseError:0x0000000100c8ca20
    @location=(4..4),
    @message="Expected a ')' to close the argument list.">]> was expected to be empty.
```
</details>

And finally, the positional parameters also do not support being newline-separated at the moment:

```ruby
  test "def with required parameter separated by newlines" do
    expected = nil # TODO

    assert_parses expected, "def a(b,\nc)\nend"
  end
```

<details>
<summary>Parse error</summary>

```
  <[#<YARP::ParseError:0x00000001046d7180
    @location=(8..8),
    @message="Expected ')' after left parenthesis.">,
   #<YARP::ParseError:0x00000001046d7108
    @location=(10..10),
    @message="Expected `end` to close `def` statement.">]> was expected to be empty.
```
</details>

---

So the gaps/edge-cases you're mentioning reg. keyword parameters followed by newlines should IMO be tracked and solved as a separate issue, as the issue is more general as can be seen by the above test examples.

orhantoy:
@Max-Leopold Closing my PR in favor of [yours](https://github.com/Shopify/yarp/pull/240) as it covers more scenarios ✌️ 
There might also be some inspiration to draw from Adam's PR: https://github.com/Shopify/yarp/pull/243

PR #240, opened 2023-01-28
Parse optional keyword parameters

Max-Leopold:
Parse optional keyword parameters. This is a little bit more tricky than parsing optional default parameters because there is no straight forward way to tell if a keyword parameter is optional or not and the behaviour also differs depending on wether the method parameters are defined with or without parenthesis. For more context see https://github.com/Shopify/yarp/pull/239.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -848,12 +848,18 @@ nodes:
     child_nodes:
       - name: name
         type: token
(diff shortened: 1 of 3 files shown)


kddnewton (approved):
Great work on this!

PR #243, opened 2023-01-31
Add support for keyword arg defaults

HParker:
Now we can parse both,

```
def foo(bar: :baz)
end
```

and

```
hi_there default: "friend"
```


src/yarp.c
```diff
@@ -2224,7 +2237,16 @@ parse_parameters(yp_parser_t *parser) {
...
+	yp_node_t *default_value = NULL;
+	if (parser->current.type != YP_TOKEN_NEWLINE && // TODO: do we need terminator here?
+	    parser->current.type != YP_TOKEN_SEMICOLON &&
+	    parser->current.type != YP_TOKEN_COMMA &&
+	    parser->current.type != YP_TOKEN_PARENTHESIS_RIGHT &&
+	    parser->current.type != YP_TOKEN_EOF) {
+	  default_value = parse_expression(parser, BINDING_POWER_NONE, "Expected a default value after keyword argument which does not terminate.");
+	}
+
+        yp_node_t *param = yp_node_keyword_parameter_node_create(parser, &name, default_value);
```
HParker, line 2249: Are you ok reusing the `keyword_parameter` node here? I am not sure how isolated parameter and argument are intended to be from each other.

kddnewton (changes requested):
Mentioned in the inline comment, but hoping to split this into two prs.

kddnewton:
@Max-Leopold and @orhantoy could you get together with @HParker on this? It looks like we've got some duplicated efforts going on, and each of your PRs handles it in slightly different ways. I think maybe there are other edge cases in the other PRs that could be of benefit here as well.

HParker:
I read over the PRs, now just handling method call kwarg values I should overlap _less_ 😅

src/yarp.c
```diff
@@ -2037,6 +2037,65 @@ parse_statements(yp_parser_t *parser, yp_context_t context) {
...
+
+    if (hash->as.hash_node.elements.size != 0) {
+      expect(parser, YP_TOKEN_COMMA, "expected a comma between hash entries.");
+    }
+
+    while (accept(parser, YP_TOKEN_NEWLINE));
+    yp_node_t *element;
+
+    switch (parser->current.type) {
+    case YP_TOKEN_STAR_STAR: {
```
orhantoy, line 2053: `case`s are usually indented one level below the `switch`, perhaps some C linter would catch that eventually, but just mentioning as it caught my eye.

HParker: Thanks! I think I gotta change an editor config. 🤦‍♂️

Max-Leopold:
Hey 👋  sorry I only had time to look at this again now. 

My [PR](https://github.com/Shopify/yarp/pull/240) was only intended to showcase the differing behaviour when parsing optional keyword parameters with or without parenthesis. I saw that @orhantoy closed his PR now and I'm happy to change my draft to include more edge cases so the parsing is correct in all places.

> Can you split this into 2 PRs? I think you've got the keyword parameter stuff correct, and I'd love to merge that.

It seemed that your PR was dealing with the optional keyword parameters correctly so I'm also very happy to close my PR and you can create a second PR with you optional keyword parameter parsing. I unfortunately can't see how you parsed the optional keyword parameters anymore because of the force push.

Regarding the overlap: I think as this PR stands today it doesn't overlap with the changes @orhantoy or I did for the parameter parsing. It's only doing keyword arguments now so should be good to 🚢 if @kddnewton is happy with it.

HParker:
> It seemed that your PR was dealing with the optional keyword parameters correctly so I'm also very happy to close my PR and you can create a second PR with you optional keyword parameter parsing.

Yeah, sorry the force push made those changes impossible to review. After reviewing your and @orhantoy's PRs, I was not handling nearly as many edge cases as either of your PRs. I also initially made the mistake of modeling method definition kwargs as the same as method call kwargs.

I agree that there is now no real overlap.

HParker:
merged via: https://github.com/Shopify/yarp/pull/260

PR #258, opened 2023-02-08
Support symbol to proc

HParker:
foo(&:hi)

Should parse now. Things like

    foo(&hi.to_sym)

also works, just like in Ruby. There are probably some things that this supports that are invalid, but this is a start.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -668,6 +668,18 @@ nodes:
             bar(...)
             ^^^^^^^^
           end
+  - name: SymbolToProcNode
+    child_nodes:
+      - name: operator
+        type: token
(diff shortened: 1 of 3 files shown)


kddnewton:
I think you might be overlapping with https://github.com/Shopify/yarp/pull/234 here. 

Regardless, definitely don't want to call it `SymbolToProcNode` since it can be any kind of expression.

HParker:
ah, blah, thanks! I think I have another issue storing this as a regular argument since it likely needs to be in the block location in the args, so I think that is enough reason for me to close in favor of the other PR.

PR #262, opened 2023-02-14
Implement parsing operators in method definition

haldun:
This PR implements parsing method definitions where the method name is an operator.

Fixes https://github.com/Shopify/yarp/issues/259

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -1447,7 +1447,7 @@ lex_token_type(yp_parser_t *parser) {
 
         // xstring literal
         case '`': {
-          if (!lex_state_p(parser, YP_LEX_STATE_FNAME|YP_LEX_STATE_DOT)) {
+          if (!lex_state_operator_p(parser)) {
             yp_lex_mode_t lex_mode = {
               .mode = YP_LEX_STRING,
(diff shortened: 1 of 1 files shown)


src/yarp.c
```diff
@@ -541,6 +578,11 @@ lex_state_operator_p(yp_parser_t *parser) {
   return lex_state_p(parser, YP_LEX_STATE_FNAME | YP_LEX_STATE_DOT);
 }
 
+static inline void
+parser_set_lex_state(yp_parser_t *parser, yp_lex_state_t state) {
+  parser->lex_state = state;
+}
```
HParker, line 584: I think this change isn't required for this PR. Maybe we can do it elsewhere?

kddnewton (changes requested):
This PR looks great!

Just one thing: in almost all of the places where you have `parser->lex_state == YP_LEX_STATE_FNAME` I think CRuby is checking `IS_AFTER_OPERATOR` which is equivalent in our case to `YP_LEX_STATE_FNAME | YP_LEX_STATE_DOT`. Can you change that to make that equivalent?

A `static inline bool lex_state_after_operator_p` would be good for this.

PR #264, opened 2023-02-14
Add `parser_set_lex_state`

haldun:
This PR adds an inline function `parser_set_lex_state` to replace direct assignments to `parser->lex_state` in the code. This allows to be able to put a breakpoint into `parser_set_lex_state` and help debugging. Introduces no behavior change.


kddnewton (changes requested):
Yeah I'm cool with adding this method, but can you name is `lex_state_set` to be consistent with the other prefixes?

Also can you rebase?

PR #273, opened 2023-02-16
Set token.full_start that includes whitespaces

### LINKED ISSUE #158: Whitespace on tokens
Right now, we skip whitespace _before_ the tokens start. We skip past all of the content before marking the start of the token. The tokens themselves have two pointers: `start` and `end`. We should have an additional pointer on all of the tokens called `full_start` that includes whitespace.

We need this to handle the `warn_indent` directive and also will need it eventually for error recovery to better handle the `end` keyword.

mrzasa:
I'm setting `token.full_start` in the lexer, before the whitespaces are skipped.

I don't know how to test it properly - `full_start` seems not to be available in the Ruby code. For now I print output to check that it's OK, but I'd appreciate if you could advise me, how to write unit tests for this case.

Fixes #158

### DIFF
--- a/bin/templates/src/ast.h.erb
+++ b/bin/templates/src/ast.h.erb
@@ -24,6 +24,7 @@ typedef struct {
   yp_token_type_t type;
   const char *start;
   const char *end;
+  const char *full_start;
 } yp_token_t;
 
 typedef struct {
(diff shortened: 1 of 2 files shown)


kddnewton:
Thanks @mrzasa! I've gone back and forth on this one, but since we've attached tokens to the tree, we don't want to make it any bigger than we have to. We're going to revisit the way we approach this problem entirely. Thanks for the contribution!

PR #274, opened 2023-02-16
Parse keywords as method names

haldun:
TIL that keywords are allowed as method names, so this PR extends parsing them.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -951,69 +951,71 @@ lex_identifier(yp_parser_t *parser) {
     }
   }
 
-  switch (width) {
-    case 2:
-      if (lex_keyword(parser, "do", YP_LEX_STATE_BEG, false)) {
-        if (yp_state_stack_p(&parser->do_loop_stack)) {
-          return YP_TOKEN_KEYWORD_DO_LOOP;
+  if (parser->lex_state != YP_LEX_STATE_DOT) {
+    switch (width) {
+      case 2:
+        if (lex_keyword(parser, "do", YP_LEX_STATE_BEG, false)) {
+          if (yp_state_stack_p(&parser->do_loop_stack)) {
+            return YP_TOKEN_KEYWORD_DO_LOOP;
+          }
+          return YP_TOKEN_KEYWORD_DO;
         }
-        return YP_TOKEN_KEYWORD_DO;
-      }
 
-      if (lex_keyword(parser, "if", YP_LEX_STATE_BEG, true)) return YP_TOKEN_KEYWORD_IF;
-      if (lex_keyword(parser, "in", YP_LEX_STATE_BEG, false)) return YP_TOKEN_KEYWORD_IN;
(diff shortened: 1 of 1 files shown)


src/yarp.c
```diff
@@ -850,6 +900,8 @@ lex_keyword(yp_parser_t *parser, const char *value, yp_lex_state_t state, bool m
   if (strncmp(parser->current.start, value, strlen(value)) == 0) {
     if (parser->lex_state & YP_LEX_STATE_FNAME) {
       lex_state_set(parser, YP_LEX_STATE_ENDFN);
+    } else if (parser->lex_state == YP_LEX_STATE_DOT) {
```
kddnewton, line 903: I'm not understanding where this change is coming from. This is the code from CRuby:

```c
if (mb == ENC_CODERANGE_7BIT && (!IS_lex_state(EXPR_DOT) || enforce_keyword_end)) {
	const struct kwtable *kw;

	/* See if it is a reserved word.  */
	kw = rb_reserved_word(tok(p), toklen(p));
	if (kw) {
	    enum lex_state_e state = p->lex.state;
	    if (IS_lex_state_for(state, EXPR_FNAME)) {
		SET_LEX_STATE(EXPR_ENDFN);
		set_yylval_name(rb_intern2(tok(p), toklen(p)));
		return kw->id[0];
	    }
	    SET_LEX_STATE(kw->state);
	    if (IS_lex_state(EXPR_BEG)) {
		p->command_start = TRUE;
	    }
	    if (kw->id[0] == keyword_do) {
		if (lambda_beginning_p()) {
		    p->lex.lpar_beg = -1; /* make lambda_beginning_p() == FALSE in the body of "-> do ... end" */
		    return keyword_do_LAMBDA;
		}
		if (COND_P()) return keyword_do_cond;
		if (CMDARG_P() && !IS_lex_state_for(state, EXPR_CMDARG))
		    return keyword_do_block;
		return keyword_do;
	    }
	    if (IS_lex_state_for(state, (EXPR_BEG | EXPR_LABELED | EXPR_CLASS)))
		return kw->id[0];
	    else {
		if (kw->id[0] != kw->id[1])
		    SET_LEX_STATE(EXPR_BEG | EXPR_LABEL);
		return kw->id[1];
	    }
	}
    }
```

Can you explain this state change?

kddnewton: I've put it into a PR.

We need to set that state before we lex the subsequent token, as per the production rule. We need to do it in any places where we're about to reduce. So for example

https://github.com/Shopify/yarp/pull/275/files#diff-71ae829a627932e547cf825850d9a6b4773b009b0b68ae8eb1d8ef02e6d98987R4081
https://github.com/Shopify/yarp/pull/275/files#diff-71ae829a627932e547cf825850d9a6b4773b009b0b68ae8eb1d8ef02e6d98987R4112
https://github.com/Shopify/yarp/pull/275/files#diff-71ae829a627932e547cf825850d9a6b4773b009b0b68ae8eb1d8ef02e6d98987R4172

PR #309, opened 2023-02-24
Dedent tilde heredocs

### LINKED ISSUE #310: Dedenting heredocs (e.g., <<~FOO)

jemmaissroff:
This PR implements the correct indentation in tilde heredocs. It calculates the minimum whitespace indentation, and then applies that to all lines within the heredoc.

Closes #310

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -3377,11 +3377,22 @@ lex_token_type(yp_parser_t *parser) {
         }
       }
 
+      // If we're in a tilde heredoc, we need to calculate the minimum amount of leading
+      // whitespace so that we can "ignore" this amount of whitespace on all lines in the
+      // heredoc.
+      //
+      // In order to calculate the minimum amount of whitespace, we find the end of the
+      // heredoc, and then iterate from the start until the end, counting the whitespace
+      // on every line and using the `ignored_whitespace` field on the heredoc struct to
+      // track the minimum whitespace we see.
       if (heredoc_struct.indent == YP_HEREDOC_INDENT_TILDE &&
               heredoc_struct.ignored_whitespace == -1) {
         int min_non_newline_whitespace = heredoc_struct.ignored_whitespace;
 
         const char *end_of_heredoc = parser->current.end - 1;
+
+        // Increment the end_of_heredoc pointer until we find something that matches the
+        // start
         while(strncmp(end_of_heredoc, ident_start, ident_length) != 0) {
             end_of_heredoc++;
             if (end_of_heredoc >= parser->end) {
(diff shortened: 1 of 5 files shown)


src/yarp.c
```diff
@@ -3262,21 +3284,90 @@ lex_token_type(yp_parser_t *parser) {
...
+              heredoc_struct.ignored_whitespace == -1) {
+        int min_non_newline_whitespace = heredoc_struct.ignored_whitespace;
+
+        const char *end_of_heredoc = parser->current.end - 1;
+
+        // Increment the end_of_heredoc pointer until we find something that matches the
+        // start
+        while(strncmp(end_of_heredoc, ident_start, ident_length) != 0) {
+            end_of_heredoc++;
+            if (end_of_heredoc >= parser->end) {
```
jemmaissroff, line 3305: This is the case where the heredoc isn't properly closed. I put this line in temporarily so it doesn't segfault, before we decide what do to here.

Mirror ripper and lex all lines as strings?

src/yarp.c
```diff
@@ -3355,21 +3377,90 @@ lex_token_type(yp_parser_t *parser) {
...
 
+      // If we're in a tilde heredoc, we need to calculate the minimum amount of leading
+      // whitespace so that we can "ignore" this amount of whitespace on all lines in the
+      // heredoc.
+      //
+      // In order to calculate the minimum amount of whitespace, we find the end of the
+      // heredoc, and then iterate from the start until the end, counting the whitespace
+      // on every line and using the `ignored_whitespace` field on the heredoc struct to
+      // track the minimum whitespace we see.
+      if (heredoc_struct.indent == YP_HEREDOC_INDENT_TILDE &&
```
kddnewton, line 3388: Originally, I thought this was correct, but looking at it more, I'm not so sure now.

Consider this very annoying example:

```ruby
p <<~FOO
    #{<<~FOO}
        inner
    FOO
  outer
FOO
```

In the snippet above, this block will loop until it finds the first `FOO` and then it will be very confused.

I don't think we can eagerly look forward and just match blindly without running the parser.

The other example doesn't involve nesting, just confusing interpolation:

```ruby
p <<~FOO
    #{
FOO
    }
FOO
```

I think we're going to have to have the context of the parser in order to know what we're allowed to consider as a terminator.

I don't really know the solution here, or how CRuby does it. Maybe it doesn't emit any of the string content events until it has parsed the full heredoc? I'm not sure.

kddnewton (changes requested):

jemmaissroff:
Closing in favor of #342 which has a better solution

PR #313, opened 2023-02-25
Support case statements

HParker:
This supports both case when and case in statements

```
case :hi
when :hi
end
```

becomes,

```
CaseNode(
  KEYWORD_CASE("case"),
  SymbolNode(SYMBOL_BEGIN(":"), IDENTIFIER("hi"), nil),
  [CaseConditionNode(
     KEYWORD_WHEN("when"),
     [SymbolNode(SYMBOL_BEGIN(":"), IDENTIFIER("hi"), nil)],
     nil
   )],
  nil,
  KEYWORD_END("end")
)
```

A stab at: https://github.com/Shopify/yarp/issues/137

I see there is another open PR, so feel free to close in favor of that one, but this is here as an option.

This handles both `when` and `in` forms of case statements with else clauses.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -795,6 +795,39 @@ nodes:
 
           if foo then bar end
           ^^^^^^^^^^^^^^^^^^^
+  - name: CaseNode
+    child_nodes:
+      - name: case_keyword
+        type: token
+      - name: predicate
+        type: node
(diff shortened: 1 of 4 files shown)


kddnewton (approved):
Looks great!

PR #315, opened 2023-02-25
Fix lexer state for the question mark

haldun:
Fixes the lexer state when we see a question mark, i.e. form like:

```ruby
a = foo ? bar : baz
```

#### `rake lex` results

**Before**
```
PASSING=2591
FAILING=2063
```

**After**
```
PASSING=2605
FAILING=2049
```


kddnewton:
Thanks @haldun I ended up grabbing this in #328

PR #319, opened 2023-02-25
Fix lex state for the symbol arguments

haldun:
Fixes the lex state for the symbols in arguments, e.g.

```ruby
a.should_receive(:+).and_return(:result)
```

#### `rake lex` results

**Before**
```
PASSING=2591
FAILING=2063
```

**After**
```
PASSING=2708
FAILING=1946
```

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -3442,7 +3442,11 @@ lex_token_type(yp_parser_t *parser) {
         return YP_TOKEN_INVALID;
       }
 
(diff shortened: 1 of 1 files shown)


kddnewton:
You're totally right, a bunch of stuff needed to change here. 

I've updated it in https://github.com/Shopify/yarp/pull/327 and I think I'm going to go with that one, which strips out this mode entirely.

PR #323, opened 2023-02-26
Support rescue, else and ensure on def nodes

HParker:
this adds support for the syntax:

```
def a
rescue
else
ensure
end
```

I think this proves out a pattern that can apply to other places which support rescue clauses. If we like this, I think it should be easy to apply to the rest of the nodes which support trailing rescue clauses.


HParker:
@kddnewton how do you like the new structure?

```
class A
rescue 
else
ensure
end"
```

```
ClassNode(
  Scope([]),
  KEYWORD_CLASS("class"),
  ConstantRead(CONSTANT("A")),
  nil,
  nil,
  BeginNode(
    NOT_PROVIDED(""),
    Statements([]),
    RescueNode(KEYWORD_RESCUE("rescue"), [], nil, nil, Statements([]), nil),
    ElseNode(KEYWORD_ELSE("else"), Statements([]), SEMICOLON(";")),
    EnsureNode(KEYWORD_ENSURE("ensure"), Statements([]), KEYWORD_END("end")),
    nil
  ),
  KEYWORD_END("end")
)
```

PR #335, opened 2023-02-28
Fix case-else with else containing statements

haldun:
This PR fixes parsing

```ruby
case arg
when foo
  :a
else
  :b
end
```

However, I am not sure about what should be the `end_keyword` of the `ElseNode` here. Currently it is `newline` but I don't know whether this is OK. On the other hand I see that `CaseNode` also has an `end_keyword` child, so in this case I don't know what the last `end` token should be part of, should it be part of `CaseNode` or `ElseNode`?

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -3653,7 +3653,7 @@ context_terminator(yp_context_t context, yp_token_t *token) {
     case YP_CONTEXT_ENSURE:
       return token->type == YP_TOKEN_KEYWORD_END;
     case YP_CONTEXT_CASE_WHEN:
-      return token->type == YP_TOKEN_KEYWORD_WHEN || token->type == YP_TOKEN_KEYWORD_END;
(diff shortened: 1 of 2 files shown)


kddnewton (changes requested):
I think we should attach the `end` keyword to both, I don't think it needs to be exclusively owned by one or the other. Can you add that to this PR?

src/yarp.c
```diff
@@ -5201,7 +5201,7 @@ parse_expression_prefix(yp_parser_t *parser) {
         yp_node_t *else_node;
 
         if (!match_type_p(parser, YP_TOKEN_KEYWORD_END)) {
-          else_node = yp_node_else_node_create(parser, &else_keyword, parse_statements(parser, YP_CONTEXT_CASE_WHEN), &parser->previous);
+          else_node = yp_node_else_node_create(parser, &else_keyword, parse_statements(parser, YP_CONTEXT_CASE_WHEN), &parser->current);
```
kddnewton, line 5204: This is definitely the right change, but we shouldn't be using `YP_CONTEXT_CASE_WHEN` here, we should be passing `YP_CONTEXT_ELSE`. Can you change that?

kddnewton (approved):

PR #352, opened 2023-03-01
Mirror Ripper's expected regexp_end state on Ruby layer

jemmaissroff:
On Regexp end, Ripper scans tokens and then sets the lex state. This means that in Ripper, "regexp_end" has an inaccurate end state. We accommodate this on the Ruby layer when comparing Ripper and YARP lexed output


jemmaissroff:
This PR takes `rake lex` percentage correct from 80.25% to 85.24%

PR #354, opened 2023-03-01
Add a DEBUG_MODE macro, and automated debugging prints on state change

jemmaissroff:
When DEBUG_MODE is toggled to 1, then on any state change, print information about where the state was changed, and what it was changed to.

Hopefully, as we continue debugging, we can make this have more info for things other than state changes, and make the DEBUG_MODE more fine grained.

For example, when running `bin/lex test.rb` on this file:

```ruby
#
1
```

output with `DEBUG_MODE` toggled to 1 is:

```
Caller: lex_token_type:2787
Previous: STATE: YP_LEX_STATE_BEG
Now: STATE: YP_LEX_STATE_END

Caller: lex_newline:2214
Previous: STATE: YP_LEX_STATE_END
Now: STATE: YP_LEX_STATE_BEG
```

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -8,6 +8,10 @@ char* yp_version(void) {
   return YP_VERSION_MACRO;
 }
 
+#define DEBUG_MODE 0
+
+#if DEBUG_MODE
+
 /******************************************************************************/
 /* Debugging                                                                  */
 /******************************************************************************/
@@ -142,6 +146,8 @@ debug_scope(yp_parser_t *parser) {
 
   fprintf(stderr, "\n");
 }
+
+#endif
 
 /******************************************************************************/
 /* Node initializers                                                          */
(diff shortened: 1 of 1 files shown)


src/yarp.c
```diff
@@ -8,6 +8,10 @@ char* yp_version(void) {
   return YP_VERSION_MACRO;
 }
 
+#define DEBUG_MODE 0
```
kddnewton, line 11: Can you use `ifndef` here? That way we could pass `-D` if we wanted to turn it on (I think?)

jemmaissroff: Changed! But I couldn't figure out the correct make incantation to change the flag. Was trying the suggestions [here](https://stackoverflow.com/questions/54094963/passing-parameters-to-make-to-change-define-variable). Any thoughts?

PR #358, opened 2023-03-02
Fix parsing arrays with newlines and without trailing comma

haldun:
Fixes parsing arrays with newlines but without trailing comma:

```ruby
[:a,
 :b
]
```

Takes `rake lex` from `88.87%` to `90.22%`.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -5208,6 +5208,9 @@ parse_expression_prefix(yp_parser_t *parser, binding_power_t binding_power) {
       yp_node_t *array = yp_array_node_create(parser, &opening, &opening);
 
       while (!match_any_type_p(parser, 2, YP_TOKEN_BRACKET_RIGHT, YP_TOKEN_EOF)) {
+        // Handle the case where we don't have a comma and we have a newline followed by a right bracket.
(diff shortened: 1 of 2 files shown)


src/yarp.c
```diff
@@ -5208,6 +5208,15 @@ parse_expression_prefix(yp_parser_t *parser, binding_power_t binding_power) {
       yp_node_t *array = yp_array_node_create(parser, &opening, &opening);
 
       while (!match_any_type_p(parser, 2, YP_TOKEN_BRACKET_RIGHT, YP_TOKEN_EOF)) {
+        // Handle the case where we don't have a comma and we have a newline followed by a right bracket.
+        bool seen_newline = false;
```
kddnewton, line 5212:
```suggestion
                if (accept(parser, YP_TOKEN_NEWLINE) && match_type_p(parser, YP_TOKEN_BRACKET_RIGHT)) break;
```

Only the first newline is going to be `YP_TOKEN_NEWLINE`, the rest will be ignored newlines. So in this case you should be able to just check for one.

kddnewton (changes requested):

PR #371, opened 2023-03-03
Fix parsing module names starting with `::`

haldun:
Fixes parsing:

```ruby
module ::Foo
end
```

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -5990,10 +5990,7 @@ parse_expression_prefix(yp_parser_t *parser, binding_power_t binding_power) {
       parser_lex(parser);
 
       yp_token_t module_keyword = parser->previous;
-      yp_node_t *name;
-
-      expect(parser, YP_TOKEN_CONSTANT, "Expected to find a module name after `module`.");
-      name = yp_node_constant_read_create(parser, &parser->previous);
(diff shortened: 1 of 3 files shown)


src/yarp.c
```diff
@@ -5990,10 +5990,7 @@ parse_expression_prefix(yp_parser_t *parser, binding_power_t binding_power) {
       parser_lex(parser);
 
       yp_token_t module_keyword = parser->previous;
-      yp_node_t *name;
-
-      expect(parser, YP_TOKEN_CONSTANT, "Expected to find a module name after `module`.");
-      name = yp_node_constant_read_create(parser, &parser->previous);
+      yp_node_t *name = parse_expression(parser, BINDING_POWER_CALL, "Expected to find a module name after `module`.");
```
kddnewton, line 5993: I don't want to call `parse_expression` here because that will gather up stuff like

```ruby
foo.bar.baz
```

I think in this PR we should explicitly support checking for a `::` here instead.

kddnewton (changes requested):

src/yarp.c
kddnewton: Yes please!

kddnewton (approved):

PR #380, opened 2023-03-03
Skip comment state when lexing

jemmaissroff:
This PR slightly refactors how we compare lexed tokens in order to set up a nicer API for ignoring state on comment tokens.


lib/yarp.rb
```diff
@@ -165,8 +167,64 @@ def visit_child_nodes(node)
...
+      inspect[index]
+    end
+
+    def ==(other)
+      other in LexedToken[
+        location: ^(location),
+        event: ^(event),
+        value: ^(value),
+        state: ^(state)
+      ] or other in LexedToken[
```
jemmaissroff, line 197: Is there a better way to do this in one pattern matching comparison? Couldn't quite figure it out...

lib/yarp.rb
```diff
@@ -165,8 +167,64 @@ def visit_child_nodes(node)
...
+
+    def inspect
+      [@location, @event, @value, @state]
+    end
+
+    def deconstruct_keys(keys)
+      { location: location, event: event, value: value, state: state }
+    end
+
+    def pretty_print(q)
```
jemmaissroff, line 216: I'm new to `pretty_print` so copied other methods, let me know if this should be done differently though!

lib/yarp.rb
```diff
@@ -165,8 +167,64 @@ def visit_child_nodes(node)
   # This lexes with the Ripper lex. It drops any space events but otherwise
   # returns the same tokens.
   def self.lex_ripper(source)
-    Ripper.lex(source).each_with_object([]) do |token, tokens|
-      tokens << token unless token[1] == :on_sp
+    Ripper.lex(source).each_with_object([]) do |token_arr, tokens|
+      token = LexedToken.new(*token_arr)
```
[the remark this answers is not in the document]
jemmaissroff: Oh, I wasn't understanding that at all, thanks for explaining.

Happy to do this if you prefer, but to me I think it makes sense to just have a `LexedToken` (or `Token` -- just the name is used elsewhere) type, and do the comparison on that type. It's a little odd to me to have the tokens be sometimes arrays, and sometimes `CommentTokens`. I think _not_ special casing `CommentToken`s also makes the API a little nicer with `LexedToken#is_tilde_heredoc?`. 

I pushed one more stab at this, getting rid of most of the `LexedToken` methods since it's `SimpleDelegator`, but still preserving the named methods. Let me know how that looks to you. If you'd still prefer the arrays, with just a `CommentToken` class, I can definitely change to that too.

kddnewton (approved):
Looks good!

PR #407, opened 2023-03-04
Fix parsing empty %w list

### LINKED ISSUE #402: Empty `%w` lists with whitespace
```ruby
%w{
                        }
```

DmitryTsepelev:
Fixes #402.

Please let me know if I miss something.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -6352,6 +6352,9 @@ parse_expression_prefix(yp_parser_t *parser, binding_power_t binding_power) {
       yp_token_t opening = parser->previous;
       yp_node_t *array = yp_array_node_create(parser, &opening, &opening);
 
+      // skip all leading whitespaces
+      accept(parser, YP_TOKEN_WORDS_SEP);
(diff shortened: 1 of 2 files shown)


src/yarp.c
```diff
@@ -6352,6 +6352,11 @@ parse_expression_prefix(yp_parser_t *parser, binding_power_t binding_power) {
       yp_token_t opening = parser->previous;
       yp_node_t *array = yp_array_node_create(parser, &opening, &opening);
 
+      // skip all leading whitespaces
+      while (parser->current.type == YP_TOKEN_WORDS_SEP) {
```
kddnewton, line 6356: I think there should only be a single token here, so you should be able to do just `accept(parser, YP_TOKEN_WORDS_SEP);` without the look

kddnewton (changes requested):

DmitryTsepelev:
fixed

kddnewton (approved):

PR #408, opened 2023-03-06
Emit error message for module definition in method body

haldun:
Ruby does not allow defining a module in a method body https://github.com/ruby/ruby/blob/22d944c8b76be04dc437100c46626db5fe9dd7f0/parse.y#L3514

This PR emits an error message for this case.


kddnewton (changes requested):
Ahh nice!

We'll need to do a little more though, because we could be inside a block inside a `def`. So we'll need to walk the linked list all of the way up to the top to check.

PR #439, opened 2023-03-08
%= should come before %string

### LINKED ISSUE #392: `defined?` with parentheses has different operator precedence
```ruby
defined?(x %= 2).should == "assignment"
```

jemmaissroff:
Closes #392

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -3075,7 +3075,11 @@ lex_token_type(yp_parser_t *parser) {
             return YP_TOKEN_INVALID;
           }
 
-          if (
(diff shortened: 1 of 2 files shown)


src/yarp.c
```diff
@@ -3063,7 +3063,11 @@ lex_token_type(yp_parser_t *parser) {
             return YP_TOKEN_INVALID;
           }
 
-          if (
+          if (match(parser, '=')) {
```
kddnewton, line 3066:
```suggestion
          if (!lex_state_beg_p(parser) && match(parser, '=')) {
```

PR #460, opened 2023-03-10
Improvements to the generated Java code

eregon:
The toString() is quite simplistic but it's a start and for sure better than `org.yarp.Nodes$Program@6818d900`.
For `a=1; a+2` it looks like this:
```
Program
  Scope
  Statements
    LocalVariableWrite
      IntegerNode
    CallNode
      LocalVariableRead
      ArgumentsNode
        IntegerNode
```
For comparison the JRuby AST looks like this:
```
JRuby:
RootNode line: 0
  BlockNode line: 0
    LocalAsgnNode*:a line: 0
      FixnumNode line: 0, long: 1
    OperatorCallNode*:+ line: 0
      LocalVarNode:a line: 0
      ArrayNode line: 0
        FixnumNode line: 0, long: 2

TruffleRuby's printing of the JRuby AST:
RootParseNode
  BlockParseNode
    LocalAsgnParseNode:a
      FixnumParseNode[value=1]
    CallParseNode:+
      LocalVarParseNode:a
      ArrayParseNode
        FixnumParseNode[value=2]

TruffleRuby's Truffle AST:
  RubyLambdaRootNode
    body = SequenceNode
      body[0] = WriteLocalVariableNode
        valueNode = ProfileArgumentNodeGen
          childNode_ = ReadSelfNode
      body[1] = WriteLocalVariableNode
        valueNode = IntegerFixnumLiteralNode
      body[2] = InlinedAddNodeGen
        leftNode_ = ReadLocalVariableNode
        rightNode_ = IntegerFixnumLiteralNode
```

cc FYI @enebo @andrykonchin


eregon:
BTW, I noticed Ruby nodes have `child_nodes` already, so it seems nice to have it for Java too (JRuby AST has `children()`).

For a fancier `toString()` it's tempting to actually generate a custom toString per class. The trade-off is that would increase the amount of logic/bytecode per node class for toString which is mostly for debugging. But that increase may be negligible and is probably worth the flexibility of it. Just wanted to note the advantage of using `childNodes()`.

bin/templates/java/org/yarp/Nodes.java.erb
```diff
@@ -43,27 +53,67 @@ public abstract class Nodes {
...
     }
 
     <%- nodes.each do |node| -%>
 
     <%= "#{node.comment.split("\n").map { |line| "// #{line}" }.join("\n    ")}\n" if node.comment -%>
     public static final class <%= node.name -%> extends Node {
-        <%-node.params.each do |param| -%>
-        public final <%= param.java_type %> <%= param.name %>;
+        <%- node.params.each do |param| -%>
+        public final <%= param.java_type %> <%= param.name %>;<%= ' // optional' if param.class.name.start_with?('Optional') %>
```
nirvdrum, line 81: Would it make sense to use `Optional<T>` here?

bin/template.rb
```diff
@@ -20,6 +20,8 @@ def rbs_class = "Node?"
   def java_type = "Node"
 end
 
+SingleNodeParam = -> node { NodeParam === node or OptionalNodeParam === node }
```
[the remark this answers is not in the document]
eregon: What if I use it in `node.rb.erb` too and maybe some other files?

bin/templates/java/org/yarp/Nodes.java.erb
eregon: It means they can be `null`. We don't use Java's `Optional` because that's a lot of extra allocations and indirections.

PR #499, opened 2023-03-13
Fix parsing singular bare hash argument

### LINKED ISSUE #494: Singular bare hash arguments
```ruby
outbuf = IO.popen(cmd, :err => [:child, :out], &:read)
```

haldun:
Fixes #494

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -4923,7 +4923,10 @@ parse_arguments(yp_parser_t *parser, yp_node_t *arguments, yp_token_type_t termi
           argument = bare_hash;
 
           // Then parse more if we have a comma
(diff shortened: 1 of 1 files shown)


src/yarp.c
```diff
@@ -4923,7 +4923,9 @@ parse_arguments(yp_parser_t *parser, yp_node_t *arguments, yp_token_type_t termi
           argument = bare_hash;
 
           // Then parse more if we have a comma
-          if (accept(parser, YP_TOKEN_COMMA)) parse_assocs(parser, argument);
+          if (accept(parser, YP_TOKEN_COMMA) && !match_type_p(parser, YP_TOKEN_AMPERSAND)) {
```
kddnewton, line 4926: Instead of checking for just ampersand, can you check `token_begins_expression_p`? It should catch more.

kddnewton (changes requested):

PR #506, opened 2023-03-13
Fix parsing method calls with constant paths

### LINKED ISSUE #479: Constants and local reads can have blocks and become calls
```ruby
foo = 1
foo do
end
```

```ruby
Const do
end
```

### LINKED ISSUE #496: Constant paths can be method calls
```ruby
Bundler::URI uri
```

haldun:
Fixes #496
Fixes #479

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -6180,7 +6180,8 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
       // fact a method call, not a constant read.
       if (
         match_type_p(parser, YP_TOKEN_PARENTHESIS_LEFT) ||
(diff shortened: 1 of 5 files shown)


test/parse_test.rb
```diff
@@ -2053,6 +2053,23 @@ def a(b:
     assert_parses expected, "foo(\n:a,\nb: :c,\n)"
   end
 
+  test "method call with constant paths" do
+    expected = CallNode(
+      ConstantReadNode(CONSTANT("Bundler")),
+      COLON_COLON("::"),
+      IDENTIFIER("uri"),
```
kddnewton, line 2060: This token seems wrong. It should be `CONSTANT("URI")` not the identifier. I think something may have gotten messed up here.

kddnewton (changes requested):

PR #531, opened 2023-03-15
Do not read anything off the end

kddnewton:
@eregon this should fix the failures you were seeing. Thanks for the repro branch!


kddnewton:
Yeah @eregon for sure

PR #534, opened 2023-03-15
Emit error messages for "bad" parameter types

haldun:
Emit an error message when we see a 

- constant 
- instance variable
- global variable
- class variable

in the parameters list like CRuby:

```
ruby -e "def foo(A);end"                                      
-e:1: formal argument cannot be a constant
def foo(A);end
```

Ref https://github.com/ruby/ruby/blob/56b38fdd69d56e8ff2ae9032fa55f2eda8bb5d63/parse.y#L5599


kddnewton:
Nice this looks great

PR #547, opened 2023-03-17
Fix segfault for invalid syntax

haldun:
yarp creates a segfault when we see an unexpected token after method call:

```ruby
foo() @foo
foo() 123
foo() "bar"
```

This PR fixes the segfault but it basically just eats up the unexpected tokens after the method call. I think we should emit some kind of error message for those cases.


kddnewton:
Thanks @haldun! I found the issue in #567.

PR #548, opened 2023-03-17
Emit an error message when assigning to a reserved numbered parameter

haldun:
Emits an error message when we see an assignment to a reserved numbered parameter:

```ruby
_1 = :a
```


kddnewton:
Looks great!

PR #572, opened 2023-03-19
Fix calling `parse_expression_infix` for unary operators

### LINKED ISSUE #516: Unary minus can begin expression in labelled bare hash
```ruby
sort("ygData.msgId": -1)
```

haldun:
Fixes #516

TBH I am not 100% sure about the way, but the tests are OK and I don't see any regression in the targets but instead a positive change in a very big Ruby codebase.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -4347,11 +4347,13 @@ typedef enum {
 typedef struct {
   yp_binding_power_t left;
   yp_binding_power_t right;
+  bool binary;
 } yp_binding_powers_t;
 
-#define BINDING_POWER_ASSIGNMENT { YP_BINDING_POWER_UNARY, YP_BINDING_POWER_ASSIGNMENT }
-#define LEFT_ASSOCIATIVE(precedence) { precedence, precedence + 1 }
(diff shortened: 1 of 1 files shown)


src/yarp.c
```diff
@@ -8304,7 +8304,12 @@ parse_expression(yp_parser_t *parser, yp_binding_power_t binding_power, const ch
   // Otherwise we'll look and see if the next token can be parsed as an infix
   // operator. If it can, then we'll parse it using parse_expression_infix.
   yp_binding_powers_t current_binding_powers;
-  while (current_binding_powers = yp_binding_powers[parser->current.type], binding_power <= current_binding_powers.left) {
+  while (
+    current_binding_powers = yp_binding_powers[parser->current.type],
+    binding_power <= current_binding_powers.left &&
+    current_binding_powers.right != YP_BINDING_POWER_UNARY &&
```
kddnewton, line 8310: I definitely agree this is the right solution but I'm nervous about this accidentally introducing something we're going to have to remember to change.

What do you think about introducing a `binary` attribute to the `binding_powers_t` struct instead? Then here we would check `&& current_binding_powers.binary` instead.

PR #573, opened 2023-03-19
Fix block params

haldun:
This PR fixes:

1. Allow newlines after the final block parameter.

```ruby
foo lambda { |
  a: 1,
  b: 2
  |
}
```

~2. Emit an error message for the trailing commas in block parameters:~

```ruby
foo lambda { |
  a: 1,
  b: 2,
  |
}
```

This turned out to be a bit trickier to solve, and this is not only for block parameters but also method definitions too. So, as far as I understand Ruby does not allow trailing commas after label and rest arguments:

```ruby
def foo(a: 1,)
end


def foo(a: 1, **c,)
end


def foo(...,)
end
```

All these are not allowed but yarp does not emit an error message.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -5532,7 +5532,7 @@ parse_block(yp_parser_t *parser) {
   if (accept(parser, YP_TOKEN_PIPE)) {
     parameters = parse_block_parameters(parser);
     parser->command_start = true;
-    while (accept(parser, YP_TOKEN_NEWLINE));
+    accept(parser, YP_TOKEN_NEWLINE);
     expect(parser, YP_TOKEN_PIPE, "Expected block parameters to end with '|'.");
   }
 


src/yarp.c
```diff
@@ -5532,6 +5532,7 @@ parse_block(yp_parser_t *parser) {
   if (accept(parser, YP_TOKEN_PIPE)) {
     parameters = parse_block_parameters(parser);
     parser->command_start = true;
+    while (accept(parser, YP_TOKEN_NEWLINE));
```
kddnewton, line 5535: There should only be one here, the rest should be ignored newlines. So this can just be `accept`.

kddnewton (changes requested):

kddnewton (approved):

PR #611, opened 2023-03-21
parse_dup and parse_file_dup APIs

kddnewton:
These APIs dup the input string coming in in order to provide a hard limit to give valgrind a better chance of finding any errors where we may read off the end of the string.


eregon:
Great, thank you!

src/yarp.c
```diff
@@ -2081,7 +2081,7 @@ lex_optional_float_suffix(yp_parser_t *parser) {
 
   // Here we're going to attempt to parse the optional decimal portion of a
   // float. If it's not there, then it's okay and we'll just continue on.
-  if (*parser->current.end == '.') {
+  if ((parser->current.end < parser->end) && *parser->current.end == '.') {
```
eregon, line 2084: I wonder if some helper function would be useful here, something like `at(parser, parser->current.end)` which returns 0 if `pointer >= parser->end` and reads the pointer otherwise (JRuby's lexer has a `p()` helper, although that one lets the JVM check bounds vs checking explicitly).

Also maybe some `peek` function like JRuby's RubyLexer has:
```java
    public boolean peek(int c) {
        return peek(c, 0);
    }

    protected boolean peek(int c, int n) {
        return lex_p + n < lex_pend && p(lex_p + n) == c;
    }
````

PR #650, opened 2023-03-23
Refactor locations default

jemmaissroff:
Set a default location instead of using `YP_LOCATION_NULL_VALUE`


jemmaissroff:
Closed in favor of removing locations entirely in #656

PR #652, opened 2023-03-23
Ensure enough space in the serialization buffer

### LINKED ISSUE #646: Corrupted buffer length during dump.
Something is corrupting a buffer length field while file dumping.   The output before the realloc error shows at some point 
buffer->length is clearly overwritten with a bad value. 

```text
mri31 -Ilib -ryarp -e 'YARP.dump_file("/home/enebo/work/jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb")'
LENGTH: 1, BL: 1024, CAP: 1024, 2xCAP: 2048
LENGTH: 4749, BL: 1065, CAP: 2048, 2xCAP: 4096
LENGTH: 1, BL: 5814, CAP: 4096, 2xCAP: 8192
realloc(): invalid next size
Aborted (core dumped)
```

```diff
diff --git a/src/util/yp_buffer.c b/src/util/yp_buffer.c
index 6766db8..d13bf9b 100644
--- a/src/util/yp_buffer.c
+++ b/src/util/yp_buffer.c
@@ -1,3 +1,4 @@
+#include <stdio.h>
 #include "util/yp_buffer.h"
 
 #define YP_BUFFER_INITIAL_SIZE 1024
@@ -20,8 +21,16 @@ yp_buffer_init(yp_buffer_t *buffer) {
 static inline void
 yp_buffer_append(yp_buffer_t *buffer, const void *source, size_t length) {
   if (buffer->length + length > buffer->capacity) {
+  fprintf(stderr, "LENGTH: %ld, BL: %ld, CAP: %ld, 2xCAP: %ld\n", length, buffer->length, buffer->capacity, buffer->capacity * 2);
     buffer->capacity = buffer->capacity * 2;
-    buffer->value = realloc(buffer->value, buffer->capacity);
+
+
+    char *t = realloc(buffer->value, buffer->capacity);
+    if (t == NULL) {
+      fprintf(stderr, "REALLOC FAILED in yp_buffer_append\n");
+    }
+
+    buffer->value = t;
   }
   memcpy(buffer->value + buffer->length, source, length);
   buffer->length += length;
```

kernel_require.rb
```text
# frozen_string_literal: true
#--
# Copyright 2006 by Chad Fowler, Rich Kilmer, Jim Weirich and others.
# All rights reserved.
# See LICENSE.txt for permissions.
#++

require "monitor"

module Kernel

  RUBYGEMS_ACTIVATION_MONITOR = Monitor.new # :nodoc:

  # Make sure we have a reference to Ruby's original Kernel#require
  unless defined?(gem_original_require)
    alias gem_original_require require
    private :gem_original_require
  end

  file = Gem::KERNEL_WARN_IGNORES_INTERNAL_ENTRIES ? "<internal:#{__FILE__}>" : __FILE__
  module_eval <<'RUBY', file, __LINE__ + 1 # rubocop:disable Style/EvalWithLocation
  ##
  # When RubyGems is required, Kernel#require is replaced with our own which
  # is capable of loading gems on demand.
  #
  # When you call <tt>require 'x'</tt>, this is what happens:
  # * If the file can be loaded from the existing Ruby loadpath, it
  #   is.
  # * Otherwise, installed gems are searched for a file that matches.
  #   If it's found in gem 'y', that gem is activated (added to the
  #   loadpath).
  #
  # The normal <tt>require</tt> functionality of returning false if
  # that file has already been loaded is preserved.

  def require(path)
    if RUBYGEMS_ACTIVATION_MONITOR.respond_to?(:mon_owned?)
      monitor_owned = RUBYGEMS_ACTIVATION_MONITOR.mon_owned?
    end
    RUBYGEMS_ACTIVATION_MONITOR.enter

    path = path.to_path if path.respond_to? :to_path

    if spec = Gem.find_unresolved_default_spec(path)
      # Ensure -I beats a default gem
      resolved_path = begin
        rp = nil
        load_path_check_index = Gem.load_path_insert_index - Gem.activated_gem_paths
        Gem.suffixes.each do |s|
          $LOAD_PATH[0...load_path_check_index].each do |lp|
            safe_lp = lp.dup.tap(&Gem::UNTAINT)
            begin
              if File.symlink? safe_lp # for backward compatibility
                next
              end
            rescue SecurityError
              RUBYGEMS_ACTIVATION_MONITOR.exit
              raise
            end

            full_path = File.expand_path(File.join(safe_lp, "#{path}#{s}"))
            if File.file?(full_path)
              rp = full_path
              break
            end
          end
          break if rp
        end
        rp
      end

      begin
        Kernel.send(:gem, spec.name, Gem::Requirement.default_prerelease)
      rescue Exception
        RUBYGEMS_ACTIVATION_MONITOR.exit
        raise
      end unless resolved_path
    end

    # If there are no unresolved deps, then we can use just try
    # normal require handle loading a gem from the rescue below.

    if Gem::Specification.unresolved_deps.empty?
      RUBYGEMS_ACTIVATION_MONITOR.exit
      return gem_original_require(path)
    end

    # If +path+ is for a gem that has already been loaded, don't
    # bother trying to find it in an unresolved gem, just go straight
    # to normal require.
    #--
    # TODO request access to the C implementation of this to speed up RubyGems

    if Gem::Specification.find_active_stub_by_path(path)
      RUBYGEMS_ACTIVATION_MONITOR.exit
      return gem_original_require(path)
    end

    # Attempt to find +path+ in any unresolved gems...

    found_specs = Gem::Specification.find_in_unresolved path

    # If there are no directly unresolved gems, then try and find +path+
    # in any gems that are available via the currently unresolved gems.
    # For example, given:
    #
    #   a => b => c => d
    #
    # If a and b are currently active with c being unresolved and d.rb is
    # requested, then find_in_unresolved_tree will find d.rb in d because
    # it's a dependency of c.
    #
    if found_specs.empty?
      found_specs = Gem::Specification.find_in_unresolved_tree path

      found_specs.each do |found_spec|
        found_spec.activate
      end

    # We found +path+ directly in an unresolved gem. Now we figure out, of
    # the possible found specs, which one we should activate.
    else

      # Check that all the found specs are just different
      # versions of the same gem
      names = found_specs.map(&:name).uniq

      if names.size > 1
        RUBYGEMS_ACTIVATION_MONITOR.exit
        raise Gem::LoadError, "#{path} found in multiple gems: #{names.join ', '}"
      end

      # Ok, now find a gem that has no conflicts, starting
      # at the highest version.
      valid = found_specs.find {|s| !s.has_conflicts? }

      unless valid
        le = Gem::LoadError.new "unable to find a version of '#{names.first}' to activate"
        le.name = names.first
        RUBYGEMS_ACTIVATION_MONITOR.exit
        raise le
      end

      valid.activate
    end

    RUBYGEMS_ACTIVATION_MONITOR.exit
    return gem_original_require(path)
  rescue LoadError => load_error
    RUBYGEMS_ACTIVATION_MONITOR.enter

    begin
      if load_error.path == path and Gem.try_activate(path)
        require_again = true
      end
    ensure
      RUBYGEMS_ACTIVATION_MONITOR.exit
    end

    return gem_original_require(path) if require_again

    raise load_error
  ensure
    if RUBYGEMS_ACTIVATION_MONITOR.respond_to?(:mon_owned?)
      if monitor_owned != (ow = RUBYGEMS_ACTIVATION_MONITOR.mon_owned?)
        STDERR.puts [$$, Thread.current, $!, $!.backtrace].inspect if $!
        raise "CRITICAL: RUBYGEMS_ACTIVATION_MONITOR.owned?: before #{monitor_owned} -> after #{ow}"
      end
    end
  end
RUBY

  private :require

end
```

kddnewton:
Sorry about that @enebo! 

Fixes #646

### DIFF
--- a/src/util/yp_buffer.c
+++ b/src/util/yp_buffer.c
@@ -19,10 +19,15 @@ yp_buffer_init(yp_buffer_t *buffer) {
 // Append a generic pointer to memory to the buffer.
 static inline void
 yp_buffer_append(yp_buffer_t *buffer, const void *source, size_t length) {
-  if (buffer->length + length > buffer->capacity) {
-    buffer->capacity = buffer->capacity * 2;
+  size_t next_length = buffer->length + length;
+
+  if (next_length > buffer->capacity) {
+    while (next_length > buffer->capacity) {
+      buffer->capacity = buffer->capacity * 2;
+    }
     buffer->value = realloc(buffer->value, buffer->capacity);
   }
+
   memcpy(buffer->value + buffer->length, source, length);
   buffer->length += length;
 }


enebo:
@kddnewton it's no problem at all.    Very quick turnaround too.

I think you will need to fix this in two more places: yp_string_list_append and yp_buffer_append assuming you can get a length greater than 2*capacity in those cases.

kddnewton:
I think we're okay in those places because in those places we're just adding one at a time.

PR #658, opened 2023-03-24
Change binding powers for rescue modifier

### LINKED ISSUE #633: Binding power on modifier rescue
```ruby
foo do |x|
  bar(y) rescue ArgumentError fail "baz"
end
```

haldun:
Fixes #633

Although I am not 100% sure about this change, I haven't seen any regression in the targets.


kddnewton:
Yeah, after some testing this seems right.

PR #662, opened 2023-03-25
Compile with -Wpedantic

kddnewton:


Makefile
```diff
@@ -4,8 +4,25 @@ else
...
+
+# Check for the presence of strncasecmp
+ifeq ($(shell echo '\#include <string.h>\nint main() { strncasecmp("", "", 0); }' | $(CC) -o /dev/null -x c - 2>/dev/null && echo 1), 1)
+CFLAGS := $(CFLAGS) -DHAVE_STRNCASECMP
+endif
+
+# Check for the presence of strnstr
+ifeq ($(shell echo '\#include <string.h>\nint main() { strnstr("", "", 0); }' | $(CC) -o /dev/null -x c - 2>/dev/null && echo 1), 1)
+CFLAGS := $(CFLAGS) -DHAVE_STRNSTR
+endif
```
eregon, line 20: This doesn't seem to work, at least for me on Linux, `strnlen` is definitely there but on `bundle exec rake` the `-DHAVE_STRNLEN` is not shown.

eregon: BTW, isn't strnlen() POSIX.1-2008 and so available everywhere?

eregon: The problem seems differences in how `echo` behaves, I get:
```
<stdin>:1:20: warning: extra tokens at end of #include directive
/usr/bin/ld: /usr/lib/gcc/x86_64-redhat-linux/12/../../../../lib64/crt1.o: in function `_start':
(.text+0x1b): undefined reference to `main'
collect2: error: ld returned 1 exit status
```
with:
```diff
diff --git a/Makefile b/Makefile
index 2b066cf..ab7d524 100644
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,9 @@ endif
 CFLAGS := -O3
 
 # Check for the presence of strnlen
-ifeq ($(shell echo '\#include <string.h>\nint main() { strnlen("", 0); }' | $(CC) -o /dev/null -x c - 2>/dev/null && echo 1), 1)
+ifeq ($(shell echo '#include <string.h>\nint main() { strnlen("", 0); }' | $(CC) -o /dev/null -x c - && echo 1), 1)
 CFLAGS := $(CFLAGS) -DHAVE_STRNLEN
+STRNLEN := 1
 endif
 
 # Check for the presence of strncasecmp
@@ -24,7 +25,7 @@ endif
 all: build/librubyparser.$(SOEXT)
 
 build/librubyparser.$(SOEXT): $(shell find src -name '*.c') $(shell find src -name '*.h') Makefile build include/yarp/ast.h
-	$(CC) $(CFLAGS) $(DEBUG_FLAGS) -std=c99 -Wall -Werror -Wextra -Wpedantic -Wsign-conversion -fPIC -g -fvisibility=hidden -shared -Iinclude -o $@ $(shell find src -name '*.c')
+	$(CC) $(CFLAGS) $(STRNLEN) $(DEBUG_FLAGS) -std=c99 -Wall -Werror -Wextra -Wpedantic -Wsign-conversion -fPIC -g -fvisibility=hidden -shared -Iinclude -o $@ $(shell find src -name '*.c')
 
 build:
 	mkdir -p build
```
A solution to that would be to move these small test programs to files in the repo.

eregon:
Another issue with this approach is none of these flags will be set if this Makefile isn't used.
That's the case for TruffleRuby where we copy .h and .c files but let ninja build it.
And it would also be the case for the C extension if that statically included YARP (but currently it dynamically links to it so no problem).

PR #669, opened 2023-03-27
Improve parameter order

HParker:
This resurrects https://github.com/Shopify/yarp/pull/227 (thanks @ccocchi ! I took the liberty of co-authoring you)

and in the second commit uses this newly tracked state to split required params which are after the optional arguments into their own list.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -2156,6 +2156,7 @@ yp_parameters_node_create(yp_parser_t *parser) {
 
   yp_node_list_init(&node->requireds);
   yp_node_list_init(&node->optionals);
+  yp_node_list_init(&node->posts);
   yp_node_list_init(&node->keywords);
 
   return node;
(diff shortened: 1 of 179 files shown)


HParker:
Thank you memcheck!

```
def foo(..., ...)
end
```

was setting the `rest` node twice and therefore leaking the original `rest` node. I don't think I would have found that without valgrind.

src/yarp.c
```diff
@@ -2180,8 +2181,8 @@ yp_parameters_node_location_set(yp_parameters_node_t *params, yp_node_t *param)
 // Append a required parameter to a ParametersNode node.
 static void
 yp_parameters_node_requireds_append(yp_parameters_node_t *params, yp_node_t *param) {
-  yp_parameters_node_location_set(params, param);
-  yp_node_list_append2(&params->requireds, param);
+  yp_parameters_node_location_set(params, (yp_node_t *) param);
```
kddnewton, line 2184: It looks like you shouldn't need to cast this since you already have a `yp_node_t *` right?

src/yarp.c
```diff
@@ -2201,8 +2209,8 @@ yp_parameters_node_rest_set(yp_parameters_node_t *params, yp_rest_parameter_node
 // Append a keyword parameter to a ParametersNode node.
 static void
 yp_parameters_node_keywords_append(yp_parameters_node_t *params, yp_node_t *param) {
-  yp_parameters_node_location_set(params, param);
-  yp_node_list_append2(&params->keywords, param);
+  yp_parameters_node_location_set(params, (yp_node_t *) param);
```
kddnewton, line 2212: And here

kddnewton (changes requested):
Looking really good! Just a couple of small things.

PR #671, opened 2023-03-27
Add the filepath to SourceFileNode

jemmaissroff:
This commit adds the filepath to the SourceFileNode by first storing the filepath string on the parser, and then adding that string to any constructed SourceFileNodes


config.yml
```diff
@@ -1508,6 +1508,11 @@ nodes:
           __ENCODING__
           ^^^^^^^^^^^^
   - name: SourceFileNode
+    child_nodes:
+      - name: file_keyword
```
kddnewton, line 1512: I don't think we should add a token here, since it's going to always be the same type and have the same location as the rest of the node.

ext/yarp/extension.c
```diff
@@ -68,9 +68,9 @@ source_file_unload(source_t *source) {
 
 // Dump the AST corresponding to the given source to a string.
 static VALUE
-dump_source(source_t *source) {
+dump_source(source_t *source, char *filepath) {
```
kddnewton, line 71: Can you make this a `const char *`?

ext/yarp/extension.h
```diff
@@ -3,6 +3,7 @@
 
 #include <ruby.h>
 #include <ruby/encoding.h>
+#include <ruby/internal/special_consts.h>
```
kddnewton, line 6: What's this include do?

kddnewton (changes requested):

ext/yarp/extension.h
jemmaissroff: It allows us to use RTEST to check if filepath is nil [here](https://github.com/Shopify/yarp/blob/f2345416942939637dafd2fcb8bf444f840219d6/ext/yarp/extension.c#L96-L99)

Issue #465, opened 2023-03-12
Nodes are too big


kddnewton:
At the moment we've got everything in a big union. This means the largest node keeps all of the nodes large. 

CRuby solves this by having 3 unions in the node that each node knows how to use properly. We could also potentially just keep an array of child nodes around that we pre-allocate when they are first created.

The `bin/memsize` script currently will tell you how big each node is.

kddnewton:
Yeah, we actually do this with our comment/error lists and I think it would be a really good way to go. Once https://github.com/Shopify/yarp/pull/656 is merged I think this would be a really good next step.

kddnewton:
In your examples though I would name them `yp_node_t` and `yp_alias_node_t`.

PR #674, opened 2023-03-28
Migrate `AliasNode`

haldun:
Migrates `AliasNode` away from the union inside `yp_node_t`.

Related #465 

The approach here is the struct composition where the specific structs contain the base `yp_node_t` as their first element.

Please note that we still have the `union` inside `yp_node_t` till we migrate all the node types.

To migrate a node type, I decided to put an indicator field `is_migrated` to the config.yml.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -255,7 +255,7 @@ yp_node_create_from_token(yp_parser_t *parser, yp_node_type_t type, const yp_tok
 }
 
 // Allocate and initialize a new alias node.
-static yp_node_t *
+static yp_alias_node_t *
 yp_alias_node_create(yp_parser_t *parser, const yp_token_t *keyword, yp_node_t *new_name, yp_node_t *old_name) {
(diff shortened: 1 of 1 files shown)


src/yarp.c
```diff
@@ -249,22 +258,22 @@ yp_node_create_from_token(yp_parser_t *parser, yp_node_type_t type, const yp_tok
 static yp_node_t *
```
kddnewton, line 258: I think we decided to return `yp_alias_node_t *` here?

kddnewton (changes requested):

src/yarp.c
kddnewton: The reason I'm thinking is there will be lots of places where we create (for example) an arguments node, and then pass that into various functions that append arguments. So you'll get something back for an arguments node that is an `yp_node_t *` and then you'll have to cast it every time it gets passed into one of those functions. Can we try it for the first couple and if it gets really annoying we can change it back?

Issue #675, opened 2023-03-28
Tracking: Migrate node types


haldun:
Related #465

- [x] AliasNode #674
- [x] AlternationPatternNode #676
- [x] AndNode #677
- [x] ArgumentsNode #678
- [x] ArrayNode #681
- [x] ArrayPatternNode #707
- [x] AsPatternNode #711
- [x] AssocNode #708
- [x] AssocSplatNode #712
- [x] BeginNode #686
- [x] BlockArgumentNode #695
- [x] BlockNode #695
- [x] BlockParameterNode #695
- [x] BlockParametersNode #695
- [x] BreakNode #689
- [x] CallNode #710
- [x] CaseNode #690
- [x] ClassNode #684
- [x] ClassVariableWriteNode #725
- [x] ConstantPathNode #696
- [x] ConstantPathWriteNode #726
- [x] DefNode #683
- [x] DefinedNode #691
- [x] ElseNode #688
- [x] EnsureNode #692 
- [x] FindPatternNode #713
- [x] ForNode #697
- [x] ForwardingSuperNode #697
- [x] GlobalVariableReadNode #714
- [x] GlobalVariableWriteNode #714
- [x] HashNode #699
- [x] HashPatternNode #715
- [x] HeredocNode #716
- [x] IfNode #709
- [x] InNode #717
- [x] InstanceVariableWriteNode #718
- [x] InterpolatedRegularExpressionNode #720
- [x] InterpolatedStringNode #720
- [x] InterpolatedSymbolNode #720
- [x] InterpolatedXStringNode #720
- [x] KeywordParameterNode #719
- [x] KeywordRestParameterNode #719
- [x] LambdaNode #719
- [x] LocalVariableReadNode #704
- [x] LocalVariableWriteNode #704
- [x] MatchPredicateNode #704
- [x] MatchRequiredNode #704
- [x] ModuleNode #704
- [x] MultiWriteNode #704
- [x] NextNode #704
- [x] NoKeywordsParameterNode #704
- [x] OperatorAndAssignmentNode #702
- [x] OperatorAssignmentNode #702
- [x] OperatorOrAssignmentNode #702
- [x] OptionalParameterNode #702
- [x] OrNode #702
- [x] ParametersNode #702
- [x] ParenthesesNode #702
- [x] PinnedExpressionNode #702
- [x] PinnedVariableNode #702
- [x] PostExecutionNode #702
- [x] PreExecutionNode #702
- [x] ProgramNode #702
[...]

PR #678, opened 2023-03-28
Migrate arguments node

haldun:
Related #675

@kddnewton I think this node makes a good example of the effect of returning a specific node type from the `create` functions. I don't have a strong opinion but it looks like we're gonna need more casts.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -189,7 +189,7 @@ not_provided(yp_parser_t *parser) {
 // of the call node creation functions.
 typedef struct {
   yp_token_t opening;
-  yp_node_t *arguments;
+  yp_arguments_node_t *arguments;
   yp_token_t closing;
   yp_node_t *block;
 } yp_arguments_t;
@@ -815,7 +815,7 @@ yp_call_node_binary_create(yp_parser_t *parser, yp_node_t *receiver, yp_token_t 
 
   yp_arguments_node_t *arguments = yp_arguments_node_create(parser);
   yp_arguments_node_arguments_append(arguments, argument);
-  node->as.call_node.arguments = YP_NODE_DOWNCAST(arguments);
+  node->as.call_node.arguments = arguments;
 
   yp_string_shared_init(&node->as.call_node.name, operator->start, operator->end);
   return node;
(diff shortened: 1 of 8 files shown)


src/yarp.c
```diff
@@ -9267,7 +9267,7 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
         yp_binding_power_t binding_power = yp_binding_powers[parser->current.type].left;
 
         if (binding_power == YP_BINDING_POWER_UNSET || binding_power >= YP_BINDING_POWER_RANGE) {
-          arguments = yp_arguments_node_create(parser);
+          arguments = YP_NODE_DOWNCAST(yp_arguments_node_create(parser));
```
kddnewton, line 9270: Can you change the signature of `parse_arguments` to instead accept a `yp_arguments_node_t *`?

src/yarp.c
```diff
@@ -6584,10 +6584,10 @@ parse_target(yp_parser_t *parser, yp_node_t *target, yp_token_t *operator, yp_no
 
         if (value != NULL) {
           if (target->as.call_node.arguments == NULL) {
-            target->as.call_node.arguments = yp_arguments_node_create(parser);
+            target->as.call_node.arguments = YP_NODE_DOWNCAST(yp_arguments_node_create(parser));
```
kddnewton, line 6587: Could you create a way in `config.yml` to specify the type? Right now `arguments` on CallNode is

```yaml
      - name: arguments
        type: node?
```

what do you think of

```yaml
      - name: arguments
        type: node?
        kind: ArgumentsNode
```

In which case it would template out a `yp_arguments_node_t *` on the AST.

src/yarp.c
```diff
@@ -9274,11 +9274,11 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
 
       switch (keyword.type) {
         case YP_TOKEN_KEYWORD_BREAK:
-          return yp_break_node_create(parser, &keyword, arguments);
+          return yp_break_node_create(parser, &keyword, YP_NODE_DOWNCAST(arguments));
```
kddnewton, line 9277: Looks like these 3 nodes can have the `kind` field as well.

kddnewton (changes requested):

Issue #659, opened 2023-03-24
Looking into making sure assertions are no imposing a penalty


kddnewton:

jemmaissroff:
@kddnewton when do we _not_ want to be running assertions?

PR #700, opened 2023-03-28
Fix assertions which were doing operations

jemmaissroff:
When looking into #659, I found that if we disable assertions, we get warnings in two places. This fixes those warnings.


kddnewton (approved):
This looks good. Could you add a comment that the assertion for the regexp parsing is to indicate whether or not we could successfully parse the regexp?

PR #703, opened 2023-03-28
Add all-no-debug to Makefile

### LINKED ISSUE #659: Looking into making sure assertions are no imposing a penalty

jemmaissroff:
Add a non-debugging build to the makefile that runs without assertions. Tested by adding an `assert(false)` and ensuring it raised with `make all` but not with `make all-no-debug`

Closes #659


Makefile
```diff
@@ -41,3 +41,6 @@ clean:
 		src/util/yp_strspn.c
 
 .PHONY: clean
+
+all-no-debug: DEBUG_FLAGS := -DNDEBUG=1
```
jemmaissroff, line 45: I am definitely not super familiar with Makefiles - let me know if there's any odd syntax here....

PR #706, opened 2023-03-29
Update .gitignore

haldun:
This PR removes the exclusion of the generated source code files from the git repository. The rationale is to enable the project to be built without any Ruby setup in the developer's machine.


kddnewton:
So I think the idea is good of making it so it can be built on a machine without Ruby, but I don't want to go this way with it. We gitignored those files because it was really hard to review pull requests with so much noise any time one of those files changed.

I think another way to go would be to release the shared library itself as a package on the various native package managers (e.g., apt-get) and also maybe as a homebrew tap. That way folks could get the library with everything already templated but wouldn't have to have Ruby.

PR #731, opened 2023-03-29
Allow bundling source as you would need for a gem

kddnewton:


ext/yarp/include
```diff
@@ -0,0 +1 @@
+../../include
```
eregon, line 1: Is this symlink and the below necessary?
It causes several editors (Atom, VSCode) to show duplicate matches for global search.

PR #756, opened 2023-03-31
Typed data

### LINKED ISSUE #755: Extension speed: used typed data, not copying

kddnewton:
This PR starts using the typed data API instead of the usage of `rb_class_new_instance` function calls we have been using. It's a work-in-progress because it currently doesn't support all of the parameter types. 

Closes #755.


eregon:
> I also thought back then for Ruby nodes we could just always use dump and load

More discussion about that in https://github.com/ruby/yarp/discussions/836

kddnewton:
Closing this pending further discussion in the future.

PR #769, opened 2023-04-05
Fix trailing embdoc

### LINKED ISSUE #759: embdoc does not work properly with end of file
when we have an `=end` at the end of a file but no trailing newline the parser generates an error `"Unterminated embdoc"`.
This path *is* exercised in the tests, but there is no assertion for it so it does not trigger a test failure, but I think it should.

I.e. if I apply this patch

```
diff --git a/test/comments_test.rb b/test/comments_test.rb
index 961e046..9ebabf1 100644
--- a/test/comments_test.rb
+++ b/test/comments_test.rb
@@ -40,6 +40,8 @@ class CommentsTest < Test::Unit::TestCase
   private

   def assert_comment(source, type)
-    YARP.parse_dup(source) => YARP::ParseResult[comments: [YARP::Comment[type: type]]]
+    result = YARP.parse_dup(source)
+    assert result.errors.empty?, result.errors.map(&:message).join("\n")
+    result => YARP::ParseResult[comments: [YARP::Comment[type: type]]]
   end
 end
```

then `rake test` has two errors. 

Related, I think `bin/lex` should also output errors, so it'd behave like `rake lex`, otherwise we have a weird situation

```
rake lex FILEPATHS=foo.rb # ark, an error, bad yarp!
bin/lex foo.rb  # I mean, all is good, sorry!
```

jemmaissroff:
Asserted that the changes to comment_test made the existing test fail (since it threw an error) before the fix.

Closes #759

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -4502,12 +4502,13 @@ lex_embdoc(yp_parser_t *parser) {
 
   // Now, loop until we find the end of the embedded documentation or the end of
   // the file.
-  while (parser->current.end + 5 < parser->end) {
+  while (parser->current.end + 4 < parser->end) {
     parser->current.start = parser->current.end;
 
     // If we've hit the end of the embedded documentation then we'll return that
     // token here.
-    if (strncmp(parser->current.end, "=end", 4) == 0 && yp_char_is_whitespace(parser->current.end[4])) {
+    if (strncmp(parser->current.end, "=end", 4) == 0 &&
+	(yp_char_is_whitespace(parser->current.end[4]) || parser->current.end + 4 == parser->end)) {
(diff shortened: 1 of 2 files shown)


src/yarp.c
```diff
@@ -4502,12 +4502,13 @@ lex_embdoc(yp_parser_t *parser) {
 
   // Now, loop until we find the end of the embedded documentation or the end of
   // the file.
-  while (parser->current.end + 5 < parser->end) {
+  while (parser->current.end + 5 <= parser->end) {
```
HParker, line 4505: Should we do,

```
parser->current.end + 4 < parser->end
```

to match the check on 4510?

Not blocking.

HParker (approved):

PR #770, opened 2023-04-05
Add error output to bin/lex

jemmaissroff:
Previously, if bin/lex had errors, it wouldn't be visible to the user. This prints errors before printing the bin/lex comparison

As suggested in #759 

New output comes before the usual comparison and looks like this:

```
Errors lexing:                       
- Unterminated embdoc                                                                                                                                                                                                                                       
```


HParker (approved):
nice!

PR #773, opened 2023-04-06
try adding ASCII fast path to unicode parsing

HParker:
Using `a.rb` as the benchmark in linked issue.

I removed the part of `parse` that creates the Ruby tree nodes and set `-O3`

benchmark:
```
#!/usr/bin/env ruby
# frozen_string_literal: true
require 'benchmark/ips'
$:.unshift(File.expand_path("lib", __dir__))
require "yarp"

Benchmark.ips do |x|
  # Configure the number of seconds used during
  # the warmup phase (default 2) and calculation phase (default 5)
  x.config(:time => 5, :warmup => 2)

  # Typical mode, runs the block as many times as it can
  x.report("parse_file_dup") {
    YARP.parse_file_dup("a.rb")
  }
end
```

Before:

```
      parse_file_dup   133.000  i/100ms
      parse_file_dup      1.326k (± 1.8%) i/s -      6.650k in   5.017386s
```

After:

```
      parse_file_dup   134.000  i/100ms
      parse_file_dup      1.336k (± 1.5%) i/s -      6.700k in   5.018107s
```

`1336 / 1326 = 1.036`

I am not sure this is outside of the margin of error, but it seems clear this should be better.


HParker:
@enebo I wonder if something in your test is making the performance difference bigger. I don't see a large performance difference changing the encoding comment in `a.rb` on main without these changes.

I still think this change is probably worth doing, but I am not sure I am measuring it super effectively.

enebo:
@HParker Are you building with optimizations enabled?  YARP in my original run would take 1 minute vs 20s originally.  I suspect that will make a big difference (I am using -O3).

enebo:
I used your benchmark and I am not profiling but I am timing it:

HEAD:
```text
mri31 -Ilib bench.rb 
      parse_file_dup     5.000  i/100ms
parse_file_dup     57.896  (± 1.7%) i/s -    290.000  in   5.011949s
```

HEAD(-O3):
```text
      parse_file_dup    12.000  i/100ms
      parse_file_dup    121.190  (± 1.7%) i/s -    612.000  in   5.051568s
```

Your PR + -O3:

```text
      parse_file_dup    22.000  i/100ms
      parse_file_dup    235.896  (± 7.2%) i/s -      1.188k in   5.066008s
```

enebo:
I just tried clang-15.0.7 (which is shipped in Fedora Core):

HEAD(-O3):
```text
      parse_file_dup    11.000  i/100ms
      parse_file_dup    113.936  (± 8.8%) i/s -    572.000  in   5.064028s
```

Your PR + -O3:
```text
      parse_file_dup    23.000  i/100ms
      parse_file_dup    248.188  (± 2.0%) i/s -      1.242k in   5.006048s
```

jemmaissroff:
Convinced by @enebo's benchmarks!

enebo:
@jemmaissroff This is embarassing :smile: ... I have two separate YARP repositories: one for JRuby and one for CRuby.  I applied @HParker fix to C Ruby one and it had not been updated to include #748.  I would have sworn up and down the JRuby checkout was up to date but I now doubt that since these values are so much faster without this PR.  Without this PR:

```text
      parse_file_dup     3.073k i/100ms
      parse_file_dup     32.492k (± 2.5%) i/s -    162.869k in   5.016019s
```

With the PR:

```text
      parse_file_dup     3.210k i/100ms
      parse_file_dup     31.464k (± 4.1%) i/s -    157.290k in   5.008593s
```

So this should get reverted.   I am still at a total loss on how I saw this with JRuby on my branch since I am frequently merging.  I wonder if I messed up a merge.

HParker:
Ok, that is about what I was seeing. The change _might_ be faster or slower, but it is in the margin of error.

HParker:
Yeah, you might have been measuring

https://github.com/Shopify/yarp/pull/746

and

https://github.com/Shopify/yarp/commit/938af7af831ec228a01d891364a0a56232f75a21

which both landed after we did the first `perf` pass on the parser.

enebo:
@HParker @jemmaissroff yeah sorry about this.  I am still trying to understand how my fully merged branch in my fork ended up not seeing those changes. :frowning_face:

eregon:
Right, looking at the diff, there is basically the same 7-bit check in `utf_8_codepoint()` so perf should be similar either way (unless $CC does not optimize this well).

PR #777, opened 2023-04-07
Add nested numeric nodes for different base node types

HParker:
closes: https://github.com/Shopify/yarp/issues/768

now you can convert,

```
1.2ri
```

into,

```
ImaginaryNode(0...5)(RationalNode(0...5)(FloatNode(0...4)()
```

I tested all the combinations of rational and imaginary modifiers I could think of. Let me know if I missed any.


enebo:
0x is a missed prefix but I am not sure it matters so long as IntegerNode is tested against those values.

src/yarp.c
```diff
@@ -1548,13 +1658,88 @@ yp_if_node_ternary_create(yp_parser_t *parser, yp_node_t *predicate, const yp_to
...
+    case YP_TOKEN_INTEGER: {
+      numeric_node = (yp_node_t *)yp_integer_node_create(parser, &numeric_token);
+      break;
+    }
+    case YP_TOKEN_FLOAT: {
+      numeric_node = (yp_node_t *)yp_float_node_create(parser, &numeric_token);
+      break;
+    }
+    default: {
+      assert(false && "unreachable rational base");
```
jemmaissroff, line 1692: I _think_ you'll still have to assign a value to `numeric_node` in the default case so it doesn't complain when assertions are turned off that there's no value in this case.

src/yarp.c
```diff
@@ -3994,8 +4050,35 @@ lex_numeric(yp_parser_t *parser) {
...
+      lex_mode_push(parser, (yp_lex_mode_t) {
+          .mode = YP_LEX_NUMERIC,
+          .as.numeric.type = current,
+          .as.numeric.start = parser->current.start,
+          .as.numeric.end = parser->current.end
+        });
+      type = YP_TOKEN_RATIONAL_NUMBER;
+    }
+    if (match(parser, 'i')) {
+
```
jemmaissroff, line 4063: Can we remove the extra space here?

jemmaissroff (changes requested):
And yes, I think `0x` is the only missed prefix. Fwiw, I asked chatgpt for a sample file here and got:

```
1i

3.4i

-5i

0b101i

0o17i

0x1Ai

1ri

3.4ri

-5ri

0b101ri

0o17ri

0x1Ari
```

jemmaissroff (approved):

PR #787, opened 2023-04-07
Allow generic option collection in templating parameters

enebo:
This is a transitional PR which allows extra attributes of parameters from config.yml to get saved as 'options' to the parameter we interact with when generating code via ERB.

Why? In my current branch I have marked syntactical tokens that execution is uninterested in serializing (and then loading into an AST) as 'exclude: execute'.  My templates for serialize.c, Nodes.java, and Loader.java then will skip those elements altogether. I can think of other useful attributes we can put in to help with generation.

### DIFF
--- a/bin/template.rb
+++ b/bin/template.rb
@@ -4,9 +4,30 @@ require "erb"
 require "fileutils"
 require "yaml"
 
+class Param
+  attr_reader :name, :options
+  
+  def initialize(name:, type:, **options)
+    @name, @type, @options = name, type, options
+  end
+end
+
+module CTypes
+  def c_type
+    if options[:kind]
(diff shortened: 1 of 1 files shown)


enebo:
I have signed CLA after I saw that fail.

enebo:
I removed \#param from all the param types because nothing seems to call it.

eregon:
This sounds good to me. Could you also include the changes to the templates in this PR? I think that makes it easier to review and see how it's used.

enebo:
@eregon What I have on my playground branch is quite a bit different but I will add this to 'java' templates (since i think we both want this in some form) so you can see how it is used.

enebo:
@eregon I will add another PR for the exclude stuff to java templates so you can see this and hopefully also want the same set of excludes.

PR #793, opened 2023-04-08
Fix __ENCODING__ symbols

HParker:
This was missing from the KEYWORD list.
I also added some tests for the other double underscore methods

closes: https://github.com/Shopify/yarp/issues/791

Looks like Ripper treats these as @kw, not ident, so the structure seems good:

```
Ripper.sexp("x = :__FILE__") => 
[:program,                                                                               
 [[:assign,                                                                              
   [:var_field, [:@ident, "x", [1, 0]]],                                                 
   [:symbol_literal, [:symbol, [:@kw, "__FILE__", [1, 5]]]]]]] 
```


src/yarp.c
```diff
@@ -8037,7 +8037,8 @@ parse_conditional(yp_parser_t *parser, yp_context_t context) {
   case YP_TOKEN_KEYWORD_NOT: case YP_TOKEN_KEYWORD_OR: case YP_TOKEN_KEYWORD_REDO: case YP_TOKEN_KEYWORD_RESCUE: \
   case YP_TOKEN_KEYWORD_RETRY: case YP_TOKEN_KEYWORD_RETURN: case YP_TOKEN_KEYWORD_SELF: case YP_TOKEN_KEYWORD_SUPER: \
   case YP_TOKEN_KEYWORD_THEN: case YP_TOKEN_KEYWORD_TRUE: case YP_TOKEN_KEYWORD_UNDEF: case YP_TOKEN_KEYWORD_UNLESS: \
-  case YP_TOKEN_KEYWORD_UNTIL: case YP_TOKEN_KEYWORD_WHEN: case YP_TOKEN_KEYWORD_WHILE: case YP_TOKEN_KEYWORD_YIELD
+  case YP_TOKEN_KEYWORD_UNTIL: case YP_TOKEN_KEYWORD_WHEN: case YP_TOKEN_KEYWORD_WHILE: case YP_TOKEN_KEYWORD_YIELD: \
+  case YP_TOKEN_KEYWORD___ENCODING__
```
jemmaissroff, line 8041: Nit: Can we keep `YP_TOKEN_KEYWORD___ENCODING__ ` at the beginning of this list with `YP_TOKEN_KEYWORD___LINE__` and `YP_TOKEN_KEYWORD___FILE__`? And maybe actually alphabetize those two while we're in here so that they fit with the rest of the alphabetization?

jemmaissroff (changes requested):
Thanks for writing this! Just a small nit

jemmaissroff (approved):
LGTM, thanks!

PR #795, opened 2023-04-10
Fix range default on argument defaults

HParker:
before this change ... range default in arguments got the ENDARG state as if they where forwarding params. Now we can tell the differenced based on if the previous state was END after parsing a number


jemmaissroff (approved):
Thanks for fixing this case! Made me realize we're currently also lexing incorrectly for open ended ranges in method definitions, so I made #797

PR #796, opened 2023-04-10
Fix small embdoc off at end

HParker:
There is an embdoc off by one error which causes =end with no newline to not be returned in the lex. This fixes that error and adds a test


HParker:
Valgrind is so good! I think I need to check for end of file a little sooner to pass the memory test.

PR #799, opened 2023-04-10
Fix regex lexing comparison on the Ruby layer

jemmaissroff:
Ripper splits regex lexing into multiple content strings on `#$` and `#@`. Prior to this change, we were implicily assuming that there would only be one split per regex. This change addresses the fact that there could be multiple.


HParker (approved):
Nice! I didn't know Ripper splits like that. I wonder why.

PR #801, opened 2023-04-10
Remove unused length field from serialization.

enebo:
This appears to be skipped over and it removes 4 bytes per Node from the serialized blob.

### DIFF
--- a/bin/templates/java/org/yarp/Loader.java.erb
+++ b/bin/templates/java/org/yarp/Loader.java.erb
@@ -102,7 +102,6 @@ public class Loader {
 
     private Nodes.Node loadNode() {
         int type = buffer.get() & 0xFF;
-        int length = buffer.getInt();
         int startOffset = buffer.getInt();
         int endOffset = buffer.getInt();
 
--- a/bin/templates/lib/yarp/serialize.rb.erb
+++ b/bin/templates/lib/yarp/serialize.rb.erb
@@ -71,7 +71,7 @@ module YARP
       end
 
       def load_node
-        type, _length = io.read(5).unpack("CL")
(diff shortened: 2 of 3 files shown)


eregon:
I think we should keep it for DefNode though, for lazy parsing of methods purposes.

It could be just a regular `type: integer` field on DefNode, but serialize.c needs to fill it correctly (nothing else knows how much bytes that node consumed in serialized form), so maybe another type is better?

eregon:
We could merge this and add that back later.
OTOH it may be best to do it at the same time so there is no need to find again how to compute that length in serialized form, which is code being removed in this PR.
So I think better to address it together in this PR.

enebo:
@eregon This is the wrong approach: https://gist.github.com/enebo/00fc2add15397f69375fc96b6dfd5bb1

This will bubble through all the things and I do not think anyone but us want this field.  I am not sure what path to take.   Do you have an idea?

I will say I won't forget this was removed and I can look it up later in repo history (meaning if we land this simple PR I don't think we will have a hard time making a second one).  Although, it would be nice how to include items that we want but others don't.

eregon:
@enebo Could you push that approach as an extra commit and then we ask for @jemmaissroff's opinion?

enebo:
@eregon I quickly tried that and ran into a bunch of issues.  I had to transfer off this to do some JRuby stuff but I can probably make another quick attempt in a day or two.

jemmaissroff (approved):
LGTM, let me know if you're both comfortable merging now, and I'm happy to merge!

eregon:
Sorry for late reply, I only noticed this when I was at RubyKaigi. It looks great.

PR #804, opened 2023-04-10
Ability to exclude elements and verify Java files can compile

enebo:
The main template feature is the ability to exclude fields from serialization
and loading via (ruby_exclude: true, java_exclude: true).   I believe both
TruffleRuby and JRuby will both probably exclude the same fields so I am
starting with a simpler solution.  There is currently nothing using
ruby_exclude but it felt like a natural dual to java_exclude.

Secondly, I wanted to be able to test the generated files by compiling them right after generation so I had to update the Rakefile to be JRuby-aware.  As written this will not work from TruffleRuby to compile the Java files but, at this point, TR is pulling this source into their project.  We still have some future discussions on how we plan on packaging things.  Nonetheless, this will easily flag syntax problems with generation changes.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,20 +3,29 @@
 require "rake/extensiontask"
 require "rake/testtask"
 require "rake/clean"
-require "ruby_memcheck"
 
 Rake.add_rakelib("tasks")
-
-RubyMemcheck.config(binary_name: "yarp")
 
 task compile: :make
 task compile_no_debug: :make_no_debug
 
-Rake::ExtensionTask.new(:compile) do |ext|
-  ext.name = "yarp"
-  ext.ext_dir = "ext/yarp"
-  ext.lib_dir = "lib/yarp"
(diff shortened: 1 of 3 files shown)


enebo:
@jemmaissroff I can do that but it will lead to me having to replicate the entire file's structure and then merge two similar trees into NodeParams/Nodes.  When node design changes then my file will be out of sync and no one will consider it broken because that is JRuby's file (this is more of a near term issue).

In the bigger picture I plan on doing WAY more since we are passing so much unused info that execution never uses perhaps your suggestion is for the best (if ignoring extra fields is not desirable).  Part of the issue is config.yml is defined for being capable of full analysis and rewriting of Ruby syntax but so much of that is not needed for execution.

eregon:
Related: https://github.com/Shopify/yarp/issues/807

enebo:
This is replaced by whatever addresses #807.

PR #806, opened 2023-04-11
Fix end of capture groups being fooled by escaped close parens

HParker:


src/regexp.c
```diff
@@ -1,5 +1,5 @@
 #include "yarp/regexp.h"
-
+#include <stdio.h>
```
jemmaissroff, line 2: What do we need this for?

src/regexp.c
```diff
@@ -454,8 +460,10 @@ yp_regexp_parse_item(yp_regexp_parser_t *parser) {
         parser->cursor++;
       }
       return yp_regexp_parse_quantifier(parser);
-    case '(':
-      return yp_regexp_parse_group(parser) && yp_regexp_parse_quantifier(parser);
+  case '(': {
```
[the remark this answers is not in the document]
HParker: blah, thanks! This was also a debugging change. This is what I get for forgetting to use `git add -p` 🤦

src/regexp.c
```diff
@@ -342,8 +342,14 @@ yp_regexp_parse_group(yp_regexp_parser_t *parser) {
     yp_regexp_options_init(&options);
 
     switch (*parser->cursor) {
-      case '#': // inline comments
-        return yp_regexp_char_find(parser, ')');
+      case '#': { // inline comments
+	bool found = yp_regexp_char_find(parser, ')');
+	// the close paren we found is escaped, we need to find another
+	while (parser->start <= parser->cursor - 2 && *(parser->cursor - 2) == '\\') {
```
nobu, line 348: The RHS expression isn't considering encodings that `\\` can be a trailing byte, e.g. Shift_JIS.

PR #816, opened 2023-04-12
Lex against top 100 gems as part of CI

### LINKED ISSUE #788: Lex against top 100 gems as part of CI
Currently taking ~1 minute to find gems, download them, and lex against their Ruby files locally. We should be able to add this to CI.

[Here's](https://gist.github.com/jemmaissroff/f0dbee4ff8e05bb20823257a05b4d8e1) a script that does it locally, borrowing heavily from `LexTask` and [Kevin's gist](https://gist.github.com/kddnewton/aa8ff90c9383caa5d3b7d80e2f3d1f4a). We should productionize this and add it to CI. Also, consider using the RubyGems API instead of scraping the website

jemmaissroff:
This commit introduces the `lex:topgems` rake task, and adds it to CI. The rake task lexes against the top 100 gems, and compares their failures to the known failures in `tasks/top-100-gems.yml`. If they are not equivalent, the rake task will fail.

This will ensure we have no regressions when lexing the top 100 gems

Closes #788

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,6 +69,18 @@ jobs:
     - name: Lex discourse/discourse
       run: bundle exec rake lex:discourse
 
+  lex-top-100:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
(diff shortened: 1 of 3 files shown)


HParker:
We can probably add the script, but also don't know if the top 100 changes enough that it will be a big deal.

PR #817, opened 2023-04-12
Fix places where dos newlines create the wrong lex

HParker:
This fixes some places where we assume that newlines are always unix newlines. Instead now we can handle newlines that are dos style as well.

I also removed a few tabs that probably shouldn't be there.


jemmaissroff (approved):
Looks great, thanks!

jemmaissroff (changes requested):
Sorry, just realized, can we also remove `seattlerb/heredoc__backslash_dos_format.rb` from `known_failures` in `test/parse_test.rb`?

jemmaissroff (approved):

PR #818, opened 2023-04-12
Add new context within default parameters, fix open range arg lexing

### LINKED ISSUE #797: Lex wrong state for open ended ranges in method definitions
Currently failing:

```
def bar(a: (...10))
end

def bar(a: (1...))
end
```

jemmaissroff:
This commit adds a `YP_CONTEXT_DEFAULT_PARAMS` context which allows us to accurately differentiate between lexing `(...)` params and lexing ranges as default params.

Closes #797


HParker (approved):
👏 Significant improvement giving it its own context.

PR #830, opened 2023-04-13
improve encoding comment support

HParker:
this adds support for cp125{1,2} and extended utf-8 formats


jemmaissroff:
> my corpus has `koi8-r, gbk, sjis`.

Can you add issues for these?

PR #840, opened 2023-04-14
Allow octothorpe list terminators

HParker:
%i#one two three#

thinks that it might be interpolation, this causes the correct terminator behavior in that case.


HParker:
Thanks @haldun and @jemmaissroff I think I resolved the comments thanks so much!

> Thanks for writing this, and sorry I missed this case when fixing regex. I just checked but want to confirm, this can't happen for anything else, correct?

I can't think of any other places you can define your own terminators. It is possible there are more, but I can't think of them.

jemmaissroff (approved):

PR #843, opened 2023-04-14
Fix whitespace delim strings

HParker:
This fixes Ruby strings which use whitespace as the delim including some fun special cases for dos newline mixing.

For example ruby is totally ok with,

```
%\r\nstring\n
```

or

```
%\nstring\r\n
```

This also integrates the rubylexer tests with a list of known failures since that is where I found this issue and the tests for it are good there.

closes: https://github.com/Shopify/yarp/issues/839


HParker:
Closing to reopen as more then 1 pr

PR #847, opened 2023-04-15
Allow whitespace after not in expressions

HParker:
Small fix that came up from examples in zonify-0.4.9

### DIFF
--- a/test/fixtures/not.rb
+++ b/test/fixtures/not.rb
@@ -3,3 +3,18 @@ not foo and not bar
 not(foo and bar)
 
 not foo
+
+not foo and not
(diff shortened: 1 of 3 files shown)


test/fixtures/not.rb
```diff
@@ -3,3 +3,11 @@
 not(foo and bar)
 
 not foo
+
```
jemmaissroff, line 6: Can you also add a test case with multiple newlines after not? Something like

```
not


foo
```

jemmaissroff (changes requested):
Thanks for fixing this, minor testing comment!

PR #870, opened 2023-04-17
Emit error for else nodes without rescue

haldun:
Emit error for:

```ruby
begin; 1; else bar; end
```

```
ruby -c -e "begin; 1; else bar; end"
-e:1: else without rescue is useless
begin; 1; else bar; end
```

I unfortunately cannot add a test because `Ripper.sexp_raw` returns non-nil for the code.


jemmaissroff:
> I unfortunately cannot add a test because `Ripper.sexp_raw` returns non-nil for the code.

Interesting. This will presumably be the case for other errors, right? Should we build a harness that allows us to test this correctly?

jemmaissroff:
Okay, I created #889 as a place to discuss, and will merge this in the interim

PR #902, opened 2023-04-24
Check that the start_offset <= end_offset for all locations

eregon:
This shows clearly where it does not hold and gives more context.


eregon:
Failures can be seen at https://github.com/Shopify/yarp/actions/runs/4790229653/jobs/8519067347?pr=902#step:5:33

PR #904, opened 2023-04-25
Fix location for the `WhileNode`

haldun:
- Handle where the statements are before the `while` keyword like:

```ruby
foo = bar while foo != baz
```

~⚠️ Since I am not 100% sure whether we should set the end location of the `WhileNode` to be the end location of the `end` keyword, I haven't implemented that in this PR.~

After a second thought, I've decided to take the `end` keyword into consideration for the location of the enclosing `WhileNode`.


eregon:
Agreed, the location of the `while` node, if there is an `end` keyword for it should cover the `end`.
Thank you for the fixes, I will let someone else review though.

PR #916, opened 2023-05-19
Extract downloading the top 100 gems and add a task to parse the files

eregon:


eregon:
This needs to have https://github.com/ruby/yarp/pull/917 merged first and then it's ready for review.

tasks/lex.rake
```diff
@@ -230,13 +230,83 @@ task "lex:rubygems": :compile do
...
+              end
+            end
+          end
+        end
+      end
+    end
+
+  workers.each(&:join)
+end
+
```
jemmaissroff, line 280: Can we add documentation for this task including in which cases it'll fail?

tasks/lex.rake
```diff
@@ -230,13 +230,83 @@ task "lex:rubygems": :compile do
...
+  class ParseTop100GemsTest < Test::Unit::TestCase
+    Dir["#{TOP_100_GEMS_DIR}/**/*.rb"].each do |filepath|
+      test filepath do
+        result = YARP.parse_file_dup(filepath)
+        unless result.success?
+          if system(RbConfig.ruby, "-c", filepath, out: File::NULL, err: File::NULL)
+            raise "#{filepath} had errors but CRuby parsed it fine: #{result.errors}"
+          else
+            puts "\n#{filepath} was ignored as it is not valid syntax"
```
jemmaissroff, line 294: What's the purpose of this `puts`? Is it helpful for us to know which files were ignored?

src/yarp.c
jemmaissroff, line 1: Can we separate out the rescue node changes into their own PR?

test/test_helper.rb
```diff
@@ -62,6 +62,23 @@ def assert_equal_nodes(expected, actual, compare_location: true, parent: nil)
         assert_equal expected, actual
       end
     end
+
+    def assert_valid_locations(value, parent: nil)
```
jemmaissroff, line 66: This is great, thanks so much!

jemmaissroff (changes requested):

src/yarp.c
eregon: Ah I thought GitHub would automatically update the diff once https://github.com/ruby/yarp/pull/917 got merged but it seems not, I'll rebase.

tasks/lex.rake
eregon: Done

eregon:
Everything addressed, ready for another review/to merge.

jemmaissroff (approved):

PR #917, opened 2023-05-19
Fix location of the rescue node

### LINKED ISSUE #891: Assert start <= end for all locations in snapshots
As part of our test infrastructure we should assert that locations in everything we output in test/snapshots/* all have ends that are _after_ the start.

haldun:
Closes #891


jemmaissroff (approved):
LGTM, thanks!

jemmaissroff (approved):
LGTM, thanks!

PR #919, opened 2023-05-20
Remove `ScopeNode`

### LINKED ISSUE #467: ScopeNode shouldn't be its own node
It should just be a token list on the node that owns it. Having it as its own node confuses other nodes, and removing it would simplify the tree.

haldun:
Removes `ScopeNode` and store the locals as `yp_token_list_t` for the nodes that owns the scope.

Closes #467


jemmaissroff (approved):
LGTM, thanks for refactoring this!

jemmaissroff (approved):
LGTM, thanks for refactoring this!

PR #924, opened 2023-05-22
Check back and nth reference in assignments

### LINKED ISSUE #920: Return error when Regexp related global variables are assigned some value
There are special variables with result of the last Regexp matching: `$``, `$+`, `$'`, `$&`, `${n}`, `$~` that aren't allowed to be assigned. It leads to `SyntaxError`:

```
ruby -e '$+ = nil'
-e:1: Can't set variable $+
```

But YARP doesn't return errors right now:

```
bin/parse -e '$+ = nil'
#<YARP::ParseResult:0x000000010f3f1230
 @comments=[],
 @errors=[],
 @value=
  ProgramNode(0...8)(
    ScopeNode(0...0)([]),
    StatementsNode(0...8)([GlobalVariableWriteNode(0...8)(BACK_REFERENCE(0...2)("$+"), EQUAL(3...4)("="), NilNode(5...8)())])
  ),
 @warnings=[]>
```

haldun:
Closes #920

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -6898,8 +6898,9 @@ parse_target(yp_parser_t *parser, yp_node_t *target, yp_token_t *operator, yp_no
       yp_token_t name = ((yp_global_variable_read_node_t *) target)->name;
       yp_global_variable_write_node_t *result = yp_global_variable_write_node_create(parser, &name, operator, value);
       if ((name.type == YP_TOKEN_BACK_REFERENCE) || (name.type == YP_TOKEN_NTH_REFERENCE)) {
-        char *message = malloc(strlen("Can't set variable $1") * sizeof(char));
-        snprintf(message, strlen("Can't set variable $1") + 1, "Can't set variable %.*s", 2, name.start);
+        size_t message_len = strlen("Can't set variable ") + (unsigned long)(name.end - name.start) + 1;
+        char *message = malloc(message_len * sizeof(char));
+        snprintf(message, message_len, "Can't set variable %.*s", (int)(name.end - name.start) + 1, name.start);
         yp_diagnostic_list_append(
           &parser->error_list,
           name.start,
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -6898,14 +6898,11 @@ parse_target(yp_parser_t *parser, yp_node_t *target, yp_token_t *operator, yp_no
       yp_token_t name = ((yp_global_variable_read_node_t *) target)->name;
       yp_global_variable_write_node_t *result = yp_global_variable_write_node_create(parser, &name, operator, value);
       if ((name.type == YP_TOKEN_BACK_REFERENCE) || (name.type == YP_TOKEN_NTH_REFERENCE)) {
-        size_t message_len = strlen("Can't set variable ") + (unsigned long)(name.end - name.start) + 1;
-        char *message = malloc(message_len * sizeof(char));
-        snprintf(message, message_len, "Can't set variable %.*s", (int)(name.end - name.start) + 1, name.start);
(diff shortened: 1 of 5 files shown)


src/yarp.c
```diff
@@ -6895,7 +6895,11 @@ parse_target(yp_parser_t *parser, yp_node_t *target, yp_token_t *operator, yp_no
     case YP_NODE_CONSTANT_READ_NODE:
       return (yp_node_t *) yp_constant_path_write_node_create(parser, target, operator, value);
     case YP_NODE_GLOBAL_VARIABLE_READ_NODE: {
-      yp_global_variable_write_node_t *result = yp_global_variable_write_node_create(parser, &((yp_global_variable_read_node_t *) target)->name, operator, value);
+      yp_token_t name = ((yp_global_variable_read_node_t *) target)->name;
+      yp_global_variable_write_node_t *result = yp_global_variable_write_node_create(parser, &name, operator, value);
+      if ((name.type == YP_TOKEN_BACK_REFERENCE) || (name.type == YP_TOKEN_NTH_REFERENCE)) {
+        yp_diagnostic_list_append(&parser->error_list, name.start, name.end, "Can't set variable");
```
eregon, line 6901: Shouldn't the error message include the variable name, like CRuby seems to do (from the issue description)?

src/yarp.c
```diff
@@ -6895,7 +6895,18 @@ parse_target(yp_parser_t *parser, yp_node_t *target, yp_token_t *operator, yp_no
     case YP_NODE_CONSTANT_READ_NODE:
       return (yp_node_t *) yp_constant_path_write_node_create(parser, target, operator, value);
     case YP_NODE_GLOBAL_VARIABLE_READ_NODE: {
-      yp_global_variable_write_node_t *result = yp_global_variable_write_node_create(parser, &((yp_global_variable_read_node_t *) target)->name, operator, value);
+      yp_token_t name = ((yp_global_variable_read_node_t *) target)->name;
+      yp_global_variable_write_node_t *result = yp_global_variable_write_node_create(parser, &name, operator, value);
+      if ((name.type == YP_TOKEN_BACK_REFERENCE) || (name.type == YP_TOKEN_NTH_REFERENCE)) {
+        char *message = malloc(strlen("Can't set variable $1") * sizeof(char));
+        snprintf(message, strlen("Can't set variable $1") + 1, "Can't set variable %.*s", 2, name.start);
```
andrykonchin, line 6902: I suppose an Nth reference name could take more than 2 characters, e.g. `$10` etc

eregon:
Looks good to me, thank you!
I've asked @jemmaissroff to review it too.

src/yarp.c
```diff
@@ -6898,8 +6898,9 @@ parse_target(yp_parser_t *parser, yp_node_t *target, yp_token_t *operator, yp_no
       yp_token_t name = ((yp_global_variable_read_node_t *) target)->name;
       yp_global_variable_write_node_t *result = yp_global_variable_write_node_create(parser, &name, operator, value);
       if ((name.type == YP_TOKEN_BACK_REFERENCE) || (name.type == YP_TOKEN_NTH_REFERENCE)) {
-        char *message = malloc(strlen("Can't set variable $1") * sizeof(char));
-        snprintf(message, strlen("Can't set variable $1") + 1, "Can't set variable %.*s", 2, name.start);
+        size_t message_len = strlen("Can't set variable ") + (unsigned long)(name.end - name.start) + 1;
```
andrykonchin: I suppose `message_len` should also take into account a terminating `\0` byte.

andrykonchin: Ah, I see.

jemmaissroff (approved):

PR #925, opened 2023-05-23
Introduce `KeywordHashNode`

### LINKED ISSUE #781: Differentiate HashNode types
If you look at this syntax:

```ruby
foo(1, b: 1)
foo(1, {b: 1})
```

They both generate the same node types:

```text
#<YARP::ParseResult:0x00007f4073732840
 @comments=[],
 @errors=[],
 @value=
  ProgramNode(0...27)(
    ScopeNode(0...0)([]),
    StatementsNode(0...27)(
      [CallNode(0...12)(
         nil,
         nil,
         IDENTIFIER(0...3)("foo"),
         PARENTHESIS_LEFT(3...4)("("),
         ArgumentsNode(4...11)(
           [IntegerNode(4...5)(),
            HashNode(7...11)(
              nil,
              [AssocNode(7...11)(
                 SymbolNode(7...9)(
                   nil,
                   LABEL(7...8)("b"),
                   LABEL_END(8...9)(":"),
                   "b"
                 ),
                 IntegerNode(10...11)(),
                 nil
               )],
              nil
            )]
         ),
         PARENTHESIS_RIGHT(11...12)(")"),
         nil,
         "foo"
       ),
       CallNode(13...27)(
         nil,
         nil,
         IDENTIFIER(13...16)("foo"),
         PARENTHESIS_LEFT(16...17)("("),
         ArgumentsNode(17...25)(
           [IntegerNode(17...18)(),
            HashNode(21...25)(
              BRACE_LEFT(20...21)("{"),
              [AssocNode(21...25)(
                 SymbolNode(21...23)(
                   nil,
                   LABEL(21...22)("b"),
                   LABEL_END(22...23)(":"),
                   "b"
                 ),
                 IntegerNode(24...25)(),
                 nil
               )],
              BRACE_RIGHT(25...26)("}")
            )]
         ),
         PARENTHESIS_RIGHT(26...27)(")"),
         nil,
         "foo"
       )]
    )
  ),
 @warnings=[]>
```

If we examine opening for non-null we can know which type of hash style was passed but:

  1. These semantically are two different things
  2. I don't want to populate AST with syntactal info for execution.  For syntax tree uses this also saves space for keywords arg case.

I propose we make an additional type KeywordHashNode:

```text
  - name: KeywordHashNode
    child_nodes:
      - name: elements
        type: node[]
    comment: |
      Represents a hash literal.

           foo(1, a: b) 
                     ^^^^
```

haldun:
Closes #781

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1025,10 +1025,10 @@ nodes:
       - name: elements
         type: node[]
     comment: |
-      Represents a hash literal.
+      Represents a hash literal without an opening and closing braces.
 
-          a => b
-          ^^^^^^^^^^
+          foo(a: b)
+              ^^^^
   - name: KeywordParameterNode
     child_nodes:
       - name: name
--- a/config.yml
+++ b/config.yml
@@ -1020,6 +1020,15 @@ nodes:
 
           `foo #{bar} baz`
           ^^^^^^^^^^^^^^^^
+  - name: KeywordHashNode
+    child_nodes:
+      - name: elements
+        type: node[]
+    comment: |
+      Represents a hash literal.
+
+          a => b
+          ^^^^^^^^^^
   - name: KeywordParameterNode
     child_nodes:
       - name: name
--- a/config.yml
+++ b/config.yml
@@ -867,11 +867,11 @@ nodes:
   - name: HashNode
     child_nodes:
       - name: opening
-        type: token?
+        type: token
       - name: elements
         type: node[]
       - name: closing
-        type: token?
+        type: token
     comment: |
       Represents a hash literal.
 
--- a/config.yml
+++ b/config.yml
@@ -1025,7 +1025,7 @@ nodes:
       - name: elements
         type: node[]
     comment: |
-      Represents a hash literal without an opening and closing braces.
+      Represents a hash literal without opening and closing braces.
 
           foo(a: b)
               ^^^^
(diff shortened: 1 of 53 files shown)


config.yml
```diff
@@ -1020,6 +1020,15 @@ nodes:
 
           `foo #{bar} baz`
           ^^^^^^^^^^^^^^^^
+  - name: KeywordHashNode
+    child_nodes:
+      - name: elements
+        type: node[]
+    comment: |
+      Represents a hash literal without an opening and closing braces.
```
jemmaissroff, line 1028: Nit `Represents a hash literal without an opening and closing braces.` -> `Represents a hash literal without opening and closing braces.`

src/yarp.c
```diff
@@ -1977,6 +1979,35 @@ yp_interpolated_xstring_node_closing_set(yp_interpolated_x_string_node_t *node,
...
+
+  yp_node_list_init(&node->elements);
+  return node;
+}
+
+// Append an element to a KeywordHashNode node.
+static void
+yp_keyword_hash_node_elements_append(yp_keyword_hash_node_t *hash, yp_node_t *element) {
+  yp_node_list_append(&hash->elements, element);
+  if (hash->elements.size == 1) {
```
jemmaissroff, line 2005: Is this check more straightforward as `hash->base.location.start == NULL`?

jemmaissroff (changes requested):

src/yarp.c
jemmaissroff: Yup, they are definitely logically the same, I think just more clear to read with the NULL check, thanks!

jemmaissroff (approved):

PR #926, opened 2023-05-23
Check argument names for duplication

### LINKED ISSUE #888: Should emit errors for duplicated argument names
```
> ruby -c -e "def foo(a,a);end"
-e:1: duplicated argument name
```

haldun:
Closes #888


test/errors_test.rb
jemmaissroff, line 1: Can we add a few more test cases? Explicitly, I'd like us to test the underscore cases too

jemmaissroff (changes requested):

jemmaissroff (approved):

PR #928, opened 2023-05-24
Deserialization Optimizations

eregon:
This likely best reviewed commit-by-commit.

It's overall fairly trivial changes, except [9c57958604d87eff3641c672bdbf6316349b291c](https://github.com/ruby/yarp/pull/928/commits/9c57958604d87eff3641c672bdbf6316349b291c) which is bigger.
That also affects node's `initialize` arguments, i.e., `FloatNode.new(0, 0)` instead of `FloatNode.new(Location.null)`, but the arguments of `YARP::DSL` methods is unchanged.


bin/templates/lib/yarp/serialize.rb.erb
```diff
@@ -12,9 +12,11 @@ module YARP
       attr_reader :source, :io, :encoding
 
       def initialize(source, io)
-        @source = source.b
-        @io = io
+        # This is wrong, we should be dumping and loading the encoding. For now,
```
enebo, line 15: Does this project have a convention of using FIXME: for things like this?

eregon: In case it's not clear, it was an existing comment, I just moved it to a better place.
I can prepend it with FIXME/TODO.

eregon: Ah this is already tracked as https://github.com/ruby/yarp/issues/643

jemmaissroff (approved):

PR #932, opened 2023-05-31
Add length() and endOffset() to migrate to (startOffset, length) in a compatible way

eregon:
For https://github.com/ruby/yarp/issues/872
By adding these methods, and TruffleRuby migrating to them, we can avoid breaking the `Test Loader.java with TruffleRuby` CI job.

### DIFF
--- a/bin/template.rb
+++ b/bin/template.rb
@@ -136,7 +136,7 @@ class NodeType
 
   # Should emit serialized length of node so implementations can skip
   # the node to enable lazy parsing.
-  def needs_length?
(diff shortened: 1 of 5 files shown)


eregon:
Corresponding TruffleRuby PR: https://github.com/oracle/truffleruby/pull/3081

BTW in this case since it was easy to not break the job I did it, but I think it's totally fine to break it if we know how to fix it and there is no easy way to do it in a non-breaking manner (e.g., renaming a node class or a node field).

PR #934, opened 2023-05-31
Introduce flags, parse regular expression options

### LINKED ISSUE #760: Regexp Node should encode options
Given ```/foo/i``` the current node structure is:

```text
#<YARP::ParseResult:0x00007f95a1ade6a0
 @comments=[],
 @errors=[],
 @value=
  ProgramNode(0...6)(
    ScopeNode(0...0)([]),
    StatementsNode(0...6)(
      [RegularExpressionNode(0...6)(
         REGEXP_BEGIN(0...1)("/"),
         STRING_CONTENT(1...4)("foo"),
         REGEXP_END(4...6)("/i"),
         "foo"
       )]
    )
  ),
 @warnings=[]>
```

In order to determine what options exist consumers need to make a string processing method for REGEXP_END (execution will likely end up stripping syntax fields out of nodes at some point).  I talked last week to @kddnewton about allowing node types to have enums defined.  This may be a good target node to try that on.

kddnewton:
Fixes #760.

This introduces a new concept to `config.yml` which is sets of flags. Flags are meant to be used in conjunction with `uint32_t` parameters and function as bitfields.

The first place we introduce this is for regular expressions and interpolated regular expressions. We now parse each of the options given on the end of the string into a bitfield. Since we're now serializing variable-length integers, for the most part this only adds a single byte per regular expression to the serialized output.

It does add 4 bytes to each regular expression in memory on the C side, but I think it's worth it to make it easier on the consumer. Future optimizations could recognize the maximum value of the integer and shrink to fit.

Since I'm just using `uint32_t` parameters to hold the actual bitfields, there's no direct link from the enum to the value. So I tried to be nice and add an enum in Java and constants in Ruby to make it more obvious to the consumer how to interact with it. I don't really know what I'm doing on the Java side so @eregon or @enebo let me know if there's a better way to express that.

### DIFF
--- a/bin/templates/java/org/yarp/Nodes.java.erb
+++ b/bin/templates/java/org/yarp/Nodes.java.erb
@@ -18,6 +18,12 @@ public abstract class Nodes {
     }
 
     static final TokenType[] TOKEN_TYPES = TokenType.values();
+<%- flags.each do |flag| -%>
+
+    <%- flag.values.each_with_index do |value, index| -%>
+    static final int <%= flag.human.upcase %>_<%= value.upcase %> = 1 << <%= index %>;
+    <%- end -%>
+<%- end -%>
 
     public static final class Token {
         public final TokenType type;
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -337,6 +337,30 @@ yp_alloc(__attribute__((unused)) yp_parser_t *parser, size_t size) {
 /* Node creation functions                                                    */
 /******************************************************************************/
 
+// Parse out the options for a regular expression.
+static inline uint32_t
+yp_regular_expression_flags_create(const yp_token_t *closing) {
+  uint32_t flags = 0;
+
+  if (closing->type == YP_TOKEN_REGEXP_END) {
+    for (const char *flag = closing->start + 1; flag < closing->end; flag++) {
+      switch (*flag) {
+        case 'i': flags |= YP_REGULAR_EXPRESSION_FLAGS_IGNORECASE; break;
+        case 'm': flags |= YP_REGULAR_EXPRESSION_FLAGS_MULTILINE; break;
+        case 'x': flags |= YP_REGULAR_EXPRESSION_FLAGS_EXTEND; break;
+        case 'e': flags |= YP_REGULAR_EXPRESSION_FLAGS_EUCJP; break;
+        case 'n': flags |= YP_REGULAR_EXPRESSION_FLAGS_ASCII8BIT; break;
+        case 's': flags |= YP_REGULAR_EXPRESSION_FLAGS_WINDOWS31J; break;
+        case 'u': flags |= YP_REGULAR_EXPRESSION_FLAGS_UTF8; break;
+        case 'o': flags |= YP_REGULAR_EXPRESSION_FLAGS_ONCE; break;
+        default: assert(false && "unreachable");
+      }
+    }
+  }
+
+  return flags;
+}
+
 // Allocate and initialize a new StatementsNode node.
 static yp_statements_node_t *
 yp_statements_node_create(yp_parser_t *parser);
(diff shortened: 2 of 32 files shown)


src/yarp.c
```diff
@@ -1837,19 +1861,20 @@ yp_instance_variable_write_node_create(yp_parser_t *parser, yp_instance_variable
...
+yp_interpolated_regular_expression_node_create(yp_parser_t *parser, const yp_token_t *opening) {
   yp_interpolated_regular_expression_node_t *node = yp_alloc(parser, sizeof(yp_interpolated_regular_expression_node_t));
 
   *node = (yp_interpolated_regular_expression_node_t) {
     {
       .type = YP_NODE_INTERPOLATED_REGULAR_EXPRESSION_NODE,
       .location = {
         .start = opening->start,
-        .end = closing->end
+        .end = opening->end
```
jemmaissroff, line 1872: Nit but I wonder if we should set `.end = NULL` here. It should be overwritten, but in the off chance this method is ever called without a subsequent `yp_interpolated_regular_expression_node_closing_set` call, it'd be more clear to have the end as NULL

jemmaissroff (approved):
Minor nit, feel free to address or not! Otherwise LGTM

jemmaissroff (approved):
Minor nit, feel free to address or not! Otherwise LGTM

enebo:
@kddnewton I think I got confused in the small only considering the individual value (e.g. MULTILINE) but I think it is that you are constructing an enum of values but regexp flags are an OR'd set.  In order to represent that in Java you then need to use an EnumSet.  In JRuby,  our regexp code is an OR'd value of the same integers you encode so we wouldn't use the enum at all (technically this is what our oniguruma port joni uses).   This may not be true of other flags but for anything OR'd we almost never use enums.  If those values are "correct" values then we have an int field somewhere where that value could just be copied.

I cannot speak for TruffleRuby though.  They may like that.

I would love to set up another chat or even know if there is a place to IM to talk to you about some broader design issues.

enebo:
@kddnewton An added note regexp optoins on JRuby is we do effectively make a custom enum set of sorts called RegexpOptions (each flag value is a boolean).  It has two forms (java embedding + ruby) of converting between ints and this object but it is actually an enum set in form.  With that said we have construction method which takes OR'd int to make it.  So OR'd int is still preferable (and joni does use an OR'd int which this knows how to make).

I guess the notion maybe is yarp provides a representation and we make the same representation as something else.  This makes load generic but it also makes a sliver of extra work translating one value to another (e.g. RubyRegexp will never use a YARP enum directly).

LOL.  Sorry a bit tired.  I will reread this on the plane tomorrow :)

eregon:
I'll review this this weekend or Monday, looks good from a quick look on my phone but I need to check the generated code to get a better idea

kddnewton:
@enebo the value on the node is still an integer. I was mostly putting these things there for convenience so you can use it to decode the value. I think the value should still be an integer on the node at the end of the day, but we could create custom flag objects later.

I dropped the enum since it is a little weird for it to be an enum. Instead it's just a bunch of integers now. I'm happy to change the representation to whatever is easier for y'all going forward. I'm going to go ahead and merge this for now because I want to put flags on a couple of other nodes, but definitely don't feel like this is set in stone at all.

bin/templates/java/org/yarp/Nodes.java.erb
```diff
@@ -18,6 +18,12 @@ public abstract class Nodes {
     }
 
     static final TokenType[] TOKEN_TYPES = TokenType.values();
+<%- flags.each do |flag| -%>
+
+    <%- flag.values.each_with_index do |value, index| -%>
+    static final int <%= flag.human.upcase %>_<%= value.upcase %> = 1 << <%= index %>;
```
eregon, line 24: It should be `public` to be accessible outside of this package.
I'll make a small class for this, based on https://github.com/oracle/truffleruby/blob/master/src/main/java/org/truffleruby/core/regexp/RegexpOptions.java

eregon: https://github.com/ruby/yarp/pull/948

PR #940, opened 2023-06-02
Remove tokens from the tree

kddnewton:
For the most part, there's no need to attach tokens to the AST. Instead they can be represented as a source range. The actual type of token holds no value to the consumer in almost all cases. In the few cases where it does contain useful information, it can be better represented using an explicit set of flags. By removing tokens from the tree we save on memory on both the C side and the serialized side, as the type byte gets dropped.

This PR removes _most_ of the tokens from the AST, but not all. There are a couple that are trickier than a simple find/replace so I figured I'd split this into a couple of PRs for the last couple. While there are a lot of changes in this PR, each node is split into its own commit. Almost all of the changes are inside the snapshots, where you'll see something like `IDENTIFIER(0...5)` --> `(0...5)`. I tried to keep it as consistent as possible so it's really just dropping that one type byte.

The remaining work _after_ this is done is summarized below:

1. Switch the `locals` sets to use strings instead of tokens. Right now they're just comparing on their bounds and ignoring their types anyway. We can save on memory here. This will also pave the way for a potential constant pool/string interning in the future. Those places are `BlockNode#locals`, `ClassNode#locals`, `DefNode#locals`, `LambdaNode#locals`, `ModuleNode#locals`, `ProgramNode#locals`, and `SingletonClassNode#locals`.
2. Switch `BlockParametersNode#locals` over to using a string list. This is different from the scope locals in that this is the block-local list, as in `-> (; foo) {}`.
3. Switch `CallNode` over to using locations. This includes `#call_operator`, `#message`, `#opening`, and `#closing`. Since calls are everywhere, this is going to be its own PR.
4. Switch `GlobalVariableReadNode#name` and `GlobalVariableWriteNode#name` over to using locations. I wasn't 100% sure if we should introduce flags or not for this. `#name` is polymorphic (as opposed to the monomorphic changes in this PR) because it can be a back reference, an nth reference, or a global variable. This will need more investigation so it goes in its own PR.
5. Switch `OperatorAssignmentNode#operator` to a location. For this one we should definitely introduce an enum for the various operations so that consumers don't have to re-parse the operator.
6. Switch `InterpolatedSymbolNode` and `SymbolNode` to using locations. The only reason this wasn't in this PR was because there's some things we do with whether or not a symbol is in fact a label that we might need to look at more in-depth.

Overall I'm pretty happy with the state of this PR, because it saves us a lot of memory, and makes it more clear in the tree where the relevant content is. Once we finish all of the other work, we'll be able to remove the token types entirely from the tree so that consumers don't have to know about them at all.


jemmaissroff (approved):
LGTM! Thanks for doing this. Tried going through it 1 by 1, and mostly looks pretty plug in so I think all good!

PR #949, opened 2023-06-05
Fixing 3.3 errors

jemmaissroff:
This PR has 3 commits which fix separate issues from bumping to run against Ruby head, and then one final commit which bumps our GitHub CI to run against Ruby head. This is all in preparation for mirroring to ruby/ruby

### DIFF
--- a/.github/workflows/truffleruby.yml
+++ b/.github/workflows/truffleruby.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Setup system Ruby
       uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 3.2
+        ruby-version: head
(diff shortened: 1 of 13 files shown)


.github/workflows/truffleruby.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     - name: Setup system Ruby
       uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 3.2
```
eregon, line 20: This should still be 3.2 or at least a release (`ruby-version: ruby`), this ruby will notably be used to run `jt`, which might not work on ruby-head.

kddnewton (approved):
Looks great!

PR #950, opened 2023-06-05
Use the constant pool for local variables and parameters

kddnewton:
This pull request introduces using the constant pool into the AST. There are a couple of ways that it does this.

1. We add a constant pool onto the parser.
2. We introduce two new parameter types: `constant` and `constant[]`. The first is the id of an interned constant string in the parser's constant pool. The second is a buffered list of ids of interned constant strings.
3. We change almost all places where we're using `token[]` to use `constant[]` instead. This changes our list of locals at all scopes to use id lists, which saves on space and comparison speed.
4. We add `constant_id` to `LocalVariableReadNode`, `LocalVariableWriteNode`, `RequiredParameterNode`, and `OptionalParameterNode`. This makes compiling these nodes much easier. (In the future this will get added to more nodes, I just did the minimal to get the tests passing at the moment.)
5. Loaders of the serialization API now return an object that contains both the root node and the constant pool. In Ruby this means the node and an array of strings. In Java this means the node and a `byte[][]`. To do this I needed to keep around the source and I'm definitely not 100% sure on it so @enebo and @eregon it could use your eyes.

I'm going to do some benchmarking to prove this, but my intuition is that the more identifiers present in the file the better this is going to perform.

## UPDATES

I've changed it now so that when reifying into Ruby objects or deserializing into Ruby objects, it uses the constant pool to create symbol objects instead. So now the constant pool is relatively transparent.


enebo:
Just a note on how we have done this for our persistence (used in AOT) in JRuby.  I put an offset field in the prologue of the serialized header which contains the location of where the constant pool starts.  I write out the contents then once I am at end I go back and add that as the offset then write the pool at the end.  Doing it this way would not require changing the serialization signature but it also keep the foreign function calling code as simple as possible.

Not part of this PR but something tangential to think about.  I am hoping not to have the serialized blob AND source in memory as this will be heavier than what we do today.   Also, if we consider a notion of precompilation it may be convenient/faster to only load a single file.  Whether strings are a new pool, part of this one, or just embedded into nodes is obviously a different issue/PR but it felt worth a mention.

eregon:
Thanks, I'll take care of the Java side.

eregon (approved):

PR #952, opened 2023-06-05
More mirroring changes

jemmaissroff:
A few more changes, mostly minor refactors, to ensure that the mirroring will work. I tried to separate them into bite-sized commits, hopefully makes it easier to see.


eregon (approved):

lib/yarp.rb
```diff
@@ -234,5 +234,5 @@ def self.load(source, serialized)
 require_relative "yarp/node"
 require_relative "yarp/ripper_compat"
 require_relative "yarp/serialize"
-require_relative "yarp/yarp"
 require_relative "yarp/pack"
+require "yarp.so"
```
kddnewton, line 238: Is this extension correct on all platforms? (Honestly asking I don't know)

jemmaissroff: As far as I understand, yes!

PR #954, opened 2023-06-06
More cleanup for mirroring to ruby/ruby

jemmaissroff:
This is again split commit by commit. I'll likely keep adding PRs like this incrementally as I'm going through the mirroring process.


test/fixtures/seattlerb/heredoc_squiggly_visually_blank_lines.rb
```diff
@@ -1,6 +1,6 @@
 a = <<~EOF
   x
- 
+
```
kddnewton, line 3: I think this one is intentional.

kddnewton (approved):
One small change then good.

kddnewton:
Ehh, it's probably fine. The snapshot doesn't change much, so go ahead and make it a blank line.

PR #956, opened 2023-06-06
Use unpack1 offset to avoid intermediate string

kddnewton:


rafaelfranca (approved):

eregon (approved):
Nice, I didn't know about unpack's offset: kwarg

PR #958, opened 2023-06-07
Rename snapshot, fixture files to have `.txt` suffixes

jemmaissroff:
The fixture files are sometimes invalid Ruby code (for instance a series of `break`s in a row). The snapshot files are not Ruby files. This PR changes their suffixes to `.txt` to reflect that they're not Ruby files. It also makes these files work with the ruby/ruby test suite which has some tests which will try to parse them as valid Ruby files.


kddnewton (approved):
This is good, but can you also update the .gitattributes file?

PR #959, opened 2023-06-07
Add default location to parameters node

jemmaissroff:
This is currently breaking on 32 bit machines because in the case where we only have errors in the parameters, we never set the location, and then have underflow if the file is sufficiently large

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -2458,7 +2458,7 @@ yp_parameters_node_create(yp_parser_t *parser) {
     *node = (yp_parameters_node_t) {
         {
             .type = YP_NODE_PARAMETERS_NODE,
-            .location = { .start = NULL, .end = NULL },
+            .location = { .start = parser->current.start, .end = parser->current.start },
         },
         .rest = NULL,
         .keyword_rest = NULL,


kddnewton (approved):
I'm approving this, but I think we shouldn't be creating parameters nodes in the case that they're entirely empty. Could you find out in where in the tree we're creating these extra nodes?

jemmaissroff:
> I'm approving this, but I think we shouldn't be creating parameters nodes in the case that they're entirely empty. Could you find out in where in the tree we're creating these extra nodes?

Not sure I'm following this question. We create the parameters nodes [here](https://github.com/ruby/yarp/blob/aa6148adbdb828718da3cd037b807b200429b067/src/yarp.c#L7752) before we've actually parsed the parameters. We then go through and parse parameters, and if they're all invalid / errors, we'll still have the parameters node.

This is a snippet from [our errors test](https://github.com/ruby/yarp/blob/aa6148adbdb828718da3cd037b807b200429b067/test/errors_test.rb#L434-L439) which triggered the code path where we create a parameters node with an invalid location: `"def foo(A, @a, $A, @@a);end"`
and here's the parsed tree (prior to this commit):

```
#<YARP::ParseResult:0x000000010733b140
 @comments=[],
 @errors=
  [#<YARP::ParseError:0x0000000107474750 @location=(8...17), @message="Formal argument cannot be a constant">,
   #<YARP::ParseError:0x00000001074746d8 @location=(11...24), @message="Formal argument cannot be an instance variable">,
   #<YARP::ParseError:0x0000000107474660 @location=(13...27), @message="Unexpected ','.">,
   #<YARP::ParseError:0x0000000107474610 @location=(14...28), @message="Expected ')' after left parenthesis.">,
   #<YARP::ParseError:0x00000001074745c0 @location=(14...28), @message="Expected to be able to parse an expression.">,
   #<YARP::ParseError:0x0000000107474570 @location=(20...40), @message="Expected a newline or semicolon after statement.">,
   #<YARP::ParseError:0x0000000107474520 @location=(20...40), @message="Expected to be able to parse an expression.">],
 @value=
  ProgramNode(0...25)(
    [],
    StatementsNode(0...25)(
      [DefNode(0...25)(
         (4...7),
         nil,
         ParametersNode(-4391280584...-4391280584)([], [], [], nil, [], nil, nil),
         StatementsNode(14...20)([MissingNode(14...14)(), ClassVariableReadNode(17...20)(), MissingNode(20...20)()]),
         [],
         (0...3),
         nil,
         (7...8),
         (14...14),
         nil,
         (22...25)
       )]
    )
  ),
 @warnings=[]>
 ```

kddnewton:
Yeah I'm saying we shouldn't be calling `parse_parameters` in places where we know the next token can't possibly start a set of parameters.

jemmaissroff:
Okay, merging now, but created #961 to fix this.

PR #960, opened 2023-06-07
Profiling and unicode work

kddnewton:
This commit adds two things. The first is a build/profile make target which is a small script that walks through a directory recursively and parses every Ruby file that it finds. I use this in combination with dappprof and dtrace to do some profiling to determine where we're spending the most time.

The second thing this adds is a complete refactor of our UTF-8 decoder. Because UTF-8 is the default encoding and we need to know if characters are alphanumeric or not very commonly, it's important that this path is very fast. Previously we were always decoding the unicode codepoint and then comparing it to a list of known values, but that's incredibly inefficient when the character is < 128 (i.e., almost all Ruby code). Instead we can use a lookup table like we do for all the other encodings.

The UTF-8 decoding is actually significantly faster as well for the few files that actually do need it. It now uses Bjoern Hoehrmann's DFA-based UTF-8 decoder (modified to fit our use case).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,9 @@ all: build/librubyparser.$(SOEXT)
 build/librubyparser.$(SOEXT): $(shell find src -name '*.c') $(shell find src -name '*.h') Makefile build include/yarp/ast.h
 	$(CC) $(OPTFLAGS) $(DEBUG_FLAGS) $(CFLAGS) -std=c99 -Wall -Werror -Wextra -Wpedantic -Wsign-conversion -fPIC -g -fvisibility=hidden -shared -Iinclude -o $@ $(shell find src -name '*.c')
 
+build/profile: $(shell find src -name '*.c') $(shell find src -name '*.h') Makefile build include/yarp/ast.h bin/profile.c
+	$(CC) $(CFLAGS) -O3 -std=c99 -Iinclude -o $@ $(shell find src -name '*.c') bin/profile.c
+
 build:
 	mkdir -p build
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -22,6 +22,9 @@ all: build/librubyparser.$(SOEXT)
 build/librubyparser.$(SOEXT): $(shell find src -name '*.c') $(shell find src -name '*.h') Makefile build include/yarp/ast.h
 	$(CC) $(OPTFLAGS) $(DEBUG_FLAGS) $(CFLAGS) -std=c99 -Wall -Werror -Wextra -Wpedantic -Wsign-conversion -fPIC -g -fvisibility=hidden -shared -Iinclude -o $@ $(shell find src -name '*.c')
 
+build/profile: $(shell find src -name '*.c') $(shell find src -name '*.h') Makefile build include/yarp/ast.h bin/profile.c
+	$(CC) $(CFLAGS) -O3 -std=c99 -Iinclude -o $@ $(shell find src -name '*.c') bin/profile.c
```
eregon, line 26: Instead of recompiling all C files, one way would be to do something like `be rake clobber && be rake compile_no_debug && be rake compile`, or an easier rake command to get librubyparser compiled as `-O3` with no asserts.
And then link profile.c to that.

This line seems to not disable asserts (`-DNDEBUG=1`).

eregon: Yes, but specifically the non-debug, `-O3` shared library.
Maybe we should store both libraries in different paths, that might be clearer.

In any case, feel free to merge this PR as-is, we can look at that later.

PR #962, opened 2023-06-08
Return NULL for empty ParametersNode

### LINKED ISSUE #961: Fix cases where we create a ParamtersNode with no parameters
We shouldn't create a parameters node if there are no valid parameters. For instance, in this case `"def foo(A, @a, $A, @@a);end"`. See #959 for more context

haldun:
Fixes #961

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -3409,7 +3409,6 @@ yp_required_destructured_parameter_node_closing_set(yp_required_destructured_par
 // Allocate a new RequiredParameterNode node.
 static yp_required_parameter_node_t *
 yp_required_parameter_node_create(yp_parser_t *parser, const yp_token_t *token) {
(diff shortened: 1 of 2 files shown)


kddnewton:
A couple of thoughts on this. 

1. I think the case where we have invalid parameters should still add parameters to the list. For example `def foo(A)`, you'd still expect there to be a node in the list of required parameters, even though it's incorrect. That way there's a place to attach the error message and formatters/linters have something to point to.
2. Instead of allocating the parameters node and then immediately deallocating, I would much prefer that we simply never allocate in the first place. I think this is probably going to be solved by the first point above, but there may be other cases where we need to look forward instead.

kddnewton:
@haldun I've come around on this. Could you rebase and make sure the tests pass?

PR #964, opened 2023-06-08
Track a list of newline locations as we parse

kddnewton:


ext/yarp/extension.c
```diff
@@ -436,6 +436,24 @@ compile(VALUE self, VALUE string) {
     return result;
 }
 
+static VALUE
+newlines(VALUE self, VALUE string) {
```
jemmaissroff, line 440: Can we add a comment describing this method? Thinking specifically it'd be helpful to say that `string` is the actual text of the file, not the filename

include/yarp/util/yp_newline_list.h
```diff
@@ -0,0 +1,52 @@
...
+#include <stddef.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
+// A list of offsets of newlines in a string. The offsets are assumed to be
+// sorted/inserted in ascending order.
+typedef struct {
+    const char *start;
+
+    size_t *offsets;
```
jemmaissroff, line 22: Are we going to try trend away from `size_t`s?

src/util/yp_newline_list.c
```diff
@@ -0,0 +1,117 @@
...
+
+    list->start = start;
+
+    // This is 1 instead of 0 because we want to include the first line of the
+    // file as having offset 0, which is set because of calloc.
+    list->size = 1;
+    list->capacity = capacity;
+
+    list->last_index = 0;
+    list->last_offset = 0;
```
jemmaissroff, line 18: Not sure where I land on this, but as is, we're missing the optimization on the very first line. We could consider setting it to `-1` or something so we keep the optimization there. Again though, not sure if the signed int is worth getting this extra optimization...

jemmaissroff (changes requested):

src/util/yp_newline_list.c
kddnewton: Yeah I figured it was fine to lose it at the beginning honestly.

test/parse_test.rb
```diff
@@ -73,6 +73,11 @@ def normalize_printed(printed)
       # changing the shape of the tree.
       assert_equal_nodes(value, YARP.load(source, YARP.dump(source, filepath)))
 
+      # Next, assert that the newlines are in the expected places.
+      expected_newlines = [0]
+      source.b.scan("\n") { expected_newlines << $~.offset(0)[0] }
+      assert_equal expected_newlines, YARP.newlines(File.read(filepath))
```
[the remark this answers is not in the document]
kddnewton: 👍

include/yarp/util/yp_newline_list.h
kddnewton: I think this is fine because it won't be serialized. When moving away from `size_t` I was mostly talking about it in the context of having to change them all to `uint32_t` for explicit widths in the serialization.

ext/yarp/extension.c
kddnewton: 👍

jemmaissroff (approved):

PR #965, opened 2023-06-08
Make it so we can release as a gem

kddnewton:


ext/yarp/extconf.rb
```diff
@@ -10,11 +10,12 @@
...
 
-  # Tell Ruby where to find the headers for YARP, specify the C standard, and make
-  # sure all symbols are hidden by default.
+  # Tell Ruby where to find the headers for YARP, specify the C standard, and
+  # make sure all symbols are hidden by default.
+  find_header("yarp.h", File.expand_path("include", __dir__))
   $CFLAGS << " -I#{__dir__}/include -std=gnu99 -fvisibility=hidden"
 
-  $objs = %w[extension.o yarp.o node.o pack.o]
+  $objs = %w[api_node.o api_pack.o compile.o extension.o yarp.o]
```
eregon, line 18: Do we need to specify this list? It seems rather error-prone

ext/yarp/extension.h
```diff
@@ -3,7 +3,7 @@
 
 #include <ruby.h>
 #include <ruby/encoding.h>
-#include "yarp.h"
+#include <yarp.h>
```
jemmaissroff, line 6: Why this change? I had changed it to `"yarp.h"` so that it still worked for mirroring when this file was in a slightly different place

ext/yarp/extconf.rb
kddnewton: Honestly I don't know. I _think_ so? I am really struggling to understand mkmf

yarp.gemspec
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+Gem::Specification.new do |spec|
+  spec.name = "yarp"
+  spec.version = "0.4.0"
+  spec.authors = ["Shopify"]
+  spec.email = ["ruby@shopify.com"]
```
jemmaissroff, line 7: This is funny, didn't know we had this email address!

jemmaissroff (changes requested):

ext/yarp/extconf.rb
kddnewton: The issue is, `yarp.c` _won't_ be in `ext/yarp` because we create it immediately above. So maybe I just need `$objs << "yarp.o"`?

eregon: Do we need this BUNDLE mode? Or could we just always use the case below?
Concatenating seems not so nice if one ever has to debug those files, and we could use some LTO flags if they prove effective.

flavorjones: @kddnewton It shouldn't matter -- the code I pointed to above is in `create_makefile` and the source files are determined immediately above the code I linked.

Try removing the `$objs` line and see if it breaks? It seems to work OK for me (but my dev env may not be set up correctly).

yarp.gemspec
rafaelfranca: To all of you https://groups.google.com/a/shopify.com/g/ruby/members

yarp.gemspec
```diff
@@ -0,0 +1,33 @@
...
+  spec.files = Dir.glob([
+    "CODE_OF_CONDUCT.md",
+    "config.yml",
+    "CONTRIBUTING.md",
+    "Gemfile",
+    "LICENSE.md",
+    "Makefile",
+    "Rakefile",
+    "yarp.gemspec",
+    "bin/**/*",
```
rafaelfranca, line 23: Should bin be in the gem. It looks like internal scripts to develop this project.

ext/yarp/extconf.rb
flavorjones: Oh, I see, you have source files in $srcdir *and* the build directory. :thinking:

flavorjones, line 18:
```suggestion
  srcdir = RbConfig.expand(+"$(srcdir)")
  $srcs = Dir[File.join(srcdir, "*.c"), "*.c"]
```

flavorjones: I believe this change addresses @eregon's concern.

kddnewton:
Going to close this for now in favor of the PR @flavorjones is going to post

ext/yarp/extconf.rb
nobu: > So maybe I just need `$objs << "yarp.o"`?

It should work.

PR #966, opened 2023-06-09
Try adding windows workflow

jemmaissroff:
just for looking at CI

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, macos-latest ]
+        os: [ ubuntu-latest, macos-latest, windows-latest ]
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -20,6 +20,11 @@ jobs:
       with:
         ruby-version: head
         bundler-cache: true
+    - name: Install MinGW (Windows)
```
flavorjones, line 23: I see this is a draft but I wanted to note that I think setup-ruby comes with mingw installed on Windows. This step might not be necessary?

eregon: Yes, it shouldn't be

eregon:
https://github.com/ruby/yarp/actions/runs/5225354819/jobs/9434710232?pr=966#step:6:26
`cc ... -o build/librubyparser.so` so that needs to be fixed most likely, I think it should be `rubyparser.dll` or so

jemmaissroff:
> I think it should be `rubyparser.dll` or so

@eregon do you mind explaining why?

eregon:
On Windows SOEXT is I think `dll` and not `so`
Also the "libprefix" on Windows is "", vs "lib" for most other OS.
@MSP-Greg you know a lot more about this, maybe you could advice?

MSP-Greg:
If the file is loaded in Ruby via `require`, it would normally have an `.so` extension.  If it's loaded by fiddle/ffi, it would have a `.dll` extension.

I can't do so right now, but later I'll look at this.

MSP-Greg:
Re Windows:

I changed `Makefile` to start with the following:
```make
ifeq ($(shell uname), Darwin)
SOEXT := dylib
else ifeq ($(shell uname), Linux)
SOEXT := so
else
SOEXT := dll
endif
```

In `yarp/.rb`, loaded `librubyparser.dll` with Fiddle.

Running `parser_test.rb` causes a SEGV, but all other tests pass.

jemmaissroff:
Punting this for later work! Created #980

PR #970, opened 2023-06-13
Fix line and file information in generated files

tenderlove:
When we generate files, error messages (including compilation errors and stack traces) would point to the generated .c file when the error actually originates from the source ERB file.  When fixing these generated files, we need to fix the source ERB file rather than the generated file, so this commit changes the generated file to use the source file for line and file reporting information

For example, if there is a syntax error in the source ERB file like this:

yarp — ~:g:yarp — -fish — 284×76 2023-06-13 11-34-35

When you compile YARP, the error will be reported in the wrong place like this:

wrong

This patch fixes the error message so that it's reported as part of the source ERB file:

fixed

This PR only fixes `node.c`.  We need to fix the other files, but there also might be a more generic way to do this (maybe like a custom ERB generator that automatically emits the `#line` directives).

### DIFF
--- a/bin/templates/src/node.c.erb
+++ b/bin/templates/src/node.c.erb
@@ -1,3 +1,4 @@
+#line <%= __LINE__ + 1 %> "<%= File.basename(__FILE__) %>"
 #include "yarp/node.h"
 
 // Clear the node but preserves the location.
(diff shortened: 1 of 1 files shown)


kddnewton (approved):
Wow I had no idea you could do this. Nice!

PR #972, opened 2023-06-13
Increase default constant id list size

kddnewton:
Seeing a lot of reallocations when adding lists of constants because it does one for the 1st constant, then another for the 2nd constant, then another for the 3rd. Setting this to 8 by default doesn't significantly increase memory usage (all of the constant ids are still within a single cache line) but does avoid a bunch of unnecessary syscalls.

### DIFF
--- a/src/util/yp_constant_pool.c
+++ b/src/util/yp_constant_pool.c
@@ -13,8 +13,8 @@ yp_constant_id_list_init(yp_constant_id_list_t *list) {
 bool
 yp_constant_id_list_append(yp_constant_id_list_t *list, yp_constant_id_t id) {
     if (list->size >= list->capacity) {
-        list->capacity = list->capacity == 0 ? 1 : list->capacity * 2;
-        list->ids = realloc(list->ids, sizeof(yp_constant_id_list_t) * list->capacity);
+        list->capacity = list->capacity == 0 ? 8 : list->capacity * 2;
+        list->ids = (yp_constant_id_t *) realloc(list->ids, sizeof(yp_constant_id_t) * list->capacity);
         if (list->ids == NULL) return false;
     }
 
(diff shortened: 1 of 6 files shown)


src/util/yp_constant_pool.c
```diff
@@ -13,7 +13,7 @@ yp_constant_id_list_init(yp_constant_id_list_t *list) {
 bool
 yp_constant_id_list_append(yp_constant_id_list_t *list, yp_constant_id_t id) {
     if (list->size >= list->capacity) {
-        list->capacity = list->capacity == 0 ? 1 : list->capacity * 2;
+        list->capacity = list->capacity == 0 ? 8 : list->capacity * 2;
         list->ids = realloc(list->ids, sizeof(yp_constant_id_list_t) * list->capacity);
```
eregon, line 17:
```suggestion
        list->ids = realloc(list->ids, sizeof(yp_constant_id_t) * list->capacity);
```
It would cause allocating more than necessary.
(and it's a bit sad a C compiler doesn't warn about that)

eregon:
BTW, I noticed https://github.com/ruby/yarp/blob/0e050a3c663bebf9f6d751ce03ed12d6374644a3/include/yarp/util/yp_constant_pool.h#L14
Maybe it should be `uint32_t`? That would go 8->4 bytes on non-Windows (on Windows 64-bit, both int and long are 4 bytes). And anyway serialization requires it to fit in `uint32_t`: https://github.com/ruby/yarp/blob/0e050a3c663bebf9f6d751ce03ed12d6374644a3/bin/templates/src/serialize.c.erb#L81

src/util/yp_constant_pool.c
kddnewton: Ugh, good catch. The compiler catches it if you cast, but not otherwise.

kddnewton:
@eregon yeah I've changed it to `uint32_t`

PR #975, opened 2023-06-14
Windows changes again

jemmaissroff:


test/availability/strncasecmp.c
```diff
@@ -1,6 +1,13 @@
 #include "yarp/defines.h"
+#ifndef _WIN32
 #include <strings.h>
+#endif
 
 int main() {
+#if defined(_WIN32)
+    return 1;
```
kddnewton, line 8: I'm assuming this means that `strncasecmp` is always available on windows no matter what?

kddnewton (approved):

test/availability/strncasecmp.c
jemmaissroff: I had done this wrong, fixed it now to only include `strings.h` if we're not in windows. `strings.h` and relatedly `strncasecmp` aren't in windows!

bin/templates/ext/yarp/api_node.c.erb
```diff
@@ -90,8 +92,9 @@ yp_node_new(yp_parser_t *parser, yp_node_t *node, rb_encoding *encoding, ID *con
     }
 }
 
+#line <%= __LINE__ + 1 %> "<%= File.basename(__FILE__) %>"
 VALUE yp_ast_new(yp_parser_t *parser, yp_node_t *node, rb_encoding *encoding) {
-    ID constants[parser->constant_pool.size];
+    ID * constants = calloc(parser->constant_pool.size, sizeof(ID));
```
kddnewton, line 97: Super surprised this can't be used, I thought this was a C99 feature?

kddnewton (approved):

bin/templates/ext/yarp/api_node.c.erb
jemmaissroff: I think on windows machines we can't dynamically allocate the size  of the array

PR #976, opened 2023-06-14
Remove more tokens from the tree

### LINKED ISSUE #809: CallNode needs to denote operator call type
Item for  #807.  CallNode.call_operatior is a token but we need to know whether the call is '.' or '&.'.

kddnewton:
There are a couple of changes in this, split out over the various commits. The overarching effort is to replace tokens in the tree in favor of locations. In service of that, we have the following commits:

1. Drop `token[]` in favor of `location[]`. This is exclusively used in the context of block-local variables. There's no need to store the extra type byte for all of these. This is the last usage of `token[]` in the list, but unfortunately it means we need to introduce `location[]`. At least this one is smaller and doesn't require knowledge of tokens.
2. Use `YP_EMPTY_NODE_LIST` instead of `yp_node_list_init`. This uses a macro instead of a function call. It's effectively the same thing, and the creation nodes end up being a little cleaner.
3. `InterpolatedSymbolNode#opening` -> `InterpolatedSymbolNode#opening_loc`, replaces the optional token with an optional location.
4. `SymbolNode#opening` -> `SymbolNode#opening_loc` and `SymbolNode#value` -> `SymbolNode#value_loc`. These always have the same kinds, so it's duplicate information.
5. `InterpolatedSymbolNode#closing` -> `InterpolatedSymbolNode#closing_loc`, replaces the optional token with an optional location.
6. `SymbolNode#closing -> SymbolNode#closing_loc`, same.
7. `CallNode#call_operator` -> `CallNode#operator_loc` + `CallNode#flags`. `call_operator` was previously a token, and the type was necessary to delineate between `.` and `&.`. Instead we're now keeping around a location for the operator and adding an additional set of flags to the `CallNode` that represent whether or not `&.` was used. This closes #809.
8. `CallNode#message` -> `CallNode#message_loc`, since we already have the actual method name as a string, this was just unnecessary information.
9. `CallNode#opening` -> `CallNode#opening_loc` and `CallNode#closing` -> `CallNode#closing_loc`. These were parentheses or brackets, but it's really not necessary for the consumer to know, and if they want to know they can reference the source.

The three remaining tokens in the tree are:

1. `GlobalVariableReadNode#name`
2. `GlobalVariableWriteNode#name`
3. `OperatorAssignmentNode#operator`

These require a little more thought. I think `GlobalVariableReadNode` should actually be split up into a couple of nodes because it can contain different types. It can be a back reference, and nth reference, or a global variable. We should have individual nodes for these, especially because only global variable can be set, the others can only be read.

For the operator assignment, we should have an enum for the operator so that folks don't have to reference the source.


jemmaissroff (approved):
Looks good to me, thanks!

kddnewton:
@jemmaissroff looking for your approval again since this completely changed since the last time you reviewed it 😅

jemmaissroff (approved):
Still lgtm!

PR #977, opened 2023-06-14
Refactor to remove parse_file_dup, use debug_mode build

jemmaissroff:
Instead of exposing parse_file_dup as a debugging utility, this commit adds a debug mode build that will perform the same checks without the additional complexity of parse_file_dup

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,6 +41,24 @@ jobs:
     - name: Run Ruby tests
       run: bundle exec rake compile_no_debug
 
+  build-debug-mode:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-latest, macos-latest ]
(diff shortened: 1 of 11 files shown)


.github/workflows/main.yml
```diff
@@ -41,6 +41,24 @@ jobs:
...
+    - uses: actions/checkout@v3
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: head
+        bundler-cache: true
+    - name: Run Ruby tests
+      run: bundle exec rake
+      env:
+        YARP_DEBUG: "1"
```
kddnewton, line 60: I think the general approach here is totally fine, but I'm unclear on `YP_DEBUG` versus `YARP_DEBUG` now. Could we either name this something different or make `YP_DEBUG` have other options?

kddnewton (approved):

PR #978, opened 2023-06-14
Refactor YARP.parse to optionally take a filepath

jemmaissroff:


ext/yarp/extension.c
```diff
@@ -294,15 +294,15 @@ parse_source(source_t *source, char *filepath) {
...
+parse(VALUE self, VALUE string, VALUE filepath) {
     source_t source;
     source_string_load(&source, string);
 #ifdef YARP_DEBUG_MODE_BUILD
     char* dup = malloc(source.size);
     memcpy(dup, source.source, source.size);
     source.source = dup;
 #endif
-    VALUE value = parse_source(&source, NULL);
+    VALUE value = parse_source(&source, NIL_P(filepath) ? NULL : StringValueCStr(filepath));
```
eregon, line 305: I think this might need `RB_GC_GUARD(filepath)`. Because otherwise there could be a GC, the register that used to hold filepath might be pointing to something else now, filepath is cleaned which involves freeings its char*, the char* passed here is now pointing to reclaimed memory (or worse, reused by something else).

https://github.com/ruby/yarp/discussions/836 would avoid this kind of shenanigans, thoughts? We need it for JRuby and it'd be faster for TruffleRuby, so maybe do it for CRuby too for simplicity and reliability?

kddnewton: I'm reluctant to use the dump/load approach for absolutely everything since it is slower. 

What would you think about checking at runtime, i.e.,

```ruby
if jruby? || truffleruby?
  require "yarp/yarp_ruby"
else
  require "yarp/yarp"
end
```

That way we could get the best of both worlds? Of course we'd test both paths on main as well.

PR #979, opened 2023-06-14
Introduce autotools

tenderlove:
This PR introduces autotools for detecting whether or not some
    functions exist.  The benefit of using autotools is that we don't need
    to write our own feature detection C files, and that we'll use the same
    `HAVE_*` macros that Ruby uses.  That means we can copy our C files in
    to Ruby and piggyback off Ruby's configure scripts when Ruby is built.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,0 +1,8 @@
+AC_INIT
+# Check that we have a C compiler
+AC_PROG_CC
+AC_DEFINE(_XOPEN_SOURCE, 700)
+AC_CHECK_FUNCS(mmap)
+AC_CHECK_FUNCS(strncasecmp)
+AC_CONFIG_FILES([Makefile])
+AC_OUTPUT
--- a/.gitignore
+++ b/.gitignore
@@ -32,4 +32,7 @@ test-native/run-one
 compile_commands.json
 .cache/
 
+configure
+Makefile
+
 tags
--- a/Makefile.in
+++ b/Makefile.in
@@ -1,29 +1,20 @@
 ifeq ($(shell uname), Darwin)
-SOEXT := dylib
+	SOEXT := dylib
 else
-SOEXT := so
+	SOEXT := so
 endif
 
-OPTFLAGS :=
-CFLAGS :=
-
-# Check for the presence of strncasecmp
-ifeq ($(shell $(CC) -Iinclude -o /dev/null test/availability/strncasecmp.c 2>/dev/null && echo 1), 1)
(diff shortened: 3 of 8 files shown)


kddnewton:
Thank you Aaron!

configure.ac
```diff
@@ -0,0 +1,8 @@
+AC_INIT
+# Check that we have a C compiler
+AC_PROG_CC
+AC_DEFINE(_XOPEN_SOURCE, 700)
```
eregon, line 4: Does this end up being defined when compiling yarp (in this repo, outside of ruby/ruby)?
I don't see anything that would include the file create by autotools, am I missing it?
It's done via `@DEFS@` maybe?

eregon:
We need to figure out how to get the proper `HAVE_` (HAVE_STRNCASECMP, HAVE_STRNSTR, etc) defined for JRuby & TruffleRuby.

That already didn't work before for TruffleRuby since we don't use the Makefile from the repo but let `mx+ninja` do everything, so it was just using the fallbacks.

configure.ac
eregon: Indeed, that's how it works:
```
log:
gcc -g -O2 -std=c99 -Wall -Werror -Wextra -Wpedantic -Wsign-conversion -fPIC	-fvisibility=hidden -DPACKAGE_NAME=\"\" -DPACKAGE_TARNAME=\"\" -DPACKAGE_VERSION=\"\" -DPACKAGE_STRING=\"\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -D_XOPEN_SOURCE=700 -DHAVE_MMAP=1 -DHAVE_STRNCASECMP=1 -Iinclude  -c -o src/enc/shift_jis.o src/enc/shift_jis.c
in Makefile:
DEFS := -DPACKAGE_NAME=\"\" -DPACKAGE_TARNAME=\"\" -DPACKAGE_VERSION=\"\" -DPACKAGE_STRING=\"\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -D_XOPEN_SOURCE=700 -DHAVE_MMAP=1 -DHAVE_STRNCASECMP=1
```

eregon:
This has the side effect to create `.o` files under `src/`. That's rather inconvenient to import the project, but I can workaround. Maybe there is an easy fix to avoid mixing sources and build results?

eregon:
Ah my bad, that's unrelated to this PR as @kddnewton told me.

PR #982, opened 2023-06-15
Lots of cleanup

kddnewton:
* Remove unused availability test now that we have autotools
* Removing profile.c and the make target since we have `YARP.profile_file`
* Move `yarp-lsp` into its own `exe` directory to match expected gem structure
* Move templates to the top level, it was weird under `bin`
* Move `bin/template.rb` to `templates/template.rb`
* Update the README.md with more accurate information and compilation instructions
* Update the target SHAs in the `rake lex` task
* Update the documentation to reflect latest
* Remove accidentally committed `yp_allocator.h`


jemmaissroff (approved):
LGTM, and glad we're doing this before mirroring

PR #984, opened 2023-06-15
Deploy RDoc to GitHub pages

kddnewton:


.github/workflows/github-pages.yml
```diff
@@ -0,0 +1,54 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.2"
```
jemmaissroff, line 33: Why 3.2 instead of head?

jemmaissroff (changes requested):

.github/workflows/github-pages.yml
kddnewton: Because the gems are cached, so the workflow will be faster. Otherwise with head every time head changes it busts the cache. Since we're only running rdoc I figured it was fine.

jemmaissroff (approved):

Issue #985, opened 2023-06-15
Fix StringNode location checks on windows


jemmaissroff:
Location checks for StringNodes are broken on windows machines. See [this build](https://github.com/jemmaissroff/ruby/actions/runs/5282067751/jobs/9556505725) for more information.

This looks like it has to do with carriage return characters

flavorjones:
I believe this problem is reproducible in CI vi #1001

kddnewton:
I think this is now fixed.

PR #986, opened 2023-06-15
Disable StringNode location checks on Windows machines

jemmaissroff:
Disabling these checks until we fix them. See #985 for more context.


test/yarp_test_helper.rb
```diff
@@ -58,18 +58,19 @@ def assert_equal_nodes(expected, actual, compare_location: true, parent: nil)
...
-            actual.start_offset,
-            -> { "Start locations were different. Parent: #{parent.pretty_inspect}" }
-          )
-
-          assert_equal(
-            expected.end_offset,
-            actual.end_offset,
-            -> { "End locations were different. Parent: #{parent.pretty_inspect}" }
-          )
+          if !(parent.is_a?(YARP::StringNode) && (RbConfig::CONFIG['host_os'] =~ /mswin|mingw32|windows/))
```
kddnewton, line 61: This looks good but would you mind moving `RbConfig::CONFIG['host_os'] =~ /mswin|mingw32|windows/)` into a constant? This function gets called a ton in the test suite.

PR #989, opened 2023-06-15
Allow for NULL StatementsNode

jemmaissroff:
Previously, we were always returning `StatementsNodes`, even if they were empty, which meant the nodes themselves and their locations were incorrect. This fix allows us to return `NULL` `StatementNodes` which means both the nodes and their locations are correct


kddnewton (approved):
Great work!

PR #990, opened 2023-06-15
Use demand paging on windows, and do not call mmap or malloc for an empty file

kddnewton:
And a couple of other small fixes (don't leak file descriptor and a little more documentation).


jemmaissroff (changes requested):
Will do a full review once I get it passing on ruby/ruby, the `file_size` typo is just the first thing the build broke on

PR #997, opened 2023-06-16
Split up operator assignment

kddnewton:
This is a relatively big change in order to make operator assignments easier to compile.

Right now in order to tell what is actually happening within an `OperatorAssignmentNode`, `OperatorAndAssignmentNode`, and `OperatorOrAssignmentNode`, you have to descend into the target node and figure out its type. This is not great. Furthermore for the `OperatorAssignmentNode` you have to switch on the token type of the operator to figure out the operation to perform.

To remedy all of this, all three of these nodes are getting split up, and the various operator assignment nodes are getting a constant id that contains the operation to perform. So, this means we now have:

* `CallOperatorAndWriteNode` - `foo.bar &&= value`
* `CallOperatorOrWriteNode` - `foo.bar ||= value`
* `CallOperatorWriteNode` - `foo.bar += baz`
* `ClassVariableOperatorAndWriteNode` - `@@target &&= value`
* `ClassVariableOperatorOrWriteNode` - `@@target ||= value`
* `ClassVariableOperatorWriteNode` - `@@target += value`
* `ConstantOperatorAndWriteNode` - `Target &&= value`
* `ConstantOperatorOrWriteNode` - `Target ||= value`
* `ConstantOperatorWriteNode` - `Target += value`
* `ConstantPathOperatorAndWriteNode` - `Parent::Child &&= value`
* `ConstantPathOperatorOrWriteNode` - `Parent::Child ||= value`
* `ConstantPathOperatorWriteNode` - `Parent::Child += value`
* `GlobalVariableOperatorAndWriteNode` - `$target &&= value`
* `GlobalVariableOperatorOrWriteNode` - `$target ||= value`
* `GlobalVariableOperatorWriteNode` - `$target += value`
* `InstanceVariableOperatorAndWriteNode` - `@target &&= value`
* `InstanceVariableOperatorOrWriteNode` - `@target ||= value`
* `InstanceVariableOperatorWriteNode` - `@target += value`
* `LocalVariableOperatorAndWriteNode` - `target &&= value`
* `LocalVariableOperatorOrWriteNode` - `target ||= value`
* `LocalVariableOperatorWriteNode` - `target += value`

The tree ends up being smaller now, because there are fewer nodes that end up staying in the tree. Previously `target += value` would have been represented as `OperatorAssignmentNode(LocalVariableWriteNode, CallNode)` but is now just `LocalVariableOperatorWriteNode(CallNode)`. Hopefully this makes it much easier to compile.

This also is the last token on the tree, so this means none of the serialization consumers need to know about the token types!

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -594,6 +594,50 @@ nodes:
 
           foo&.bar
           ^^^^^^^^
+  - name: CallOperatorAndWriteNode
+    child_nodes:
+      - name: target
+        type: node
+        kind: CallNode
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+    comment: |
+      Represents the use of the `&&=` operator on a call.
+
+          foo.bar &&= value
+          ^^^^^^^^^^^^^^^^^
+  - name: CallOperatorOrWriteNode
+    child_nodes:
+      - name: target
+        type: node
+        kind: CallNode
+      - name: value
+        type: node
+      - name: operator_loc
+        type: location
+    comment: |
+      Represents the use of the `||=` operator on a call.
+
+          foo.bar ||= value
+          ^^^^^^^^^^^^^^^^^
+  - name: CallOperatorWriteNode
+    child_nodes:
+      - name: target
+        type: node
+        kind: CallNode
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+      - name: operator_id
+        type: constant
+    comment: |
+      Represents the use of an assignment operator on a call.
+
+          foo.bar += baz
+          ^^^^^^^^^^^^^^
   - name: CapturePatternNode
     child_nodes:
       - name: value
@@ -648,6 +692,47 @@ nodes:
 
           class Foo end
           ^^^^^^^^^^^^^
+  - name: ClassVariableOperatorAndWriteNode
+    child_nodes:
+      - name: name_loc
+        type: location
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+    comment: |
+      Represents the use of the `&&=` operator for assignment to a class variable.
+
+          @@target &&= value
+          ^^^^^^^^^^^^^^^^
+  - name: ClassVariableOperatorOrWriteNode
+    child_nodes:
+      - name: name_loc
+        type: location
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+    comment: |
+      Represents the use of the `||=` operator for assignment to a class variable.
+
+          @@target ||= value
+          ^^^^^^^^^^^^^^^^^^
+  - name: ClassVariableOperatorWriteNode
+    child_nodes:
+      - name: name_loc
+        type: location
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+      - name: operator
+        type: constant
+    comment: |
+      Represents assigning to a class variable using an operator that isn't `=`.
+
+          @@target += value
+          ^^^^^^^^^^^^^^^^^
   - name: ClassVariableReadNode
     comment: |
       Represents referencing a class variable.
@@ -667,6 +752,47 @@ nodes:
 
           @@foo = 1
           ^^^^^^^^^
+  - name: ConstantOperatorAndWriteNode
+    child_nodes:
+      - name: name_loc
+        type: location
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+    comment: |
+      Represents the use of the `&&=` operator for assignment to a constant.
+
+          Target &&= value
+          ^^^^^^^^^^^^^^^^
+  - name: ConstantOperatorOrWriteNode
+    child_nodes:
+      - name: name_loc
+        type: location
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+    comment: |
+      Represents the use of the `||=` operator for assignment to a constant.
+
+          Target ||= value
+          ^^^^^^^^^^^^^^^^
+  - name: ConstantOperatorWriteNode
+    child_nodes:
+      - name: name_loc
+        type: location
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+      - name: operator
+        type: constant
+    comment: |
+      Represents assigning to a constant using an operator that isn't `=`.
+
+          Target += value
+          ^^^^^^^^^^^^^^^
   - name: ConstantPathNode
     child_nodes:
       - name: parent
@@ -680,6 +806,50 @@ nodes:
 
           Foo::Bar
           ^^^^^^^^
+  - name: ConstantPathOperatorAndWriteNode
+    child_nodes:
+      - name: target
+        type: node
+        kind: ConstantPathNode
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+    comment: |
+      Represents the use of the `&&=` operator for assignment to a constant path.
+
+          Parent::Child &&= value
+          ^^^^^^^^^^^^^^^^^^^^^^^
+  - name: ConstantPathOperatorOrWriteNode
+    child_nodes:
+      - name: target
+        type: node
+        kind: ConstantPathNode
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+    comment: |
+      Represents the use of the `||=` operator for assignment to a constant path.
+
+          Parent::Child ||= value
+          ^^^^^^^^^^^^^^^^^^^^^^^
+  - name: ConstantPathOperatorWriteNode
+    child_nodes:
+      - name: target
+        type: node
+        kind: ConstantPathNode
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+      - name: operator
+        type: constant
+    comment: |
+      Represents assigning to a constant path using an operator that isn't `=`.
+
+          Parent::Child += value
+          ^^^^^^^^^^^^^^^^^^^^^^
   - name: ConstantPathWriteNode
     child_nodes:
       - name: target
@@ -864,6 +1034,47 @@ nodes:
 
           super
           ^^^^^
+  - name: GlobalVariableOperatorAndWriteNode
+    child_nodes:
+      - name: name_loc
+        type: location
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+    comment: |
+      Represents the use of the `&&=` operator for assignment to a global variable.
+
+          $target &&= value
+          ^^^^^^^^^^^^^^^^^
+  - name: GlobalVariableOperatorOrWriteNode
+    child_nodes:
+      - name: name_loc
+        type: location
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+    comment: |
+      Represents the use of the `||=` operator for assignment to a global variable.
+
+          $target ||= value
+          ^^^^^^^^^^^^^^^^^
+  - name: GlobalVariableOperatorWriteNode
+    child_nodes:
+      - name: name_loc
+        type: location
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+      - name: operator
+        type: constant
+    comment: |
+      Represents assigning to a global variable using an operator that isn't `=`.
+
+          $target += value
+          ^^^^^^^^^^^^^^^^
   - name: GlobalVariableReadNode
     comment: |
       Represents referencing a global variable.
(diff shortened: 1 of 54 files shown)


jemmaissroff (approved):
🚀 great cleanup

eregon:
> This also is the last token on the tree, so this means none of the serialization consumers need to know about the token types!

Great!

> To remedy all of this, all three of these nodes are getting split up

Mmh, that's a lot of nodes, I wonder if it's better or not. I think it may be worse.
For instance it makes it harder to share logic for `OperatorOrAssignmentNode` (`||=`), which could otherwise be fully expressed through a generic read node and a generic write node.

For example this is how JRuby AST represents it:
```
$ ruby tool/parse_ast.rb '@a ||= 2+3'   
Source:
@a ||= 2+3

AST:
RootNode line: 0
  OpAsgnOrNode* line: 0
    InstVarNode:@a line: 0
    InstAsgnNode:@a line: 0
      OperatorCallNode*:+ line: 0
        FixnumNode line: 0, long: 2
        ArrayNode line: 0
          FixnumNode line: 0, long: 3
```

And the semantics are expressed by a single node for all `||=` in TruffleRuby, `OrLazyValueDefinedNode`, which is very simple, execute the first child, if truthy return it, otherwise execute the second child, done.

eregon:
I think the same applies for `&&=` and other operators, i.e., the logic and compilation can be shared regardless of the variable type. But probably not so easily when there is a method call on LHS.

&&=
```
Source:
@a &&= 2+3

AST:
RootNode line: 0
  OpAsgnAndNode* line: 0
    InstVarNode:@a line: 0
    InstAsgnNode:@a line: 0
      OperatorCallNode*:+ line: 0
        FixnumNode line: 0, long: 2
        ArrayNode line: 0
          FixnumNode line: 0, long: 3
```

+= on local var
```
Source:
@a += 2+3

AST:
RootNode line: 0
  InstAsgnNode*:@a line: 0
    OperatorCallNode*:+ line: 0
      InstVarNode:@a line: 0
      ArrayNode line: 0
        OperatorCallNode*:+ line: 0
          FixnumNode line: 0, long: 2
          ArrayNode line: 0
            FixnumNode line: 0, long: 3
```
This is fully desugared in JRuby AST, quite convenient for compilation, but maybe not for tooling.

+= with call:
```
Source:
foo.a += 42

AST:
RootNode line: 0
  OpAsgnNode* line: 0
    VCallNode*:foo line: 0
    FixnumNode line: 0, long: 42
```
That one is treated specially in JRuby AST, and if `foo` is a method call it must not execute it twice.
To desugar it one would need a temporary variable to store the receiver (like local variables but not visible to Ruby).

I would suggest to merge ||=/&&=/+= for all variable types, and only have the 3 CallOperator* because those cannot really be done easily otherwise.

eregon:
cc @enebo WDYT?

eregon:
FWIW in the parser gem:
```
$ ruby -rparser/current -e 'pp Parser::CurrentRuby.parse("@foo ||= 1")'     
s(:or_asgn,
  s(:ivasgn, :@foo),
  s(:int, 1))
$ ruby -rparser/current -e 'pp Parser::CurrentRuby.parse("foo ||= 1")' 
s(:or_asgn,
  s(:lvasgn, :foo),
  s(:int, 1))
$ ruby -rparser/current -e 'pp Parser::CurrentRuby.parse("foo.bar ||= 1")'
s(:or_asgn,
  s(:send,
    s(:send, nil, :foo), :bar),
  s(:int, 1))
$ ruby -rparser/current -e 'pp Parser::CurrentRuby.parse("@a += 1")'      
s(:op_asgn,
  s(:ivasgn, :@a), :+,
  s(:int, 1))
$ ruby -rparser/current -e 'pp Parser::CurrentRuby.parse("@a &&= 1")'
s(:and_asgn,
  s(:ivasgn, :@a),
  s(:int, 1))
```
which is similar to before this PR.
I think having the read node too in the AST is very helpful.

eregon:
> That one is treated specially in JRuby AST, and if `foo` is a method call it must not execute it twice.
> To desugar it one would need a temporary variable to store the receiver (like local variables but not visible to Ruby).

Another way could be to only give a receiver to the "read" node, and no receiver for the write node:
```
foo.a += 42
OperatorAssignmentNode(
        readNode = CallNode(0...5)(
           receiver=CallNode(0...3)(nil, nil, IDENTIFIER(0...3)("foo"), nil, nil, nil, nil, "foo"),
           method="a",
           args=[]
        ),
        writeNode = CallNode(
          receiver=null, // to not have the receiver tree twice and to not evaluate it twice (would be incorrect)
          method="a=",
          args=[] // the resulting value will be added as an argument implicitly
        )
        operator="+",
        value=IntegerNode(9...11)()
       )  
```

That seems the cleanest to me, then we only need OperatorAndAssignmentNode, OperatorOrAssignmentNode, OperatorAssignmentNode and they should be easy to compile/translate and also easy to analyse for semantics, with minimal duplication in the handling of these nodes.

kddnewton:
Thanks for taking the time to write these options out. I have a lot of thoughts on this and have been going back and forth with lots of folks over the last couple of days trying to come to a conclusion. Here's where I've landed:

In general with this tree we've tried to make it so that every node has its own semantics. We've split stuff up specifically because of this before - in previous iterations of the tree `FalseNode`, `SelfNode`, `TrueNode`, and `NilNode` were all part of the same `KeywordNode`. Previously we also split `AssignNode` into `LocalVariableWriteNode`, `InstanceVariableWriteNode`, etc. The only time we've unified nodes has been when they represent the exact same semantics, as in all of the `CallNode` variants, heredocs and string literals being the same semantic thing, if and elsif being the same semantic thing, etc.

These nodes are all different semantically - though I grant you that they have similar patterns of execution (read something, do something, write something). But an `OperatorAssignmentNode` has no actual semantic meaning on its own. It would be the first node in the tree that doesn't actually mean anything unless you go and look at child nodes. So this feels very inconsistent to me. Especially because we split up the other write nodes into their own types.

There are other benefits as well - consuming libraries will be able to have fewer nested switches, which already makes it hard for compilers and various tools like semantic highlighting to function without extra unnecessary steps.

I'm sorry if this makes it more difficult for TruffleRuby to share logic. Maybe we could put something into the serialization that marks some of these nodes with interfaces such that you could treat them all the same? I think marker interfaces in Java and marker modules in Ruby could be a nice way to provide extra information to the consumers of both.

eregon:
> These nodes are all different semantically - though I grant you that they have similar patterns of execution (read something, do something, write something). But an `OperatorAssignmentNode` has no actual semantic meaning on its own.

I think `OperatorAssignmentNode` has actual semantics, just as much as IfNode.
Let's compare:
`IfNode(condition, foo, a * 2)` is `if condition; foo; else a * 2; end`.
`IfNode(condition, foo, @a * 2)` is `if condition; foo; else @a * 2; end`.
`OperatorAssignmentNode(op=+, read=LocalVariableReadNode(a), write=LocalVariableWriteNode(a), value=2)` is `a = a + 2`.
`OperatorAssignmentNode(op=+, read=InstanceVariableReadNode(@a), write=InstanceVariableWriteNode(@a), value=2)` is `@a = @a + 2`.
(for the special case of a non-variable/constant receiver, see the previous comment, that's also handled the same way)

In both cases, the semantics are expressed by the node type (IfNode/OperatorAssignmentNode), the child nodes can be completely generic and **do not** need to be looked at to compile OperatorAssignmentNode.
The proof is the JRuby AST does not split such assignment node per child node/variable type, and the compilation (at least in TruffleRuby, probably on JRuby too) does not switch on it either.
See https://github.com/oracle/truffleruby/blob/0e633b8d014fe8cd328d11028b86b8af44b46fe9/src/main/java/org/truffleruby/language/control/OrLazyValueDefinedNode.java
https://github.com/oracle/truffleruby/blob/0e633b8d014fe8cd328d11028b86b8af44b46fe9/src/main/java/org/truffleruby/parser/BodyTranslator.java#L2333-L2343
https://github.com/oracle/truffleruby/blob/0e633b8d014fe8cd328d11028b86b8af44b46fe9/src/main/java/org/truffleruby/parser/BodyTranslator.java#L2319-L2331
https://github.com/oracle/truffleruby/blob/master/src/main/java/org/truffleruby/parser/ast/OpAsgnOrParseNode.java#L39
Does the CRuby compilation really need to switch on the variable type, do you have a link to that? It seems needless code duplication to me if that's the case.
In fact looking at instructions it's pretty obvious there is no need to split e.g. `LocalVariableOperatorOrWriteNode` and `InstanceVariableOperatorOrWriteNode`, the bytecodes are the same, and the logic to emit can be fully generic, at the very least for all variable types:
```
$ ruby --dump=insns -e 'foo ||= 1'
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,9)> (catch: FALSE)
local table (size: 1, argc: 0 [opts: 0, rest: -1, post: 0, block: -1, kw: -1@-1, kwrest: -1])
[ 1] foo@0
0000 getlocal_WC_0                          foo@0                     (   1)[Li]
0002 dup
0003 branchif                               10
0005 pop
0006 putobject_INT2FIX_1_
0007 dup
0008 setlocal_WC_0                          foo@0
0010 leave
$ ruby --dump=insns -e '@foo ||= 1'
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,10)> (catch: FALSE)
0000 getinstancevariable                    :@foo, <is:0>             (   1)[Li]
0003 dup
0004 branchif                               12
0006 pop
0007 putobject_INT2FIX_1_
0008 dup
0009 setinstancevariable                    :@foo, <is:0>
0012 leave
```
i.e. it's literally:
```
0000 READ
0003 dup
0004 branchif                               12
0006 pop
0007 VALUE
0008 dup
0009 WRITE
0012 leave
```

And the same thing for `foo += 1` vs `@foo += 1` (OperatorAssignmentNode).

The only thing compiling OperatorAssignmentNode needs is: the operator name, a node to read the receiver, a node to write to the receiver and a generic value node.
Just like The only thing compiling IfNode needs is: a condition node to evaluate, a generic thenBody node, a generic elseBody node.

And this is even clearer for OperatorOrAssignmentNode as shown in https://github.com/ruby/yarp/pull/997#issuecomment-1597044892.
`OperatorOrAssignmentNode(read=LocalVariableReadNode(a), write=LocalVariableWriteNode(a), value = 42)` is `a ||= 42` which is `a || a = 42` and generically `defined?(receiver) || receiver = value`.
One might think the `defined?` there is easier to handle with `LocalVariableOperatorOrWriteNode`, but not really, one needs to implement `defined?` for any node in Ruby anyway. Some nodes are always `defined?`, e.g., local variable reads, and so as an optimization the `defined?` can be omitted. JRuby AST tracks that via [needsDefinitionCheck()](https://github.com/oracle/truffleruby/blob/0e633b8d014fe8cd328d11028b86b8af44b46fe9/src/main/java/org/truffleruby/parser/ast/ParseNode.java#L239-L244). The full logic is [a few lines of code](https://github.com/oracle/truffleruby/blob/0e633b8d014fe8cd328d11028b86b8af44b46fe9/src/main/java/org/truffleruby/parser/BodyTranslator.java#L2319-L2331).

> consuming libraries will be able to have fewer nested switches

Why would they need nested switches? Compilers don't need it (except maybe for `needsDefinitionCheck()` if they choose to implement it that way, but then it can just be a separate function for that). Syntax highlighters don't need it, in fact they will have to handle all these extra node types while they wouldn't before. Semantics analyzers will likely have more work to deal with all these.

> I think marker interfaces in Java and marker modules in Ruby could be a nice way to provide extra information to the consumers of both.

This would still be a problem, because it means e.g. I need to duplicate the logic for local variable read & write to handle LocalVariableOperatorWriteNode, and for instance variable read and write for InstanceVariableOperatorWriteNode, etc. Or I need to create `OperatorAssignmentNode` with the proper read node and write node myself which is then a non-trivial overhead.

I see only one disadvantage of OperatorAssignmentNode, we can't specify the true receiver for both the read and write nodes in https://github.com/ruby/yarp/pull/997#issuecomment-1600563384. But that is a typical problem for write nodes, it's similar for multiple assignment where we cannot give the value node for `foo, bar = baz`. Some amount of that (maybe all? unsure) could probably be solved if YARP has "temporary variables" and desugars more:
`foo.a += 42` would become `_tmp_receiver = foo; _tmp_receiver.a = _tmp_receiver.a + 42`.
For compilation and semantic analysis it would be awesome and simplify a lot.
Though for syntax highlighting that would need to be ignored and could make things harder.

I think it's  mistake to split these per variable type. It looks [very messy in the docs](https://ruby.github.io/yarp/) (takes a lot of vertical space when listing nodes and the names are extremely long and hard to read), it blows up the number of node types which hurts performance (https://github.com/ruby/yarp/issues/1099#issuecomment-1614512917), it's harder to process for everyone (from the output of compiling these nodes, and the fact JRuby AST and parser gem AST unify those), or at least not easier and also causes duplication. It also stands out clearly from all other nodes because it's a matrix combination of nodes.
We do need LocalVariableReadNode/LocalVariableWriteNode because that compiles very differently than e.g. instance variable read/write. But OperatorAssignmentNode, OperatorOrAssignmentNode, OperatorAndAssignmentNode compile exactly the same for local/instance/global variables and constants and very similarly for operator methods (need to care about evaluating the receiver only once, could be fully unified with temporary variables).

tenderlove:
> `OperatorOrAssignmentNode(read=LocalVariableReadNode(a), write=LocalVariableWriteNode(a), value = 42)` is `a ||= 42` which is `a || a = 42`

Interesting.  I never consider `a ||= 42` to be `a || a = 42`, rather `a = a || 42` which I think would eliminate the need for specialized assignment nodes, no?

>  and generically `defined?(receiver) || receiver = value`.

I don't think this is accurate because `receiver` could be `false` or `nil` (`a = false; a ||= 42`, `a` is defined but not true).

> I think it's mistake to split these per variable type.

Aren't we doing this with other nodes?

eregon:
> Interesting. I never consider `a ||= 42` to be `a || a = 42`, rather `a = a || 42` which I think would eliminate the need for specialized assignment nodes, no?

Indeed, if YARP desugars `||=`, `&&=` and `+=` then there is no need for many of these `Operator*WriteNode`.
Probably only the `CallOperator*` and `ConstantPath*` are needed then (to avoid evaluating the receiver twice, could be addressed with temporary variables).
I think that could be a nice solution to avoid node types explosion.
For semantics it's great to desugar, for syntactical analysis probably less but maybe still OK?

The original `Operator*AssignmentNode` are IMO the best, because they are easy to deal with for semantics, for syntax, they are clear and don't create 21 node types when only 3 are needed.

> > and generically `defined?(receiver) || receiver = value`.
> 
> I don't think this is accurate because `receiver` could be `false` or `nil` (`a = false; a ||= 42`, `a` is defined but not true).

Right, it's `(defined?(receiver) && receiver) || receiver = value`, I'll fix it above as well.

> > I think it's mistake to split these per variable type.
> 
> Aren't we doing this with other nodes?

AFAIK no, the only per-variable-type nodes we have are simple read and write nodes, which we do need per variable type.
The parser gem and JRuby parser both only have simple read and write nodes per variable type, I think we should follow them here.

PR #1000, opened 2023-06-17
Add Windows CI

MSP-Greg:
Adds a GitHub Actions workflow that only runs `bundle exec rake compile`.  It has code to always pass, so it won't 'fail' a PR.

Several errors/warnings are shown, some were fixed in #966.

Notes:
1. This is only using a Ruby built with the MSYS2 UCRT64 gcc tool chain.  I believe additional code is needed to even try to build with a mswin (MSVC) Ruby.
2. First time this runs, installing/compiling Nokogiri takes several minutes.  Subsequent builds can make use of the cache, so they're much quicker.
3. Once it compiles, addtional code is needed to load `librubyparser.dll`.  Something similar to the below:
```ruby
require 'fiddle'
kernel32 = Fiddle.dlopen 'kernel32'
load_library = Fiddle::Function.new(
  kernel32['LoadLibraryW'], [Fiddle::TYPE_VOIDP], Fiddle::TYPE_INT,
)
dll_path = File.absolute_path('../build/librubyparser.dll', __dir__)
if load_library.call(dll_path.encode('utf-16le')).zero?
  abort "Failed to load librubyparser.dll from\n  #{dll_path}"
end
```

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,0 +1,39 @@
+name: Windows
+
+concurrency:
+  group: "${{github.workflow}}-${{github.ref}}"
+  cancel-in-progress: true
+
+on:
+  push:
+    #branches:
+    #  - main
+  pull_request:
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ windows-latest ]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Ruby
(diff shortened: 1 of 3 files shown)


flavorjones:
FWIW I also have a branch in progress with windows support that is also going to introduce static linking of the ruby C extension.

kddnewton:
Thank you so much for the help @MSP-Greg! I see that Mike credited you on some of the commits he added recently. I'm going to close this one in favor of that other one.

PR #1002, opened 2023-06-18
Add ExcludeEnd flag to RangeNode

FnControlOption:
Discussed in #996:

> Currently, the only way to tell the difference between an inclusive range `a..b` and an exclusive range `a...b` is to check whether `operator_loc.end - operator_loc.start` is either two or three.

This PR adds a new `flags` field to `RangeNode` and a `RangeNodeFlags` enum to go with it. The `ExcludeEnd` flag is set if the range excludes its end value.


kddnewton (approved):
Yeah this looks good. I like what you did with the comment field

PR #1017, opened 2023-06-20
Revert "Partially fix `rake test` on Mac"

jemmaissroff:
This reverts commit 7cc29c2d1230d6b503ebd6739431f58468988afb.

This commit was unnecessary, and breaks builds on ruby/ruby. I think this commit is necessary if using an old ruby version. But if running `rake` with latest ruby, it's not necessary to have this config suffix and `.so` should suffice. Tested on mac


kddnewton (approved):

eregon:
Right, Ruby's `require` automatically declines `.so` into the right extension when searching the filesystem.

PR #1023, opened 2023-06-20
Fix location on BeginNode

jemmaissroff:
Prior to this commit, the location was incorrect whenever we called `parse_rescues_as_begin` because it was being set to the empty begin_token. This commit sets it to the first non-null child of the begin node (could be statements, rescue clause, else clause or ensure clause)


kddnewton (approved):
Looks great!

PR #1027, opened 2023-06-21
Final fixes to mirror to ruby/ruby

jemmaissroff:


kddnewton (approved):

include/yarp/util/yp_newline_list.h
```diff
@@ -14,6 +14,8 @@
 #include <stdbool.h>
 #include <stdlib.h>
 
+#include "yarp/defines.h"
```
eregon, line 17: `#include "yarp/defines.h"` should always be before includes, because it affects which functions/macros these other includes define (same above).
Like in `include/yarp/enc/yp_encoding.h` for instance

eregon: For example `strnlen` is only defined on Linux if:
```
       strnlen():
           Since glibc 2.10:
               _POSIX_C_SOURCE >= 200809L
```
So if `./include/yarp/defines.h` was included before `#include <string.h>` (which is actually problematic in the file above).

eregon: Right, although it still seems best to have `defines` first, just like CRuby always includes `#include "ruby/internal/config.h"` first.

PR #1028, opened 2023-06-21
Updates to polyfills

kddnewton:
* `yp_strnstr` was never being replaced by a system function, so splitting it out into its own util file to be more clear.
* Moving `yp_strncasecmp` into its own file, but putting its redefinition into `defines.h` to make it first.
* Adding `yp_snprintf` as a polyfill for `snprintf` on systems that don't have it, putting its redefinition into `defines.h`.

`snprintf` was previously the only warning generated when compiling with clang so that's gone now.


jemmaissroff (changes requested):
Small nit. Also: I wonder if we should check this on a branch off of the `mirror-yarp` branch before merging it just to double check we won't break a build there?

jemmaissroff (approved):

include/yarp/util/yp_strnstr.h
```diff
@@ -0,0 +1,14 @@
...
+#define YP_STRNSTR_H
+
+#include "yarp/defines.h"
+
+#include <stddef.h>
+#include <string.h>
+
+// Here we have rolled our own version of strnstr because for the most part it
+// is not available unless on BSD, and even then -std=c99 sometimes does not
+// include it.
```
eregon, line 11: It's available on Linux and macOS, so maybe the comment should just be clear it's really just indows missing it?

eregon: Ah sorry, this is `strnstr`, not `snprintf`, nevermind

src/util/yp_snprintf.c
```diff
@@ -0,0 +1,12 @@
+#include "yarp/defines.h"
+
+// In case snprintf isn't present on the system, we provide our own that simply
+// forwards to the less-safe sprintf.
+int
+yp_snprintf(char *dest, YP_ATTRIBUTE_UNUSED size_t size, const char *format, ...) {
```
eregon, line 6: FWIW it seems snprintf might be defined on Windows, at least UCRT: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-snprintf-snprintf-l-snwprintf-snwprintf-l?view=msvc-170
But maybe not on older Windows versions?
From https://stackoverflow.com/questions/9052224/error4error-c3861-snprintf-identifier-not-found#comment11358163_9052232 snprintf is standard in C99, so Windows should have it since CRuby requires C99 nowadays (AFAIK)

eregon (approved):

nobu:
If `yp_snprintf` just ignores the `size`, why not define a variadic macro?

```C
# define yp_snprintf(dest, size, ...) sprintf((dest), __VA_ARGS__)
```

PR #1036, opened 2023-06-21
Makefile: fix object dependencies on headers

flavorjones:
Two issues being fixed in this PR:

- headers are not located in `src`, they're in `include`
- the objects' dependencies need to be declared with `$(OBJECTS):` not `OBJECTS:`

So now, changes to header files will actually cause a recompilation.


kddnewton:
Thank you!!

kddnewton (approved):

PR #1037, opened 2023-06-21
Fix snapshot checking

kddnewton:


test/parse_test.rb
```diff
@@ -75,12 +75,15 @@ def test_parse_takes_file_path
       assert_empty result.errors, value
 
       if File.exist?(snapshot)
-        expected = File.read(snapshot)
-        normalized = normalize_printed(expected)
-        if expected != normalized
+        normalized = normalize_printed(File.binread(snapshot))
```
flavorjones, line 78: Why binread instead of plain old read?

kddnewton: Because `normalize_printed` is going to call `.b` anyway so this just saves a step. Also because I'm nervous about CLRF. Though maybe I should have used `File.read(filepath, binmode: true, external_encoding: Encoding::UTF_8)` to be consistent.

PR #1049, opened 2023-06-22
Use serialization for snapshots

kddnewton:
This gets rid of the `LANG` issue and makes it so that we don't have to deal with encoding problems.

To get the nicer snapshot diffs, run:

```bash
git config diff.snapshot.textconv bin/deserialize
git config diff.snapshot.binary true
```


jemmaissroff (approved):
One small thing we could do is make a helper script to get nicer git diffs. But since it's only two lines, definitely don't think that's necessary

kddnewton:
Yeah I was thinking it might be good to have a `bin/setup` script for installing dependencies and other stuff. We could add this stuff to that.

eregon:
This means it's no longer possible to review the snapshots diff on GitHub.
What's this `LANG` issue? Wouldn't it be enough to dump the snapshots as UTF-8 or binary to solve the issue?

PR #1050, opened 2023-06-22
Add appropriate cases to follow `::` syntax

jemmaissroff:


kddnewton:
This looks good, but I'm confused by a couple of the others. I see for example you have `LOWER_I` but I don't know how you could get into that situation?

jemmaissroff:
> This looks good, but I'm confused by a couple of the others. I see for example you have `LOWER_I` but I don't know how you could get into that situation?

You're right, removed all of the extra ones. Unsure why the ubuntu builds are failing. Will look into them later today

kddnewton (approved):

PR #1051, opened 2023-06-22
Generate `configure-macros`

### LINKED ISSUE #1041: Figure out how to set the HAVE_ macros when using YARP in JRuby & TruffleRuby
See https://github.com/ruby/yarp/pull/979#issuecomment-1602495850
cc @enebo

kddnewton:
Fixes #1041 

We now use a `configure-macros.in` file that generates a `configure-macros` text file containing the definitions generated by autotools. On my machine that means:

```
-DPACKAGE_NAME=\"YARP\" -DPACKAGE_TARNAME=\"yarp\" -DPACKAGE_VERSION=\"0.4.0\" -DPACKAGE_STRING=\"YARP\ 0.4.0\" -DPACKAGE_BUGREPORT=\"https://github.com/ruby/yarp/issues/new\" -DPACKAGE_URL=\"https://github.com/ruby/yarp\" -DYP_VERSION_MAJOR=0 -DYP_VERSION_MINOR=4 -DYP_VERSION_PATCH=0 -D_XOPEN_SOURCE=700 -DHAVE_MMAP=1 -DHAVE_STRNCASECMP=1
```

You can get this by running `autoconf && ./configure`.

This PR also centralizes _some_ stuff around versioning. For example we now reference `PACKAGE_VERSION` to try to minimize the places we're stating the version numbers.


eregon:
I'll take a look next week how we can use this in TruffleRuby's build process.

eregon:
This is turning out to be quite difficult from a quick look.
Currently, TruffleRuby only imports `include/` and `src/` (and `java/`): https://github.com/oracle/truffleruby/blob/master/tool/import-yarp.sh
And `mx` generates a ninja file to compile everything, and also compile the `yarp_bindings.c` JNI file all as one `.so`.

But the autotools stuff is at the root, so it won't be copied and there is no way to run in the TruffleRuby repo while only importing those. We could import more, but it starts to feel more brittle. We could import the whole repo, but that seems a lot of stuff not needed.
I suppose we could import `configure*` or so, and hope that doesn't need other files.

eregon:
Generating a header seems nicer and more consistent, I tried that in https://github.com/ruby/yarp/pull/1088

eregon:
https://github.com/ruby/yarp/pull/1088 works so I'll close this.
TruffleRuby can now use YARP's Makefile, use the static library, and then creates a dynamic library with the JNI bindings + librubyparser.a.

PR #1055, opened 2023-06-23
Let git ignore snapshots of fixtures

nobu:


.gitignore
```diff
@@ -29,6 +29,7 @@ test.rb
 /src/serialize.c
 /src/token_type.c
 /src/**/*.o
+/test/snapshots/fixtures/
```
jemmaissroff, line 32: Hmmm, I don't think there's anything in the `/test/snapshots/fixtures/` dir. Did you mean the `/test/snapshots/` dir?

jemmaissroff (changes requested):
@nobu can you tell me more about the goal in this PR? I find looking at the git diff useful as a reviewer.

.gitignore
nobu: Seems generated by a mistake at #1054.

PR #1074, opened 2023-06-26
Fix a potential leak in `parse_block_parameters`

haldun:
Fixes a potential leak in `parse_block_parameters` where `yp_block_parameters_node_create` fails to allocate memory.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -8544,6 +8544,12 @@ parse_block_parameters(
     }
 
     yp_block_parameters_node_t *block_parameters = yp_block_parameters_node_create(parser, parameters, opening);
+
+    if (!block_parameters && parameters) {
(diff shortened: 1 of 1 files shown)


kddnewton:
Okay so this brings up a really good point, and something we need to design away from.

I think we actually shouldn't handle this case explicitly right now. What _should_ happen is we use our own arena allocation. So these allocations would never fail unless we ran out of room in our arena.

If we run out of room in our arena we should use `malloc`, and if it returns `NULL` we can crash the process like you do here. But if we are integrated into CRuby we should use `xmalloc`, which will attempt to run a GC before crashing in an attempt to make more room.

So all of that is to say, you've definitely found a bug, but I want to fix it in a different way.

PR #1085, opened 2023-06-26
Centralize allocation functions

kddnewton:
This PR does a bunch of work to centralize where we allocate memory in preparation for an arena allocator.

Previously we were calling `malloc`, `calloc`, and `realloc` in a bunch of places. Now we always call `yp_malloc` or `yp_realloc`. Those functions are the same except that they accept a `yp_parser_t *`. `yp_parser_t` now has `malloc_callback` and `realloc_callback` fields that can be set through a public API. It is assumed that these callbacks will never return `NULL`.

In the case that the callbacks were not set, we use the system `malloc` and `realloc` and assert that they are not `NULL`.

The other pieces of this PR are moving all of the files around in order to get all of the allocation functions to flow through these new functions. Mostly this involves changing function signatures to accept a parser and pass it all of the way through.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ a.out
 /java/org/yarp/Nodes.java
 /lib/yarp/node.rb
 /lib/yarp/serialize.rb
+/src/memsize.c
 /src/node.c
 /src/prettyprint.c
 /src/serialize.c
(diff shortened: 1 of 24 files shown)


eregon:
Any reason this was not merged?
It seems nice to not have to care about malloc() failure.

PR #1087, opened 2023-06-27
Introduce Source

kddnewton:
This PR adds a bunch of functionality to the Ruby extension.

Every `YARP::Node`, `YARP::Token`, `YARP::Comment`, `YARP::Error`, and `YARP::Warning` now have a `YARP::Location` instance. (We were previously relatively inconsistent between using a location vs using start_offset + length. Now we just have a single object.)

`YARP::Location` instances now have the following instance variables. They are all given in the initializer/constructed by the parser in the extension:

* `@source` - a `YARP::Source` object
* `@start_offset` - the start offset of the node in the source
* `@length` - the length of the node in the source

`YARP::Location` instances now have the following methods:

* `#start_offset` - returns the instance variable
* `#end_offset` - returns the instance variable + the length
* `#length` - returns the instance variable
* `#start_line` - queries the source range object for a line number with the start offset
* `#end_line` - queries the source range object for a line number with the end offset
* `#start_column` - queries the source range object for a column number with the start offset
* `#end_column` - queries the source range object for a column number with the end offset
* `#slice` - returns a slice of the source string represented by the location

`YARP::SourceRange` instances are generally shared between all of the locations within a tree. They are constructed by the parser in the extension. They have the following instance variables:

* `@source` - the source string
* `@offsets` - an array of line offsets (the offset of the start of each line)

`YARP::Source` instances have the following methods:

* `#line(value)` - returns the line number of the given offset
* `#column(value)` - returns the column number of the given offset
* `#slice(offset, length)` - returns a slice of the source string

Built on top of all of this are some nice convenience APIs that we template onto each node now whenever you have location parameters. For example, now you can call `IfNode#if_keyword` to receive back a string that is sliced from the source (typically either `"if"` or `"elsif"`). This is also nice in places like `SymbolNode#content` that returns a string that is sliced from the source.

These APIs could be improved by using a better data structure like a succinct bit vector to store the line offsets (https://techlife.cookpad.com/entry/2018/10/17/115126), but fortunately the user-facing side would not need to change. cc @mame

### DIFF
--- a/templates/lib/yarp/node.rb.erb
+++ b/templates/lib/yarp/node.rb.erb
@@ -6,13 +6,12 @@ module YARP
     attr_reader :<%= param.name %>
 
     <%- end -%>
-    # def initialize: (<%= (node.params.map { |param| "#{param.name}: #{param.rbs_class}" } + ["start_offset: Integer", "length: Integer"]).join(", ") %>) -> void
-    def initialize(<%= (node.params.map(&:name) + ["start_offset", "length"]).join(", ") %>)
+    # def initialize: (<%= (node.params.map { |param| "#{param.name}: #{param.rbs_class}" } + ["location: Location"]).join(", ") %>) -> void
+    def initialize(<%= (node.params.map(&:name) + ["location"]).join(", ") %>)
       <%- node.params.each do |param| -%>
       @<%= param.name %> = <%= param.name %>
       <%- end -%>
-      @start_offset = start_offset
-      @length = length
+      @location = location
     end
 
     # def accept: (visitor: Visitor) -> void
@@ -62,14 +61,14 @@ module YARP
     private
 
     # Create a new Location object
-    def Location(start_offset = 0, length = 0)
-      Location.new(start_offset, length)
+    def Location(source = nil, start_offset = 0, length = 0)
+      Location.new(source, start_offset, length)
     end
     <%- nodes.each do |node| -%>
 
     # Create a new <%= node.name %> node
-    def <%= node.name %>(<%= node.params.map(&:name).join(", ") %>)
-      <%= node.name %>.new(<%= (node.params.map(&:name) + ["0", "0"]).join(", ") %>)
+    def <%= node.name %>(<%= (node.params.map(&:name) + ["location = Location()"]).join(", ") %>)
+      <%= node.name %>.new(<%= (node.params.map(&:name) + ["location"]).join(", ") %>)
     end
     <%- end -%>
   end
--- a/templates/lib/yarp/node.rb.erb
+++ b/templates/lib/yarp/node.rb.erb
@@ -69,6 +69,24 @@ module YARP
   end
 
   <%- end -%>
+  # A class that knows how to walk down the tree. None of the individual visit
+  # methods are implemented on this visitor, so it forces the consumer to
+  # implement each one that they need. For a default implementation that
+  # continues walking the tree, see the Visitor class.
+  class BasicVisitor
+    def visit(node)
+      node&.accept(self)
+    end
+
+    def visit_all(nodes)
+      nodes.map { |node| visit(node) }
+    end
+
+    def visit_child_nodes(node)
+      visit_all(node.child_nodes)
+    end
+  end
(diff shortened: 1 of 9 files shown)


eregon:
One thought about naming, maybe it should be just [Source](https://www.graalvm.org/truffle/javadoc/com/oracle/truffle/api/source/Source.html) (`YARP::SourceRange` -> `YARP::Source`)?

For instance in Truffle that is named Source and a Location is called SourceSection or SourceLocation (YARP::Location is great as-is).

templates/lib/yarp/node.rb.erb
```diff
@@ -6,13 +6,12 @@ module YARP
...
-    # def initialize: (<%= (node.params.map { |param| "#{param.name}: #{param.rbs_class}" } + ["start_offset: Integer", "length: Integer"]).join(", ") %>) -> void
-    def initialize(<%= (node.params.map(&:name) + ["start_offset", "length"]).join(", ") %>)
+    # def initialize: (<%= (node.params.map { |param| "#{param.name}: #{param.rbs_class}" } + ["location: Location"]).join(", ") %>) -> void
+    def initialize(<%= (node.params.map(&:name) + ["location"]).join(", ") %>)
       <%- node.params.each do |param| -%>
       @<%= param.name %> = <%= param.name %>
       <%- end -%>
-      @start_offset = start_offset
-      @length = length
+      @location = location
```
eregon, line 14: This means the Location object is created eagerly, which is likely significantly slower when creating Ruby nodes.
Basically it reverts https://github.com/ruby/yarp/pull/928/commits/9c57958604d87eff3641c672bdbf6316349b291c
I would suggest to store `start_offset, length, source` instead, and only create the Location object when it's asked for.

eregon: BTW I was thinking we should also do this for location fields, to object-inline them, in Java and Ruby, since unlike C structs, these languages don't have a way to object-inline in the syntax, it needs to be done manually as multiple fields in the parent object.

kddnewton:
> One thought about naming, maybe it should be just `Source` (`YARP::SourceRange` -> `YARP::Source`)? What does the `Range` mean in `SourceRange`?
> 
> For instance in Truffle that is named [Source](https://www.graalvm.org/truffle/javadoc/com/oracle/truffle/api/source/Source.html) and a Location is called SourceSection or SourceLocation (YARP::Location is great as-is).

Done!

templates/lib/yarp/node.rb.erb
kddnewton: So I tried inlining it, and it degraded performance for parse and dump/load. I wonder if tripped some threshold where the number of ivars is too large to be using a smaller representation by adding source?

Here are the benchmarks I found:

## 1000.times { YARP.parse_file("lib/yarp/node.rb") }

No inlining:

```
Rehearsal -----------------------------------------
parse   5.587947   0.104401   5.692348 (  5.694816)
-------------------------------- total: 5.692348sec

            user     system      total        real
parse   5.580172   0.096831   5.677003 (  5.677191)
```

Inlining:

```
Rehearsal -----------------------------------------
parse   5.917879   0.181696   6.099575 (  6.162626)
-------------------------------- total: 6.099575sec

            user     system      total        real
parse   6.205026   0.157306   6.362332 (  6.429058)
```

## 1000.times { YARP.load(source, YARP.dump(source)) }

No inlining:

```
Rehearsal -----------------------------------------
parse  18.519872   0.169504  18.689376 ( 18.703276)
------------------------------- total: 18.689376sec

            user     system      total        real
parse  18.937560   0.276980  19.214540 ( 19.350734)
```

Inlining:

```
Rehearsal -----------------------------------------
parse  20.784706   0.186370  20.971076 ( 20.988403)
------------------------------- total: 20.971076sec

            user     system      total        real
parse  20.899019   0.164351  21.063370 ( 21.068951)
```

If you want to mess around with the patch and maybe we can get it faster, I'd be happy to take another look. Here's the patch I was using:

[inline.patch](https://github.com/ruby/yarp/files/11899117/inline.patch)

PR #1088, opened 2023-06-28
Generate and use a config.h header

eregon:
Useful for #1041

Related to https://github.com/ruby/yarp/pull/1051


eregon:
This works with TruffleRuby with https://github.com/oracle/truffleruby/pull/3140, so let's wait that merges and then we can merge this.

kddnewton (changes requested):
Yeah this looks good. I actually changed my mind on the `src/yarp.c` changes though, would you be able to revert those only? Other than that, I think this is good to go. Let me know when the truffleruby PR is merged.

eregon:
> I actually changed my mind on the `src/yarp.c` changes though, would you be able to revert those only?

Do you mean defining `YP_VERSION_MAJOR` in `include/yarp.h` instead of in `config.h` (via autoconf)? What's the rationale?

(BTW this PR should remove those `YP_VERSION*` from `include/yarp.h` if we keep it in autoconf, it seems https://github.com/ruby/yarp/pull/1051 missed that cleanup)

kddnewton:
Specifically I mean using the `PACKAGE_VERSION` macro. I'm nervous about it conflicting since it's not prefixed with anything.

eregon:
Ah sure, happy to change that back to `YP_VERSION`

eregon:
I remove the arguments to AC_INIT as well then to not define any `PACKAGE_*` macros?

kddnewton:
> Should I remove the arguments to AC_INIT as well then to not define any `PACKAGE_*` macros?

I think they get defined as empty in that case. But yeah feel free.

kddnewton:
@eregon is this good to go or are we waiting on truffleruby?

eregon:
Let's wait the [TruffleRuby PR](https://github.com/oracle/truffleruby/pull/3140) merges to not break the CI here.

eregon:
https://github.com/oracle/truffleruby/actions/runs/5415357250/jobs/9843509793?pr=3140
```
src/enc/unicode.c: In function ‘utf_8_codepoint’:
src/enc/unicode.c:2233:28: error: conversion to ‘uint32_t’ {aka ‘unsigned int’} from ‘int’ may change the sign of the result [-Werror=sign-conversion]
 2233 |             (0xff >> type) & (byte);
      |                            ^
```
Not sure why it this error happens there and not in other places. I'll try a trivial fix.

kddnewton:
@eregon strange - probably needs to be `0xffu`

eregon:
Yeah, that goes further, but there are more errors: https://github.com/oracle/truffleruby/actions/runs/5416543973/jobs/9846399623?pr=3140
I'm just using `make all-no-debug`, so really weird it doesn't happen in YARP's CI.
One difference is that job runs on Ubuntu 20.04 vs 22.04 here.
I'll try to add a 20.04 job here to see if that repros.

eregon:
OK it repros on 20.04: https://github.com/ruby/yarp/actions/runs/5416596186/jobs/9846527321?pr=1088
@kddnewton Could you address these warnings?

kddnewton:
@eregon yeah can you rebase off https://github.com/ruby/yarp/pull/1103

eregon:
Thanks!
We'll need a bit longer before merging the TruffleRuby PR, because that needs a small change in `mx`, and that's currently blocked due to some unrelated CI issue.

eregon:
The TruffleRuby PR merged, CI is green on rerun, I'll merge

PR #1089, opened 2023-06-28
Add fuzzing

stevenjohnstone:
Fuzzing for yp_parse_serialize, yp_regexp_named_capture_group_names and yp_unescape_manipulate_string. Automation of common tasks via make and docker plus some helper scripts.


fuzz/parse.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+OUTPUT_DIR=$1
+
+screen -S grammar_fuzzer -d -m  /bin/bash -c "AFL_CUSTOM_MUTATOR_LIBRARY=/usr/local/lib/libgrammarmutator-ruby.so AFL_CUSTOM_MUTATOR_ONLY=1 afl-fuzz -i fuzz/corpus/parse -o $OUTPUT_DIR -S grammar -- ./build/fuzz.parse"
```
kddnewton, line 5: Seems like the parse corpus is a direct copy of the tests from the top level. Could we do something to remove this duplication like mount that as a directory in docker instead?

stevenjohnstone: Yes, that'd work. Or a symlink. I was wondering if there's a better collection of programs I can use as a corpus? This was the laziest possible thing I could do...

kddnewton: We should probably use our fixtures. They're the best corpus we have. Everything under `test/fixtures/**/*.txt` is a Ruby program. We have them listed as text only to not confuse tools looking for all Ruby programs.

stevenjohnstone: That's far more sensible! I'll try that out.

fuzz/dict
```diff
@@ -0,0 +1,31 @@
+"__ENCODING__"
```
kddnewton, line 1: Could we expand this dictionary to include at least one of every token? For the most part you can just copy the examples out of `config.yml`

kddnewton:
This looks really good @stevenjohnstone, thanks so much! I'm going to merge this and I'd like to see if we can iterate on their grammar.json file because it looks relatively bare-bones.

PR #1105, opened 2023-06-30
Improve error message format

HParker:


config.yml
```diff
@@ -2094,3 +2094,6 @@ nodes:
 
           yield 1
           ^^^^^^^
+errors:
+  - name: duplicate_parameter
+    message: duplicate parameter `{location}` in parameters
```
HParker, line 2099: add type and pull each arg out into `name`, `type`

ext/yarp/extension.c
```diff
@@ -280,15 +280,19 @@ parser_errors(yp_parser_t *parser, rb_encoding *encoding, VALUE source) {
...
             LONG2FIX(error->start - parser->start),
-            LONG2FIX(error->end - parser->start)
+            LONG2FIX(error->end - error->start)
         };
 
         VALUE error_argv[] = {
             rb_enc_str_new_cstr(error->message, encoding),
-            rb_class_new_instance(3, location_argv, rb_cYARPLocation)
+            rb_class_new_instance(3, location_argv, rb_cYARPLocation),
+            rb_enc_str_new_cstr(error->line, encoding),
```
HParker, line 289: Not needed anymore since `location` has more information now! 🎉

PR #1110, opened 2023-06-30
Package up YARP as a gem

flavorjones:
__Summary__

- Wrap YARP as a gem
  - add a gemspec
  - add a "rake build" task to package the gem
  - add a "rake check_manifest" task to validate the gem packaging
  - extconf.rb knows how to make librubyparser
  - add CI jobs to test gem packaging and installation
- Add test coverage for symbol visibility in static archives and shared objects
- Switch the compile-time default to NOT export symbols on non-windows platforms, superseding #1111.

__Notable change__

The 'compile' rake task is no longer a dependency of the 'test' task, to allow us to run the test suite against the installed gem in CI. CONTRIBUTING.md has been updated. Developers can still run `bundle exec rake` to do all the things.

__Notable design choice__

"autoconf" is run during gem packaging, and the "configure" script is in the gem. This reduces the requirements on the installation system.


ext/yarp/extconf.rb
```diff
@@ -19,32 +21,59 @@ def configure_debug
...
+            build_static_rubyparser
           end
           append_ldflags(static_archive_path)
         else
           unless find_library("rubyparser", "yp_parser_init", build_dir)
-            raise "Please run make to build librubyparser.so"
+            build_shared_rubyparser
+            unless find_library("rubyparser", "yp_parser_init", build_dir)
+              raise "could not build librubyparser.so"
+            end
```
tenderlove, line 47: Are we ever going to use this code path?  I'd like to change `RUBY_FUNC_EXPORTED` such that it's empty _unless_ we're building `librubyparser`.  Currently Ruby is leaking YARP symbols, and if you try to compile and use YARP with Ruby head, it'll end up using the functions defined in Ruby rather than the ones in here.

Basically I want to do [this](https://github.com/tenderlove/ruby/commit/32101689902d233f062fb31cb7b0a34ee4136286), then add `-D YP_EXPORTED_FUNCTION=extern` to the CFLAGS when building the `.so`

flavorjones: I expect you're right that we don't need it and that would simplify some of the hoops I jumped through in https://github.com/ruby/yarp/pull/1069

Can we leave that for a separate PR, though?

eregon:
I'd like to merge https://github.com/ruby/yarp/pull/1088 first to avoid conflicts if possible, I fixed more things in the Makefile.
Also I'd be happy to review this PR next week

flavorjones:
Right now the failing tests on ruby head are because we don't have a way to specify "load the gem's yarp.rb and yarp.so" and the stdlib versions are being loaded (which causes tests to fail). @k0kubun mentioned he has some ideas on how to handle this.

ext/yarp/extconf.rb
tenderlove: Let's keep this branch for now, but I think we need to revisit, document, and most importantly test all YARP build possibilities.

For the gem, JRuby will need the shared object along with an FFI integration.  I'm not sure if that's true for Truffle Ruby.  I'd like it if CRuby didn't need to use the serialization interface in the gem, but we probably should test performance.  IOW if all Ruby implementations used the same interface, that would make our lives easier.  I just don't want it to be at the cost of performance on CRuby.

eregon: I think for the Ruby API on TruffleRuby the deserialization will be much better than running through the C extension, but I need to benchmark. I plan to make a PR to implement `YARP.load` with Fiddle (or FFI) on JRuby/TruffleRuby.
I would prefer if CRuby would use that too, but indeed if the performance is significantly worse then we'll have to keep both ways (they used to be pretty close last time I looked at it for CRuby).

eregon: > I plan to make a PR to implement `YARP.parse` with Fiddle (or FFI) on JRuby/TruffleRuby.

Early draft PR for that: https://github.com/ruby/yarp/pull/1112

eregon: > Currently Ruby is leaking YARP symbols, and if you try to compile and use YARP with Ruby head, it'll end up using the functions defined in Ruby rather than the ones in here.

How does that happen? Does CRuby use the static library? That one doesn't export anything, so that should work fine.

eregon:
> A notable design choice:
> 
>     * "autoconf" is run during gem packaging, and the "configure" script is in the gem. This reduces the requirements on the installation system.

Definitely the best :) I did the same for TruffleRuby BTW because our macOS CI doesn't have autotools installed/in PATH by default.

yarp.gemspec
```diff
@@ -0,0 +1,47 @@
...
+    "Makefile.in",
+    "README.md",
+    "config.h.in",
+    "config.yml",
+    "configure",
+    "ext/yarp/api_node.c",
+    "include/yarp/ast.h",
+    "java/org/yarp/AbstractNodeVisitor.java",
+    "java/org/yarp/Loader.java",
+    "java/org/yarp/Nodes.java",
```
eregon, line 27: We don't need those in the gem, at least not for TruffleRuby.

eregon (approved):
Looks great!

tasks/check_manifest.rake
```diff
@@ -0,0 +1,63 @@
...
+  missing_files = intended_files - spec_files
+  extra_files = spec_files - intended_files
+
+  unless missing_files.empty?
+    puts "missing:"
+    missing_files.sort.each { |f| puts "- #{f}" }
+  end
+  unless extra_files.empty?
+    puts "unexpected:"
+    extra_files.sort.each { |f| puts "+ #{f}" }
```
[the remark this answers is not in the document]
flavorjones: Probably! Will make it so.

yarp.gemspec
flavorjones: @enebo Do these need to be in the gem for JRuby?

eregon: I would say not for now for JRuby, because there is nothing compiling these `.java` files in the gem right now.
We can always include them later if it's useful.

flavorjones: OK, will remove.

k0kubun:
> @k0kubun mentioned he has some ideas on how to handle this.

Today I locally confirmed that non-builtin `yarp.rb` and `yarp.so` files could be loaded if you modify `$LOAD_PATH`. It means that _technically_ it's not mandatory to gemify YARP in the interpreter to replace it with a gem. 

I was thinking of somehow doing that somewhere, but since a gemspec is not evaluated when Gemfile loads yarp.gem IIUC, there's no way to make it happen automatically on Gemfile unless you add some hack to the interpreter, rubygems, or bundler.

So, just like the existing convention, I think the YARP that is shipped as part of CRuby should be gemified to allow replacing the YARP implementation with a gem.

flavorjones:
@tenderlove @k0kubun @jemmaissroff I've updated this branch with:

- test coverage of symbol visibility,
- switch the compile-time default to hide symbols (superseding #1111)

Note that ruby-head is still failing because of the gem loading issue described above. That failure should go away once ruby stdlib yarp is gemified and contains these commits.

Please let me know what you think.

tenderlove (approved):
This seems great to me!  Same question as @jemmaissroff  though (not a show stopper obviously).

yarp.gemspec
```diff
@@ -0,0 +1,44 @@
...
+    "lib/yarp/node.rb",
+    "lib/yarp/serialize.rb",
+    "src/node.c",
+    "src/prettyprint.c",
+    "src/serialize.c",
+    "src/token_type.c",
+    "yarp.gemspec",
+  ]
+
+  spec.files += Dir.glob([
```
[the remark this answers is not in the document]
flavorjones: Thanks for asking this question. Where there is overlap, the files listed explicitly are _generated_ files, and are not tracked in source control (and so may not exist in a particular work tree). The goal is to ensure an error if we try to package the gem without those files.

If I'm being honest, I'm not entirely comfortable using globs for a gem manifest; but files are still being added pretty often, and I didn't want to create a lot of work for folks (and, because the other files are in source control, we can generally rely on them existing).

Once the manifest stabilizes I would _prefer_ to list every file individually ([like this](https://github.com/sparklemotion/nokogiri/blob/main/nokogiri.gemspec#L55)).

flavorjones:
@tenderlove I answered above, but for posterity: the files listed explicitly in the gemspec which overlap with the globs are _generated_ files, and are not in source control -- so we can't assume they exist. Explicitly listing them is a safety check, and attempting to package the gem without them will fail with an error.

Though, I'm noticing now that `include/yarp/config.h` is being packaged up because it matches a glob; but we shouldn't be packaging it -- so I'd like to switch to listing every file out individually.

yarp.gemspec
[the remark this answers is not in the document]
flavorjones: Force-pushed an update that lists every file individually.

flavorjones:
I've force-pushed an update that lists every file in the gemspec individually, which also now correctly does _not_ package `include/yarp/config.h`.

jemmaissroff (approved):
Thanks for your work here, Mike!

PR #1112, opened 2023-07-02
Add FFI backend for JRuby and TruffleRuby

eregon:
There are still many things to do before it's fully functional (see comments in the diff), but it does run a fair number of tests.

I used Fiddle instead of FFI because that way we can test it (and maybe use it if fast enough) on CRuby without any additional dependency.


lib/yarp.rb
```diff
@@ -227,4 +227,84 @@ def self.load(source, serialized)
...
+    module LibRubyParser
+      extend Fiddle::Importer
+
+      POINTER_SIZE = sizeof("void*")
+
+      Buffer = struct ['char *value', 'size_t length', 'size_t capacity']
+
+      dlload File.expand_path("../build/librubyparser.so", __dir__)
+
+      typealias 'bool', 'char' # OK? https://github.com/ruby/fiddle/issues/130
```
eregon, line 249: @tenderlove Maybe you know a good way to do this?

lib/yarp.rb
```diff
@@ -227,4 +227,84 @@ def self.load(source, serialized)
...
+            unless excludes.any? { |exclude| exclude =~ line }
+              extern line
+            end
+          end
+        end
+      end
+
+      load_expored_functions_from("yarp.h", [/callback/, /yp_token_type_to_str/])
+      load_expored_functions_from("yarp/util/yp_buffer.h")
+      load_expored_functions_from("yarp/unescape.h")
```
eregon, line 269: This seems nice to avoid duplication and out-of-sync function declarations.

lib/yarp.rb
```diff
@@ -227,4 +227,84 @@ def self.load(source, serialized)
...
+
+    end
+
+
+    VERSION = LibRubyParser.yp_version.to_s
+
+    def self.dump(code, filepath = nil)
+      buffer = LibRubyParser::Buffer.malloc
+      raise unless LibRubyParser.yp_buffer_init(buffer) == 1
+      LibRubyParser.yp_parse_serialize(code, code.bytesize, buffer)
```
eregon, line 286: `yp_parse_serialize()` doesn't take a `filepath` but it should

lib/yarp.rb
```diff
@@ -227,4 +227,84 @@ def self.load(source, serialized)
...
+      dump(code, filepath)
+    end
+
+    def self.parse(code, filepath = nil)
+      serialized = dump(code, filepath)
+      node = load(code, serialized)
+      ParseResult.new(node, [], [], [])
+    end
+
+    # def self.unescape(source, type)
```
eregon, line 301: This seems messy to implement because
```c
YP_EXPORTED_FUNCTION void yp_unescape_manipulate_string(yp_parser_t *parser, const char *value, size_t length, yp_string_t *string, yp_unescape_type_t unescape_type, yp_list_t *error_list);
```
is quite complicated to call.

I think in general we should have minimum amount of code in the C extension, move the logic in librubyparser, and expose easy-to-call functions.

lib/yarp.rb
```diff
@@ -227,4 +227,84 @@ def self.load(source, serialized)
...
+
+    def self.dump(code, filepath = nil)
+      buffer = LibRubyParser::Buffer.malloc
+      raise unless LibRubyParser.yp_buffer_init(buffer) == 1
+      LibRubyParser.yp_parse_serialize(code, code.bytesize, buffer)
+      buffer.value.to_s(buffer.length)
+    end
+
+    def self.dump_file(filepath)
+      code = File.binread(filepath)
```
eregon, line 291: It'd be nice to use mmap here, which requires the mmap stuff to move to librubyparser.

lib/yarp.rb
```diff
@@ -227,4 +227,84 @@ def self.load(source, serialized)
...
+      Buffer = struct ['char *value', 'size_t length', 'size_t capacity']
+
+      dlload File.expand_path("../build/librubyparser.so", __dir__)
+
+      typealias 'bool', 'char' # OK? https://github.com/ruby/fiddle/issues/130
+
+      typealias 'yp_unescape_type_t', 'int'
+      YP_UNESCAPE_NONE = 0
+      YP_UNESCAPE_MINIMAL = 1
+      YP_UNESCAPE_ALL = 2
```
eregon, line 254: It seems Fiddle has not support for `enum` so this is the closest thing to it.

lib/yarp.rb
```diff
@@ -227,4 +227,84 @@ def self.load(source, serialized)
...
+
+    def self.dump_file(filepath)
+      code = File.binread(filepath)
+      dump(code, filepath)
+    end
+
+    def self.parse(code, filepath = nil)
+      serialized = dump(code, filepath)
+      node = load(code, serialized)
+      ParseResult.new(node, [], [], [])
```
eregon, line 298: Deserialization should return a `ParseResult`, otherwise we lose warnings, errors and comments.
(alternatively those could be stored in ProgramNode but that just seems hacky).
We need to serialize those too then of course, currently they are not serialized.

lib/yarp.rb
tenderlove: This is probably fine.  I'll try to fix the issue upstream (tbh I don't really use the C parser functionality in Fiddle very often, and typically manually declare sizes 😅 )

tenderlove:
This is very cool.

Probably off topic for this PR, but one thing I've considered doing some day is using clang to generate Fiddle bindings when we build the gem.  So when we build the gem for the `-java` target, just have a program that uses clang to generate Fiddle bindings and package that along with the gem.

RJIT is using clang to generate its bindings, maybe we could do something similar?

lib/yarp.rb
eregon: kou fixed it already: https://github.com/ruby/fiddle/issues/130
But of course we need this to work on older Fiddle too, so probably this workaround is fine.

lib/yarp.rb
eregon: https://github.com/ruby/yarp/pull/1172 to fix that

lib/yarp.rb
eregon: https://github.com/ruby/yarp/pull/1172 to fix that

lib/yarp.rb
```diff
@@ -313,4 +319,120 @@ def self.newlines(source)
...
+      ParseResult.new(node, cheat.comments, cheat.errors, cheat.warnings, source)
+    end
+
+    def self.parse_file(filepath)
+      code = File.binread(filepath)
+      parse(code, filepath)
+    end
+
+    # Should we even implement these? Probably they shouldn't be part of public YARP API
+    # We could guard the corresponding tests with `if YARP::BACKEND == :CExtension`
```
eregon, line 412: https://github.com/ruby/yarp/issues/1173

eregon:
This is actually working without loading the C extension and passes the test suite now :tada: 
Still needs to adapt some docs and update CI to test it on JRuby/TruffleRuby and do some cleanups.

eregon:
I'm trying to run this on JRuby and TruffleRuby but both have limited support for Fiddle.
For example, TruffleRuby doesn't implement `Fiddle::Pointer#to_s` & `Fiddle::Pointer.malloc` and JRuby breaks on `Buffer = struct ['char *value', 'size_t length', 'size_t capacity']`: https://github.com/jruby/jruby/issues/7883.

So maybe I should use FFI after all, because both support FFI better.
The alternative is waiting that both fix Fiddle enough.

eregon:
CI passing now, both for `bundle exec rake` and for `gem install`, using FFI.
Just waiting that https://github.com/ruby/yarp/pull/1245 gets merged, then I'll rebase this and it will be ready for review.

enebo:
@eregon thanks for the FFI change.  It is a bummer we are having that issue with Fiddle in JRuby.  We will fix it but obviously that takes time and a release.

eregon:
Yeah, exactly I'm testing in CI here on jruby and truffleruby releases (for improved stability, and because it's valuable that the YARP gem can be installed on jruby and truffleruby releases already) so Fiddle is not an option.

lib/yarp.rb
```diff
@@ -334,4 +340,148 @@ def self.parse_serialize_file(filepath)
 require_relative "yarp/serialize"
 require_relative "yarp/pack"
 
-require "yarp/yarp"
+if RUBY_ENGINE == 'ruby' and ENV["YARP_FFI_BACKEND"] != "true"
+  require "yarp/yarp.so"
+else
+  require "rbconfig"
```
kddnewton, line 346: Can you move this to another file? Something like `yarp/ffi`?

templates/include/yarp/ast.h.erb
```diff
@@ -84,14 +84,15 @@ typedef struct yp_<%= node.human %> {
...
     <%= case param
-    in NodeParam | OptionalNodeParam then "struct #{param.c_type} *#{param.name}"
-    in NodeListParam then "struct yp_node_list #{param.name}"
-    in LocationListParam then "yp_location_list_t #{param.name}"
-    in ConstantParam then "yp_constant_id_t #{param.name}"
-    in ConstantListParam then "yp_constant_id_list_t #{param.name}"
-    in StringParam then "yp_string_t #{param.name}"
-    in LocationParam | OptionalLocationParam then "yp_location_t #{param.name}"
-    in UInt32Param then "uint32_t #{param.name}"
+    when NodeParam, OptionalNodeParam then "struct #{param.c_type} *#{param.name}"
```
kddnewton, line 87: It looks like you're removing all pattern matching and single-line method definitions. I'm assuming this is to get CI running for JRuby/TruffleRuby. In this case can you document what Ruby features/version we can use for internal development?

templates/src/serialize.c.erb
```diff
@@ -1,7 +1,4 @@
-#include "yarp/ast.h"
-#include "yarp/diagnostic.h"
-#include "yarp/parser.h"
-#include "yarp/util/yp_buffer.h"
+#include "yarp.h"
```
kddnewton, line 1: Why this change?

templates/include/yarp/ast.h.erb
enebo: @kddnewton until TR implements YARP they have almost no support for pattern matching.  JRuby supports it but we need both to work with the same code.

kddnewton: Yeah for sure, makes sense. I just wasn't sure if it was just pattern-matching and single-line methods or if it was anything else as well.

kddnewton: Does JRuby or TruffleRuby not have support for single-line methods?

.github/workflows/main.yml
```diff
@@ -176,8 +197,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ["3.0", "3.1", "3.2", "head"]
+        ruby: ["3.0", "3.1", "3.2", "head", "jruby", "truffleruby"]
         os: [ubuntu-latest, macos-latest, windows-latest]
+        exclude:
+        - { ruby: truffleruby, os: windows-latest }
```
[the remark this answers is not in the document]
eregon: TruffleRuby doesn't support Windows

Gemfile
```diff
@@ -8,3 +8,7 @@ gem "rake"
 gem "rake-compiler"
 gem "test-unit"
 gem "ruby_memcheck", platform: %i[mri truffleruby mswin mingw x64_mingw]
+
+if RUBY_ENGINE == 'ruby' and ENV["YARP_FFI_BACKEND"] == "true"
```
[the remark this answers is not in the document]
eregon: OK

ext/yarp/extconf.rb
```diff
@@ -34,7 +42,7 @@ def configure_rubyparser
           end
           $LOCAL_LIBS << " #{static_archive_path}"
         else
-          shared_library_path = File.join(build_dir, "librubyparser.#{RbConfig::CONFIG["DLEXT"]}")
+          shared_library_path = File.join(build_dir, "librubyparser.#{RbConfig::CONFIG["SOEXT"]}")
```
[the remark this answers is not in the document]
eregon: Yes, the shared library is SOEXT, not DLEXT, that's only for C extensions (differs on darwin, .bundle vs .dylib)

ext/yarp/extconf.rb
```diff
@@ -119,7 +127,7 @@ def debug_mode_build?
   end
 end
 
-if arg_config("--help")
+if ARGV.delete("--help")
```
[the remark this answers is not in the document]
eregon: We don't require `mkmf` on non-CRuby, so we can't use `arg_config`

lib/yarp/ripper_compat.rb
```diff
@@ -47,9 +47,9 @@ def _dispatch_event_push(list, item)
       Ripper::PARSER_EVENT_TABLE.each do |event, arity|
         case event
         when /_new\z/
-          alias :"on_#{event}" :_dispatch_event_new if arity == 0
+          alias_method :"on_#{event}", :_dispatch_event_new if arity == 0
```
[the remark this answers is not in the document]
eregon: Unfortunately `alias` with interpolation is not supported on current TruffleRuby, it's fixed on -dev but not on latest release. Interestingly this is the first time this comes up, we only found out before due to our translator testing (while migrating to YARP).

templates/include/yarp/ast.h.erb
eregon: > It looks like you're removing all pattern matching and single-line method definitions. I'm assuming this is to get CI running for JRuby/TruffleRuby.

Correct.

> In this case can you document what Ruby features/version we can use for internal development?

Where should I document that? Not that the CI will catch such issues and of course a newer release of jruby/truffleruby might support some of that.

> Does JRuby or TruffleRuby not have support for single-line methods?

No for TruffleRuby, we're counting on adopting YARP to solve it: https://github.com/oracle/truffleruby/issues/3038

enebo: JRuby supports all 3.1 syntax features sans fixing argument order execution and flip-flops.  The argument order execution I do not think is going to be a source of problems for years but we will get to it.  flip-flops no one uses (at least in libraries).

templates/src/serialize.c.erb
eregon: Otherwise:
```
serialize.c.erb: In function ‘yp_lex_serialize’:
serialize.c.erb:215:5: error: implicit declaration of function ‘yp_parser_init’; did you mean ‘yp_buffer_init’? [-Werror=implicit-function-declaration]
serialize.c.erb:223:23: error: implicit declaration of function ‘yp_parse’ [-Werror=implicit-function-declaration]
serialize.c.erb:223:23: error: initialization of ‘yp_node_t *’ {aka ‘struct yp_node *’} from ‘int’ makes pointer from integer without a cast [-Werror=int-conversion]
serialize.c.erb:237:5: error: implicit declaration of function ‘yp_node_destroy’ [-Werror=implicit-function-declaration]
serialize.c.erb:238:5: error: implicit declaration of function ‘yp_parser_free’; did you mean ‘yp_buffer_free’? [-Werror=implicit-function-declaration]
...
```
Because `yp_parser_init` is in yarp.h we have to include it, and that includes everything else.

templates/include/yarp/ast.h.erb
kddnewton: `CONTRIBUTING.md` would be a good place.

templates/src/serialize.c.erb
kddnewton: Got it.

ext/yarp/extconf.rb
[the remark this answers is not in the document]
eregon: It's already correct in other places.

eregon:
@kddnewton All done, thanks for the review!

kddnewton (approved):

PR #1116, opened 2023-07-03
Fix header dependency in Makefile.in

flavorjones:
Fix header dependency to ensure that `include/yarp/ast.h` is generated when running 'make'.

Also remove template files from the Makefile `clean` target because these aren't generated by the Makefile, and are more robustly removed by `rake clobber` anyway.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -25,6 +25,9 @@ all: shared static
 shared: build/librubyparser.$(SOEXT)
 static: build/librubyparser.a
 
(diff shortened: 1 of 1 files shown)


flavorjones:
The failing truffleruby test seems to be from https://github.com/oracle/truffleruby/commit/5dabddf9c365b3cf2817bbed2139d963f57b5380 and not from this change.

/cc @eregon

flavorjones:
Closing, this was already done as part of #1088

PR #1123, opened 2023-07-05
Set newline flag for Java nodes by walking the AST after deserialization

eregon:
For https://github.com/ruby/yarp/issues/1099, useful until YARP provides this information in C struct and/or in the serialized form.


eregon:
We should merge after the truffleruby PR https://github.com/oracle/truffleruby/pull/3154 merges, to not break the CI check.

PR #1126, opened 2023-07-05
Remove strncasecmp, call yp_snprintf

jemmaissroff:


tenderlove:
Great.  Just to provide some context around these changes:

1. Ruby doesn't use `configure` from YARP, but YARP wants "HAS_MMAP" etc to exist.  Those headers need to be calculated on the machine where Ruby is being compiled, so we need YARP to use Ruby's `config.h`.  This is handled by a change to the sync script.  The sync script will write out a `yarp/config.h` that simply includes Ruby's `ruby/config.h`.
2. We'd like to use `strncasecmp` if it exists on the platform, but detection seems broken on Windows.  I don't think this function is really a bottleneck, and mirroring upstream is a much higher priority, so lets use our version of `strncasecmp` until we can figure out how to get detection working well on Windows.
3. To simplify mirroring, we pulled the YARP version constants back in to their own header file.  We could keep them part of `configure.ac`, but they need to be defined somewhere besides `yarp/config.h` because of the fact that Ruby's build system doesn't execute YARP's `configure`

tenderlove (approved):

eregon:
Makes sense, thanks for the summary.

Yes, from when I looked how YARP is built in CRuby I saw the YARP Makefile is not used nor any of YARP's build system, instead it's directly CRuby's Makefile (IIRC).
And indeed it makes sense to have those check in CRuby's `configure.ac` and reuse CRuby's `config.h`.

> To simplify mirroring, we pulled the YARP version constants back in to their own header file.

Yeah, that sounds fine.

> 2. We'd like to use strncasecmp if it exists on the platform, but detection seems broken on Windows. I don't think this function is really a bottleneck, and mirroring upstream is a much higher priority, so lets use our version of strncasecmp until we can figure out how to get detection working well on Windows.

Right, yp_strncasecmp seems only used for comparing encoding names for the magic encodign comment, so that's not perf-critical. We could also just not use strncasecmp on Windows but doesn't seem worth the trouble given the usage.

include/yarp/version.h
```diff
@@ -0,0 +1,5 @@
+#define YP_VERSION_MAJOR 0
+#define YP_VERSION_MINOR 4
+#define YP_VERSION_PATCH 0
+
+#define YP_VERSION "0.4.0"
```
eregon, line 5: We should probably reuse the old logic to create this string from the components, so they can't get out of sync: https://github.com/ruby/yarp/pull/1088/files#diff-71ae829a627932e547cf825850d9a6b4773b009b0b68ae8eb1d8ef02e6d98987

PR #1131, opened 2023-07-07
Expose location method in the C API for ruby/ruby to use

jemmaissroff:


include/yarp.h
```diff
@@ -66,4 +66,23 @@ YP_EXPORTED_FUNCTION void yp_parse_serialize(const char *source, size_t size, yp
...
+// rb_code_position_struct in ruby/ruby which allows us to type pun
+typedef struct yp_code_position_struct {
+    int lineno;
+    int column;
+} yp_code_position_t;
+
+// This reprsents a begin and end of anything in the AST that has a
+// location from the parsed source code.
+// The layouts of this struct should not change because they mirror
+// rb_code_position_struct in ruby/ruby which allows us to type pun
```
eregon, line 80: rb_code_position_struct->yp_code_location_t

kddnewton:
I think this mostly duplicates `yp_newline_list_line_column` (https://github.com/ruby/yarp/blob/main/src/util/yp_newline_list.c#L92-L96). Could you use that instead? It's got a couple of optimizations like using a binary search or using the last lookup value to make it faster.

PR #1135, opened 2023-07-09
Fallback variadic macro `yp_snprintf`

nobu:


src/util/yp_snprintf.c
jemmaissroff, line 1: Why delete this file?

jemmaissroff (approved):

PR #1140, opened 2023-07-14
YARP gem sets `required_ruby_version` to `>= 3.0.0`

flavorjones:
This PR suggests that YARP gem should support being run by Ruby >= 3.0. It also adds test coverage to ensure compatibility.

I'm not sure what the YARP project committers want to support, so (hopefully obviously) I'm not expressing a strong opinion here.

_Note: this PR was originally attempting support for running with 2.7, I changed to 3.0 to attempt a more modest goal._

### DIFF
--- a/templates/lib/yarp/serialize.rb.erb
+++ b/templates/lib/yarp/serialize.rb.erb
@@ -89,8 +89,13 @@ module YARP
         unless constant
           offset = constant_pool_offset + index * 8
 
-          start = serialized.unpack1("L", offset: offset)
-          length = serialized.unpack1("L", offset: offset + 4)
+          if RUBY_VERSION >= "3.1"
+            start = serialized.unpack1("L", offset: offset)
+            length = serialized.unpack1("L", offset: offset + 4)
+          else
+            start = serialized[offset..].unpack1("L")
+            length = serialized[offset+4..].unpack1("L")
+          end
 
           constant = input.byteslice(start, length).to_sym
           constant_pool[index] = constant
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -144,7 +144,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ["3.2", "head"]
+        ruby: ["2.7", "3.0", "3.1", "3.2", "head"]
         os: [ ubuntu-latest, macos-latest, windows-latest ]
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -144,7 +144,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ["2.7", "3.0", "3.1", "3.2", "head"]
+        ruby: ["3.0", "3.1", "3.2", "head"]
         os: [ ubuntu-latest, macos-latest, windows-latest ]
     runs-on: ${{ matrix.os }}
     steps:
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@
 
 source "https://rubygems.org"
 
-ruby ">= 2.7.0"
+ruby ">= 3.0.0"
 
 gemspec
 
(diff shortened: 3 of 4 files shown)


flavorjones:
The parse tests seem to require 3.2:

```
  def test_Ruby_3_2_plus
    assert_operator RUBY_VERSION, :>=, "3.2.0", "ParseTest requires Ruby 3.2+"
  end
```

and there are 21 other failures as well on Ruby 3.1.

Earlier rubies fail because the Gemfile specifies `>= 3.1.0`.

What policy does the YARP team want to see implemented?

flavorjones:
Ruby 3.0 support will require avoiding unsupported pattern matching:

```
/home/flavorjones/code/oss/yarp/test/yarp_test_helper.rb:3:in `require': /home/flavorjones/code/oss/yarp/lib/yarp.rb:92: syntax error, unexpected (, expecting local variable or method (SyntaxError)
...er in Location[start_offset: ^(start_offset), end_offset: ^(...
...                              ^
/home/flavorjones/code/oss/yarp/lib/yarp.rb:92: syntax error, unexpected ',', expecting `end'
...[start_offset: ^(start_offset), end_offset: ^(end_offset)]
...                              ^
/home/flavorjones/code/oss/yarp/lib/yarp.rb:92: syntax error, unexpected ']', expecting `end'
...et), end_offset: ^(end_offset)]
...                              ^
/home/flavorjones/code/oss/yarp/lib/yarp.rb:198: syntax error, unexpected (, expecting local variable or method
      other in Token[type: ^(type), value: ^(value)]
                            ^
/home/flavorjones/code/oss/yarp/lib/yarp.rb:198: syntax error, unexpected ',', expecting `end'
...  other in Token[type: ^(type), value: ^(value)]
...                              ^
/home/flavorjones/code/oss/yarp/lib/yarp.rb:198: syntax error, unexpected ']', expecting `end'
...type: ^(type), value: ^(value)]
...                              ^
```

And there's some additional pattern matching we'd need to make to support Ruby 2.7. It seems possible, but it's clearly more complicated than just bumping the versions.

Closing.

flavorjones:
Reopening - I guess it's useful to see what's failing if this goal is aspirational.

flavorjones:
Ruby 3.0 is now running CI, and in addition to the ParseTest failures, this test is failing:

https://github.com/ruby/yarp/actions/runs/5558589699/jobs/10153864762?pr=1140#step:6:60

```
===============================================================================
Failure: test_duplicated_parameter_names(ErrorsTest):
  <[:program,
   [:stmts_add,
    [:stmts_new],
    [:def,
     [:@ident, "foo", [1, 4]],
     [:paren,
      [:params,
       [[:@ident, "a", [1, 8]],
        [:@ident, "b", [1, 10]],
        [:@ident, "a", [1, 12]]],
       nil,
       nil,
       nil,
       nil,
       nil,
       nil]],
     [:bodystmt, [:stmts_add, [:stmts_new], [:void_stmt]], nil, nil, nil]]]]> was expected to be nil.
/home/runner/work/yarp/yarp/test/errors_test.rb:970:in `assert_errors'
/home/runner/work/yarp/yarp/test/errors_test.rb:916:in `test_duplicated_parameter_names'
     913:       Location()
     914:     )
     915: 
  => 916:     assert_errors expected, "def foo(a,b,a);end", ["Duplicated parameter name."]
     917: 
     918:     expected = DefNode(
     919:       Location(),
```

templates/lib/yarp/serialize.rb.erb
```diff
@@ -89,8 +89,13 @@ module YARP
         unless constant
           offset = constant_pool_offset + index * 8
 
-          start = serialized.unpack1("L", offset: offset)
-          length = serialized.unpack1("L", offset: offset + 4)
+          if RUBY_VERSION >= "3.1"
```
eregon, line 92: This check is slow (so it slows downs the fast path on >= 3.1).

kddnewton: Agreed. I would prefer a polyfill to a version check here.

kddnewton (changes requested):

PR #1141, opened 2023-07-14
Remove pattern matching pinning to enable support for <= Ruby 3.0

jemmaissroff:
Pattern matching variable pinning was introduced in Ruby 3.1. We need to remove it from YARP to support earlier rubies.


paracycle (approved):
Thanks

PR #1145, opened 2023-07-18
Fix up the gemspec and the manifest check

flavorjones:
- ignore lib/yarp.bundle
- :check_manifest task is a dependency of :test
- remove yp_snprintf.c from the gemspec (see 4793b1ce / #962)


jemmaissroff (approved):
Thanks!

PR #1146, opened 2023-07-18
Move the gem's C extension into lib/yarp/

flavorjones:
and change the require from "yarp.so" to "yarp/yarp", which is more aligned with existing community conventions for gems.


lib/yarp.rb
```diff
@@ -240,4 +240,5 @@ def self.newlines(source)
 require_relative "yarp/ripper_compat"
 require_relative "yarp/serialize"
 require_relative "yarp/pack"
-require "yarp.so"
+
+require "yarp/yarp"
```
eregon, line 244:
```suggestion
require "yarp/yarp.so"
```
Seems clearer

tasks/check_manifest.rake
```diff
@@ -36,7 +36,7 @@ task :check_manifest => [:templates, "configure"] do
     java/org/yarp/Loader.java
     java/org/yarp/Nodes.java
     java/org/yarp/Parser.java
-    lib/yarp.{jar,so,bundle}
+    lib/yarp/yarp.*
```
eregon, line 39: This could ignore a `lib/yarp/yarp.rb` or `lib/yarp/yarp.config` or so, it's probably safer to keep the extensions there

eregon (approved):

tasks/check_manifest.rake
flavorjones: Fair, will change it.

PR #1149, opened 2023-07-19
Allow installation of discoverable C API development files

flavorjones:
## The problem I'm trying to address

The use case I'm trying to address here is that of a downstream C extension that wishes to integrate with the yarp gem's C API.

We don't export the C API from `yarp.{so,bundle}` and so (for example) downstream C extensions can't resolve the symbol `yp_parse` from that shared object.

## The implementation of this solution

This PR presents one possible solution:

- support for a `--install-c-api` flag at gem install time, which installs header files and librubyparser.a as part of the gem installation footprint
- a standalone ruby file `yarp/c_api` which provides CPPFLAGS and LDFLAGS to allow compilation and linking to happen

So usage might look like this:

```
$ gem install yarp -- --install-c-api
Building native extensions with: '--install-c-api'
This could take a while...
Successfully installed yarp-0.4.0
```

and then the downstream consumer's `extconf.rb` could look something like this:

```ruby
# frozen_string_literal: true
require "mkmf"

require "yarp/c_api"
append_cppflags(*Yarp::CAPI::CPPFLAGS)
append_ldflags(*Yarp::CAPI::LDFLAGS)

append_cflags("-fvisibility=hidden")

create_makefile("ruby_indexer/ruby_indexer")
```

If the gem is installed without `--install-c-api` then requiring `yarp/c_api` will raise a `LoadError` exception:

```
warning: LoadError: Yarp's C API is not installed. Please re-install with the --install-c-api flag.
```

## Alternative solutions

One idea is to ship a separate gem which contains the C API development files, but version it the same as the `yarp` gem so there's a one-to-one correspondence between them. So users of the C API would do:

```
gem "yarp-c-api"
```

## Reasons not to do this?

Is the C API going to be stable enough to package and version like this? @eregon brings up some valid points [below](https://github.com/ruby/yarp/pull/1149#issuecomment-1642833806):

> Node structs might be public currently but I'm not sure they should stay so (except for CRuby of course), I would see them as implementation details and it's definitely much harder to evolve in a compatible way than a Ruby API.
> (note: exposing struct members in a C API is often considered a mistake in terms of API evolution)

### DIFF
--- a/ext/yarp/extconf.rb
+++ b/ext/yarp/extconf.rb
@@ -56,7 +56,6 @@ module Yarp
       end
 
       def build_target_rubyparser(target)
-        # explicit "sh" is used for Windows where shebangs are not supported
         Dir.chdir(root_dir) do
           if !File.exist?("configure") && Dir.exist?(".git")
             # this block only exists to support building the gem from a "git" source,
--- a/ext/yarp/extconf.rb
+++ b/ext/yarp/extconf.rb
@@ -7,9 +7,12 @@ require "rake"
 module Yarp
   module ExtConf
     class << self
+      DEV_INSTALL_DIR = "c_api"
+
       def configure
         configure_c_extension
         configure_rubyparser
+        install_development_files
 
         create_makefile("yarp")
       end
@@ -55,6 +58,21 @@ module Yarp
         build_target_rubyparser "build/librubyparser.a"
       end
 
+      def install_development_files
+        return unless arg_config("--install-c-api", false)
+
+        # copy the static archive into "c_api/lib"
+        lib_dir = File.join(DEV_INSTALL_DIR, "lib")
+        FileUtils.mkdir_p lib_dir
+        FileUtils.cp File.join(build_dir, "librubyparser.a"), lib_dir
+
+        # copy header files into "c_api/include"
+        FileUtils.cp_r File.join(include_dir), DEV_INSTALL_DIR
+
+        $INSTALLFILES << ["#{DEV_INSTALL_DIR}/lib/*", "$(RUBYLIBDIR)/yarp"]
+        $INSTALLFILES << ["#{DEV_INSTALL_DIR}/include/**/*.h", "$(RUBYLIBDIR)/yarp"]
+      end
+
       def build_target_rubyparser(target)
         Dir.chdir(root_dir) do
           if !File.exist?("configure") && Dir.exist?(".git")
@@ -62,7 +80,7 @@ module Yarp
             # normally we package up the configure and other files in the gem itself
             Rake.sh("autoconf")
             Rake.sh("autoheader")
-            Rake.sh("rake", "templates")
+            Rake.sh("templates/template.rb")
           end
           Rake.sh("sh", "configure") # explicit "sh" for Windows where shebangs are not supported
           Rake.sh("make", target)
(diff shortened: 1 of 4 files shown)


vinistock:
One question about the `--install-c-api` flag, how would a gem pass that to add it as a dependency?

E.g.:

```ruby
# I don't think this exists, but you get my idea. How would we be able to do something like
# this from a dependency?
spec.add_dependency("yarp", "~> 0.5.0", "--install-c-api")
```

flavorjones:
> how would a gem pass that to add it as a dependency

Yeaaaah, that's the weak part of this approach, there's no easy way to do that today in a Gemfile (though you can do it with a bundler configuration setting).

One alternative is to install the C development dependencies by default (and flip the flag to `--no-install-c-api`). Maybe that would be better?

Some people get very upset at a gem installation footprint being bigger than it needs to be! Otherwise I'd just install those files and not worry about it.

vinistock:
Is there a way for the gem that wants the dependency to specify the bundler setting? If not, then indeed it'll be difficult to define that a gem depends on the C API.

Would setting an environment variable in the gemspec work? E.g.: `ENV["YARP_INCLUDE_C_API"]=true`.

flavorjones:
I'm debugging an issue with the location of the installed C development files.

eregon:
Are there gems that plan to use the (public) C API of YARP?
That's pretty small in terms of public functions, I don't think it's very useful/usable.
Node structs might be public currently but I'm not sure they should stay so (except for CRuby of course), I would see them as implementation details and it's definitely much harder to evolve in a compatible way than a Ruby API.

Why not use the Ruby API?

flavorjones:
> Are there gems that plan to use the (public) C API of YARP?

Yes, a team I work with within Shopify is experimenting with this for performance reasons (running across a large codebase to build an index). @vinistock has more context.

flavorjones:
I've update the description with a bit more information. This branch works, for whatever that's worth (I have a private project using the C API that is green).

But the maintainers should discuss:

- whether shipping a C API is something they _want_ to do,
- and decide whether _this_ is the right solution.

vinistock:
As Mike mentioned, we made an exploration for indexing codebases. If we hook directly into C, we can avoid allocating Ruby objects for each node in the tree and instead only produce objects for the things we're interested in maintaining on the index. We parse in C, go through the nodes in C and then produce objects only when we find what we need.

This has proved to have **very significantly better performance** than the approaches using only the Ruby API and allows tools such as the Ruby LSP to achieve the desired experience even on super large projects. The performance is so good, it allows us to not have to take on the complexity of using parallelism or caching for maintaining the index.

In terms of whether the node structs are a private API, that's a decision for the maintainers, so I'll defer to them. But in terms of usefulness, we have everything we need to analyze an AST: the ability to parse and the node structures.

Keeping the ability for gems to hook into the C part directly like this is definitely an exception. The only reason we explored it was to achieve our performance goals. But for these specific scenarios, it produces amazing results, so I hope that the team is open to allowing for something like this.

vinistock:
We are continuing to explore this and it turns out we made an unfair comparison between the Ruby API and the C API. We're able to achieve similar performance using the Ruby API after fixing some discrepancies so this may not be needed after all.

eregon:
That's great news.
I think the Ruby API has also some room to be optimized further (e.g. delaying Location allocations, I did some of that in https://github.com/ruby/yarp/pull/928 but it changed since then).
It's also worth trying with YJIT, as it might optimize the allocations and ivar sets in `initialize` quite a bit better.

flavorjones:
That is good news. I'm going to close this until/unless we have a use case.

PR #1150, opened 2023-07-19
Allow the gem to be used from a Gemfile "git:" source

flavorjones:
This, naturally, means that we need to run autoconf and generate the template files, which increases the requirements on the installation target system.

### DIFF
--- a/ext/yarp/extconf.rb
+++ b/ext/yarp/extconf.rb
@@ -61,11 +61,11 @@ module Yarp
           if !File.exist?("configure") && Dir.exist?(".git")
             # this block only exists to support building the gem from a "git" source,
             # normally we package up the configure and other files in the gem itself
-            Rake.sh("sh autoconf")
-            Rake.sh("sh autoheader")
-            Rake.sh("rake templates")
+            Rake.sh("sh", "autoconf")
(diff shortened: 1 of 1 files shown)


ext/yarp/extconf.rb
```diff
@@ -56,8 +56,16 @@ def build_static_rubyparser
...
         Dir.chdir(root_dir) do
-          Rake.sh("sh configure") # explicit "sh" for Windows where shebangs are not supported
+          if !File.exist?("configure") && Dir.exist?(".git")
+            # this block only exists to support building the gem from a "git" source,
+            # normally we package up the configure and other files in the gem itself
+            Rake.sh("sh autoconf")
+            Rake.sh("sh autoheader")
+            Rake.sh("rake templates")
+          end
+          Rake.sh("sh configure")
```
casperisfine, line 68:
```suggestion
            Rake.sh("sh", "autoconf")
            Rake.sh("sh", "autoheader")
            Rake.sh("rake","templates")
          end
          Rake.sh("sh", "configure")
```

We can avoid a subshell here. Not the end of the world but it's one of my pet peeves.

jemmaissroff (approved):

PR #1153, opened 2023-07-19
extconf.rb: using "sh" when unnecessary won't work on some systems

flavorjones:
While running commands with "sh" works on MacOS and Windows, it won't work on some Linux systems (where the command won't be found).

Let's remove it for "autoconf" and "autoheader" until we actually see a need for it.


flavorjones:
Please don't merge this yet - working through another issue installing the gem via a git source.

eregon (approved):

PR #1156, opened 2023-07-20
Set newline flag using a visitor for both Ruby and Java nodes

eregon:
* These 3 places is what the CRuby and JRuby parser use.

### DIFF
--- a/bin/parse
+++ b/bin/parse
@@ -9,7 +9,10 @@ $:.unshift(File.expand_path("../lib", __dir__))
 require "yarp"
 
 if ARGV[0] == "-e"
-  pp YARP.parse(ARGV[1])
+  result = YARP.parse(ARGV[1])
 else
-  pp YARP.parse_file(ARGV[0])
+  result = YARP.parse_file(ARGV[0])
 end
+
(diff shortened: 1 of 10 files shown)


eregon:
> 1. Can you explain the decision not to implement this in the C API as well?

I think it's not needed, because this stuff is done (at least partly) by `compile.c`.
`parse.y` also marks every child of `stmt` and `if/unless` with `newline_node()`, but I'd guess that can likely be done in `compile.c` too.

> 2. Why not turn on this behavior by default?

Because it's extra work and causes an extra visit of the tree, at least the way it's done currently.
Also it's not super-optimized for Ruby due to e.g. calling `child_nodes`.
And most Ruby API users are probably unlikely to need this information.

Maybe we should do this in the parser directly (mentioned in https://github.com/ruby/yarp/issues/1099#issuecomment-1643877726).
I needed to make progress in TruffleRuby so I finished this way as I know it works.
Doing it directly in the parser might lead to some different nodes being marked due to the order of visiting nodes (needs to be checked). OTOH it would avoid duplication of the logic and it would be available in C too (at the cost of adding a field per node struct).

PR #1162, opened 2023-07-21
Symbols with Strings (:"foo") are SymbolNodes unless interpolated

### LINKED ISSUE #1120: InterpolatedSymbolNode even when not interpolated
```
$ bin/parse -e ':"foobar"'
#<YARP::ParseResult:0x00007f44a7833ef0
 @comments=[],
 @errors=[],
 @value=
  ProgramNode(0...9)(
    [],
    StatementsNode(0...9)(
      [InterpolatedSymbolNode(0...9)((0...2), [StringNode(2...8)(nil, (2...8), nil, "foobar")], (8...9))]
    )
  ),
 @warnings=[]>
```

But there is no actual interpolation here, it's just a Symbol literal using `"`.
Currently we're working around by detecting the pattern in TruffleRuby, but I think this should just be a SymbolNode.

jemmaissroff:
Prior to this commit, all strings within symbols would create InterpolatedSymbolNodes regardless of whether they were interpolated or not. With this commit, only actually interpolated nodes create InterpolatedSymbolNodes.

Fixes #1120


eregon (approved):
Thanks!
The CI seems to have caught some leak though.

src/yarp.c
```diff
@@ -9073,26 +9100,39 @@ parse_symbol(yp_parser_t *parser, yp_lex_mode_t *lex_mode, yp_lex_state_t next_s
...
 
-    // If we weren't in a string in the previous check then we have to be now.
-    assert(lex_string);
-
-    if (lex_interpolation) {
-        yp_interpolated_symbol_node_t *interpolated = yp_interpolated_symbol_node_create(parser, &opening, NULL, &opening);
+    if (can_be_interpolated) {
+        // Create a node_list first. We'll use this to check if it should be an InterpolatedSymbolNode
+        // or a SymbolNode
+        yp_node_list_t *node_list = malloc(sizeof(yp_node_list_t));
```
eregon, line 9106: This is not freed. Maybe it could be stack-allocated instead?

PR #1164, opened 2023-07-21
Eliminate warning: '_XOPEN_SOURCE' macro redefined

tenderlove:
This commit eliminates the _XOPEN_SOURCE redefined warning.

YARP has 3 different ways to be compiled:

1. As part of CRuby
2. As a C extension gem
3. As a library to be linked against

CRuby defines XOPEN_SOURCE as well as HAVE_MMAP etc as part of its configuration process.  In other words, when you run `./configure` in Ruby on a target system, it will check for mmap etc and put those macros in ruby/config.h.  When we mirror YARP to Ruby, we rewrite `yarp/config.h` to simply include `ruby/config.h` so that YARP will use the same HAVE_* macros

When C extensions are compiled, Ruby will pass compilation flags to the C extension that is being compiled.  One of those compilations flags is `-D_XOPEN_SOURCE`.  This clashes with the header file we generate with via the M4 scripts.  Actually this is surfacing a bug because we were generating `yarp/config.h` on the system building the YARP gem and *not* on the target system compiling the gem.  This means that if we build the gem on a system that supports mmap, but someone installs the gem on a system that doesn't have mmap, there would be a bug.  To handle this correctly, we _must_ test for the existence of mmap on the target system at gem compile time.  To accommodate this, I've added `have_func("mmap")` etc to `ext/yarp/extconf.rb`.  Those definitions are passed to the compiler at gem compile time.  The gem is still compiled with _XOPEN_SOURCE enabled because as I mentioned earlier: Ruby will automatically pass that flag (as Ruby is compiled with that flag).

Finally, I changed the M4 scripts to pass XOPEN_SOURCE as well as HAVE_MMAP, etc via the command line as well.  This means when we build the library, the library will have XOPEN_SOURCE defined, but it doesn't conflict with Ruby's XOPEN_SOURCE definition.


eregon:
> Actually this is surfacing a bug because we were generating `yarp/config.h` on the system building the YARP gem and _not_ on the target system compiling the gem. This means that if we build the gem on a system that supports mmap, but someone installs the gem on a system that doesn't have mmap, there would be a bug. To handle this correctly, we _must_ test for the existence of mmap on the target system at gem compile time.

AFAIK this should already work fine because running `./configure` will regenerate that config.h header, isn't it?

eregon (changes requested):
I think this change breaks compilation outside of CRuby, i.e., when using the Makefile here to build librubyparser none of the HAVE_* macros will be set.

eregon:
What defines _XOPEN_SOURCE when building YARP in CRuby, besides CRuby itself?
AFAIK the `./configure` is not used there (nor any autotools stuff from YARP), so it should not be defined by YARP in that case.

Do you have a failing log? What's in `config.h` for that case?

eregon:
Is the issue when doing `gem install yarp` on a already-built CRuby?
And that would cause the warning?

`_XOPEN_SOURCE` was added in https://github.com/ruby/yarp/commit/dc856cd9d854ed180c695c6c3efe8185b23f9d3b#diff-2fe1922dfc942181cf09b204d19fd538391e31feaa5ebdc0a453cac4b32316cc for strnlen() and strncasecmp().
We don't use `strncasecmp` anymore.
For strnlen() we still need it I think when building in this repo from the Rakefile+Makefile.
But we can do
```c
#ifndef _XOPEN_SOURCE
#define _XOPEN_SOURCE 700
#endif
```
in `defines.h` as it used to be, that should avoid the warning and still define it before `<string.h>` is included.

For the two functions we check:
* `mmap` doesn't need _XOPEN_SOURCE.
* `snprintf` probably doesn't need _XOPEN_SOURCE because it's C99.

eregon:
Given we only check for mmap and snprintf maybe we should simplify by removing autotools entirely.
mmap() should be available on all systems except Windows, where we could just use `#ifndef _WIN32`.
Probably `snprintf` is always defined since we compile with `-std=c99`. If it's not on Windows we could exclude it there the same way.

eregon:
In hindsight I think autotools is clearly not worth all the complexity and headaches it brings for YARP.
So I think the best solution for YARP is to remove autotools and function availability checks entirely, and just use `#ifndef _WIN32`.

tenderlove:
I'm not against removing autotools, but from my perspective we have 3 different ways to compile YARP:

1. As part of Ruby
2. As a gem
3. As a C library

I would like YARP to leverage the configuration tools for each of these environments:

1. The configuration header file generated by Ruby
2. The configuration detected by extconf.rb
3. The configuration generated by autotools

As I said, I'm not against removing autotools, but it seems like the appropriate tool for that particular compilation environment.

eregon:
I made a PR to simply not define _XOPEN_SOURCE anymore, since that is no longer needed: https://github.com/ruby/yarp/pull/1165

Regarding the 3 ways, one issue with that is we need to duplicate the list of functions to check in 3 places (YARP configure.ac, YARP extconf.rb and CRuby configure.ac), that seems prone to get out of sync.
Also it simplifies quite a few things if there is no configure step for YARP.

eregon:
> AFAIK this should already work fine because running `./configure` will regenerate that config.h header, isn't it? (I'm thinking gem installing YARP here)

Indeed, because the config.h is not included in the packaged gem:
https://github.com/ruby/yarp/blob/c23db977a5f831ddc9548bd18b26638d77a9dc4e/tasks/check_manifest.rake#L36
And `./configure` is run when compiling the gem C extension:
https://github.com/ruby/yarp/blob/main/ext/yarp/extconf.rb#L67

tenderlove:
> AFAIK this should already work fine because running ./configure will regenerate that config.h header, isn't it? (I'm thinking gem installing YARP here)

We don't run `./configure` when we install the gem, we just use `extconf.rb`

eregon:
@tenderlove We do, specifically the extconf.rb runs `configure`, see just above:
https://github.com/ruby/yarp/blob/c23db977a5f831ddc9548bd18b26638d77a9dc4e/ext/yarp/extconf.rb#L67

But only when really installing the gem, not on `bundle exec rake` in this repo, because in that case `librubyparser.a` and `librubyparser.so` are already built and so these checks skip the configure + building:
https://github.com/ruby/yarp/blob/c23db977a5f831ddc9548bd18b26638d77a9dc4e/ext/yarp/extconf.rb#L25
https://github.com/ruby/yarp/blob/c23db977a5f831ddc9548bd18b26638d77a9dc4e/ext/yarp/extconf.rb#L34

Which is explains why we don't see the configure output a second time on `bundle exec rake`.

I also verified on real gem installation: `gem install --local --verbose pkg/yarp-*.gem`
That does contain:
```
sh configure
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether the compiler supports GNU C... yes
checking whether gcc accepts -g... yes
checking for gcc option to enable C11 features... none needed
checking for mmap... yes
checking for snprintf... yes
configure: creating ./config.status
config.status: creating Makefile
config.status: creating include/yarp/config.h
```

tenderlove:
> @tenderlove We do, specifically the extconf.rb runs `configure`, see just above:

I see.  I should have read the code more closely 🤦🏻‍♀️.

Is this actually a good idea though?  I'm worried that the `configure` script we generate at build time could end up being platform specific (in fact the extconf already has comments wrt Windows and running `configure`).

Regardless, I'm going to close this because the other PR removed `XOPEN_SOURCE` which eliminates the warning (and that's what I really cared about).

eregon:
> Is this actually a good idea though? I'm worried that the `configure` script we generate at build time could end up being platform specific (in fact the extconf already has comments wrt Windows and running `configure`).

Ruby release tarballs do the same (e.g. https://cache.ruby-lang.org/pub/ruby/3.2/ruby-3.2.2.tar.gz), they ship `configure`. So that must mean `configure` is portable.
(I had the same question initially)

PR #1169, opened 2023-07-25
Fix locations for comments, warnings, and errors

haldun:
Fixes https://github.com/ruby/yarp/issues/1167


jemmaissroff (approved):
Thanks for the fix!

PR #1170, opened 2023-07-25
Allow JRuby to run Rake

enebo:
This PR allows:
  1. JRuby to run rake tasks
  2. Allow Java to compile any generated template code

For near term use this means for development work in YARP I can change a template and know with a rake run that I generated something which is at least valid Java.  Note: The yarp.jar created is just put into tmp since there is no immediate plan for JRuby to consume this code as a jar (it will be imported into JRuby repo).

Longer term (when fiddle impl of the gem side of YARP has landed) it will allow JRuby to test like MRI does.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,11 +5,13 @@ require "rake/extensiontask"
 require "rake/testtask"
 require "rake/clean"
 require "rdoc/task"
-require "ruby_memcheck"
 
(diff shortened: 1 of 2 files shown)


kddnewton:
@enebo could you add a CI check for JRuby that just builds?

PR #1171, opened 2023-07-27
WIP - Introduce contextually parsing programs vs evals

enebo:
This is more or less the code I used in my POC in JRuby to parse evals.  Evals depend on parent variable scopes and will produce a different syntax tree.

Questions:
  1. How does MRI compile evals currently?  I cannot find anything.
  2. This passes in a char * of data.  It does not encode the variables we pass in because the system calling this already knows.  Is this adequate though?
  3. Can I get guidance on how best to test this?

### DIFF
--- a/include/yarp.h
+++ b/include/yarp.h
@@ -51,7 +51,7 @@ YP_EXPORTED_FUNCTION void yp_parser_register_encoding_decode_callback(yp_parser_
 YP_EXPORTED_FUNCTION void yp_parser_free(yp_parser_t *parser);
 
 // Parse the Ruby source associated with the given parser and return the tree.
-YP_EXPORTED_FUNCTION yp_node_t * yp_parse(yp_parser_t *parser, bool eval);
+YP_EXPORTED_FUNCTION yp_node_t * yp_parse(yp_parser_t *parser);
 
 // Pretty-prints the AST represented by the given node to the given buffer.
 YP_EXPORTED_FUNCTION void yp_prettyprint(yp_parser_t *parser, yp_node_t *node, yp_buffer_t *buffer);
--- a/include/yarp.h
+++ b/include/yarp.h
@@ -61,7 +61,7 @@ YP_EXPORTED_FUNCTION void yp_serialize(yp_parser_t *parser, yp_node_t *node, yp_
 
 // Parse and serialize the AST represented by the given source to the given
 // buffer.
-YP_EXPORTED_FUNCTION void yp_parse_serialize(const char *source, size_t size, yp_buffer_t *buffer, const char *parent_scopes);
+YP_EXPORTED_FUNCTION void yp_parse_serialize(const char *source, size_t size, yp_buffer_t *buffer, const char *metadata);
 
 // Returns a string representation of the given token type.
 YP_EXPORTED_FUNCTION const char * yp_token_type_to_str(yp_token_type_t token_type);
--- a/include/yarp.h
+++ b/include/yarp.h
@@ -51,7 +51,7 @@ YP_EXPORTED_FUNCTION void yp_parser_register_encoding_decode_callback(yp_parser_
 YP_EXPORTED_FUNCTION void yp_parser_free(yp_parser_t *parser);
 
 // Parse the Ruby source associated with the given parser and return the tree.
-YP_EXPORTED_FUNCTION yp_node_t * yp_parse(yp_parser_t *parser);
+YP_EXPORTED_FUNCTION yp_node_t * yp_parse(yp_parser_t *parser, bool eval);
 
 // Pretty-prints the AST represented by the given node to the given buffer.
 YP_EXPORTED_FUNCTION void yp_prettyprint(yp_parser_t *parser, yp_node_t *node, yp_buffer_t *buffer);
@@ -61,7 +61,7 @@ YP_EXPORTED_FUNCTION void yp_serialize(yp_parser_t *parser, yp_node_t *node, yp_
 
 // Parse and serialize the AST represented by the given source to the given
 // buffer.
-YP_EXPORTED_FUNCTION void yp_parse_serialize(const char *source, size_t size, yp_buffer_t *buffer);
+YP_EXPORTED_FUNCTION void yp_parse_serialize(const char *source, size_t size, yp_buffer_t *buffer, const char *parent_scopes);
 
 // Returns a string representation of the given token type.
 YP_EXPORTED_FUNCTION const char * yp_token_type_to_str(yp_token_type_t token_type);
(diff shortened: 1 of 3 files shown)


eregon:
Interesting, passing a "serialized" list of local variables in as a single `char*`.
Seems a nice approach in that it's easy to use via FFI/Fiddle/JNI.

enebo:
@eregon  I had only done it this way because I had assumed we wanted the most bare signature possible (I did this a couple of months ago).  Using multi-dimensional array is obviously an option here too.  If there is a need for knowing these variables encoding then this will need something more also.  I am trying to avoid that though :)

src/yarp.c
```diff
@@ -12773,8 +12773,8 @@ parse_expression(yp_parser_t *parser, yp_binding_power_t binding_power, const ch
 }
 
 static yp_node_t *
-parse_program(yp_parser_t *parser) {
-    yp_parser_scope_push(parser, true);
+parse_program(yp_parser_t *parser, bool eval) {
+    yp_parser_scope_push(parser, !eval);
```
kddnewton, line 12777: Instead of passing this in and passing it all of the way through all the `yp_parse` calls, I would prefer if this derived it from the number of scopes currently active. That way we don't have to change the signature everywhere.

kddnewton (changes requested):
I think the general idea of this is great.

I think a serialized format makes a lot of sense for the scopes. However, I would like to extend it to generally just be metadata. We already don't support the filepath, and we need to. We're going to need to support line number eventually as well, in addition to various pragmas like frozen string literal.

In terms of the serialized form, I'm also wary of using platform-specific-width integers. Ideally I think we would use the same serialization techniques we use for the AST, in that we would use variable-length integers. Additionally, instead of null-terminated identifiers, I would prefer we use length+content. For the most part, it should end up being the same size in memory anyway.

You should be able to use `yp_parser_local_add_location` instead of `yp_parser_local_add_token` as well so you can create a location instead of a token.

Since this is going to be the dumping ground for most of the configuration coming into the parser, if you wouldn't mind adding a small markdown file explaining the format, I think that would help a lot.

eregon:
> in that we would use variable-length integers

That would require  newcode in each caller language to serialize to a protobuf varint. That doesn't exist for Java and Ruby for instance, only deserialization currently.
That could be added, but I think varints make sense when trying to save space, but do not make much sense when calling with a small temporary buffer, the memory savings here would be insignificant while making the function much harder to call.

Agreed we shouldn't just use native int/long, how about `uint32_t` then?

enebo:
> I think the general idea of this is great.
> 
> I think a serialized format makes a lot of sense for the scopes. However, I would like to extend it to generally just be metadata. We already don't support the filepath, and we need to. We're going to need to support line number eventually as well, in addition to various pragmas like frozen string literal.

Ok I will also change the name at the same time to reflect the notion this will be more than just extra scope data.  metadata sounds good.
 
> In terms of the serialized form, I'm also wary of using platform-specific-width integers. Ideally I think we would use the same serialization techniques we use for the AST, in that we would use variable-length integers. Additionally, instead of null-terminated identifiers, I would prefer we use length+content. For the most part, it should end up being the same size in memory anyway.

 1. as @eregon said I think uint32_t is reasonable here since it is fixed and I cannot imagine a platform which cannot process that.
 2. I will change the format to be len, data.  I should have done that initially since I already knew some encodings will use 2 bytes to terminate a string (and I like Rust :) ).
 3. As per @eregon comment variable width is a bigger ask than just picking a value.  Also this incoming data size is not a concern so I think uint32_t would simplify the requirements.
 
> 
> You should be able to use `yp_parser_local_add_location` instead of `yp_parser_local_add_token` as well so you can create a location instead of a token.

ok.  I did not know about location.  Will change that. 


> Since this is going to be the dumping ground for most of the configuration coming into the parser, if you wouldn't mind adding a small markdown file explaining the format, I think that would help a lot.

Yeah my comment was a bit sparse.  Markdown coming up.

kddnewton:
That all sounds good. I'm fine with `uint32_t`.

enebo:
@kddnewton I did not make a separate markdown file but it seemed like we either need it more local (like I did in this commit) or broader (document how a new implementation would fully do the whole process of using serialization).  I am happy to do a start on the broader document but I also feel like we are iterating on a design so it perhaps is not worth drawing that up until we are done-ish.

src/yarp.c
```diff
@@ -12931,7 +12974,7 @@ yp_parser_free(yp_parser_t *parser) {
 // Parse the Ruby source associated with the given parser and return the tree.
 YP_EXPORTED_FUNCTION yp_node_t *
 yp_parse(yp_parser_t *parser) {
-    return parse_program(parser);
+  return parse_program(parser);
```
kddnewton, line 12977: I think this was probably unintentional?

src/yarp.c
```diff
@@ -12794,6 +12794,49 @@ parse_program(yp_parser_t *parser) {
...
+//
+// ```text
+// [number_of_variable_scopes: uint32_t,
+//   [number_of_variables: uint32_t,
+//      [data_length: uint32_t, data: char*]*
+//   ]*
+// ]
+// ```
+static void
+yp_process_metadata(yp_parser_t *parser, const char *metadata) {
```
kddnewton, line 12818: Just in terms of naming, could you call this `yp_parser_metadata`?

include/yarp.h
```diff
@@ -61,7 +61,7 @@ YP_EXPORTED_FUNCTION void yp_serialize(yp_parser_t *parser, yp_node_t *node, yp_
 
 // Parse and serialize the AST represented by the given source to the given
 // buffer.
-YP_EXPORTED_FUNCTION void yp_parse_serialize(const char *source, size_t size, yp_buffer_t *buffer);
+YP_EXPORTED_FUNCTION void yp_parse_serialize(const char *source, size_t size, yp_buffer_t *buffer, const char *parent_scopes);
```
kddnewton, line 64: I think this should be called `metadata`.

kddnewton (changes requested):
Great! This is really good, not too many changes left.

src/yarp.c
```diff
@@ -12794,6 +12794,49 @@ parse_program(yp_parser_t *parser) {
...
+    const char *p = metadata;
+    uint32_t number_of_scopes = (uint32_t) *p;
```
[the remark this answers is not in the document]
enebo: I can add this but does this also mean when I am making this blob of data I have to guarantee alignment?  It was not how I originally wrote it (although to be fair I was only using 1 char for the int value so alignment didn't matter).

src/yarp.c
enebo: I was using emacs before (I do not write C very much) and I switched to an IDE.  I expected it would just indent.  I guess not.

include/yarp.h
enebo: doh!

src/yarp.c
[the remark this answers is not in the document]
enebo: I just wrote a question but I realize now that your method shows how to read from aligned and unaligned so my question made no sense.  I will add this.  Is it ok for this to be in same file or should this be in util/ some place?

kddnewton (approved):
This looks really good @enebo, thanks!

PR #1172, opened 2023-07-29
Move unescape and file reading logic to librubyparser

eregon:
* So it can be shared by the various users of librubyparser.
* Needed for https://github.com/ruby/yarp/pull/1112

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -16,7 +16,7 @@ end
 task compile: :make
 task compile_no_debug: :make_no_debug
 
-task default: [:clobber, :compile, :test]
+task default: [:compile, :test]
 
 require_relative "templates/template"
 
--- a/ext/yarp/extension.c
+++ b/ext/yarp/extension.c
@@ -529,30 +529,16 @@ named_captures(VALUE self, VALUE source) {
 // version.
 static VALUE
 unescape(VALUE source, yp_unescape_type_t unescape_type) {
-    yp_string_t string;
-    VALUE result;
+    yp_string_t result;
 
-    yp_list_t error_list;
-    yp_list_init(&error_list);
-
-    const char *start = RSTRING_PTR(source);
-    size_t length = RSTRING_LEN(source);
-
-    yp_parser_t parser;
-    yp_parser_init(&parser, start, length, "");
-
-    yp_unescape_manipulate_string(&parser, start, length, &string, unescape_type, &error_list);
-    if (yp_list_empty_p(&error_list)) {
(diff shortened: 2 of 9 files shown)


Rakefile
```diff
@@ -16,7 +16,7 @@ end
 task compile: :make
 task compile_no_debug: :make_no_debug
 
-task default: [:clobber, :compile, :test]
```
eregon, line 19: If one wants to always clean before, it's possible via `bundle exec rake clobber default`

eregon:
I'll merge this to avoid more conflicts.
FYI @kddnewton we should avoid having logic in extension.c and it should be in librubyparser so it can be reused by Fiddle/FFI/JNI.

kddnewton:
Can you wait a bit for me to review? This is a bigger change.

eregon:
OK. It's mostly moving logic around and replacing `input_t` by `yp_string_t` since that's already part of the public API and so easier to use from e.g. Fiddle.

kddnewton (dismissed):
So it looks like there are a couple of things happening here.

1. Removing `clobber` from the default rake task - I'm fine with making this change, but it's a little arbitrary to be a part of this changeset.
2. Making the extension use strings for mapped files - It looks like you're using owned strings, but that they don't actually release their contents in the `yp_string_free` function. I think this would be a lot easier to understand if you added a new string type: `YP_STRING_MAPPED`. That way you could stick with the existing conventions and add `yp_string_mapped_init(yp_string_t *, const char *filepath)` and then update `yp_string_free` appropriately. It would be ideal if this were a separate PR that you could then rebase off of for this one.
3. Moving the file reading into the library
4. Moving the escaping into the library

Could you move the first 2 into separate PRs and make those changes to the string utility?

kddnewton:
> 1. It's just fixing [7ff99d0](https://github.com/ruby/yarp/commit/7ff99d00fdabad8c2439cf3d570463ab789a7249) ;)

Ahh, I didn't see that go in.

> Re `yp_string_mapped_init`, sounds good, but I don't want a separate PR, that would be a lot of work to rebase this, I'll do the change as an extra commit in this PR.

In that case I'm happy to do the rebase for you if you don't mind giving me access to your fork. Or I can push the branch on this repository. I know it's annoying, but I'd like to keep the history as clean as possible.

eregon:
I've pushed commits to use `yp_string_mapped_init` and fix a couple string-related issues.
Feel free to force push to this branch, I have a backup of the branch.

PR #1174, opened 2023-07-31
Add test files to manifest ignore

kddnewton:


tasks/check_manifest.rake
```diff
@@ -40,6 +40,8 @@ task :check_manifest => [:templates, "configure"] do
     include/yarp/config.h
     lib/yarp.{jar,so,bundle}
     tags
+    test.c
+    test.rb
```
eregon, line 44: test.rb was already at line 60 below.
The reason I ignored it there is because `ignore_files` means ignore in any directory, not just at the root, which I'm not sure is desirable.

PR #1178, opened 2023-07-31
Various newline stuff

kddnewton:
This PR starts work to track newline flags inside the parser. It does this by tagging the type field on the node enum with a bit that indicates the field could potentially trigger a `line` event. This is the same way CRuby does it currently. Fortunately this adds no memory overhead since we had plenty of room in the enum.

This PR does _not_ expose this flag in any way at the moment. That's going to be a follow-up PR.

It also splits out the newline test, as that is going to be changed once we start using this and I wanted to be able to run it separately. It also refactors it to remove creating classes on the fly.


templates/include/yarp/ast.h.erb
```diff
@@ -52,6 +52,24 @@ typedef enum {
 <%- end -%>
 } yp_node_type_t;
 
+// We store the node type enum in every node in the tree. We don't have nearly
+// as many node types as we do bits in an enum, so we're going to use the top
+// half of the enum to store flags about the node.
+static const unsigned int YP_NODE_FLAGS_SHIFT = sizeof(yp_node_type_t) * 8 / 2;
```
eregon, line 58: Maybe bit fields could be used for this? Then you wouldn't need to mask manually

eregon:
Just to be sure, you saw https://github.com/ruby/yarp/issues/1099, right?
Maybe we should have a call to sync on this & ideas for it.

eregon:
I was thinking if as in this PR it's as simple as every StatementsNode, IfNode and UnlessNode are marked, then we wouldn't need to mark them explicitly, we could just have a `is_potential_newline()` or so function which just check the type among those 3.

kddnewton:
It looks to be the children of every statements node, all if nodes, and all unless nodes. Then in the compiler it dedups the lines to only put it on the first of the line.

kddnewton:
Honestly I'm iffy about putting this into the parser at all at this point. It's so specific to the bytecode it feels weird having it here.

eregon:
Yeah exactly, let's have some call to figure out what makes sense.
I kind of already tried a bit of everything for https://github.com/ruby/yarp/issues/1099.

templates/include/yarp/ast.h.erb
eregon: This is causing compiler errors on linux-aarch64 like, as found by @andrykonchin:
```
In file included from include/yarp.h:5:0,
                 from src/unescape.c:1:
include/yarp/ast.h:349:50: error: conversion to ‘unsigned int’ from ‘int’ may change the sign of the result [-Werror=sign-conversion]
 static const unsigned int YP_NODE_FLAG_NEWLINE = 1 << YP_NODE_FLAGS_SHIFT;
                                                  ^
include/yarp/ast.h:349:50: error: initializer element is not constant
cc1: all warnings being treated as errors
```

Changing them to `#define` probably works around that.

But I feel it's really unsafe and most likely undefined behavior to set bits of an `enum` field like done in this PR.
And undefined behavior means the program can do random bad things, like segfaulting and assuming wrong things.
Could you try to use [C bit fields](https://www.tutorialspoint.com/cprogramming/c_bit_fields.htm) instead?

PR #1182, opened 2023-08-01
Add JRuby rake CI task

enebo:
Actually runs #1170 any time anything changes so we know the Java generated at least compiles. 

 A later PR will expand this to also run JRuby via the fiddle backend to also verify the gem part is also working on JRuby.


kddnewton:
Great work, thanks @enebo!

PR #1184, opened 2023-08-01
Simplify build script

kddnewton:
There are lots of problems with the way we currently build things. Specifically the interaction between `make`, `rake`, and `mkmf` is very difficult to understand. The dependencies are almost always in some kind of messed up state such that folks end up needing to `rake clobber` randomly to ensure the compiled objects are the actual ones that they want. In `extconf.rb` it only attempts to generate the archive/dylib if the file doesn't exist, which means it's not taking advantage of dependency tracking, so it ends up getting stale frequently. We're using `rake` to shell out to `make`, so it's not clear where you should put various logic. Autotools itself is taking care of some stuff, while we're doing the rest in various other places.

This PR standardizes all of the build logic into `rake` and removes autotools entirely. To accomplish this it no longer shells out to `make` to build the archive/dylib, instead it puts the .c files into `ext/yarp` and copies all of the .h files into `ext/yarp/include`. They are gitignored by default.

## Things we're gaining

* Because everything is in `rake`, all of the dependencies are lined up properly. This means if you change a single file, only that file gets recompiled. It's a much nicer experience.
* Because we have fewer tools, it's a much simpler mental model to deal with. Generally this should make it easier for contributors.
* Because all of the files are present in the `ext/yarp` directory at the time the gem gets built, it's effectively being packaged as a standalone extension. This simplifies the entire build process on client machines.

## Things we're losing

* You can no longer build this on its own without Ruby. (We're not exactly losing something here, you already had to have Ruby to run the templating.)
* This PR doesn't build dynamic libraries at all. To my knowledge no one is using that functionality anyway yet, so we should be okay. The only ways to consume this library now are therefore through CRuby as a standard library, through this project as a gem, or through JRuby/TruffleRuby packaged up with whatever build system they have going. (I think TruffleRuby is using CMake?)
* The `config.h` file is gone. This was causing problems because it was generated on the machine that packaged the gem and not on the machine that was building the gem, meaning the headers could get the wrong values. In investigating this, I realized you could use `mkmf` to do this in a single line anyway, as in: `ruby -rmkmf -e "have_func('mmap'); have_func('snprintf'); print \$defs.join(' ')"`.

## Further work

* I want to remove `Makefile` since it's now only being used for fuzzing and move it to a rake library. It shouldn't be too bad, there are analogous constructs in rake for almost everything in the file today. @stevenjohnstone maybe you could take a look?
* If we need to, we can bring back the shared library support. But we should investigate if that's actually necessary.


tenderlove:
> * The `config.h` file is gone. 

We need `config.h` because when YARP is part of CRuby, it doesn't have a chance to run any `extconf.rb` but still needs to be able to detect the existence of mmap.  @jemmaissroff and I solved this by [writing a `config.h` that simply included Ruby's](https://github.com/ruby/ruby/blob/3b88a0bee841aee77bee306d9d34e587561515cf/tool/sync_default_gems.rb#L422-L423).

I suggest that we either check in an empty `config.h`, or [ask `mkmf` to write a `config.h`](https://github.com/tenderlove/ruby/blob/8b390a456ce7dc7ef942a8a2cfba2a944c3d25bd/lib/mkmf.rb#L1744) that we use.

> This was causing problems because it was generated on the machine that packaged the gem and not on the machine that was building the gem, meaning the headers could get the wrong values. In investigating this, I realized you could use `mkmf` to do this in a single line anyway, as in: `ruby -rmkmf -e "have_func('mmap'); have_func('snprintf'); print \$defs.join(' ')"`.

@jemmaissroff and I fixed this problem via the strategy above, and the `extconf.rb` reran `configure` on the target machine, so I don't think this is actually a problem at the moment.

eregon:
This is from a quick look, I will review in details tomorrow.

My impression is current build system is complex but actually satisfies everything needed and works well.
It likely deserves some documentation to understand e.g. how it works in the various situations (built from `rake`, built from `gem install`, build in CRuby, build in TruffleRuby/JRuby).

> The dependencies are almost always in some kind of messed up state such that folks end up needing to rake clobber randomly to ensure the compiled objects are the actual ones that they want.

I think they are actually fine and we got most/all dependencies right by now.
It used to be a problem at the beginning but not recently in my experience.
Except the recent move of lib/yarp.so to lib/yarp/yarp.so was an issue, but I don't think any build system handles that well. It's good check_manifest catches such an issue.

> This PR doesn't build dynamic libraries at all.

That's a problem. It will be needed when installing this gem on JRuby and TruffleRuby.
Everything is already there, and it already works to a good extent, see #1112, so please let's not break it.

I like the idea of removing autotools (as mentioned in details on https://github.com/ruby/yarp/pull/1164).
However then we need to rely on 0 `HAVE_*` (also in the future), otherwise this breaks feature checking in CRuby and in TruffleRuby/JRuby.
Currently the feature checking in TruffleRuby uses autotools and runs `./configure` and `make`, which is great because there is no duplication, it uses the existing build system (I had to do some amount of work for that in truffleruby but it works well).

> You can no longer build this on its own without Ruby. (We're not exactly losing something here, you already had to have Ruby to run the templating.)

This is a problem I think. Templating is no issue because it can be done when importing, running on existing older CRuby.
But e.g. we cannot afford to run a `extconf.rb` in TruffleRuby to build YARP, because we would have no parser at that point.
I'd much rather use the same Makefile as currently on master, to guarantee same compilation flags, etc.

eregon:
I'll make a PR to document the current build system in a .md file, that should help to clarify the requirements and how it works for the many different ways to build YARP.

I can also make a PR to remove autotools, that's trivial, we just replace the HAVE_* by `#ifndef _WIN32` and it removes an entire layer in the build system + its consequences like having a config.h that's platform-specific. All the headers would then be static headers, no magic.

flavorjones:
@eregon Please let me know if I can help document the build system -- I feel like I inadvertently contributed much of the complexity.

I don't think the extconf.rb should ever be responsible for rebuilding librubyparser if something changed. This file primarily exists to build the gem during installation (although it also is run during development). For this case, instead, developers should be running `bundle exec rake compile`:

```
$ touch src/diagnostic.c 

$ be rake compile
make
compiling build/shared/diagnostic.o
compiling build/static/diagnostic.o
linking build/librubyparser.so
building build/librubyparser.a
cp src/diagnostic.c tmp/x86_64-linux/stage/src/diagnostic.c
/usr/bin/gmake install target_prefix=
/usr/bin/install -c -m 0755 yarp.so /home/flavorjones/code/oss/yarp/lib
cp tmp/x86_64-linux/yarp/3.2.2/yarp.so tmp/x86_64-linux/stage/lib/yarp.so
```

eregon:
> Updated: Now I see what's going on with the static archive not causing a rebuild of the C extension, I think that's fixable.

Indeed. And fixed in https://github.com/ruby/yarp/pull/1195

eregon:
I think one thing we don't need BTW is to support compiling the C extension against the dynamic library, we can always use the static library. The dynamic library is only useful for Fiddle/FFI.

kddnewton:
There are some dependency issues with templates as well. The template files are known only to ruby and not to make. So frequently when you change a template you end up having to clobber to get it correct. I've fixed that in this PR.

eregon:
Documentation for the current build system is at https://github.com/ruby/yarp/pull/1196
That seems also very useful if we change it to quickly understand what changes and how.

eregon:
> There are some dependency issues with templates as well. The template files are known only to ruby and not to make. So frequently when you change a template you end up having to clobber to get it correct. I've fixed that in this PR.

There is no such problem on master AFAIK. For example adding `puts("Hi");` in `templates/src/serialize.c.erb` recompiles everything needed. It works via:
```ruby
task make: [:templates, "Makefile"] do
  sh "make"
end

task templates: TEMPLATES

TEMPLATES.each do |filepath|
  desc "Generate #{filepath}"
  file filepath => ["templates/#{filepath}.erb", "templates/template.rb", "config.yml"] do |t|
```
which regenerates the .c/.h/etc, and the Makefile notices the updated files.

eregon:
> I think one thing we don't need BTW is to support compiling the C extension against the dynamic library, we can always use the static library. The dynamic library is only useful for Fiddle/FFI.

Actually no, on JRuby/TruffleRuby this code path is needed for Fiddle/FFI, and it's likely the simplest if it's part of `extconf.rb`, just the `extconf.rb` would generate a [dummy makefile](https://github.com/rubyjs/mini_racer/blob/c7ed6720fe76da615efbc70a90fd9a85c1f7e69d/ext/mini_racer_extension/extconf.rb#L4) there, and call `make shared` on the root Makefile (notably because there is no gem platform for TruffleRuby, it's the same gem platform as CRuby).

eregon:
> I've also brought back the shared library creation, which can be done now with `rake shared`.

The shared library should also be created on `gem install yarp` on non-CRuby (i.e. on JRuby/TruffleRuby).
And that's probably very hard to do from the extconf.rb.
For that reason I think keeping the Makefile is the cleanest solution, plus the Makefile is reusable by all 6 ways to build YARP (#1196), while a Rakefile basically only works when building from the repository.
An example advantage is for instance identical compilation flags including warnings and optimization levels, which is certainly nice.

README.md
```diff
@@ -42,14 +41,7 @@ The repository contains the infrastructure for both a shared library (librubypar
...
 
-To compile the shared library, you will need:
-
-* A C99 compiler
-* autotools (autoconf, automake, libtool)
-* make
-* Ruby 3.3.0-preview1 or later
-
-Once you have these dependencies, run:
+To compile YARP, you will need a C99 compiler and Ruby 3.0.0 or later. Once you have these dependencies, run:
```
nobu, line 44: Seems to need a gcc-compliant compiler, and GNU make.

ParadoxV5:
May I query about the status of the build system in 2024?
(besides that it’s renamed from `yarp` to `prism` now)

> The shared library should also be created on `gem install yarp` on non-CRuby (i.e. on JRuby/TruffleRuby). And that's probably hard to do from the extconf.rb. For that reason I think keeping the Makefile is the cleanest solution, plus the Makefile is reusable by all 6 ways to build YARP (#1196), while a Rakefile basically only works when building from the repository.

[The `.gemspec` accepts `Rakefile` as a build script.](https://github.com/rubygems/rubygems/blob/v3.5.7/lib/rubygems/ext/builder.rb#L151-L153) From there, you can use [rake-compiler/rake-compiler](https://github.com/rake-compiler/rake-compiler) to continue with `extconf.rb`.

kddnewton:
Hi @ParadoxV5 — the build system is documented here: https://github.com/ruby/prism/blob/main/docs/build_system.md

PR #1192, opened 2023-08-02
Loop post flag

### LINKED ISSUE #1118: Explicit flag for `begin end while (cond)` while loop form
There are two semantically different ways to use `while` loop:
- condition is evaluated first, then loop body
- loop body is executed first, and then condition

It would be helpful to have it reflected explicitly in WhileNode, e.g. with a new attribute.

Examples:

```ruby
while false
  puts 1
end
# => nil
```

```ruby
begin
  puts 1
end while false
# 1
# => nil
```

kddnewton:
Explicit flag for handling `begin; end while foo` and `begin; end until foo`.

Fixes #1118.


config.yml
```diff
@@ -333,6 +333,10 @@ flags:
         comment: "&. operator"
       - name: VARIABLE_CALL
         comment: "a call that could have been a local variable"
+  - name: LoopFlags
+    values:
+      - name: BEGIN_MODIFIER
+        comment: "a loop after a begin statement"
```
eregon, line 339:
```suggestion
        comment: "a loop after a begin statement, so the body is executed first before the condition"
```

eregon (approved):
Looks great

PR #1195, opened 2023-08-02
Automatically rebuild the C extension if librubyparser.a changed

eregon:
* And naturally librubyparser.a is rebuilt if any transitive dependency of it is changed like *.c and *.h.


eregon:
I checked this worked manually by editing yarp.c and adding e.g. `puts("Hi");`, and indeed doesn't recompile before this PR, recompiles after (with `bundle exec rake`).

PR #1196, opened 2023-08-02
Document the build system of YARP

eregon:
cc @flavorjones


docs/build_system.md
```diff
@@ -0,0 +1,70 @@
...
+Then when `mx build` builds TruffleRuby and the `yarp` mx project inside, it:
+* runs `./configure`
+* runs `make`
+
+Then the `yarp bindings` mx project is built, which contains the [bindings](https://github.com/oracle/truffleruby/blob/master/src/main/c/yarp_bindings/src/yarp_bindings.c)
+and links to librubyparser.a (to avoid exporting symbols, so no conflict when installing the yarp gem).
+
+### Building YARP as part of JRuby
+
+TODO, probably similar to TruffleRuby.
```
eregon, line 70: @enebo Could you fill this section? (can be just a comment that I copy/paste for simplicity)

docs/build_system.md
```diff
@@ -0,0 +1,70 @@
+# Build System
+
+There are many ways to build YARP, which means the build system is a bit more complicated than usual.
+
+## General Design
+
+1. Templates are generated by `templates/template.rb`
+2. `autoconf` create `./configure` and `autoheader` creates `config.h.in` (both files are platform-independent)
+3. `./configure` creates `include/yarp/config.h` (which contains `HAVE_*` macros, platform-specific) and the `Makefile`
```
eregon, line 9: Removing autotools would remove these 2 layers, so that's the easiest change to simplify the build system.
Then the only generated files are the templates, that's it, and there is no need to know about autotools and which autotools-related files are platform-independent or plaftform-specific.

eregon:
I'd like to merge this soon-ish (this week ideally), so then any changes we make to the build system is reflected with changes in this document.

docs/build_system.md
```diff
@@ -0,0 +1,74 @@
...
+### Building the yarp gem by `gem install/bundle install`
+
+The gem contains the pre-generated templates, as well as `configure` and `config.h.in`
+When installing the gem, `extconf.rb` is used and that:
+* runs `./configure` which creates the `Makefile` and `include/yarp/config.h`
+* runs `make build/librubyparser.a`
+* compiles the C extension with mkmf
+
+When installing the gem on JRuby and TruffleRuby, no C extension is built, so instead of the last step,
+there is Ruby code using Fiddle which uses `librubyparser.so/dylib/dll`
```
flavorjones, line 35: The presence of "/" may be confused with path separators, instead let's try this:

```suggestion
there is Ruby code using Fiddle which uses `librubyparser.{so,dylib,dll}`
```

docs/build_system.md
```diff
@@ -0,0 +1,74 @@
+# Build System
+
+There are many ways to build YARP, which means the build system is a bit more complicated than usual.
+
+## General Design
+
+1. Templates are generated by `templates/template.rb`
+2. `autoconf` creates `./configure` and `autoheader` creates `config.h.in` (both files are platform-independent)
+3. `./configure` creates `include/yarp/config.h` (which contains `HAVE_*` macros, platform-specific) and the `Makefile`
+4. The `Makefile` compiles both `librubyparser.a` and `librubyparser.so/dylib/dll` from the `*.c` and `*.h` files
```
flavorjones, line 10:
```suggestion
4. The `Makefile` compiles both `librubyparser.a` and `librubyparser.{so,dylib,dll}` from the `src/**/*.c` and `include/**/*.h` files
```

docs/build_system.md
enebo: @eregon I will once I have something relevant to put here.  My playground work was manually copying files.  I probably will be doing something similar with importing of org.yarp as a tool but it doesn't exist yet (I will be restarting from main once metadata PR lands).

PR #1198, opened 2023-08-03
Serialize errors and warnings

andrykonchin:
Extend serialization format to support dumping syntax errors and warnings:

```
<count of errors>
<length in bytes> <message bytes> <location start> <location length>
...
<count of warnings>
<length in bytes> <message bytes> <location start> <location length>
...
```

So now only error/warning messages are dumped without location.

cc @eregon


templates/java/org/yarp/Loader.java.erb
```diff
@@ -90,6 +94,23 @@ public class Loader {
         return string;
     }
 
+    private String[] loadStringsSequence() {
+        int count = loadVarInt();
+
+        assert count >= 0;
+
+        String[] strings = new String[count];
+        Charset charset = Charset.forName("US-ASCII");
```
eregon, line 103: There is `StandardCharsets.US_ASCII` to avoid the extra lookup.
This probably deserves a comment like `// all warning and error messages only contain ASCII characters`

docs/serialization.md
```diff
@@ -15,7 +15,15 @@ The header is structured like the following table:
...
-| | the encoding name |
+| string | the encoding name |
+| varint | number of errors |
+| varint | byte length of error |
+| string | error string, as byte[] in source encoding |
+| ... | more errors |
+| varint | number of warnings |
+| varint | byte length of warning |
+| string | warning string, as byte[] in source encoding |
+| ... | more warnings |
```
eregon, line 26: I have been thinking a nicer way to document this would be to have some "meta types" for the 1st column that we would declare above this table.
So for example we could have `string` which is `varint size + bytes` and we could then describe errors as `varint number of errors + string* errors`.
And once we dump location too, we could have `diagnostic` which is `location location, string message` and location is `varint start, varint length`.

eregon:
> I'd be fine moving it to the front of the buffer, or with moving these things to the end. Either way I'd just rather they lived together. Thoughts?

The idea is if there are errors we might potentially want to not even deserialize the AST at all, and not even look at the constant table. So for that reason it feels like errors should come before AST and contants table.
To be fair, that's not how it works currently with this PR, we return a ParseResult which includes the AST YARP gives us, but maybe we should change that so e.g. the AST is null in that case, or even throws a Java exception if any error.

I suppose we could move the constant table before the AST, but that's then a second breaking change at the same time, which might make it harder to adapt to this change.

eregon:
We need the location as well for warnings and errors, could you serialize them as `varint startOffset`, `varint length`?
And then getErrors()/getWarnings() return `ParseError[]`, `ParseWarning[]` like in the Ruby API?

eregon (approved):

PR #1204, opened 2023-08-04
Rename RescueNode#exception to RescueNode#reference

### LINKED ISSUE #1160: The `exception` and `exceptions` fields in `RescueNode` are confusing a bit
`RescueNode` represents a `rescue` section and has `exception` and `exceptions` fields for optional exception classes
 and optional exception variable:

```
...
rescue <exceptions> => <exception>
```

These names look a bit confusing - it might be not obvious for someone what exactly they mean and it will be helpful to make them more precise.

Not sure if there are common used terms for these things. I could propose the following alternatives only:

### `exceptions`:
- exception classes (but it might be not only classes but some expression, e.g. `*exceptions_array`)
- exceptions list
- rescued exceptions
- expected exceptions

### `exception`:
- exception variable (but it might be also constant or method call (`a.foo` or `a[:foo]`))
- just "assignable" as far as semantic is to assign exception to the provided name

### Declaration of the `RescueNode` node:

```yaml
  - name: RescueNode
    child_nodes:
      - name: keyword_loc
        type: location
      - name: exceptions
        type: node[]
      - name: operator_loc
        type: location?
      - name: exception
        type: node?
      - name: statements
        type: node?
        kind: StatementsNode
      - name: consequent
        type: node?
        kind: RescueNode
```

kddnewton:
Fixes #1160

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1830,7 +1830,7 @@ nodes:
         type: node[]
       - name: operator_loc
         type: location?
-      - name: exception
+      - name: reference
         type: node?
       - name: statements
         type: node?
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -3517,7 +3517,7 @@ yp_rescue_node_create(yp_parser_t *parser, const yp_token_t *keyword) {
         },
         .keyword_loc = YP_LOCATION_TOKEN_VALUE(keyword),
         .operator_loc = YP_OPTIONAL_LOCATION_NOT_PROVIDED_VALUE,
-        .exception = NULL,
+        .reference = NULL,
         .statements = NULL,
         .consequent = NULL,
         .exceptions = YP_EMPTY_NODE_LIST
@@ -3531,11 +3531,11 @@ yp_rescue_node_operator_set(yp_rescue_node_t *node, const yp_token_t *operator) 
     node->operator_loc = YP_OPTIONAL_LOCATION_TOKEN_VALUE(operator);
 }
 
-// Set the exception of a rescue node, and update the location of the node.
+// Set the reference of a rescue node, and update the location of the node.
 static void
-yp_rescue_node_exception_set(yp_rescue_node_t *node, yp_node_t *exception) {
-    node->exception = exception;
-    node->base.location.end = exception->location.end;
+yp_rescue_node_reference_set(yp_rescue_node_t *node, yp_node_t *reference) {
+    node->reference = reference;
+    node->base.location.end = reference->location.end;
 }
 
 // Set the statements of a rescue node, and update the location of the node.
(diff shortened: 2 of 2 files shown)


eregon (approved):

kddnewton:
Okay, let me know when it's ready.

kddnewton:
I'm closing out for the weekend so going to merge this. I really want to figure this out going forward, otherwise it's impossible to make breaking changes without waiting for y'all to make changes. Maybe we could mark the check as optional.

enebo:
> @kddnewton The usual strategy for this is to wait that TruffleRuby adopts the change, and then when that's merged we can merge here and no CI breaks. So if this is not urgent I would wait for that.

@eregon @kddnewton I can see the benefits of wanting people to see breaking changes on TR for bug fixes, but for development work this creates a backwards dependency.  

If I submit a PR here (X) to change something then if we followed your process and I was willing to submit a PR (Y) to your repo first (and let's just pretend that is no work but just something which just happens); then I think this assumes YARP development is some single linear process.

So Y is landed, then someone in yarp fixes a different bug with a PR (Z), then we land X.  In this case Z is red and has nothing to do with the PR you fixed with Y.  That will be confusing to the author of Z. Other scenarios (WIP PRs, language level changes needing githubaction changes depending on language level TR has) but right now this one feels like it would be common. 

If we adopted the same process for JRuby,  then we have to have two consuming repositories do PRs before the PR can land on YARP.  That ignores some of YARPs CI will catch things which would be hard to test locally (so it would be nice to see that before two PRs land in other projects).

I wonder if there is some other way of achieving the goals here.  Correct me if I am wrong @eregon but you want to make sure if someone breaks TR from a fix there is a chance for them to notice that and reject the PR (or at least consult about what happened).

eregon:
I think the main issue is once it's red nobody will notice if there is another incompatible change or a bug slipping in.
And so it might be hard to get back to green if that happens (e.g. if Java deserialization is broken and has multiple bugs it quickly becomes challenging to fix it, that already happened before the check was there).

I think the existing process of merging in TruffleRuby first and in YARP right after works well for most cases.

The main alternative I see is to avoid breaking that check, for instance in this case by keeping the old field or so. It often feels hacky or overhead though or it might not be reasonably possible.

I also thought at some point to only do the "can it deserialize cleanly" check (maybe running it on JRuby if it doesn't depend on specific YARP node names and field names). But that's weaker in terms of testing than actually checking it works for a real usage of YARP, i.e., as TruffleRuby uses it. I think so far this usage is the only one that really checks the AST makes sense and has the relevant information. Nothing else in CI really uses YARP nodes for execution yet, e.g., if there was an incorrect node as the receiver of a CallNode, I think nothing would notice it except that check.

If we do a bunch of field/node renames then it probably makes sense to group them, and similarly for other related breaking changes, so there is less often the need to merge things in a particular order.

PR #1209, opened 2023-08-07
Add filepath to yp_parse_serialize metadata

kddnewton:
This adds the ability to pass the filepath through the yp_parse_serialize function in the metadata. It also adds documentation about this new API and fixes a couple of places that were calling it with the old signature. Finally, it adds a specific test to ensure yp_parse_serialize doesn't break.

@enebo I've added the filepath to the beginning of the metadata. Let me know if you have any questions on it!

### DIFF
--- a/docs/serialization.md
+++ b/docs/serialization.md
@@ -75,7 +75,7 @@ void yp_buffer_free(yp_buffer_t *);
 
 // Parse and serialize the AST represented by the given source to the given
 // buffer.
-void yp_parse_serialize(const char *, size_t, yp_buffer_t *);
+void yp_parse_serialize(const char *, size_t, yp_buffer_t *, const char *);
 ```
 
 Typically you would use a stack-allocated `yp_buffer_t` and call `yp_parse_serialize`, as in:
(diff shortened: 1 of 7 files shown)


eregon:
In https://github.com/ruby/yarp/pull/1112/files#diff-d096326212628ad464933e1a1976189b989f7498ff1a24065a19791b03d8b496 I added the filepath as the 3rd argument, which is consistent with `yp_parser_init()`.
I think that's so much simpler, the filepath is also needed for non-eval parsing, i.e., it should always be provided.
A default of empty filepath also seems worse than `<unknown>` (e.g. much harder to understand if the output is like `:3 ParseError`).

kddnewton:
We're going to change `yp_parser_init` as well to accept a struct of options because we're going to need much more than just the filepath. We'll need filepath, line number, encoding, frozen string literal, any other compilation options that impact parsing, the local scopes (for the non serialized path), etc. I'd rather be consistent and have a single parameter that contains all of the metadata.

As for the filepath, it's on the node for convenience with the client libraries. If you're not using it for truffleruby that's fine!

enebo:
@kddnewton late to the game but this looks good.

PR #1210, opened 2023-08-07
Changes necessary for syncing to CRuby

jemmaissroff:
A few minor changes necessary for syncing this to CRuby

### DIFF
--- a/src/util/yp_string.c
+++ b/src/util/yp_string.c
@@ -125,7 +125,8 @@ yp_string_mapped_init(yp_string_t *string, const char *filepath) {
     // the source to a constant empty string and return.
     if (file_size == 0) {
         CloseHandle(file);
-        yp_string_mapped_init_internal(string, "", 0);
+        char empty_string[] = "";
+        yp_string_mapped_init_internal(string, empty_string, 0);
         return true;
     }
 
@@ -171,7 +172,8 @@ yp_string_mapped_init(yp_string_t *string, const char *filepath) {
 
     if (size == 0) {
         close(fd);
-        yp_string_mapped_init_internal(string, "", 0);
+        char empty_string[] = "";
+        yp_string_mapped_init_internal(string, empty_string, 0);
         return true;
     }
 
--- a/templates/include/yarp/ast.h.erb
+++ b/templates/include/yarp/ast.h.erb
@@ -55,7 +55,7 @@ typedef enum {
 // We store the node type enum in every node in the tree. We don't have nearly
 // as many node types as we do bits in an enum, so we're going to use the top
 // half of the enum to store flags about the node.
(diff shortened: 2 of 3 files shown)


src/util/yp_string.c
```diff
@@ -125,7 +125,8 @@ yp_string_mapped_init(yp_string_t *string, const char *filepath) {
     // the source to a constant empty string and return.
     if (file_size == 0) {
         CloseHandle(file);
-        yp_string_mapped_init_internal(string, "", 0);
```
jemmaissroff, line 128: CRuby warns that `""` is a `const char[1]` not a `const char*`

templates/include/yarp/ast.h.erb
```diff
@@ -55,7 +55,7 @@ typedef enum {
 // We store the node type enum in every node in the tree. We don't have nearly
 // as many node types as we do bits in an enum, so we're going to use the top
 // half of the enum to store flags about the node.
-static const unsigned int YP_NODE_FLAGS_SHIFT = sizeof(yp_node_type_t) * 8 / 2;
```
jemmaissroff, line 58: `sizeof(yp_node_type_t) * 8 / 2` wasn't being evaluated at compile time. Accordingly, `YP_NODE_FLAG_NEWLINE` wasn't being evaluated at compile time, making it not const. 

Changing to use a macro means `YP_NODE_FLAGS_SHIFT` will be replaced at compile time in `YP_NODE_FLAG_NEWLINE` so it'll still be a `const `

test/parse_test.rb
```diff
@@ -3,7 +3,11 @@
 require "yarp_test_helper"
 
 # It is useful to have a diff even if the strings to compare are big
-Test::Unit::Assertions::AssertionMessage.max_diff_target_string_size = 5000
+# However, ruby/ruby does not have a version of Test::Unit with access to
+# max_diff_target_string_size
+if defined?(Test::Unit::Assertions::AssertionMessage)
```
jemmaissroff, line 8: Seems like CRuby doesn't have access to `AssertionMessage`. Not sure why `tool/lib/test/unit/assertions.rb` seems dated, but this snippet is a debugging optimization anyway, so not necessary in the sync

kddnewton (approved):

templates/include/yarp/ast.h.erb
eregon: https://github.com/ruby/yarp/pull/1178#discussion_r1285872849

src/util/yp_string.c
eregon: What if we make `yp_string_mapped_init_internal` take `const char*`? Currently it takes `char*`

templates/include/yarp/ast.h.erb
eregon: So I think YP_NODE_FLAG_NEWLINE needs to be a macro too

PR #1211, opened 2023-08-07
Fix string concat parsing

### LINKED ISSUE #894: Concatenating thousands strings breaks parsing
YARP throws a stack level too deep with many strings concatenated together: 

```
"a""a""a"... etc
```

Running this file breaks:

```
#!/usr/bin/env ruby

$:.unshift(File.expand_path("lib", __dir__))
require "yarp"

YARP.parse('"a"' * 100_000)
```

kddnewton:
Loop instead of recursing. Fixes #894.

This is possible because of #1207, which removed the recursion when creating the tree.


test/snapshots/unparser/corpus/semantic/dstr.txt
```diff
@@ -198,11 +198,11 @@ ProgramNode(0...608)(
        )
      ),
      StringConcatNode(552...560)(
-       StringNode(552...554)((552...553), (553...553), (553...554), ""),
-       StringConcatNode(555...560)(
-         StringNode(555...557)((555...556), (556...556), (556...557), ""),
-         StringNode(558...560)((558...559), (559...559), (559...560), "")
-       )
+       StringConcatNode(552...557)(
```
jemmaissroff, line 201: Not sure I follow why this changed? It was incorrect before?

kddnewton: It was not necessarily incorrect so much as just another way to represent it (right-associative vs left-associative). It's moot at the end of the day because everyone that consumes these nodes is going to concat them together anyway.

It's the difference between parsing:

```ruby
"a" + "b" + "c"
```

as:

```ruby
("a" + "b") + "c"
```

vs

```ruby
"a" + ("b" + "c")
```

Because of the way the loop works, it happened to change the associativity, but I don't think it matters for our use case. It might even be better, because it's more consistent with how calls work.

jemmaissroff (approved):

PR #1214, opened 2023-08-08
Prep for gem release

kddnewton:
* Drop requirement to Ruby 3.0.0
* Bump YARP version to 0.6.0
* Polyfill `unpack1` with the offset parameter for Ruby 3.0
* Remove single-line pattern matching in `serialize.rb` to get rid of warning in Ruby 3.0
* Fix tests so that they pass on Ruby 3.0, 3.1, 3.2, and 3.3

Incorporated work from #1140 so we can close that after this is out.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,6 @@
 
 source "https://rubygems.org"
 
-ruby ">= 3.1.0"
-
 gemspec
 
 gem "rake"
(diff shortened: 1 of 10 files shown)


jemmaissroff (approved):

kddnewton:
Yeah - I was trying to avoid blowing up the number of checks that we have, but you're probably right.

PR #1216, opened 2023-08-08
Fix JRuby build

kddnewton:
Didn't see that this was failing because the TruffleRuby build marked the whole thing as red.

### DIFF
--- a/templates/java/org/yarp/Loader.java.erb
+++ b/templates/java/org/yarp/Loader.java.erb
@@ -96,7 +96,7 @@ public class Loader {
 
     private ParseResult.Error[] loadSyntaxErrors() {
         int count = loadVarInt();
-        var errors = new ParseResult.Error[count];
+        ParseResult.Error[] errors = new ParseResult.Error[count];
 
         // error messages only contain ASCII characters
         for (int i = 0; i < count; i++) {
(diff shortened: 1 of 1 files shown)


eregon:
@enebo Is there a way to configure that check so it uses something less ancient than [Java 8](https://github.com/ruby/yarp/actions/runs/5801244829/job/15725086302#step:4:84)?
We use Java 17+ currently for TruffleRuby, that would be nice, and if that's too recent then 11.

Fine to merge this PR anyway to fix things.

eregon:
Apparently JRuby still uses Java 8 from what I can see here: https://github.com/jruby/jruby/blob/master/.github/workflows/ci.yml
OK so we'll tweak the config in TruffleRuby to compile with `-source 8 -target 8` too then.

enebo:
@eregon until 9.5 we are still on 8 and we should jump up minimally to 11 if not higher.  YARP if things go to plan will be default for 9.5 but I also want this as an optional feature for 9.4 (to get feedback and help debug YARP) so the Java 8 stuff is required still.

I am confused how this breakage occurred. It seems the previous PR submitted by TR had not committed to TR first so that PR was red which then people overlooked that JRuby job was also red for that PR.   I thought committing to TR first was the best practice?  If not then the TR job really is not doing us many favors.  Perhaps a nightly would be a compromise here?

eregon:
Yes, we're going to change how that test in done (discussed in details on truffleruby-yarp Slack), and a daily seems a good idea. Part of the plan is to make that JRuby check also a Required check so it can't be overlooked.

PR #1217, opened 2023-08-08
Separate yp_node_flags_t from yp_node_type_t

jemmaissroff:
Prior to this commit, we folded the flags into the type. This created extra overhead when calculating the type and setting the flags. This commit separates them.

Found this when debugging a direct access to `node->type` working on the compiler. I'm now using `YP_NODE_TYPE` over there regardless, but it seems to add unnecessary overhead to store the flags with the type.

### DIFF
--- a/templates/include/yarp/ast.h.erb
+++ b/templates/include/yarp/ast.h.erb
@@ -55,25 +55,22 @@ enum yp_node_type {
 typedef uint16_t yp_node_type_t;
 typedef uint16_t yp_node_flags_t;
 
-// We store the node type enum in every node in the tree. We don't have nearly
-// as many node types as we do bits in an enum, so we're going to use the top
-// half of the enum to store flags about the node.
-static const unsigned int YP_NODE_FLAG_NEWLINE = 0x1;
+// We store the flags enum in every node in the tree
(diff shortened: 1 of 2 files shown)


eregon (approved):
This looks definitely cleaner, thanks!

templates/include/yarp/ast.h.erb
```diff
@@ -46,26 +46,31 @@ typedef struct yp_node_list {
...
 
 // We store the node type enum in every node in the tree. We don't have nearly
 // as many node types as we do bits in an enum, so we're going to use the top
 // half of the enum to store flags about the node.
-#define YP_NODE_FLAGS_SHIFT (sizeof(yp_node_type_t) * 8 / 2)
-#define YP_NODE_FLAG_NEWLINE (1U << YP_NODE_FLAGS_SHIFT)
+static const unsigned int YP_NODE_FLAG_NEWLINE = 0x1;
 
 // For easy access, we define some macros that manipulate only the bottom half
 // of the node type enum.
```
eregon, line 64: It would be best to update these comments to reflect the new way to represent flags

kddnewton:
You should be able to rebase off https://github.com/ruby/yarp/pull/1218 now.

kddnewton (approved):
Looks great!

PR #1219, opened 2023-08-09
Split out lint task, increase matrix for gem

kddnewton:
I have begrudgingly been convinced that we need these checks in CI for the gem. Also split out the lint task, because it doesn't need to run on all of the different OSs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,11 +11,23 @@ on:
   pull_request:
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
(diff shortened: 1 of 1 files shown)


jemmaissroff (approved):

.github/workflows/main.yml
```diff
@@ -159,8 +168,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ["3.2", "head"]
-        os: [ ubuntu-latest, macos-latest, windows-latest ]
+        ruby: ["3.0", "3.1", "3.2", "head"]
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
eregon, line 172: This 3x3 matrix is a lot of jobs, if it proves annoying I think it'd be fine to test only one of the versions on all platforms, and test all versions on ubuntu.

PR #1223, opened 2023-08-09
Fix method calls with double colon and identifier

kddnewton:
The method name was incorrect because `parser->previous` has already changed at that point.


rakelib/check_manifest.rake
```diff
@@ -36,6 +36,7 @@ task :check_manifest => [:templates, "configure"] do
     config.log
     config.status
     configure.ac
+    configure~
```
jemmaissroff, line 39: What's this?

kddnewton: Some random thing that gets generated by autotools sometimes. I'm not even 100% sure what it is to be honest but I'm just ignoring it so that the manifest check stops breaking.

enebo: @kddnewton @jemmaissroff That is pretty funny.  Emacs by default backs up files by appending ~ to the end (and if you have multiple versions ~2, ~3, ...).  You actually see this proliferate around to other tools (like HEAD~12 in git).  I guess configure does it too?

jemmaissroff (approved):

PR #1224, opened 2023-08-10
Remove autotools to simplify the build system

eregon:
As you can see in the changes to `docs/build_system.md` this is a drastic simplification.

We were only really using autotools for 2 things:
* HAVE_SNPRINTF but in C99 that's guaranteed true, so it's useless to check
* HAVE_MMAP which I replaced by `!defined(_WIN32)`

I think this is by far the most important step to simplify YARP's build system.

cc @flavorjones


eregon:
@tenderlove @flavorjones Are you OK with this change, any comments?

If yes I'll merge after #1225 so CI will be green.

kddnewton:
I talked to Aaron and he said it was fine.

PR #1225, opened 2023-08-10
Test Java Loader in build-java job

eregon:
As we discussed on Slack.
This change in CI means e.g. node names and field names and other similar breaking changes can be merged with a green CI here.
The modified CI job now runs Loader.java and ensures it can deserialize fine, in addition to just javac-building `.java` files before.
That uses JRuby for convenience, so the file walking logic can be shared in Ruby instead of having to duplicate it in Java, cc @enebo

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -74,20 +74,28 @@ jobs:
     - name: Run Ruby tests
       run: bundle exec rake compile_no_debug
 
-  build-jruby:
+  build-java:
     runs-on: ubuntu-latest
     env:
       JRUBY_OPTS: "--dev"
     steps:
     - uses: actions/checkout@v3
+    - name: Set up CRuby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 3.2
+        bundler-cache: true
+    - name: Serialize fixtures on CRuby
+      run: bundle exec rake compile test:serialize_fixtures
     - name: Set up JRuby
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: jruby
         bundler-cache: true
     - name: Compile generated Java files
-      run: jruby --dev -S bundle exec rake compile
-      shell: bash
+      run: bundle exec rake compile
+    - name: Run Java Loader test
+      run: JRUBY_OPTS="-J-ea" bundle exec rake test:java_loader
 
   lex-ruby:
     runs-on: ubuntu-latest
--- a/.github/workflows/truffleruby.yml
+++ b/.github/workflows/truffleruby.yml
@@ -5,10 +5,9 @@ concurrency:
   cancel-in-progress: true
 
 on:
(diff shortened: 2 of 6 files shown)


eregon:
Ci seems fine now. @kddnewton Could you make `build-java` a `Required` check in branch protection?
That way we ensure that never breaks accidentally, no matter if there are other failing builds like currently memcheck.

Also `build-java` is quite down the list, maybe we should move it higher, e.g. just before `build-debug-mode`?

.github/workflows/main.yml
```diff
@@ -74,20 +74,28 @@ jobs:
...
       JRUBY_OPTS: "--dev"
     steps:
     - uses: actions/checkout@v3
+    - name: Set up CRuby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 3.2
+        bundler-cache: true
+    - name: Serialize fixtures on CRuby
+      run: bundle && bundle exec rake compile test:serialize_fixtures
```
kddnewton, line 89: You shouldn't need `bundle &&` here because setup ruby does that.

.github/workflows/truffleruby.yml
```diff
@@ -5,10 +5,9 @@ concurrency:
   cancel-in-progress: true
 
 on:
-  push:
-    branches:
-      - main
-  pull_request:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 8 * * *'
```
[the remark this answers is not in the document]
kddnewton: I think it's fine to run regularly. Can you make it `0 13 * * *` though? That's when I log on :)

rakelib/lex.rake
```diff
@@ -399,3 +399,44 @@ task "lex:topgems": ["download:topgems", :compile] do
     exit(1)
   end
 end
+
+known_failures = %w[seattlerb/heredoc_nested.txt]
```
kddnewton, line 403: The tasks themselves are fine, but can you move them into their own rake file? Like `loader.rake` or something.

kddnewton (changes requested):
A couple of small changes and then this is good to go

.github/workflows/main.yml
eregon: Oh true and I should know that ahah (I just copy/pasted the command I used locally)

rakelib/lex.rake
eregon: OK

.github/workflows/main.yml
eregon: Ah it's because I tried to use the system Ruby on GHA but that didn't work (doesn't have bundler)

eregon:
All done, ready to be merged.
And BTW the `build-java` job is really fast, it takes about 1 minute :)

kddnewton (approved):

PR #1228, opened 2023-08-10
Refactor flags

kddnewton:
A couple of things are happening in this commit. We now have a special param type for flags (previously these were all uint32 params). The param type accepts a "kind" field that should link up 1:1 with one of the flag types.

This commit changes all of the previous params known as flags to be of the flags type.

The flags type is special in that it doesn't have its own field in the struct. Instead we embed the value within the flags field (previously the top half of the type enum) shifted by the number of common flags (currently 1). This ends up saving us 4 bytes per call node, range node, until node, while node, regular expression node, and interpolated regular expression node.

When serializing or converting to Ruby, we unshift the value and return it as normal. As such only the C API has changed slightly.

On the Ruby API side, now that we have more information about the kind of flags, we can provide more convenience methods on the nodes based on the names of the flags. This means we now have:

CallNode#safe_navigation?
CallNode#variable_call?

InterpolatedRegularExpressionNode#ignore_case?
InterpolatedRegularExpressionNode#multi_line?
InterpolatedRegularExpressionNode#extended?
InterpolatedRegularExpressionNode#euc_jp?
InterpolatedRegularExpressionNode#ascii_8bit?
InterpolatedRegularExpressionNode#windows_31j?
InterpolatedRegularExpressionNode#utf_8?
InterpolatedRegularExpressionNode#once?

RangeNode#exclude_end?
RangeNode#flip_flop?

RegularExpressionNode#ignore_case?
RegularExpressionNode#multi_line?
RegularExpressionNode#extended?
RegularExpressionNode#euc_jp?
RegularExpressionNode#ascii_8bit?
RegularExpressionNode#windows_31j?
RegularExpressionNode#utf_8?
RegularExpressionNode#once?

UntilNode#begin_modifier?

WhileNode#begin_modifier?

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -459,7 +459,7 @@ yp_flip_flop(yp_node_t *node) {
         }
         case YP_NODE_RANGE_NODE: {
             yp_range_node_t *cast = (yp_range_node_t *) node;
-            cast->flags |= YP_RANGE_NODE_FLAGS_FLIP_FLOP;
+            cast->base.flags |= YP_RANGE_NODE_FLAGS_FLIP_FLOP;
             break;
         }
         default:
@@ -530,9 +530,9 @@ yp_arguments_validate(yp_parser_t *parser, yp_arguments_t *arguments) {
 /******************************************************************************/
 
 // Parse out the options for a regular expression.
-static inline uint32_t
+static inline uint16_t
 yp_regular_expression_flags_create(const yp_token_t *closing) {
-    uint32_t flags = 0;
+    uint16_t flags = 0;
 
     if (closing->type == YP_TOKEN_REGEXP_END) {
         for (const char *flag = closing->start + 1; flag < closing->end; flag++) {
@@ -1123,8 +1123,7 @@ yp_call_node_create(yp_parser_t *parser) {
         .opening_loc = YP_OPTIONAL_LOCATION_NOT_PROVIDED_VALUE,
         .arguments = NULL,
         .closing_loc = YP_OPTIONAL_LOCATION_NOT_PROVIDED_VALUE,
-        .block = NULL,
-        .flags = 0
+        .block = NULL
     };
 
     return node;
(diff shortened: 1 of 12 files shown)


eregon:
Nice, I was thinking indeed such methods would be very convenient for Java nodes too. I can add them in a later PR

eregon:
I'm thinking this might mean flags on the Java side have the wrong value (e.g., they might use `1` for non-newline flags). I'll push a commit to deal with the Java side of things here.

eregon:
I pushed the Java changes.
The CI is red because of another PR merged in between and so there is some issue on the C side.

eregon:
> When serializing or converting to Ruby, we unshift the value and return it as normal. As such only the C API has changed slightly.

Giving it more thoughts, in Ruby it's probably best as you did (it saves memory if there is no flags field, and `@newline` there is only set when needed, plus we need not forget that `YP_NODE_FLAG_NEWLINE` is only a potential newline so it's not the same as the newline flag in Ruby & Java). And in Java it's best to let Java pack fields neatly by packing the `short flags` and `boolean newLineFlag` in the same 4 bytes than manually doing so (plus again, not the same newline flag meaning, at least currently).

kddnewton:
Ahh sorry @eregon I lost your changes because I was rebasing my branch and didn't realize you had pushed directly onto it. Could you open another PR?

src/yarp.c
```diff
@@ -530,9 +530,9 @@ yp_arguments_validate(yp_parser_t *parser, yp_arguments_t *arguments) {
 /******************************************************************************/
 
 // Parse out the options for a regular expression.
-static inline uint32_t
+static inline uint16_t
```
jemmaissroff, line 533: No practical difference, but could we make this a `yp_node_flags_t`?

src/yarp.c
```diff
@@ -4125,7 +4123,7 @@ yp_unless_node_end_keyword_loc_set(yp_unless_node_t *node, const yp_token_t *end
 
 // Allocate a new UntilNode node.
 static yp_until_node_t *
-yp_until_node_create(yp_parser_t *parser, const yp_token_t *keyword, yp_node_t *predicate, yp_statements_node_t *statements, uint32_t flags) {
+yp_until_node_create(yp_parser_t *parser, const yp_token_t *keyword, yp_node_t *predicate, yp_statements_node_t *statements, uint16_t flags) {
```
jemmaissroff, line 4126: Same as above

jemmaissroff (approved):

eregon (approved):

PR #1233, opened 2023-08-10
More flip flop flags

kddnewton:
Whenever you see a `not` or a `!`, the receiver of that method should potentially be marked as a flip-flop.


eregon:
Looks good. Looking at this I realize we should probably do https://github.com/ruby/yarp/issues/1238

jemmaissroff (approved):

PR #1239, opened 2023-08-11
Check locals

kddnewton:
We want to ensure that we have the same set of locals for any given scope as CRuby does so that we can compile them correctly.

Currently we need to skip "special" locals like the keyword arguments hash and numbered parameters, but this gets us to a place we can iterate.

I had to add a bit of stuff to make the newline test work again. That's what the extra noise is for.

### DIFF
--- a/bin/locals
+++ b/bin/locals
@@ -1,0 +1,13 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+$:.unshift(File.expand_path("../lib", __dir__))
+require "yarp"
+
+source = ARGV[0] == "-e" ? ARGV[1] : File.read(ARGV[0])
+
+puts "CRuby:"
+p YARP.const_get(:Debug).cruby_locals(source)
+
+puts "YARP:"
+p YARP.const_get(:Debug).yarp_locals(source)
--- a/lib/yarp.rb
+++ b/lib/yarp.rb
@@ -311,6 +311,138 @@ module YARP
   # This module is used for testing and debugging and is not meant to be used by
   # consumers of this library.
   module Debug
+    class ISeq
+      attr_reader :parts
+
+      def initialize(parts)
+        @parts = parts
+      end
+
+      def type = parts[0]
+      def local_table = parts[10]
+      def instructions = parts[13]
+
+      def each_child
+        instructions.each do |instruction|
+          # Only look at arrays. Other instructions are line numbers or
+          # tracepoint events.
+          next unless instruction.is_a?(Array)
+
+          instruction.each do |opnd|
+            # Only look at arrays. Other operands are literals.
+            next unless opnd.is_a?(Array)
+
+            # Only look at instruction sequences. Other operands are literals.
+            next unless opnd[0] == "YARVInstructionSequence/SimpleDataFormat"
+
+            yield ISeq.new(opnd)
+          end
+        end
+      end
+    end
+
+    # For the given source, compiles with CRuby and returns a list of all of the
+    # sets of local variables that were encountered.
+    def self.cruby_locals(source)
+      verbose = $VERBOSE
+      $VERBOSE = nil
+
+      begin
+        locals = []
+        stack = [ISeq.new(RubyVM::InstructionSequence.compile(source).to_a)]
+
+        while (iseq = stack.pop)
+          if iseq.type != :once
+            names = iseq.local_table
+
+            # CRuby will push on a special local variable when there are keyword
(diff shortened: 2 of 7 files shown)


bin/locals
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+$:.unshift(File.expand_path("../lib", __dir__))
+require "yarp"
+
+source = ARGV[0] == "-e" ? ARGV[1] : File.read(ARGV[0])
+
+puts "CRuby:"
+p YARP.const_get(:Debug).cruby_locals(source)
```
jemmaissroff, line 10: Why a `const_get` here instead of `YARP::Debug`

lib/yarp.rb
```diff
@@ -311,6 +311,115 @@ def self.load(source, serialized)
...
+    def self.cruby_locals(source)
+      verbose = $VERBOSE
+      $VERBOSE = nil
+
+      begin
+        locals = []
+        stack = [RubyVM::InstructionSequence.compile(source).to_a]
+
+        while (iseq = stack.pop)
+          if iseq[0] != :once
```
jemmaissroff, line 325: Can we add a small top level comment around describing the layout of the iseq so `iseq[0]`, `iseq[10]` and `iseq[13]` seem less random?

lib/yarp.rb
```diff
@@ -311,6 +311,115 @@ def self.load(source, serialized)
...
+            names = iseq[10]
+
+            # CRuby will push on a special local variable when there are keyword
+            # arguments. We get rid of that here.
+            names = names.grep_v(Integer)
+
+            # We don't support numbered local variables yet, so we get rid of
```
jemmaissroff, line 332: If we're using the `TODO` convention later in this file, can we also use it here?

test/locals_test.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+# This test is going to use the RubyVM::InstructionSequence class to compile
+# local tables and compare against them to ensure we have the same locals in the
+# same order. This is important to guarantee that we compile indices correctly
+# on CRuby (in terms of compatibility).
+#
+# There have also been changes made in other versions of Ruby, so we only want
+# to test on the most recent versions.
+return if !defined?(RubyVM::InstructionSequence) || RUBY_VERSION < "3.2"
```
jemmaissroff, line 10: What are the cases where `RubyVM::InstructionSequence` isn't defined?

test/locals_test.rb
```diff
@@ -0,0 +1,86 @@
...
+
+require "yarp_test_helper"
+
+class LocalsTest < Test::Unit::TestCase
+  invalid = []
+  todos = []
+
+  # Invalid break
```
jemmaissroff, line 18: I wonder if instead of this, or maybe as an add-on PR, we should make these test snippets smaller such that they're all valid?

test/locals_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  invalid << "seattlerb/yield_arg.txt"
+  invalid << "seattlerb/yield_call_assocs.txt"
+  invalid << "seattlerb/yield_empty_parens.txt"
+  invalid << "unparser/corpus/literal/yield.txt"
+  invalid << "whitequark/args_assocs.txt"
+  invalid << "whitequark/args_assocs_legacy.txt"
+  invalid << "whitequark/yield.txt"
+  invalid << "yield.txt"
+
+  # Dead code eliminated
```
jemmaissroff, line 52: What does this mean? YARP gets rid of some code that CRuby doesn't?

jemmaissroff (changes requested):

bin/locals
kddnewton: `Debug` is a private constant, just to make it clear it's only meant for internal testing. It also means we don't impose on JRuby/TruffleRuby to implement these interfaces.

test/locals_test.rb
kddnewton: I think maybe a separate PR. It's going to be a non-trivial amount of work because a lot of them are coming directly from other imported sources.

test/locals_test.rb
kddnewton: Opposite. If you have `false ? 1 : 2` CRuby will drop the truthy path but YARP won't.

lib/yarp.rb
kddnewton: Yup!

jemmaissroff (approved):

PR #1243, opened 2023-08-11
Split out keywords from arguments

kddnewton:
Should hopefully allow easier processing for compilers. This comes at the expense of them not being in the same order anymore, which is going to make it harder on formatters, but hopefully that'll be okay.

I'm not sure if I want to merge this or not TBH. I'm going to do some measurement next week to determine if the extra pointer on every argument node is worth it.


eregon (approved):
This looks great to me.

eregon:
> This comes at the expense of them not being in the same order anymore, which is going to make it harder on formatters, but hopefully that'll be okay.

AFAIK keyword args always come after positional args, so this should not happen, right?
It does for https://github.com/ruby/yarp/pull/1243/files#diff-0a69316d72803cd5ab66cee01ea55bdd2539329d1f7782b3aaf8c8988d642f6b but that's incorrect Ruby source code, so I don't think it matters too much.

config.yml
```diff
@@ -405,6 +405,9 @@ nodes:
     child_nodes:
       - name: arguments
         type: node[]
+      - name: keywords
+        type: node?
+        kind: KeywordHashNode
```
eregon, line 410: As a note, it could just be `HashNode` and we could remove `KeywordHashNode`, because it's implied by this field.
OTOH KeywordHashNode doesn't have opening_loc/closing_loc so KeywordHashNode is better footprint-wise.

jemmaissroff (approved):

kddnewton:
I'm going to close this for now. I'm not opposed to doing it, but if we do I want to do this for blocks as well. If we do this we probably also want to change `yp_arguments_t` to have a pointer to a `yp_node_t*` that can take a couple of forms, maybe an arguments node, argumentswithkeywords node, argumentswithblock node, etc. More thinking needed.

enebo:
@kddnewton I was original person who brought up the complexity of having to check last argument and also about having blockpass and block be different things but I feel I should take another pass/look at what code I have for this.  I strongly suspect we can improve over what we have but I also want to reexamine things (I have not looked at that code in a couple of months now).  I would like to perhaps even show how the code would look if we changed design as input.

enebo:
@eregon the other point of moving block to argument node is to also combine it with block pass argument which right now is maybe last argument of arguments node.

PR #1245, opened 2023-08-12
Serialize comments and a couple Java cleanups

eregon:
* Needed for the Fiddle backend.

And there is https://github.com/oracle/truffleruby/pull/3210 adopting recent YARP changes in TruffleRuby, from the latest commit on this branch.


docs/serialization.md
```diff
@@ -1,10 +1,58 @@
...
+### comment
+
+The comment type is one of:
+* 0=`INLINE` (`# comment`)
+* 1=`EMBEDDED_DOCUMENT` (`=begin`/`=end`)
+* 2=`__END__` (after `__END__`)
+
+| # bytes | field |
+| --- | --- |
+| varint | comment type |
```
kddnewton, line 40: Can this be a single byte? Since it's just 3 types and extremely unlikely to change.

kddnewton (changes requested):
One small change. Love the documentation changes.

docs/serialization.md
eregon: In practice it only takes one byte with `varint`.
I'd rather keep it simple here since there is AFAIK no real advantage to be a single byte, and it would add complexity/inconsistency in deserialization.

eregon: Easy to change, done.

kddnewton (approved):

PR #1249, opened 2023-08-14
Run parse_test.rb on JRuby, TruffleRuby and older CRuby versions

eregon:
* Only skip the Ripper part, since that one depends on the RUBY_VERSION.

I noticed in https://github.com/ruby/yarp/actions/runs/5859209143/job/15884736966
On CRuby:
`1246 tests, 334838 assertions, 0 failures, 0 errors, 0 pendings, 1 omissions, 0 notifications`

On TruffleRuby:
`256 tests, 7106 assertions, 0 failures, 0 errors, 0 pendings, 1 omissions, 0 notifications`

With this PR on TruffleRuby:
https://github.com/ruby/yarp/actions/runs/5859462998/job/15885515840?pr=1249
`1222 tests, 309691 assertions, 0 failures, 0 errors, 0 pendings, 1 omissions, 0 notifications`

Which is much closer (not exactly the same since some tests rely on `RubyVM` and `YARP::Debug`)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,6 +8,7 @@ test/fixtures/unparser/**/*.txt linguist-vendored
 test/fixtures/whitequark/**/*.txt linguist-vendored
 test/snapshots/**/*.txt linguist-generated
 
-test/fixtures/**/*.txt -text
-
+# Preserve \n on git clone on Windows, regardless of the git config core.autocrlf value
+# See https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings
 **/*.rb text eol=lf
+**/*.txt text eol=lf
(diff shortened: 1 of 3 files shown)


eregon:
Mmh JRuby has some failures on Windows here, cc @enebo 
https://github.com/ruby/yarp/actions/runs/5859462998/job/15885524194?pr=1249
I'll skip that test on JRuby on Windows for now, so it's tested for all other cases.

eregon:
Ah the problem seems reading the snapshot produces `\r\n` and that's using `saved = File.read(snapshot)`.
Let me try a fix.

eregon:
Yeah, or at least the behavior seems to differ between CRuby and JRuby here.

eregon:
I think I can also set it in `.gitattributes`, I feel it's so wrong that `git clone` creates CRLF when there aren't any in the original files.

eregon:
OK, https://github.com/ruby/yarp/pull/1249/commits/4316f87bb28d6ceef586e7114583d39f88777c69 works, great.

kddnewton (approved):

PR #1250, opened 2023-08-14
Small style changes

kddnewton:
Nothing big here, just adding a lot of documentation to `ffi.rb` and making a couple of changes to keep the Ruby style consistent (double quotes, empty lines between methods, etc.).

The only actual changes in this are:

* `yp_string_sizeof` now returns a `size_t`. This is more consistent with other places where we call `sizeof`.
* `ffi` is always in the bundle. I like this better because it means dependabot will pick it up and it's only for local anyway. Also this gets rid of an annoyance that the Gemfile.lock differs if you pass the environment variable to trigger the test run with the FFI backend.


lib/yarp/ffi.rb
```diff
@@ -9,135 +12,200 @@ module YARP
...
-    end
   end
+
+  # Mark the LibRubyParser module as private as it should only be called through
+  # the YARP module.
   private_constant :LibRubyParser
 
-  VERSION = LibRubyParser.yp_version.to_s
+  # The version constant is set by reading the result of calling yp_version.
+  VERSION = LibRubyParser.yp_version.read_string
```
eregon, line 162: Self-note: actually we shouldn't get it from the lib but rather check it's the same and use templates or a `lib/yarp/version.rb` or so (from https://github.com/ruby/yarp/issues/765#issuecomment-1678717149)

Gemfile
```diff
@@ -8,7 +8,4 @@ gem "rake"
 gem "rake-compiler"
 gem "test-unit"
 gem "ruby_memcheck", platform: %i[mri truffleruby mswin mingw x64_mingw]
-
-if RUBY_ENGINE == 'ruby' and ENV["YARP_FFI_BACKEND"]
-  gem "ffi"
-end
+gem "ffi"
```
eregon, line 11: One side effect of this is it also installs the latest version of the ffi gem on JRuby and TruffleRuby.
That's not really representative for the `gem install` test, because ffi is not a dependency of the yarp gem, because JRuby/TruffleRuby ship with ffi anyway.
And it seems more valuable to test with the version they ship than latest, as e.g. latest might have more features but we shouldn't rely on them.
So we should add `platform: :mri` here, to test like a end user would (they wouldn't install ffi as a dependency of yarp).

lib/yarp/ffi.rb
```diff
@@ -9,135 +12,200 @@ module YARP
...
-      serialized = buffer.to_ruby_string
 
       source = Source.new(code)
-      parse_result = YARP::Serialize.load_tokens(source, serialized)
-
-      ParseResult.new(parse_result.value, parse_result.comments, parse_result.errors, parse_result.warnings, source)
-    ensure
-      LibRubyParser.yp_buffer_free(buffer)
-      buffer.pointer.free
+      Serialize.load_tokens(source, buffer.to_ruby_string).with_source(source)
```
eregon, line 191: Actually we shouldn't need the `with_source` here, it should be already correct.
I'll make a PR.

lib/yarp/ffi.rb
```diff
@@ -9,135 +12,200 @@ module YARP
...
+    LibRubyParser.with_string(filepath) { |string| lex(string.read, filepath) }
   end
 
+  # Mirror the YARP.parse API by using the serialization API.
   def self.parse(code, filepath = nil)
-    serialized = dump(code, filepath)
-    parse_result = YARP.load(code, serialized)
-    source = Source.new(code)
-    ParseResult.new(parse_result.value, parse_result.comments, parse_result.errors, parse_result.warnings, source)
+    YARP.load(code, dump(code, filepath)).with_source(Source.new(code))
```
eregon, line 202: Actually we shouldn't need the with_source here, it should be already correct.

eregon (commented):
Thank you for this cleanup, it looks much more readable

Gemfile
eregon: https://github.com/ruby/yarp/pull/1254

lib/yarp/ffi.rb
```diff
@@ -9,135 +12,200 @@ module YARP
...
       buffer.to_ruby_string
-    ensure
-      LibRubyParser.yp_buffer_free(buffer)
-      buffer.pointer.free
     end
   end
   private_class_method :dump_internal
 
+  # Mirror the YARP.dump API by using the serialization API.
   def self.dump(code, filepath = nil)
```
eregon, line 174: Maybe we should properly document these methods here, and then they would show up nicely in https://ruby.github.io/yarp/YARP.html like other methods defined in Ruby.
It certainly seems nicer than having to document in the middle of C code.
We might need to tell rdoc to ignore the C definitions somehow.

PR #1252, opened 2023-08-14
Bump to v0.7.0

kddnewton:


templates/java/org/yarp/Loader.java.erb
```diff
@@ -48,6 +48,10 @@ public class Loader {
     private ConstantPool constantPool;
     private final Nodes.Source source;
 
+    private byte MAJOR_VERSION = (byte) 0;
+    private byte MINOR_VERSION = (byte) 7;
+    private byte PATCH_VERSION = (byte) 0;
```
eregon, line 53: These should be `static final`, otherwise they actually take space for each Loader instance.
I'll fix it.

PR #1259, opened 2023-08-15
WIP: Add template to generate Sorbet RBIs

andyw8:
This adds generation of Sorbet RBI's for YARP, to assist with migrating Ruby LSP to use YARP.

The template is based on `lib/yarp/node.rb`.

TODO:
- verify all signatures are being correctly generated
- consider if we want to ship the RBI with the gem. If so, it will need to be moved to the expected location (`/rbi` ?)


templates/template.rb
```diff
@@ -38,6 +38,7 @@ class NodeParam < Param
   include KindTypes
 
   def rbs_class = "Node"
+  def sorbet_class = "YARP::Node"
```
kddnewton, line 41: I'm not a big fan of adding the `sorbet_class` methods here, because they're so specific to that one file. Honestly I would much prefer a big ERB block at the beginning of the template file to handle this kind of thing.

I would take advantage of the `kind` field here, so that you have something more specific. It should make working with it easier.

templates/template.rb
```diff
@@ -46,6 +47,7 @@ class OptionalNodeParam < Param
   include KindTypes
 
   def rbs_class = "Node?"
+  def sorbet_class = "T.nilable(YARP::Node)"
```
kddnewton, line 50: Same here.

templates/template.rb
```diff
@@ -243,6 +253,7 @@ def locals
   "java/org/yarp/ParseResult.java",
   "java/org/yarp/AbstractNodeVisitor.java",
   "lib/yarp/node.rb",
+  "lib/yarp/sorbet.rb",
```
kddnewton, line 256: I wouldn't generate this here. It's odd to have it in `lib`. I would probably make it `rbi/sorbet.rbi`.

kddnewton (changes requested):

templates/template.rb
paracycle: Totally agreed. Also, if we are generating an RBI file, which isn't the extension `.rbi`?

PR #1261, opened 2023-08-15
handle missing HEREDOC endline at start of heredoc

HParker:
Found via: https://github.com/BlakeWilliams/backbone_handlebars/blob/master/lib/backbone_handlebars/template.rb#L30

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,11 +79,11 @@ See the [CONTRIBUTING.md](CONTRIBUTING.md) file for more information. We additio
 * [Configuration](docs/configuration.md)
 * [Design](docs/design.md)
 * [Encoding](docs/encoding.md)
-* [Extension](docs/extension.md)
 * [Fuzzing](docs/fuzzing.md)
 * [Heredocs](docs/heredocs.md)
 * [Mapping](docs/mapping.md)
 * [Ripper](docs/ripper.md)
+* [Ruby API](docs/ruby_api.md)
 * [Serialization](docs/serialization.md)
 * [Testing](docs/testing.md)
 


src/util/yp_newline_list.c
```diff
@@ -32,9 +32,10 @@ yp_newline_list_append(yp_newline_list_t *list, const char *cursor) {
 
     assert(cursor >= list->start);
     size_t newline_offset = (size_t) (cursor - list->start + 1);
-    assert(list->size == 0 || newline_offset > list->offsets[list->size - 1]);
-    list->offsets[list->size++] = newline_offset;
 
+    if (list->size == 0 || newline_offset > list->offsets[list->size - 1]) {
```
kddnewton, line 36: I think we want to keep around the assertion. It looks like we're probably calling this in a place we shouldn't be.

src/yarp.c
```diff
@@ -8019,7 +8019,11 @@ parse_statements(yp_parser_t *parser, yp_context_t context) {
             while (accept_any(parser, 2, YP_TOKEN_NEWLINE, YP_TOKEN_SEMICOLON));
             if (context_terminator(context, &parser->current)) break;
         } else {
-            expect(parser, YP_TOKEN_NEWLINE, "Expected a newline or semicolon after statement.");
+            if (parser->heredoc_end) {
```
kddnewton, line 8022: I don't think we want to handle this here. I think it's probably likely that we're handling the escape somewhere and not flushing the `heredoc_end` as we should be.

Maybe take a look where we are currently handling `\` in `YP_LEX_DEFAULT`? I'd bet we aren't checking for a heredoc.

kddnewton (changes requested):

src/yarp.c
kddnewton: @HParker if you don't mind removing this section, I think it's good.

.gitattributes
```diff
@@ -11,4 +11,4 @@ test/snapshots/**/*.txt linguist-generated
 # Preserve \n on git clone on Windows, regardless of the git config core.autocrlf value
 # See https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings
 **/*.rb text eol=lf
-**/*.txt text eol=lf
```
HParker, line 14: I think this removes all `\r\n` on checkout. Let me know if there is a different configuration I am not understanding.

HParker: Ah, looks like you already did this on main, but better.

HParker:
Somehow I rebased this branch into oblivion. Re-opened here: https://github.com/ruby/yarp/pull/1278

PR #1265, opened 2023-08-15
Trim down rust bindings

kddnewton:
Expose fewer things, and don't manually maintain `yp_string_t` structure.

cc @turboladen - `yp_string_t` actually changed since you opened the PR 😓 so it's simpler to handle now. I've dropped things that we're considering "internal" for now. Tagging you just so you can see it - no pressure to review unless you have time!


turboladen:
You're so fast! :D

turboladen:
I can't "officially" approve (repo permissions ...which is fine by me), but I approve! :)

eregon:
Nice!

PR #1266, opened 2023-08-16
Initial pass at yarp crate

kddnewton:
A quick pass at what a larger yarp crate could look like. Pretty much the entire thing is generated from `config.yml`, so no need to really maintain anything. It generates:

* structs that wrap the native structs
* a single `parse` function that returns a `parse_result`
* a visitor with default implementations (the same as the `syn` crate)

cc @ianks, @turboladen, @lopopolo


rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// Returns the message associated with the diagnostic.
+    /// 
+    /// # Panics
+    /// 
+    /// Panics if the message is not able to be converted into a `CStr`.
+    /// 
+    #[must_use]
+    pub fn message(&self) -> String {
+        unsafe {
+            let message: *mut i8 = (*(self.0)).message as *mut i8;
```
lopopolo, line 55: this needs to be a cast to `*mut c_char` otherwise it will not compile on aarch64 linux.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+impl Comment {
+    /// Returns the text of the comment.
+    #[must_use]
+    pub fn text(&self) -> String {
+        unsafe {
+            let start: *mut u8 = (*(self.0)).start as *mut u8;
+            let end: *mut u8 = (*(self.0)).end as *mut u8;
+
+            let length = end as usize - start as usize;
+            String::from_raw_parts(start, length, length)
```
lopopolo, line 74: this is unsound. `String::from_raw_parts` can only be safely used on raw parts derived from a string.

Instead, create a `&[u8]`, call `std::str::from_utf8`, and then create an owned string.

However, given that the memory is already owned by the parser, it seems more acceptable to return a `&str` bound to the lifetime of the `Comment`.

One other question I have: are the byte contents guaranteed to be UTF-8 even of the source encoding of the file is e.g. shift JIS?

Based on the multi-encoding support in the parser I expected most of these APIs to be written in terms of bytes.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+    fn next(&self) -> T;
+}
+
+impl ListNode<Self> for Diagnostic {
+    fn is_null(&self) -> bool {
+        self.0.is_null()
+    }
+
+    fn next(&self) -> Self {
+        Self(unsafe { (*(self.0)).node.next.cast::<yp_diagnostic_t>() })
```
lopopolo, line 97: this API appears to be unsound because it is a safe API which can cause a null pointer dereference if `self.0` is null.

This API should return `Option<Self>`.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+/// A couple of different concepts are represented as linked lists in the
+/// parser. This includes errors, warnings, and comments. We want to be able to
+/// iterate over them in a consistent fashion. As such, we define a trait that
+/// can be used to iterate over them.
+pub trait ListNode<T> {
+    /// Returns true if the node is null.
+    fn is_null(&self) -> bool;
+
+    /// Returns the next node in the list.
+    fn next(&self) -> T;
```
lopopolo, line 88: this API should return `Option<T>`

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+
+impl ParseResult {
+    /// Returns the source string that was parsed.
+    /// 
+    /// # Panics
+    /// 
+    /// Panics if the source string is not able to be converted into a `&str`.
+    #[must_use]
+    pub fn source(&self) -> String {
+        self.source.to_str().unwrap().to_string()
```
lopopolo, line 145: does the C lib panic if the source is not UTF-8? If the underlying library doesn't operate with guarantees of UTF-8, I would not expect the Rust wrapper to make that assumption. Should this API instead return `Vec<u8>`?

In general, libraries should not panic in the face of expected errors. They should return `Result` and delegate to callers how to respond to an error condition.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+/// Parses the given source string and returns a parse result.
+/// 
+/// # Panics
+/// 
+/// Panics if the source string is not able to be converted into a `CString`.
+/// 
+#[must_use]
+pub fn parse(source: String) -> ParseResult {
```
lopopolo, line 203: I would expect a function called `parse` to accept a borrow of its input: either `&str` or `&[u8]`.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+    unsafe {
+        let source = CString::new(source).unwrap();
```
lopopolo, line 205: This means passing `String::new("abc\0xyz")` to this function will panic. This should either return an error or use the underlying yarp API that accepts a pointer + a length.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+        let mut uninit = MaybeUninit::<yp_parser_t>::uninit();
+
+        yp_parser_init(
+            uninit.as_mut_ptr(),
+            source.as_ptr(),
+            source.as_bytes().len(),
```
lopopolo, line 211: It is not necessary to go through a `CString` if all we want is a pointer to the underlying byte contents.

With source as a `String` (or `&str`), this is achievable with `source.as_bytes().as_ptr()` and `source.len()`.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+            std::ptr::null(),
+        );
+
+        let parser = NonNull::new(uninit.assume_init_mut()).unwrap();
+        let node = NonNull::new(yp_parse(parser.as_ptr())).unwrap();
```
lopopolo, line 216: if we are absolutely certain that these pointers are not null, that would be more clearly communicated by using the `NonNull::new_unchecked` unsafe API with a corresponding `// SAFETY: ` comment.

rust/yarp/Cargo.lock
turboladen, line 1: Since `yarp` is a library, I'd probably exclude the lock file (similar to how you'd exclude a Gemfile.lock for a library); was this intentional?

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+            Some(current)
+        }
+    }
+}
+
+/// The result of parsing a source string.
+pub struct ParseResult {
+    source: CString,
+    parser: NonNull<yp_parser_t>,
+    node: NonNull<yp_node_t>
```
lopopolo, line 134: this appears to be a self referential struct. if that is the case, drop order is important and I'd probably wrap each field in a `ManuallyDrop` to ensure a safe drop order in this struct's `Drop` implementation.

I'm not sure of the ergonomics, but I would probably prefer this struct to have a lifetime generic parameter so it is bound to the given source as opposed to storing the source itself. That can either be achieved by storing the source e.g. `&str` or with a `PhantomData` marker.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+impl ParseResult {
+    /// Returns the source string that was parsed.
+    /// 
+    /// # Panics
+    /// 
+    /// Panics if the source string is not able to be converted into a `&str`.
+    #[must_use]
+    pub fn source(&self) -> String {
```
lopopolo, line 144: I would expect this API to return a borrowed view of the source rather than allocating.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+        unsafe {
+            let list = &mut (*self.parser.as_ptr()).error_list;
+            ListIter { current: Diagnostic(list.head.cast::<yp_diagnostic_t>()) }
+        }
+    }
+
+    /// Returns an iterator that can be used to iterate over the warnings in the
+    /// parse result.
+    #[must_use]
+    pub fn warnings_iter(&self) -> ListIter<Diagnostic> {
```
lopopolo, line 161: generally, methods like these would be called `errors` or `warnings` and return iterator types called `Errors` and `Warnings`.

See `str::bytes` or `[_]::chunks` in `std` for examples.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+            let list = &mut (*self.parser.as_ptr()).comment_list;
+            ListIter { current: Comment(list.head.cast::<yp_comment_t>()) }
+        }
+    }
+
+    /// Returns the root node of the parse result.
+    #[must_use]
+    pub fn node(&self) -> Node {
+        Node::new(self.node.as_ptr())
+    }
```
lopopolo, line 182: is this node safely able to be kept alive if the source parse result is dropped? If not, the returned `Node` should have a lifetime parameter.

otherwise, this would be permitted in safe code:

```rust
let p = parse("1 + 2");
let n = p.node();
drop(p);
println!("{:?}", n); // <-- use after free
```

rust/yarp/build.rs
```diff
@@ -0,0 +1,507 @@
...
+
+    write_bindings(config)?;
+    Ok(())
+}
+
+/// Gets the path to the config.yml file at `[root]/config.yml`.
+///
+fn config_path() -> PathBuf {
+    cargo_manifest_path()
+        .join("../../config.yml")
```
lopopolo, line 76: for publishing this to crates.io, `config.yml` is likely going to have to exist inside the crate root.

rust/yarp/src/lib.rs
lopopolo: these iterators should also probably be bound to the lifetime of the parse result

lopopolo:
@kddnewton would you be able to post the generated bindings? it is difficult to review the code generated by the build script.

rust/yarp/build.rs
```diff
@@ -0,0 +1,507 @@
...
+struct NodeParam {
+    name: String,
+    r#type: NodeParamType,
+    kind: Option<String>
+}
+
+#[derive(Debug, Deserialize)]
+struct Node {
+    name: String,
+    child_nodes: Option<Vec<NodeParam>>,
```
turboladen, line 52: I'm not sure if it's typically considered an anti-pattern or not, but, while `Option<Vec<T>>` does convey info to the developer (sometimes there'll be nodes and sometimes not), in practice it ends up being the same amount code to work with a `Vec<T>`. I could be wrong, but I also don't believe the compiler does any optimizing if you use `Option<Vec<T>>` over `Vec<T>`.

If your concern is with deserialization (i.e. sometimes there's a `child_nodes` field, sometimes not), you could use the `#[serde(default)]` attribute... https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6ed6ea58b8d6fc86f1e634d6b152e0a4

(sorry, if you're already familiar with all this and it's all intentional; I'll stop if that's the case!)

rust/yarp/build.rs
```diff
@@ -0,0 +1,507 @@
...
+        }
+        result.push(char.to_uppercase().next().unwrap());
+    }
+
+    result
+}
+
+/// Write the generated struct for the node to the file.
+/// 
+fn write_node(file: &mut File, node: &Node) -> Result<(), Box<dyn std::error::Error>> {
```
turboladen, line 118: Any interesting in using [quote](https://docs.rs/quote/latest/quote/) for generating this code? I once had a crate where I generated a bunch of Rust code using handlebars and that got to be pretty annoying to maintain.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+        parse, Visit,
+        visit_interpolated_regular_expression_node, InterpolatedRegularExpressionNode,
+        visit_regular_expression_node, RegularExpressionNode,
+    };
+
+    struct RegularExpressionVisitor {
+        count: usize
+    }
+
+    impl Visit for RegularExpressionVisitor {
```
ianks, line 234: 🔥

kddnewton:
Thank you all so much for the review! I'm going to learn so much rust through this PR. I really appreciate you all taking the time. I'll tag you again once I implement all of your suggested changes.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+
+use std::ffi::{CStr, CString};
+use std::mem::MaybeUninit;
+use yarp_sys::{yp_parser_t, yp_parser_init, yp_parse, yp_node_destroy, yp_node_t, yp_comment_t, yp_diagnostic_t};
+
+use std::ptr::NonNull;
+
+/// A diagnostic message that came back from the parser.
+#[derive(Copy, Clone, Debug)]
+pub struct Diagnostic(*mut yp_diagnostic_t);
```
ianks, line 43: As as a tiny point of optimization, using the `NonNull` here would make sense here as well, since it would allow the [null-pointer optimization](https://www.sobyte.net/post/2022-04/rust-enum-layout/).

rust/yarp/build.rs
```diff
@@ -0,0 +1,507 @@
...
+    writeln!(file, "pub struct ConstantList(*mut yp_constant_id_list_t);")?;
+    writeln!(file)?;
+
+    for node in &config.nodes {
+        writeln!(file, "const {}: u16 = yp_node_type::{} as u16;", type_name(&node.name), type_name(&node.name))?;
+    }
+    writeln!(file)?;
+
+    writeln!(file, "/// An enum representing the different kinds of nodes that can be parsed.")?;
+    writeln!(file, "pub enum Node {{")?;
```
turboladen, line 359: I think the node types in the tuple fields here should be of the related wrapper types, not of the C pointers. (screenshooting docs for context for other reviewers). i.e. I'd expect:

```rust
enum Node {
    AliasNode(AliasNode),
    // etc
}
```

...where the `AliasNode` in the tuple is the `AliasNode` struct.

image

ianks: Agree. Ideally, any references to `*mut yp_xyz` should only be accessible through an `as_mut_ptr` method.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+    pub fn message(&self) -> String {
+        unsafe {
+            let message: *mut i8 = (*(self.0)).message as *mut i8;
+            CStr::from_ptr(message).to_str().unwrap().to_string()
+        }
+    }
+}
+
+/// A comment that was found during parsing.
+#[derive(Copy, Clone, Debug)]
```
ianks, line 62: I'm not familiar enough with YARP, but my initial feeling is that it might not make sense for this to be `Copy` + `Clone`.

Although the underlying pointer is trivially copiable, making to too easy to move the `Comment` means it becomes impossible to ensure the raw pointer is still around later on - i.e. if the `yp_comment_t` gets freed later we don't want anyone accidentally hanging onto a `Comment`.

rust/yarp/build.rs
turboladen: Since I'm seeing other cases... I think the general idea with this crate should be that most, if not all, interactions with pointers should be abstracted away; no one working with this crate should need to care about raw pointers.

rust/yarp/build.rs
```diff
@@ -0,0 +1,507 @@
...
+    let out_path = PathBuf::from(std::env::var_os("OUT_DIR").unwrap()).join("bindings.rs");
+    let mut file = std::fs::File::create(&out_path).expect("Unable to create file");
+
+    writeln!(file, "#[allow(clippy::wildcard_imports)]")?;
+    writeln!(file, "use yarp_sys::*;")?;
+    writeln!(file)?;
+
+    writeln!(file, "/// A range in the source file.")?;
+    writeln!(file, "#[derive(Debug)]")?;
+    writeln!(file, "pub struct Location(*mut yp_location_t);")?;
```
turboladen, line 297: Is the intent here (and maybe with other types as well?) to implement accessor type methods later? When I look at the docs for this (which can be helpful while you're in this stage of development; just run `cargo doc`), it doesn't look like there's any way to get the start and end points of the Location.

rust/yarp/src/lib.rs
kddnewton: The bytes in the comment message specifically are guaranteed to be UTF-8 regardless of the source because we're only using UTF-8.

rust/yarp/src/lib.rs
kddnewton: This now returns a `&[u8]`

rust/yarp/src/lib.rs
kddnewton: These are now bound to the lifetime of the parse result.

rust/yarp/build.rs
kddnewton: I attempted to use `quote` but it just returns a token stream, and getting it back to nicely formatted source appears to not be an option.

rust/yarp/Cargo.lock
kddnewton: I always check in `Gemfile.lock` just to make sure teams working on it always have the same experience. (I've been burned by this before.)

rust/yarp/build.rs
lopopolo: You can use the `prettyplease` crate for that

rust/yarp/build.rs
kddnewton: Yeah, this is going to be part of a bigger conversation - I'm not sure how we're going to publish `yarp-sys` because it references the shared library built by the makefile. I'd imagine once we get things stable we'll need to spend some time to figure out how to get all of this stuff published.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+    unreachable_pub,
+    unused_qualifications
+)]
+
+// Most of the code in this file is generated, so sometimes it generates code
+// that doesn't follow the clippy rules. We don't want to see those warnings.
+#![allow(
+    clippy::too_many_lines,
+    clippy::use_self
+)]
```
lopopolo: I would expect this bit of pragmas to only be applied to `mod bindings`. As is, with the `#!`, it applies to the whole crate.

```suggestion
// Most of the code in this file is generated, so sometimes it generates code
// that doesn't follow the clippy rules. We don't want to see those warnings.
#[allow(
    clippy::too_many_lines,
    clippy::use_self
)]
```

rust/yarp/build.rs
kddnewton: I've implemented the accessor methods now!

The main purpose of this PR is just to get something working that people can iterate on. 🤷

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+
+impl<'pr> Diagnostic<'pr> {
+    /// Returns the message associated with the diagnostic.
+    /// 
+    /// # Panics
+    ///
+    /// Panics if the message is not able to be converted into a `CStr`.
+    ///
+    #[must_use]
+    pub fn message(&self) -> String {
```
lopopolo, line 57: I think this method should return `&str`.

lopopolo:
```suggestion
    pub fn message(&self) -> &str {
```

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+        unsafe {
+            let message: *mut c_char = self.diagnostic.as_ref().message as *mut c_char;
+            CStr::from_ptr(message).to_str().unwrap().to_string()
```
lopopolo, line 60:
```suggestion
            let cstr = CStr::from_ptr(message);
            cstr.to_str().expect("yarp uses UTF-8 internally for comments")
```

lopopolo: use a binding for the `CStr` to be defensive against future refactors which may cause that temporary to be dropped.

rust/yarp/build.rs
kddnewton: I think we don't want to do that because then any of the match expressions on node are going to have to copy the wrapping struct.

rust/yarp/src/lib.rs
kddnewton: I agree, I need to figure this part out.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// 
+    /// Panics if the comment is not able to be read.
+    /// 
+    #[must_use]
+    pub fn text(&self) -> String {
+        unsafe {
+            let start: *mut u8 = self.comment.as_ref().start as *mut u8;
+            let end: *mut u8 = self.comment.as_ref().end as *mut u8;
+
+            let len = end as usize - start as usize;
```
lopopolo, line 85: Casting a pointer to `usize` is generally to be avoided and will run afoul of Rust's pointer provinance rules. The pointer type has a rich API. Use that instead.

```suggestion
            let start = self.comment.as_ref().start.cast::<u8>();
            let end = self.comment.as_ref().end.cast::<u8>();

            let len = usize::try_from(end.offset_from(start)).expect("end should point to memory after start");
```

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+    #[must_use]
+    pub const fn source(&self) -> &'pr [u8] {
+        self.source
+    }
+
+    /// Returns a slice of the source string that was parsed using the given
+    /// location range.
+    #[must_use]
+    pub fn slice(&self, location: &Location<'pr>) -> &'pr [u8] {
+        let ptr = self.source.as_ptr() as usize;
```
lopopolo, line 161: can this be reworked to avoid casting the pointer to `usize`? There should be a pointer offset API we can use instead.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+            Diagnostics { diagnostic: list.head.cast::<yp_diagnostic_t>(), marker: PhantomData }
+        }
+    }
+
+    /// Returns an iterator that can be used to iterate over the comments in the
+    /// parse result.
+    #[must_use]
+    pub fn comments(&self) -> Comments<'_> {
+        unsafe {
+            let list = &mut (*self.parser.as_ptr()).comment_list;
```
lopopolo, line 190: If `comment_list` is a raw pointer, I would avoid making a Rust reference out of it.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+/// Panics if the parser fails to initialize.
+///
+#[must_use]
+pub fn parse(source: &[u8]) -> ParseResult<'_> {
+    unsafe {
+        let mut uninit = MaybeUninit::<yp_parser_t>::uninit();
+
+        yp_parser_init(
+            uninit.as_mut_ptr(),
+            source.as_ptr().cast::<i8>(),
```
lopopolo, line 226:
```suggestion
            source.as_ptr().cast::<c_char>(),
```

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+            source.len(),
+            std::ptr::null(),
+        );
+
+        let parser = NonNull::new_unchecked(uninit.assume_init_mut());
+        let node = NonNull::new_unchecked(yp_parse(parser.as_ptr()));
```
lopopolo, line 232: for debugging purposes, I imagine it will be slightly more friendly to the debugger if the `yp_parse` invocation were on its own line.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::parse;
+
+    #[test]
+    fn comments_test() {
+        let source = "# comment 1\n# comment 2\n# comment 3\n";
```
lopopolo, line 244: does yarp accept embedded NUL bytes in comments? I believe the above code with `CStr` in `Comment::message` will behave unexpectedly if that is the case.

lopopolo (commented):
The API in the tests looks very pleasant. Most of my comments continue to be focused on the unsafe code and crate quality.

lopopolo:
I hope y'all are coming to understand why I suggested the Rust crates have CI set up that runs them under the Rust toolchain sanitizers. There is lots of unsafe code in here and it is tricky to get all of it right.

Thanks for pushing on this @kddnewton. This PR has been progressing nicely.

rust/yarp/build.rs
lopopolo: if you're concerned about that, though I wouldn't be, you can mark the wrapper structs as `#[repr(transparent)]`.

As it is, this enum is able to be accessed from safe code like this:

```rust
let mut node = Node::AliasNode(
let mut uaf = Node::AliasNode(std::ptr::null_mut());
std::mem::swap(&mut node, &mut uaf);
```

and then proceed with safe operations on `node`. Does that allow for unsound usage of this crate's APIs?

The suggestion for wrapper structs means the payload of the enum is opaque. If the wrapper struct has a private constructor, then it cannot be modified by user code.

rust/yarp/src/lib.rs
kddnewton: In this case `comment_list` is embedded directly onto the parser. So it's not a pointer, it's the struct itself. I want to get a pointer to it, though, so that I can avoid copying it.

rust/yarp/src/lib.rs
kddnewton: Yeah, it does. I've changed the API to just return `&[u8]` now.

rust/yarp/build.rs
turboladen, line 1: Because file is getting a bit long, you might consider extracting this all to a separate crate. Could give a little more freedom for code organization, etc.

rust/yarp/build.rs
turboladen: For the one `-sys` crate I've published (to a private crate server) which depended on a local C/C++ project, I pulled in the files I needed by using `include` in the `Cargo.toml`:

```toml
include = [
  "build.rs",
  "src/**/*",
  "vendor/foo/CMakeLists.txt",
  "vendor/foo/ios.toolchain.cmake",
  "vendor/foo/include/foo/api.h",
  "vendor/foo/include/foo/bar.h",
  "vendor/foo/source/bar.cpp",
  # etc
]
```

That was a couple of years ago, so it's not super fresh, but IIRC, this did require some finagling with the source tree to make `cargo` happye.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,321 @@
...
+            let list = &mut (*self.parser.as_ptr()).comment_list;
+            Comments { comment: list.head.cast::<yp_comment_t>(), marker: PhantomData }
+        }
+    }
+
+    /// Returns the root node of the parse result.
+    #[must_use]
+    pub fn node(&self) -> Node<'pr> {
+        Node::new(self.node.as_ptr())
+    }
```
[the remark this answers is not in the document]
lopopolo: Returning a reference like that without the node being a field on `self` is not possible unless you're creating a repr transparent wrapper on an unsized field (like `BStr` does).

The way this is done here with a lifetime generic parameter on the struct is the idiomatic way to do this. Similar to `std::slice::Iter`.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,321 @@
...
+                let current = Comment { comment, marker: PhantomData };
+                self.comment = unsafe { comment.as_ref().node.next.cast::<yp_comment_t>() };
+                Some(current)
+            }
+        }
+    }
+}
+
+/// The result of parsing a source string.
+pub struct ParseResult<'pr> {
```
lopopolo, line 134: this should derive Debug. Rust API guidelines suggest all public structs derive Debug.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,321 @@
...
+    marker: PhantomData<&'pr yp_comment_t>
+}
+
+impl<'pr> Comment<'pr> {
+    /// Returns the text of the comment.
+    ///
+    /// # Panics
+    /// 
+    /// Panics if the comment is not able to be read.
+    ///
```
lopopolo, line 73: this doc comment is outdated

```suggestion
```

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,321 @@
...
+            None
+        } else {
+            let current = Diagnostic {
+                diagnostic: unsafe { NonNull::new_unchecked(self.diagnostic) },
+                marker: PhantomData
+            };
+
+            self.diagnostic = unsafe { (*(self.diagnostic)).node.next.cast::<yp_diagnostic_t>() };
+            Some(current)
+        }
```
lopopolo, line 107: this can be made more maintainable and use less unsafe doing this:

```suggestion
        if let Some(diagnostic) = NonNull::new(self.diagnostic) {
            let current = Diagnostic {
                diagnostic,
                marker: PhantomData
            };

            self.diagnostic = unsafe { (*(self.diagnostic)).node.next.cast::<yp_diagnostic_t>() };
            Some(current)
        } else {
            None
        }
```

rust/yarp/src/lib.rs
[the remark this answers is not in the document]
kddnewton: The reason I did this is because a lot of the nodes hold only references to the source. For example to get the name of a method call, you would usually go back to the source and slice out the `message_loc` field. In this way the nodes themselves don't have to hold onto strings, they only hold onto pointers.

Also, we're planning on our own arena allocator, which is why the parser has the be around when you call `yp_node_destroy`. So the lifetimes have to be tied together.

One thing I think we could do would be to provide a `to_owned` or something that would copy out the source information and give you everything you needed so that you could keep the nodes around.

rust/yarp/src/lib.rs
lopopolo: it looks like this is similar to the `impl Iterator for Comment` below. It'd be good to standardize across these impls.

I prefer the `if let` style vs `match` since it removes one layer of rightward drift, but both are equivalent.

lopopolo:
apologies for the comments coming in spurts. As this PR gets more refined, it has been easier for me to spot additional areas for improvement. I'm happy to defer these at some point and instead start following up with PRs of my own if you'd prefer @kddnewton

kddnewton:
@lopopolo yeah I think at this point this is good enough to get the initial pass. I want to get this merged now so that we can iterate on it.

The only thing I want to get fixed here before we merge is making sure we're not leaking. I don't know why I can't call `yp_parser_free`. If you feel like taking a crack at it, I need to leave this for now and get back to other bugs, but I'll probably have time next week.

Other than that, follow up items from this are:

* Making sure we can publish to crates.io properly
* Moving the build generation to another crate (I love that)
* Implementing yp_string_t (right now string params just return empty strings)

Once those are there, I'd feel comfortable publishing.

rust/yarp/src/lib.rs
```diff
@@ -0,0 +1,312 @@
+//! # yarp
+//!
+//! Rustified version of Ruby's YARP parser.
+//!
+#![warn(
+    box_pointers,
```
turboladen, line 6: I see clippy is warning about box-pointer use, but we probably want to allow that, and thus this line could go away.

kddnewton:
nvm figured it out. We needed the heap allocate the parser to get it to stick around. Should be good now

rust/yarp/build.rs
```diff
@@ -0,0 +1,654 @@
...
+            }
+            writeln!(file, "/// {}", line)?;
+        }
+    }
+
+    if example {
+        writeln!(file, "/// ```")?;
+    }
+
+    writeln!(file, "pub struct {}<'pr> {{", node.name)?;
```
eregon, line 148: Any reason to not use ERB templating for this like for all other languages?
I was thinking I could peek at the generated code for Rust using `bundle exec rake templates` but no luck.
Also this would mean we likely need to modify here and `bin/template` if e.g. the format of `config.yml` changes.
That might be a barrier, e.g. my Rust is quite rusty (OTOH I have no intentions to change the format of config.yml).

kddnewton: Because when we publish to `crates.io`, this means we only need a copy of `config.yml` and don't have to have all of the mechanisms to template. If `config.yml` changes I'm happy to update it here, it's not hard to change. Plus it will fail compilation, which is a nice perk.

lopopolo:
> nvm figured it out. We needed the heap allocate the parser to get it to stick around. Should be good now

does this mean the parser struct cannot be moved?

kddnewton:
The parser struct should be able to be moved, it was the `yp_parser_t` that needed to be heap allocated. Sorry that wasn't clear.

lopopolo:
> The `ParseResult` struct should be able to be moved, it was the `yp_parser_t` that needed to be heap allocated. Sorry that wasn't clear.

My question was about the C struct. Why is the `yp_parser_t` struct unable to be moved?

kddnewton:
To be honest I'm not entirely confident with Rust's move semantics so I'm not 100% sure. The source definitely can't be moved, but if you want to move around the `yp_parser_t` it should be fine. If something is heap allocated will rust move it or not?

PR #1271, opened 2023-08-16
Use _snprintf when using Windows versions prior to 2015 (< 1900)

jemmaissroff:
Ruby CI has informed us that snprintf is not available on Windows versions, but _sprintf is supported. This commit allows us to use _sprintf where applicable.

### DIFF
--- a/include/yarp/defines.h
+++ b/include/yarp/defines.h
@@ -34,6 +34,11 @@
 #   define inline __inline
 #endif
 
+// Windows versions before 2015 use _snprintf
+#if defined(_MSC_VER) && (_MSC_VER < 1900)
+#   define snprintf _snprintf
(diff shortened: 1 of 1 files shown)


eregon (approved):

eregon:
Do you have a link to the failing log? https://rubyci.org/ doesn't seem to link builds on Windows. And GHA doesn't have such old Windows versions AFAIK.

jemmaissroff:
> Do you have a link to the failing log? https://rubyci.org/ doesn't seem to link builds on Windows. And GHA doesn't have such old Windows versions AFAIK.

The sync is out of date, the link is from an attempt at re-syncing. https://github.com/ruby/ruby/pull/8226

PR #1272, opened 2023-08-17
Fix errors on locals test caused by LANG=C

kddnewton:


k0kubun:
:+1:

k0kubun:
It looks like this PR broke this test on http://ci.rvm.jp/results/trunk-iseq_binary@ruby-sp2-docker (http://ci.rvm.jp/results/trunk-iseq_binary@ruby-sp2-docker/4689554).

PR #1274, opened 2023-08-17
fix: multiple heredocs with embedded expressions with newlines

flavorjones:
Set heredoc_end to NULL at the start of lexing a heredoc, to avoid having state from the previous heredoc confuse the parser's current location.

I'm not positive that this is the right fix.

Here's a description of the bug(s) this is fixing, though (which were found by the fuzzer).

This ruby fixture:

```ruby
<<-A + <<-B
  a
A
  b
  #{
  2}
B
```

will abort during parsing after failing [this assertion in `yp_newline_list_append`](https://github.com/ruby/yarp/blob/c57436cc48bd5d8389cc212fa61a398e7a30bb9b/src/util/yp_newline_list.c#L35) because the parser has backtracked too far and is trying to re-append newlines from earlier in the file.

And this ruby fixture:

```ruby
<<-A + <<-B
  a
A
  b
  #{2
  }
B
```

will parse, but node start and end locations are nonsensical. Serialization will abort after failing [this assertion in `serialize.c`](https://github.com/ruby/yarp/blob/c57436cc48bd5d8389cc212fa61a398e7a30bb9b/templates/src/serialize.c.erb#L21) because the node location start is after the end.


flavorjones:
Here's another code snippet that causes a similar problem, but that this branch does not address:

```ruby
<<-EOF \
  1
EOF
```

Parsing results in

> ruby: src/util/yp_newline_list.c:36: yp_newline_list_append: Assertion `list->size == 0 || newline_offset > list->offsets[list->size - 1]' failed.

where it's parsing the heredoc again, and so trying to re-append the same newlines. I'm not sure I understand what's going on, and so there may be a better fix than this PR that addresses both issues?

kddnewton:
Yes, this is right!

kddnewton (approved):

PR #1275, opened 2023-08-17
preallocate arrays when building the Ruby AST

froydnj:
This change is good for a ~10% performance improvement in the unscientific benchmark of:

```ruby
files.each do |f|
  YARP.parse_file(f)
end
```

where `files` is all the Ruby files in a large codebase.


jemmaissroff (approved):
Looks great, thanks!

PR #1278, opened 2023-08-17
handle missing HEREDOC endline at start of heredoc

HParker:
Found via: https://github.com/BlakeWilliams/backbone_handlebars/blob/master/lib/backbone_handlebars/template.rb#L30

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -5653,13 +5653,23 @@ parser_lex(yp_parser_t *parser) {
                         break;
                     case '\\':
                         if (peek_at(parser, 1) == '\n') {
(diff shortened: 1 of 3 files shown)


flavorjones:
This PR fixes the problem I mentioned in https://github.com/ruby/yarp/pull/1274#issuecomment-1682341208 :tada:

flavorjones:
@HParker I suspect that if you rebase now you'll pull in https://github.com/ruby/yarp/pull/1276 which _might_ fix the windows test failures.

HParker:
Thanks @flavorjones my branch says it is up to date, but I wonder if git already removed the CRLF. Maybe I can fix that by renaming the file.

kddnewton (approved):

PR #1280, opened 2023-08-18
Remove strange :"#arg_rest" local

kddnewton:


lib/yarp.rb
```diff
@@ -381,6 +381,10 @@ def self.cruby_locals(source)
             # of those here.
             names = names.grep_v(/^_\d$/)
 
+            # For some reason, CRuby occasionally pushes this special local
+            # variable when there are splat arguments. We get rid of that here.
+            names = names.grep_v(:"#arg_rest")
```
k0kubun, line 386: :eyes: https://github.com/ruby/ruby/blob/e913431687f2fffb1a8cc435e60c95eea887b087/iseq.c#L2852

k0kubun (approved):

lib/yarp.rb
kddnewton: Yeah I have no idea why that happens, but _doesn't_ happen on my machine.

PR #1285, opened 2023-08-18
Refactor operator write nodes

kddnewton:
* `{ClassVariable,Constant,ConstantPath,GlobalVariable,InstanceVariable,LocalVariable}OperatorAndWriteNode` instead become `AndWriteNode` with their target being the original value being written to. I decided to keep them as read nodes to save on space because they'll never end up having a value and technically you have to read them first before anything else.
* Same for `OrWrite`.
* Same for `OperatorWrite`.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -736,21 +736,6 @@ nodes:
 
           class Foo end
           ^^^^^^^^^^^^^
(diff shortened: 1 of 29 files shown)


eregon (approved):
Looks great, thank you!

PR #1286, opened 2023-08-19
Change AndWriteNode, OrWriteNode, OperatorWriteNode to contain write nodes

kddnewton:
It is going to make it more difficult to find all locations where a variable is written to if they're just read nodes. To keep things consistent we should make them write nodes.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -12112,6 +12112,10 @@ parse_expression_infix(yp_parser_t *parser, yp_node_t *node, yp_binding_power_t 
                 case YP_NODE_INSTANCE_VARIABLE_READ_NODE:
                 case YP_NODE_LOCAL_VARIABLE_READ_NODE: {
                     parser_lex(parser);
+
+                    yp_token_t operator = not_provided(parser);
+                    node = parse_target(parser, node, &operator, NULL);
+
                     yp_node_t *value = parse_expression(parser, binding_power, "Expected a value after &&=");
                     return (yp_node_t *) yp_and_write_node_create(parser, node, &token, value);
                 }
@@ -12130,11 +12134,12 @@ parse_expression_infix(yp_parser_t *parser, yp_node_t *node, yp_binding_power_t 
                         }
 
                         parser_lex(parser);
-                        yp_node_t *target = (yp_node_t *) yp_local_variable_read_node_create(parser, &(yp_token_t) { .type = YP_TOKEN_IDENTIFIER, .start = message_loc.start, .end = message_loc.end }, 0);
-                        yp_node_t *value = parse_expression(parser, binding_power, "Expected a value after &&=");
 
-                        yp_node_destroy(parser, node);
-                        return (yp_node_t *) yp_and_write_node_create(parser, target, &token, value);
+                        yp_token_t operator = not_provided(parser);
+                        node = parse_target(parser, node, &operator, NULL);
+
+                        yp_node_t *value = parse_expression(parser, binding_power, "Expected a value after &&=");
+                        return (yp_node_t *) yp_and_write_node_create(parser, node, &token, value);
                     }
 
                     parser_lex(parser);
@@ -12174,6 +12179,9 @@ parse_expression_infix(yp_parser_t *parser, yp_node_t *node, yp_binding_power_t 
                 case YP_NODE_LOCAL_VARIABLE_READ_NODE: {
                     parser_lex(parser);
 
+                    yp_token_t operator = not_provided(parser);
+                    node = parse_target(parser, node, &operator, NULL);
+
                     yp_node_t *value = parse_expression(parser, binding_power, "Expected a value after ||=");
                     return (yp_node_t *) yp_or_write_node_create(parser, node, &token, value);
                 }
@@ -12192,11 +12200,12 @@ parse_expression_infix(yp_parser_t *parser, yp_node_t *node, yp_binding_power_t 
                         }
 
                         parser_lex(parser);
-                        yp_node_t *target = (yp_node_t *) yp_local_variable_read_node_create(parser, &(yp_token_t) { .type = YP_TOKEN_IDENTIFIER, .start = message_loc.start, .end = message_loc.end }, 0);
-                        yp_node_t *value = parse_expression(parser, binding_power, "Expected a value after ||=");
 
-                        yp_node_destroy(parser, node);
-                        return (yp_node_t *) yp_or_write_node_create(parser, target, &token, value);
+                        yp_token_t operator = not_provided(parser);
+                        node = parse_target(parser, node, &operator, NULL);
+
+                        yp_node_t *value = parse_expression(parser, binding_power, "Expected a value after ||=");
+                        return (yp_node_t *) yp_or_write_node_create(parser, node, &token, value);
                     }
 
                     parser_lex(parser);
@@ -12246,6 +12255,9 @@ parse_expression_infix(yp_parser_t *parser, yp_node_t *node, yp_binding_power_t 
                 case YP_NODE_LOCAL_VARIABLE_READ_NODE: {
                     parser_lex(parser);
 
+                    yp_token_t operator = not_provided(parser);
+                    node = parse_target(parser, node, &operator, NULL);
+
                     yp_node_t *value = parse_expression(parser, binding_power, "Expected a value after the operator");
                     return (yp_node_t *) yp_operator_write_node_create(parser, node, &token, value);
                 }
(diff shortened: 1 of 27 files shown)


enebo:
@kddnewton @eregon I should just ask @eregon how should I compile this node now?

```text
./bin/parse -e '@a += 2'
#<YARP::ParseResult:0x00007f600d4afaa0
 @comments=[],
 @errors=[],
 @source=#<YARP::Source:0x00007f600d3acbf8 @offsets=[0], @source="@a += 2">,
 @value=
  ProgramNode(0...7)(
    [],
    StatementsNode(0...7)(
      [OperatorWriteNode(0...7)(
         InstanceVariableWriteNode(0...2)((0...2), nil, nil),
         (3...5),
         :+,
         IntegerNode(6...7)()
       )]
    )
  ),
 @warnings=[]>
```

Now I have to examine the write node and then determine I need to effectively make a read node by a switch?  I still need to retrieve the value before I can perform the operation.  I obviously can make a switch of all types supported by OperatorWriteNode but this feels pretty gross.  What am I missing?

enebo:
@kddnewton And the write has ephemeral values marked as final so I can't even use the writenode.  I am back to logic like this:

```java
    private Operand buildOperatorAssignment(OperatorAssignmentNode node) {
        // Strip off '=' as we don't need it.
        Token operator = new Token(node.operator.type, node.operator.startOffset, node.operator.endOffset - 1);
        RubySymbol oper = symbolFor(operator);
        byte[] name = oper.getBytes().bytes();
        Node target = node.target;
        Node arguments = new ArgumentsNode(new Node[] { node.value }, -1, -1);
        if (target instanceof InstanceVariableWriteNode) {
            Node receiver = new InstanceVariableReadNode(target.startOffset, target.endOffset);
            Node value = new CallNode(receiver, operator, arguments, null, name, -1, -1);
            return build(new InstanceVariableWriteNode(new Location(target.startOffset, target.endOffset), value, target.startOffset, target.endOffset));
        } else if (target instanceof GlobalVariableWriteNode) {
            Node receiver = new GlobalVariableReadNode(((GlobalVariableWriteNode) target).name, target.startOffset, target.endOffset);
            Node value = new CallNode(receiver, operator, arguments, null, name, -1, -1);
            return build(new GlobalVariableWriteNode(((GlobalVariableWriteNode) target).name, value, target.startOffset, target.endOffset));
        } else if (target instanceof ClassVariableWriteNode) {
            Node receiver = new ClassVariableReadNode(target.startOffset, target.endOffset);
            Node value = new CallNode(receiver, operator, arguments, null, name, -1, -1);
            return build(new ClassVariableWriteNode(((ClassVariableWriteNode) target).name_loc, value, target.startOffset, target.endOffset));
        } else if (target instanceof LocalVariableWriteNode) {
            Node receiver = new LocalVariableReadNode(((LocalVariableWriteNode) target).depth, target.startOffset, target.endOffset);
            Node value = new CallNode(receiver, operator, arguments, null, name, -1, -1);
            return build(new LocalVariableWriteNode(((LocalVariableWriteNode) target).name_loc, value, ((LocalVariableWriteNode) target).depth, target.startOffset, target.endOffset));
        } else if (target instanceof CallNode) {
            Operand receiver = build(target);
            Variable value = call(temp(), receiver, oper, new Operand[] { build(node.value) });
            RubySymbol writeName = symbol(new ByteList(((CallNode) target).name)).asWriter();
            return call(temp(), build(((CallNode) target).receiver), writeName, new Operand[]{ value });
        }

        throw notCompilable("buildOperatorAssignment node not known", target);
    }
```

kddnewton:
Yeah, we have to do a similar thing in CRuby now after this change as well.

eregon:
I think we could add a `withValue(valueNode)` method on write nodes.
Maybe even a `WriteNode#createReadNode()`.
Then it could be desugared generically like this:
```java
node.target.withValue(CallNode(node.target.createReadNode(), node.operator, node.target.value));
```

For `@a += 2` JRuby fully desugars it:
```
RootNode line: 0
  InstAsgnNode*:@a line: 0
    OperatorCallNode*:+ line: 0
      InstVarNode:@a line: 0
      ArrayNode line: 0
        FixnumNode line: 0, long: 2
```
YARP could do the same and wrap the write node in a OperatorWriteNode so syntactically it's easy to process as well.

But if we look at other operator, e.g. `@a ||= 2`:
```
RootNode line: 0
  OpAsgnOrNode* line: 0
    InstVarNode:@a line: 0
    InstAsgnNode:@a line: 0
      FixnumNode line: 0, long: 2
```
Then you can see it's very similar to YARP:
```
OrWriteNode(0...8)(
  InstanceVariableWriteNode(0...2)((0...2), nil, nil), IntegerNode(7...8)(), (3...6)
)
```

I think having the read node too might be nice, but that's clearly worse for footprint & serialized size.

In general, either we have `WriteNode#createReadNode()` in YARP or usages have a way to convert a YARP write node to their own read node/instruction. That's needed anyway for `||=`, unless we add a read node there in addition to the write node.

eregon:
> or usages have a way to convert a YARP write node to their own read node/instruction

And my point is with that the logic above in https://github.com/ruby/yarp/pull/1286#issuecomment-1686802344 would look a lot cleaner and have much less duplication (with the handling of OrWriteNode/AndWriteNode)

enebo:
@eregon so you are saying we effectively create two new nodes from two new methods during processing so we can effectively desugar without having it in the tree originally?  Yeah compared to my snippet it would be a lot cleaner but I would like to not create extra nodes at all.

I guess I don't like it.  The current solution solves not having more node types (although more nodes overall) at the cost of more post serialization execution and allocation.  The original solution had more node types but was smaller in it's serialized form.  Involves less execution (we both transform or make completely new objects from the AST without needing to make intermediate objects).   It was less polymorphic in code which will not warm up in short-executed use-cases like 'gem list'.  It did create more methods but they were really simple and could be consolidated to call into a single common method if you cared enough to do that. 

The second argument I remember was complication from processing from Ruby but I think we can solve this problem.  Since kind now is the full type name and not Node we could just make all ```InstanceVariableOperatorWriteNode < OperatorWriteNode``` and add a visitor for that generic type.  Duck-typing would provide the same methods and no one would be the wiser.

eregon:
> @eregon so you are saying we effectively create two new nodes from two new methods during processing so we can effectively desugar without having it in the tree originally? Yeah compared to my snippet it would be a lot cleaner but I would like to not create extra nodes at all.

Yes, if YARP doesn't desugar (e.g. so it's still faithful to the source syntactically, and anyway for `a.b += 2` it can't be fully desugared) then I think the execution usages should sort of desugar themselves, one way or another.

Another way to think about it is if it was desugared by YARP (like in JRuby parser) you would end up allocating those nodes anyway.

In TruffleRuby we have the concept of [AssignableNode](https://github.com/oracle/truffleruby/blob/master/src/main/java/org/truffleruby/core/array/AssignableNode.java) which can be a good way to deal with this overwriting of the value (not in the YARP AST but in the result after compilation/translation), or when there is no natural value node (e.g. multiple assignment).
Maybe a similar concept can be used for bytecode too?

eregon:
BTW a while ago I thought it could be nice if YARP would have a opt-in way to desugar nodes, so e.g. for `@a += 2` it could fully desugar, which would be very convenient for execution/compilation/translation.
That wouldn't introduce any extra node types, so I think it'd work quite well.

(Maybe it could even be used for complex things like pattern matching, although that would probably have some non-trivial overhead compared to manually dealing with it, could still be worth to support pattern with much less effort initially.)

enebo:
> Another way to think about it is if it was desugared by YARP (like in JRuby parser) you would end up allocating those nodes anyway.

The previous design in YARP did not require desugaring so it ended up not requiring these allocations.  Both of our impls plan on making other objects from what these nodes represent.   JRuby's (and MRI's) current OpAsgnNode match each other which is good for maintaining compatibility (two parsers need to be synced) but it I think fails in that is requires desugaring the write as a parent opasgnnode.  We can fix that and not have the allocations or extra execution to make those allocations.

enebo:
> BTW a while ago I thought it could be nice if YARP would have a opt-in way to desugar nodes, so e.g. for `@a += 2` it could fully desugar, which would be very convenient for execution/compilation/translation. That wouldn't introduce any extra node types, so I think it'd work quite well.

If you are interested in that then I don't care if there is such an option but personally I have no interest in something like that.  Perhaps someone could use it for a new runtime where they wanted to reduce how much code they need to write?  In sense of priority this seems like it trades off important traits like memory/non-warm execution for removing some code.

PR #1290, opened 2023-08-19
Add a convenience value method for numeric literals

eregon:
From https://github.com/ruby/yarp/issues/1098#issuecomment-1613821135

For instance this is necessary for https://github.com/ruby/yarp/issues/1237#issuecomment-1684896663

### DIFF
--- a/lib/yarp.rb
+++ b/lib/yarp.rb
@@ -313,6 +313,30 @@ module YARP
     end
   end
 
+  class FloatNode < Node
+    def value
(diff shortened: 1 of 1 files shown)


kddnewton:
Can you add some tests? Particularly just around the rational/imaginary ones.

kddnewton (changes requested):
Just marking this as "request changes" to make it easier to track.

kddnewton (approved):

PR #1291, opened 2023-08-19
WIP: Add converter from YARP AST to parser gem AST

eregon:
A very early prototype for it.
Not sure where this code should live yet, just sharing the PR in case it interests others.

See https://github.com/ruby/yarp/issues/1237#issuecomment-1684896663

### DIFF
--- a/lib/yarp/parser_gem_compat.rb
+++ b/lib/yarp/parser_gem_compat.rb
@@ -1,0 +1,259 @@
+# frozen_string_literal: true
+
+require 'ast'
+
+module YARP
+  module ParserGemCompat
+    class Visitor < ::YARP::BasicVisitor
+      include ::AST::Sexp
+
+      def visit_program_node(node)
+        convert node.statements
+      end
+
+      def visit_statements_node(node)
+        statements = node.body
+        if statements.size == 1
+          convert statements[0]
+        else
+          s(:begin, *convert_array(statements))
+        end
+      end
+
+      private def visit_statements_node_no_single(node)
+        s(:begin, *convert_array(node.body))
+      end
+
+      def visit_begin_node(node)
+        if node.rescue_clause
+          s(:kwbegin, s(:rescue, convert(node.statements), convert(node.rescue_clause)), convert_or_nil(node.else_clause))
+        else
+          super
+        end
+      end
+
+      def visit_rescue_node(node)
+        s(:resbody, s(:array, *convert_array(node.exceptions)), convert_or_nil(node.reference), convert(node.statements))
+      end
+
+      def visit_if_node(node)
+        s(:if, convert(node.predicate), convert_or_nil(node.statements), convert_or_nil(node.consequent))
+      end
+
+      def visit_unless_node(node)
+        s(:if, convert(node.predicate), convert_or_nil(node.consequent), convert_or_nil(node.statements))
+      end
+
+      def visit_else_node(node)
+        convert(node.statements)
+      end
+
+      def visit_next_node(node)
+        s(:next)
+      end
+
+      def visit_module_node(node)
+        s(:module, convert(node.constant_path), convert_or_nil(node.body))
+      end
+
+      def visit_class_node(node)
+        s(:class, convert(node.constant_path), convert_or_nil(node.superclass), convert_or_nil(node.body))
+      end
+
+      def visit_def_node(node)
+        parameters = convert_or_nil(node.parameters) || s(:args)
+        if node.receiver
+          s(:defs, convert(node.receiver), node.name.to_sym, parameters, convert_or_nil(node.body))
+        else
+          s(:def, node.name.to_sym, parameters, convert_or_nil(node.body))
+        end
+      end
+
+      def visit_block_node(node)
+        parameters = convert_or_nil(node.parameters) || s(:args)
+        [parameters, convert_or_nil(node.body)]
+      end
+
+      def visit_block_parameters_node(node)
+        raise unless node.locals.empty?
+        convert(node.parameters)
+      end
+
+      def visit_parameters_node(node)
+        children = [*convert_array(node.requireds), *convert_array(node.optionals)]
+        children << convert(node.rest) if node.rest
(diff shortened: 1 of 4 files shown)


lib/yarp/parser_gem_compat.rb
```diff
@@ -0,0 +1,80 @@
...
+      when FloatNode
+        s(:float, node.value)
+      when IntegerNode
+        s(:int, node.value)
+      when TrueNode
+        s(:true)
+      when FalseNode
+        s(:false)
+      when NilNode
+        s(:nil)
```
eregon, line 74: It would be nice to have a `LiteralNode` superclass or included module in those, and that would define a `value` method which returns the literal value.

lib/yarp/parser_gem_compat.rb
```diff
@@ -0,0 +1,80 @@
...
+      when AssocNode
+        s(:pair, convert(node.key), convert(node.value))
+      when AssocSplatNode
+        s(:kwsplat, convert(node.value))
+      when CallNode
+        arguments = node.arguments ? node.arguments.arguments.map { convert _1 } : []
+        receiver = node.receiver ? convert(node.receiver) : nil
+        s(:send, receiver, node.name.to_sym, *arguments)
+      when LocalVariableReadNode
+        s(:lvar, node.constant_id)
```
eregon, line 42: `node.constant_id` is quite weird, as mame noticed.
`node.name` would be nice.

lib/yarp/parser_gem_compat.rb
```diff
@@ -0,0 +1,80 @@
...
+      when LocalVariableWriteNode
+        if node.value
+          s(:lvasgn, node.constant_id, convert(node.value))
```
eregon, line 45: `node.constant_id` is quite weird, as mame noticed.
`node.name` here does exist but return a String and not Symbol, so that's extra confusing.

lib/yarp/parser_gem_compat.rb
```diff
@@ -0,0 +1,80 @@
...
+      when HashNode
+        s(:hash, *node.elements.map { convert _1 })
+      when AssocNode
+        s(:pair, convert(node.key), convert(node.value))
+      when AssocSplatNode
+        s(:kwsplat, convert(node.value))
+      when CallNode
+        arguments = node.arguments ? node.arguments.arguments.map { convert _1 } : []
+        receiver = node.receiver ? convert(node.receiver) : nil
+        s(:send, receiver, node.name.to_sym, *arguments)
```
eregon, line 40: Would be nice if `node.name` returns a Symbol for CallNode.
Maybe `name` fields in general should always be Symbols?

kddnewton:
Can you open any changes over on https://github.com/kddnewton/parser-yarp?

kddnewton:
@eregon can you check out the repo I posted? Want to make sure we don't duplicate effort here

eregon:
Ah just saw this now, looks like you did most/all of it already :)
It was fun to do anyway, I'll check that repo soonish.

lib/yarp/parser_gem_compat.rb
```diff
@@ -0,0 +1,259 @@
...
+      def visit_class_node(node)
+        s(:class, convert(node.constant_path), convert_or_nil(node.superclass), convert_or_nil(node.body))
+      end
+
+      def visit_def_node(node)
+        parameters = convert_or_nil(node.parameters) || s(:args)
+        if node.receiver
+          s(:defs, convert(node.receiver), node.name.to_sym, parameters, convert_or_nil(node.body))
+        else
+          s(:def, node.name.to_sym, parameters, convert_or_nil(node.body))
```
eregon, line 68: The parser gem has different types for this. It seems much nicer unified as in YARP as they are quite similar.

lib/yarp/parser_gem_compat.rb
```diff
@@ -0,0 +1,259 @@
...
+
+      def visit_assoc_splat_node(node)
+        s(:kwsplat, convert(node.value))
+      end
+
+      def visit_call_node(node)
+        arguments = node.arguments ? convert_array(node.arguments.arguments) : []
+        converted = s(:send, convert_or_nil(node.receiver), node.name.to_sym, *arguments)
+        if node.block
+          s(:block, converted, *convert(node.block))
```
eregon, line 132: This is quite weird in the parser gem wrap the `:send` with a `:block` when passing a block, much nicer in YARP.
Ignoring that, the parser gem also has a single node for CallNode, nice.

lib/yarp/parser_gem_compat.rb
```diff
@@ -0,0 +1,259 @@
...
+      def visit_array_node(node)
+        s(:array, *convert_array(node.elements))
+      end
+
+      def visit_hash_node(node)
+        s(:hash, *convert_array(node.elements))
+      end
+
+      def visit_keyword_hash_node(node)
+        visit_hash_node(node)
```
eregon, line 117: The parser gem seems to not differentiate kwargs for a last argument positional hash.
Maybe via some flag or location or something.
Probably that's for compatibility for Ruby < 2.7/3.0 which didn't differentiate them.

I think it'd be nicer if YARP CallNode would have positional args, kwargs and block in separate fields like https://github.com/ruby/yarp/pull/1243, but the current state is undeniably better than in parser gem.

lib/yarp/parser_gem_compat.rb
```diff
@@ -0,0 +1,259 @@
...
+        raise unless ConstantReadNode === node.child
+        s(:const, convert(node.parent), node.child.slice.to_sym)
+      end
+
+      def visit_constant_write_node(node)
+        s(:casgn, nil, node.name.to_sym, convert(node.value))
+      end
+
+      def visit_operator_write_node(node)
+        s(:op_asgn, convert(node.target), node.operator, convert(node.value))
```
eregon, line 181: So easy to map now that those YARP nodes got consolidated :)

eregon (commented):
Leaving some notes here comparing parser gem and YARP AST while I was playing with converting one to another.

PR #1293, opened 2023-08-20
fix: support for newline-terminated `%q`, `%Q`, `%r`

flavorjones:
Support for newline-terminated `%q`, `%Q`, `%r` is fixed in this changeset.

For `%r`, there are two fixes:

- avoid double-tracking ending newline
- correctly track the beginning newline

For the rest, only the beginning newline needed to be fixed.

Also test coverage is backfilled for newline-terminated `%`.

(These issues were detected with the fuzzer.)

A related note for posterity: I was mildly surprised to see that `%i`, `%I`, `%w`, and `%w` do not support newline-termination in CRuby, though I suppose that's sensible since newline can also be a separator.


test/parse_test.rb
```diff
@@ -141,7 +141,7 @@ def test_parse_takes_file_path
       file_contents = File.read(filepath, binmode: true, external_encoding: Encoding::UTF_8)
 
       file_contents.split(/(?<=\S)\n\n(?=\S)/).each do |snippet|
-        snippet = snippet.rstrip
+        snippet = snippet.rstrip + "\n"
```
kddnewton, line 144: This is unfortunately a super specific use case of we copy the files so that we can check for out of bounds reads with valgrind. We've caught a couple of things this way, so I think I don't want to add a newline here just in case it helps us catch things in the the future. 

Could you explicitly skip the newline terminated file you gave so that we can keep this behavior?

kddnewton (changes requested):

test/parse_test.rb
flavorjones: Sure.

kddnewton (approved):

PR #1294, opened 2023-08-20
fix: parsing a '%' expression with a CR but not a newline

flavorjones:
That is, this Ruby code: `%\r` without a trailing newline.

Previously this failed an assertion and aborted.

This was found via the fuzzer.


flavorjones:
Waiting to fix the memcheck issue until #1293 is merged, because there will be conflicts.

HParker (approved):
Nice! the early `\n` makes sense to me.

PR #1300, opened 2023-08-22
tell Cargo about the `config.yml` dependency for `yarp`

froydnj:
Otherwise we get peculiar build errors because Cargo doesn't think anything has changed.

### DIFF
--- a/rust/yarp/build.rs
+++ b/rust/yarp/build.rs
@@ -70,7 +70,9 @@ struct Config {
 /// building the library.
 /// 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let file = std::fs::File::open(config_path())?;
+    let path = config_path();
+    let file = std::fs::File::open(&path)?;
+    println!("cargo:rerun-if-changed={}", path.display());
     let config: Config = serde_yaml::from_reader(file)?;
 
     write_bindings(&config)?;


rust/yarp/build.rs
```diff
@@ -70,7 +70,9 @@ struct Config {
 /// building the library.
 /// 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let file = std::fs::File::open(config_path())?;
+    let path = config_path();
+    let file = std::fs::File::open(&path)?;
+    println!("cargo:rerun-if-changed={}", path.display());
```
lopopolo, line 75: `path.display()` escapes invalid UTF-8 bytes. I delivered extensive feedback on both of these crate build scripts previously that this code should call `path.to_str().unwrap()`.

froydnj: @lopopolo I didn't see that feedback; is it possible that you can point me to it?

lopopolo: I left several comments of this sort on the initial PRs that introduced the `yarp-sys` and `yarp` crates.

Using `Path::display` to output a `Path` with `println!` is a lossy operation because `println!` requires UTF-8 and `Path`s on Unix and windows platforms are not required to be UTF-8.

Rather than print out a cargo directive with a lossy conversion which will cause the build to fail in odd ways further down the pipe, prefer to fail fast and require a UTF-8 path at the time it is used as such.

kddnewton: https://github.com/ruby/yarp/pull/1353

PR #1301, opened 2023-08-22
[rust] add an `as_byte_slice` method to `Location`

froydnj:
This new function is slightly redundant with `ParseResult::as_slice`, but the new function is consistent with `Comment::text`.  And like `Comment::text`, `Location::as_byte_slice` is more useful when you are writing visitors and what to examine some of the source text of a node (my motivating example was examining the opening delimiter of `InterpolatedStringNode` for determining heredocs).

Open to changing the name for consistency with `ParseResult::as_slice` or `str::as_bytes`.

### DIFF
--- a/rust/yarp/build.rs
+++ b/rust/yarp/build.rs
@@ -410,15 +410,23 @@ impl<'pr> Location<'pr> {{
     pub fn end(&self) -> *const c_char {{
         unsafe {{ self.pointer.as_ref().end }}
     }}
+
+    /// Returns a byte slice for the range.
+    #[must_use]
+    pub fn as_byte_slice(&self) -> &'pr [u8] {{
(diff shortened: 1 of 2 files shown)


HParker:
I think this makes sense. I am beginning to wonder if we should have rust tests exercising code like this. @froydnj do you know what is typical in projects like this?

froydnj:
That's a good idea @HParker !  I pushed a (very) rudimentary test; let me know what you think.

HParker (approved):
Nice! I kinda like hiding the unsafe code from `fmt` as well.

kddnewton:
@froydnj would you be able to make another PR to make it consistently as_slice? I find it somewhat confusing to have two different names.

PR #1302, opened 2023-08-22
All bin/ scripts which take filepaths default to "test.rb"

jemmaissroff:
For instance, if you don't pass any args to bin/parse, it will now default to "bin/parse test.rb" This commit applies this change to bin/lex, bin/locals and bin/parse


HParker (approved):
👍

PR #1303, opened 2023-08-22
[rust] add a converter from concrete nodes to generic `Node`

froydnj:
The motivating example for this change is a visitor that wants to maintain parent node information as it traverses the AST.  The type of the stack maintained during the traversal should be `Vec<Node>`, but there's no way from the concrete nodes to upcast to the `Node` enum.


HParker (approved):
This pattern is also common in the C code, so makes sense you sometimes want a generic node.

HParker:
@froydnj What do you think about a test here as well?

froydnj:
@HParker done!  (Mostly just to make sure that `as_node()` compiles correctly.)

kddnewton (commented):
Yeah just squash and then we can merge

froydnj:
Squashed, thanks!

PR #1306, opened 2023-08-22
Match EOF after newline behavior

HParker:
in Ripper EOL after whitespace is returned as a on_nl node with the whitespace as the content.

### DIFF
--- a/lib/yarp/lex_compat.rb
+++ b/lib/yarp/lex_compat.rb
@@ -630,6 +630,14 @@ module YARP
               end
 
             Token.new([[lineno, column], event, value, lex_state])
+
+          when :on_eof
+            prev_token = result_value[index-1][0]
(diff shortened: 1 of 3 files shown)


kddnewton:
@HParker I don't think replicating it in our lexer necessarily buys us anything other than compatibility since we don't need it for the token stream. What do you think of replicating it in `lex_compat` instead?

kddnewton (changes requested):
Just marking this as "request changes"

HParker:
@kddnewton how does this look?

I think this is the only case where there is a `on_nl` without a `\n` in it, but maybe you have a better idea how to identify it.

kddnewton:
@HParker ahh I'm so sorry I misled you. I meant in our lex_compat, not changing ripper's output. I'm saying we should change our function (on the Ruby side) to return this extra token.

kddnewton (approved):

PR #1307, opened 2023-08-22
Add a ScopeNode and a way to create scope nodes from existing nodes

jemmaissroff:
This commit creates a scope node, and exposes a method to get ScopeNodes from other existing nodes. It still has TODOs around creating ScopeNodes for other types of nodes, which will be done in future commits.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1686,6 +1686,20 @@ nodes:
 
           return 1
           ^^^^^^^^
+  - name: ScopeNode
+    child_nodes:
+      - name: parameters
+        type: node?
+        kind: ParametersNode
+      - name: statements
+        type: node?
+        kind: StatementsNode
+      - name: locals
+        type: constant[]
+    comment: |
+      Used only to retrieve the scope, not an
+      actual semantically meaningful node
+
   - name: SelfNode
     comment: |
       Represents the `self` keyword.
--- a/config.yml
+++ b/config.yml
@@ -1686,20 +1686,6 @@ nodes:
 
           return 1
           ^^^^^^^^
-  - name: ScopeNode
-    child_nodes:
-      - name: parameters
-        type: node?
-        kind: ParametersNode
-      - name: statements
-        type: node?
-        kind: StatementsNode
-      - name: locals
-        type: constant[]
-    comment: |
-      Used only to retrieve the scope, not an
-      actual semantically meaningful node
-
   - name: SelfNode
     comment: |
       Represents the `self` keyword.
--- a/include/yarp/node.h
+++ b/include/yarp/node.h
@@ -41,6 +41,6 @@ YP_EXPORTED_FUNCTION const char * yp_node_type_to_str(yp_node_type_t node_type);
 typedef struct yp_scope_node {
     yp_node_t base;
     struct yp_parameters_node *parameters;
-    struct yp_statements_node *statements;
+    yp_node_t *body;
     yp_constant_id_list_t locals;
 } yp_scope_node_t;
--- a/include/yarp/node.h
+++ b/include/yarp/node.h
@@ -34,3 +34,13 @@ YP_EXPORTED_FUNCTION const char * yp_node_type_to_str(yp_node_type_t node_type);
 #define YP_EMPTY_LOCATION_LIST ((yp_location_list_t) { .locations = NULL, .size = 0, .capacity = 0 })
 
 #endif // YARP_NODE_H
+
+// ScopeNodes are helper nodes, and will never
+// be part of the AST. We manually declare them
+// here to avoid generating them
+typedef struct yp_scope_node {
+    yp_node_t base;
+    struct yp_parameters_node *parameters;
+    struct yp_statements_node *statements;
+    yp_constant_id_list_t locals;
+} yp_scope_node_t;
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -1017,7 +1017,7 @@ yp_block_argument_node_create(yp_parser_t *parser, const yp_token_t *operator, y
 }
 
 static void
-YP_ATTRIBUTE_UNUSED yp_scope_node_create(yp_parameters_node_t *parameters, yp_statements_node_t *statements, yp_constant_id_list_t locals, const char *start, const char *end, yp_scope_node_t *dest)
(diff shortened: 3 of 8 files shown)


src/yarp.c
```diff
@@ -1016,6 +1016,55 @@ yp_block_argument_node_create(yp_parser_t *parser, const yp_token_t *operator, y
...
+          break;
+      }
+
+      default:
+        assert(false && "unreachable");
+        return;
+    }
+
+    yp_scope_node_create(parameters, statements, locals, start, end, dest);
+    return;
```
HParker, line 1065: This creates the scope node, but doesn't return it?

Is the parser state going to be responsible for keeping track of this node eventually?

jemmaissroff: Yep, the caller passes the pointer to the scope node. This way, in the compiler we can allocate the memory for the scope node and not have to worry about doing that on the YARP side

HParker (approved):

eregon:
What's the purpose of it? Isn't it enough to use the fields of DefNode, etc directly?

Or is the idea to move `constant[] locals` to a ScopeNode and have DefNode have a ScopeNode field?
That seems potentially problematic for BlockNode which has a BlockParametersNode and not a ParametersNode.

eregon:
Is it maybe just a helper for C code using YARP?
If so it feels like ScopeNode shouldn't exist in Ruby & Java then, since it's never really part of the AST returned by YARP.parse.
I think it should be its own manually declared `struct` then and not end with `node` since it's not a node.

src/yarp.c
```diff
@@ -1016,6 +1016,62 @@ yp_block_argument_node_create(yp_parser_t *parser, const yp_token_t *operator, y
...
+        .parameters = parameters,
+        .statements = statements,
+        .locals = locals,
+    };
+    return;
+}
+
+// TODO: Implement this for all other nodes which can have a scope
+void
+yp_get_scope_node(yp_node_t *node, yp_scope_node_t *dest) {
```
kddnewton, line 1036: Just in terms of naming, the codebase has loosely been `yp_` + noun + verb. I haven't written that down anywhere though, so that's on me! But would you be able to rename this to `yp_scope_node_init` or something similar?

config.yml
```diff
@@ -1686,6 +1686,20 @@ nodes:
...
           return 1
           ^^^^^^^^
+  - name: ScopeNode
+    child_nodes:
+      - name: parameters
+        type: node?
+        kind: ParametersNode
+      - name: statements
+        type: node?
+        kind: StatementsNode
```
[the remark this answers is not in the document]
jemmaissroff: What's the right type here, then? I haven't implemented rescues in the compiler yet, I'm not sure if they compile totally separately to the body or not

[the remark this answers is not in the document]
kddnewton: I would imagine you would leave off `kind:` here and handle it in the compiler. In that way it would be a `yp_node_t *` which you would type check and then cast.

src/yarp.c
```diff
@@ -1016,6 +1016,66 @@ yp_block_argument_node_create(yp_parser_t *parser, const yp_token_t *operator, y
...
+      case YP_NODE_SINGLETON_CLASS_NODE: {
+          yp_singleton_class_node_t *singleton_class_node = (yp_singleton_class_node_t *) node;
+          statements = (yp_statements_node_t *)singleton_class_node->body;
+          locals = singleton_class_node->locals;
+          break;
+      }
+
+      default:
+        assert(false && "unreachable");
+        return;
```
eregon, line 1072: This seems quite dangerous, as then it leaves the `yp_scope_node_t` uninitialized.

src/yarp.c
```diff
@@ -1016,6 +1016,73 @@ yp_block_argument_node_create(yp_parser_t *parser, const yp_token_t *operator, y
...
+    yp_parameters_node_t *parameters = NULL;
+    yp_node_t *body;
+    yp_constant_id_list_t locals;
+    const char *start = node->location.start;
+    const char *end = node->location.end;
+
+    switch (node->type) {
+      case YP_NODE_BLOCK_NODE: {
+          yp_block_node_t *block_node = (yp_block_node_t *) node;
+          body = (yp_node_t *)block_node->body;
```
kddnewton, line 1046: These casts shouldn't be necessary here because they should already be `yp_node_t`

templates/src/serialize.c.erb
```diff
@@ -33,6 +33,10 @@ yp_serialize_node(yp_parser_t *parser, yp_node_t *node, yp_buffer_t *buffer) {
     serialize_location(parser, &node->location, buffer);
 
     switch (YP_NODE_TYPE(node)) {
+        // We do not need to serialize a ScopeNode ever as
+        // it is not part of the AST
+        case YP_NODE_SCOPE_NODE:
+          return;
```
kddnewton, line 39: Just small indentation things here and in the other ones.

kddnewton (changes requested):

kddnewton (approved):

include/yarp/node.h
```diff
@@ -34,3 +34,13 @@ YP_EXPORTED_FUNCTION const char * yp_node_type_to_str(yp_node_type_t node_type);
...
+
+// ScopeNodes are helper nodes, and will never
+// be part of the AST. We manually declare them
+// here to avoid generating them
+typedef struct yp_scope_node {
+    yp_node_t base;
+    struct yp_parameters_node *parameters;
+    yp_node_t *body;
+    yp_constant_id_list_t locals;
+} yp_scope_node_t;
```
eregon, line 46: This definition should still be before `#endif // YARP_NODE_H`

eregon:
> Is it maybe just a helper for C code using YARP?

It seems so: https://github.com/ruby/yarp/pull/1307/files#diff-0fde49ffd5b93b6e12016a9869a7ddccaf881a68e979c5a90bf99c89836df604R38-R40

It would be nice if such fundamental questions about the PR is answered and maybe the description improved, it would be much clearer what the PR is about.

src/yarp.c
eregon: What about this? It will be likely be a segfault if calling this function on one of the node types not handled by it and then accessing the `yp_scope_node_t`.

kddnewton: It was fixed in a follow-up PR

PR #1308, opened 2023-08-22
refactor: extract `peek_addr()`

flavorjones:
In many places in the code we use the idiom:

    x < parser->end && *x == 'y'

which is essentially an extension of an existing pattern:

- `peek()` looks at `parser->current.end`
- `peek_at()` looks at `(parser->current.end + offset)`

This commit introduces a new inline function, `peek_addr`, which accepts a pointer and encapsulates the address value check and conditional dereferencing. The result is more readable code, and more ubiquitous safety checks on pointer values, allowing us to rewrite the above as:

    peek_addr(parser, x) == 'y'

Also:

- change the type of `peek_at()`'s offset argument from `size_t` to `ptrdiff_t` so that it can accept negative offsets.
- use `current_token_starts_line` in one place where the equivalent code is inline.

These changes simplify the code and make it easier to visually spot patterns, particularly around line-endings (which will be a subject of a future pull request).

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -4157,27 +4157,30 @@ yp_do_loop_stack_p(yp_parser_t *parser) {
 /* Lexer check helpers                                                        */
 /******************************************************************************/
 
+// Get the next character in the source starting from +cursor+. If that position is beyond the end
+// of the source then return '\0'.
+static inline char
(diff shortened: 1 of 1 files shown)


HParker (approved):
This is a lot of great readability improvements. 👍

kddnewton (approved):

PR #1311, opened 2023-08-23
A whole mess of small fixes

kddnewton:
In converting the AST to the whitequark/parser gem's AST (https://github.com/kddnewton/parser-yarp) I found a bunch of very small things that were either incorrect, could be better, or were just nice to have. I bundled them up into this PR as they're all relatively small. It might be easiest to review this commit by commit. Here is a summary of the changes:

* `foo.()` syntax should result in the `message_loc` being `NULL`, since there isn't one in this case.
* `WhileNode` and `UntilNode` weren't tracking their end location properly - that's been added.
* Interpolated symbols that were converted to regular symbols (e.g., `:"abc"`) were losing their opening and closing locations. That's been fixed.
* Lambda literals weren't tracking their `{}`/`do`..`end` locations. That's been added.
* Rational nodes had a bug with their number parsing (it should be passing the string to `Rational`, otherwise it gets messed up with floats like `42.1`).
* I added a small `Location#to` API that accepts another `Location` and combines them.
* Multi write nodes had the incorrect location bounds when they had parentheses. That's been fixed.
* `YARP::parse` and `YARP::parse_file` now accept blocks which get yielded the tokens during processing. This is optional and is only done when a block is passed. This makes it so that we can get the lex and parse output in one go instead of processing the file twice.


lib/yarp.rb
```diff
@@ -102,6 +102,11 @@ def ==(other)
         other.end_offset == end_offset
     end
 
+    # Returns a new location that is the union of this location and the other.
+    def to(other)
```
jemmaissroff, line 106: `#to` is a slightly confusing name for me here, what about `#union`?

jemmaissroff (changes requested):

lib/yarp.rb
eregon: In TruffleRuby we call this `extend`: https://github.com/oracle/truffleruby/blob/4468b51ad29f7b4b197024e2baf9b0137eb75eed/src/main/java/org/truffleruby/parser/ast/ParseNode.java#L91-L98

`union(2..3, 6..7)` might be expected to yield 2..3 + 6..7 and not 2..7.

lib/yarp.rb
```diff
@@ -102,6 +102,11 @@ def ==(other)
...
+      Location.new(source, start_offset, other.end_offset - start_offset)
```
eregon, line 107: What if other.start_offset < start_offset? Then this would return an incorrect result.
The logic should be similar to https://github.com/oracle/truffleruby/blob/4468b51ad29f7b4b197024e2baf9b0137eb75eed/src/main/java/org/truffleruby/parser/ast/ParseNode.java#L91-L98
So min() of start offsets and length = max() of end offsets - the result of min()

ext/yarp/extension.c
```diff
@@ -302,16 +302,65 @@ lex_file(VALUE self, VALUE filepath) {
...
     yp_parser_init(&parser, yp_string_source(input), yp_string_length(input), filepath);
 
+    VALUE offsets;
+    VALUE source;
+
+    // If a block was given to the parse method, then we're going to register a
+    // lex callback that will yield the tokens to the block. This means you can
+    // get the lexer and the parser output in one method call instead of having
+    // to parse twice.
+    if (rb_block_given_p()) {
```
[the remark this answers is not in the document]
eregon: BTW does the parser gem needs tokens and nodes, not just nodes?

lib/yarp.rb
kddnewton: @jemmaissroff I updated it to `join`, which mirrors what the parser gem has. I didn't like `extend` only because it's already a Kernel method.

lib/yarp.rb
kddnewton: I've added tests for that now.

ext/yarp/extension.c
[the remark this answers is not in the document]
kddnewton: The parser gem doesn't need tokens, but rubocop does. I don't really know why.

I've updated it to return `[node, tokens]` in a parse result instead of a block, and it works with the FFI now.

jemmaissroff (approved):

lib/yarp.rb
eregon: Given the current unidirectional semantics, a good name might be `until`: `loc.until(other_loc)`.
`join`/`union` sounds like they would be commutative, but the impl is not.

ext/yarp/extension.c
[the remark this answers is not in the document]
eregon: Nice!

PR #1316, opened 2023-08-23
Improve handling of line endings

flavorjones:
Introduce three new inline helper functions:

- `match_line_ending`
- `match_line_ending_at`
- `match_line_ending_addr`

These functions are similar in signature to the `peek*` functions, but return the length of the line ending being inspected (or 0 if no line ending was found).

These functions are then used to simplify how we're detecting line endings throughout "src/yarp.c".

Also:
- test coverage backfilled for `__END__` comments with CRLF line endings.
- error message for invalid `%` tokens updated to not include the potential line endings.
- some small refactorings for readability along the way

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -4216,6 +4216,33 @@ next_newline(const char *cursor, ptrdiff_t length) {
     return memchr(cursor, '\n', (size_t) length);
 }
 
+// Return the length of the line ending string starting at +cursor+, or 0 if it is not a line
+// ending. This function is intended to be CRLF/LF agnostic.
+static inline size_t
+match_line_ending_addr(yp_parser_t *parser, const char *cursor) {
+    if (peek_addr(parser, cursor) == '\n') {
+        return 1;
+    }
+    if (peek_addr(parser, cursor) == '\r' && peek_addr(parser, cursor + 1) == '\n') {
+        return 2;
+    }
+    return 0;
+}
+
+// Return the length of the line ending string starting at parser->current.end + offset, or 0 if it
+// is not a line ending. This function is intended to be CRLF/LF agnostic.
+static inline size_t
+match_line_ending_at(yp_parser_t *parser, ptrdiff_t offset) {
+    return match_line_ending_addr(parser, parser->current.end + offset);
+}
+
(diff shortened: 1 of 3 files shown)


src/yarp.c
```diff
@@ -4216,6 +4216,33 @@ next_newline(const char *cursor, ptrdiff_t length) {
...
 
+// Return the length of the line ending string starting at +cursor+, or 0 if it is not a line
+// ending. This function is intended to be CRLF/LF agnostic.
+static inline size_t
+match_line_ending_addr(yp_parser_t *parser, const char *cursor) {
+    if (peek_addr(parser, cursor) == '\n') {
+        return 1;
+    }
+    if (peek_addr(parser, cursor) == '\r' && peek_addr(parser, cursor + 1) == '\n') {
+        return 2;
```
HParker, line 4227: It seems like when we use this always use it to find the location of the `\n`. Do you think we can return that `char *` location instead, or maybe there are cases I missed?

flavorjones: :thinking: Let me give it a tr.

flavorjones: There are a few places where the code branches based on whether we matched `\r` or `\r\n`, and those situations become significantly messier if we return a pointer.

HParker: Makes sense! thanks for looking into it!

HParker (approved):

src/yarp.c
flavorjones: Hmm, maybe you're right, I'll throw that work on a branch so we can compare.

flavorjones:
I've got a couple of small edits to this PR, will push shortly.

src/yarp.c
flavorjones: This ended up requiring code that was (subjectively) harder to reason about, and I actually had trouble debugging it. Offsets seem to fit the code patterns better, so I'm going to stick with that.

flavorjones:
All pushed. @kddnewton Would love your 👀 on this, too.

kddnewton (approved):
This looks great. My only feedback is stylistic (we have been putting `{` on the same line as the cases) but I'm happy to do that in a follow-up PR. Thank you very much for doing this refactor! I think this will be much clearer going forward.

PR #1317, opened 2023-08-23
Fix first method param lex failures

kddnewton:
When Ripper encounters a method parameter that is the first parameter and is an identifier and it shadows a local scope, it incorrectly marks it as END|LABEL (because it think it's a local).

We need to account for that in the lex compat in order to properly compare.


HParker (approved):
Nice! Thanks for the comment on the `on_nl` change as well!

PR #1319, opened 2023-08-23
Fix backslash-r (CR) delimited strings

HParker:
now CR can be used as a string delimiter


test/errors_test.rb
```diff
@@ -164,7 +164,7 @@ def test_unterminated_argument_expression
 
   def test_cr_without_lf_in_percent_expression
     assert_errors expression("%\r"), "%\r", [
-      ["Invalid %% token", 0..3],
+      ["Invalid %% token", 0..2],
```
HParker, line 167: kinda fun we had this test already. Technically this is a very good error test for finding that this string delimiter start was wrong.

kddnewton:
@HParker sorry now that I merged Mike's PR it made yours conflict. Could you rebase?

PR #1321, opened 2023-08-24
A couple of small stylistic changes

kddnewton:
Nothing particularly of substance here, just renaming things and moving some stuff around.

* `le_len` to `eol_length`
* Braces on the same line as switch case
* `peek_addr` -> `peek_at` (Now that we have it looking at specific places, I think it makes more sense for this API to be `at`)
* `peek_at` -> `peek_offset` (This is now more clear (I think) because it tells you it's offset from the current address)
* `match_line_ending_addr` -> `match_eol_at` (Making these match)
* `match_line_ending_at` -> `match_eol_offset`

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -4157,10 +4157,10 @@ yp_do_loop_stack_p(yp_parser_t *parser) {
 /* Lexer check helpers                                                        */
 /******************************************************************************/
 
-// Get the next character in the source starting from +cursor+. If that position is beyond the end
-// of the source then return '\0'.
+// Get the next character in the source starting from +cursor+. If that position
+// is beyond the end of the source then return '\0'.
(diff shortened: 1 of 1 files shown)


src/yarp.c
```diff
@@ -4548,7 +4550,7 @@ static yp_token_type_t
 lex_numeric_prefix(yp_parser_t *parser) {
     yp_token_type_t type = YP_TOKEN_INTEGER;
 
-    if (peek_at(parser, -1) == '0') {
+    if (peek_ahead(parser, -1) == '0') {
```
HParker, line 4553: Is it weird to "peek ahead" backwards?

kddnewton: Great point! I just updated it to `peek_offset`

PR #1322, opened 2023-08-24
Fix lex compat with BOM

kddnewton:
* BOM should not impact looking for the encoding string
* We should re-encode tokens when the encoding changes
* BOM should change the column of comments only

Fixes one more file in the remaining parse errors.


kddnewton:
> Do you have a test with the BOM?

We do now!

HParker (approved):

PR #1324, opened 2023-08-25
Add the ability to serialize shared strings

kddnewton:
Previously, we would copy the entire string every time into the serialized output. Depending on the string, this can get expensive. Now, for the most part strings should end up being about 3 bytes: 1 or 2 for the offset from the start (_maybe_ 3 for large files) and likely a 1 byte length.

### DIFF
--- a/templates/lib/yarp/serialize.rb.erb
+++ b/templates/lib/yarp/serialize.rb.erb
@@ -54,8 +54,8 @@ module YARP
         end
 
         comments = load_varint.times.map { Comment.new(Comment::TYPES.fetch(load_varint), load_location) }
-        errors = load_varint.times.map { ParseError.new(load_string, load_location) }
-        warnings = load_varint.times.map { ParseWarning.new(load_string, load_location) }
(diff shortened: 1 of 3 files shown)


templates/lib/yarp/serialize.rb.erb
```diff
@@ -110,10 +110,21 @@ module YARP
...
+      def load_message
         io.read(load_varint).force_encoding(encoding)
       end
 
+      def load_string
+        case io.getbyte
+        when 1
+          input.byteslice(load_varint, load_varint).force_encoding(encoding)
+        when 2
+          load_message
```
eregon, line 122: `load_message` feels confusing here because this might not be a "message string".
How about `load_owned_string` or `load_unshared_string`?
(same in Loader.java)

eregon: If we want to use other terms, maybe `load_source_string` vs `load_serialized_string`

kddnewton: I renamed it to `load_embedded_string` since that matches more of the Ruby codebase ideas.

eregon (approved):

PR #1325, opened 2023-08-25
Add names to classes and modules

kddnewton:
This should hopefully make it easier to compile these nodes, and also for static analysis tools for having a quick label.

As a bonus, this also fixed a bug we had where we weren't erroring on `class foo; end`.


eregon:
That's probably redundant with https://github.com/ruby/yarp/pull/1304#issuecomment-1690190274, right?
But seems fine until that one is done.

jemmaissroff (approved):
Thank you!

config.yml
```diff
@@ -731,6 +731,8 @@ nodes:
         type: node?
       - name: end_keyword_loc
         type: location
+      - name: name
+        type: string
```
eregon, line 735: It could likely use the constant pool, but it seems constants in general don't use the pool yet.

PR #1326, opened 2023-08-25
fix: unterminated regular expression with a heredoc

flavorjones:
Previously this snippet would track the same newline twice, leading to a failed assertion in yp_newline_list_append:

```
<<-END + /b
END
```

Found via the fuzzer.


kddnewton (approved):
lol wtf

PR #1327, opened 2023-08-25
Move tests from test/* to test/yarp/* to match CRuby

kddnewton:
This should make it easier on the sync to determine what changed and hopefully result in fewer merge conflicts that have to be manually resolved.


k0kubun (approved):
Thanks!

Issue #1236, opened 2023-08-10
CallOperatorWriteNode ergonomics for interpretation


enebo:
This node captures all the info needed to interpret this properly but it is not so simple to use.  If I have:

```ruby
a.b += c
```

We get:

```text
#<YARP::ParseResult:0x00007f7a624986b0
 @comments=[],
 @errors=[],
 @source=#<YARP::Source:0x00007f7a623e92f0 @offsets=[0], @source="a.b += c">,
 @value=
  ProgramNode(0...8)(
    [],
    StatementsNode(0...8)(
      [CallOperatorWriteNode(0...8)(
         CallNode(0...3)(
           CallNode(0...1)(nil, nil, (0...1), nil, nil, nil, nil, 2, "a"),
           (1...2),
           (2...3),
           nil,
           nil,
           nil,
           nil,
           0,
           "b="
         ),
         (4...6),
         CallNode(7...8)(nil, nil, (7...8), nil, nil, nil, nil, 2, "c"),
         :+
       )]
    )
  ),
 @warnings=[]>
```

In Ruby pseudo-ish code we need to do:

```text
%t1 = a
%t2 = %t1.b()
%t3 = %t2.+(c)
           %t1.b=(%t3)
```

We only call ```a``` once.  If we look up at the AST we currently have we need to do three somewhat non-obvious things:

  1. get target call node then get its receiver and call that once.
  2. get target call node but do not call ```b=``` but instead call ```b```.
  3. then when we do call ```b=``` we need to use value from 1.

This is obviously possible but I wonder if we could do it more like JRuby?  In JRuby we record this as (and I suspect this is same layout as MRI):

```text
 jruby -S ast -e 'a.b += 2'
AST:
RootNode line: 0
  OpAsgnNode* line: 0, variable_name: b, variable_assign_name: b=, operatorName: +
    VCallNode:a line: 0
    FixnumNode line: 0, long: 2
```

Is it possible to do something closer to this?  digging around in target feels less clear for interpretation but I also wonder how this looks to syntax tools since "b=" does not exist anywhere in the source.

note: in writing this up I realize how bad those fields names are: variable_name and variable_assign_name so please use better names if we do change this.

eregon:
Agreed. I also discussed this and variants in https://github.com/ruby/yarp/pull/997.
I think it would be a good time to come back to that and make the nodes for ||=, &&= and += easier to use and avoiding the full matrix of combinations.

eregon:
[...]

PR #1329, opened 2023-08-25
Write node updates

kddnewton:
This PR is the resolution of many different conversations over the last several months. For some relevant background, see #997, #1285, and #1286. Unfortunately a lot of context also happened over phone calls and slack DMs, so some of it may not have been properly recorded, but those should suffice for now. It contains 5 commits, but it might actually be easier to review as a whole.

I think it will probably be easiest to understand what this PR is doing by seeing it in a table, so that's what I've laid out below. I will lay out the changes specifically for instance variables, but this applies to: class variables, instance variables, global variables, local variables, constants, and constant paths.

| Changed | Ruby | Old | New |
| --- | --- | --- | --- |
| | `@foo` | (InstanceVariableReadNode) | (InstanceVariableReadNode) |
| | `@foo = 1` | (InstanceVariableWriteNode) | (InstanceVariableWriteNode) |
| ✅ | `@foo += 1` | (OperatorWriteNode (InstanceVariableWriteNode nil) (IntegerNode)) | (InstanceVariableOperatorWriteNode (IntegerNode)) |
| ✅ | `@foo &&= 1` | (AndWriteNode (InstanceVariableWriteNode nil) (IntegerNode)) | (InstanceVariableAndWriteNode (IntegerNode)) |
| ✅ | `@foo \|\|= 1` | (OrWriteNode (InstanceVariableWriteNode nil) (IntegerNode)) | (InstanceVariableOrWriteNode (IntegerNode)) |
| ✅ | `@foo, @bar = 1` | (MultiWriteNode [(InstanceVariableWriteNode nil) (InstanceVariableWriteNode nil)] (IntegerNode)) | (MultiWriteNode [(InstanceVariableTargetNode) (InstanceVariableTargetNode)] (IntegerNode)) |

There were many concerns in the various linked issues that have been raised over this general topic. I'll go through each one of them now and how this PR relates to them.

> I don't want to have to descend into a child node to determine how to process a parent node. It makes it difficult to compile because it requires inner switch/case statements.

This PR removes that issue. Each `*WriteNode` knows exactly what should be compiled based on the type alone.

> We're wasting memory by having parent nodes in these subtrees when we have enough information already.

Because the nodes are now as flat as possible, there is no wasted space. The `*TargetNode` variants of each write ensure we don't have wasted bytes on operators and values that will never be present.

> The node names are too long, and it's a back look for our docs.

The longest node names in the previous iteration of this were `ConstantPathOperatorAndWriteNode`. For all of the `*AndWriteNode` names we've removed `Operator` which should make it more palatable.

> Having a `nil` value in the `*WriteNode` nodes implicitly implies they will be written to in a different context. You have to know this to consistently compile. It's confusing to consumers that want to just handle variable writes.

This has been addressed by introducing the new `*TargetNode` nodes. Now, to find all places where a variable of any kind is written, you should handle the `*TargetNode` and `*WriteNode` nodes. As an added bonus, this makes it much more clear which value is going to be written - somewhat addressing the need for synthetic value nodes. (Note that this doesn't entirely solve this because I haven't addressed the call write node ergonomics detailed in #1236.)

> Now, in order to find all of the places where a variable is written, you have to handle 5 different kinds of writes to that variable type. This is a place where consumers might not handle all of the node types and could lead to bugs in consumer projects.

This has been addressed by providing a `DesugarVisitor`. If you want to handle all of the node types as they were created by YARP, you can. But if you want to handle a subset of the nodes (the subset _without_ the extra write types) you can visit the AST with an instance of the `DesugarVisitor` to get the simplified tree. In this way you will only have to handle a small number of write nodes.

> I want to be able to consistently compile write nodes, and right now I can't.

This is also addressed by this PR. Now write nodes will always have a value, and are not marked as optional. Target nodes will also be able to be compiled consistently.

---

Note that this doesn't handle the call nodes yet and there are some fixes to multi write we can make in follow ups.


tenderlove:
This seems great to me and I think will make implementing the compiler in CRuby much easier.

I have a couple questions though.

> > I don't want to have to descend into a child node to determine how to process a parent node. It makes it difficult to compile because it requires inner switch/case statements.
> 
> This PR removes that issue. Each `*WriteNode` knows exactly what should be compiled based on the type alone.

Is this true for `(InstanceVariableOperatorWriteNode (IntegerNode))`?  I assume this node encompasses all mathematical operators (also bitwise operators) and not just `@foo += 1`.  If that's the case, the compiler will still need to examine the contents of the node in order to generate the correct bytecode.

If we have to do that, what is the practical difference between `InstanceVariableOperatorWriteNode` and `InstanceVariableAndWriteNode`?  Should we not combine `InstanceVariableAndWriteNode` with `InstanceVariableOperatorWriteNode`?  Or separate all operators?

I don't understand why boolean operators are treated differently than mathematical operators.  Or, if we're grouping the nodes, why mathematical operators are grouped together, but booleans are not.

Either way, I think this is a big improvement.

> > Now, in order to find all of the places where a variable is written, you have to handle 5 different kinds of writes to that variable type. This is a place where consumers might not handle all of the node types and could lead to bugs in consumer projects.
> 
> This has been addressed by providing a `DesugarVisitor`. If you want to handle all of the node types as they were created by YARP, you can. But if you want to handle a subset of the nodes (the subset _without_ the extra write types) you can visit the AST with an instance of the `DesugarVisitor` to get the simplified tree. In this way you will only have to handle a small number of write nodes.

If "checking for writes" is a required usecase, maybe we could generate predicate methods for those node types?  If the AST were iterable, then you could do something like `root.find_all(&:variable_writer?)`.  I don't think we need this feature in the CRuby compiler though, so 🤷‍♀️

enebo:
I think this looks great. :+1:

> I don't understand why boolean operators are treated differently than mathematical operators. Or, if we're grouping the nodes, why mathematical operators are grouped together, but booleans are not.

@tenderlove I never thought about this before.  Historically they have been split up to have OR and AND as their own thing but the only advantages of keeping them separate I can think of:

  1. Saves having to record the logical operator in memory (just leverages type value)
  2. It probably makes analysis mildly easier (possibly dubious claim).

fwiw I wouldn't care very much either way.  I do tend to encourage whatever uses less memory but this will end up being the cost of recording a constant pool value.

tenderlove:
> fwiw I wouldn't care very much either way.

Same.  It's really fine to leave as-is, it just seemed like an inconsistency I thought about when reading the initial post.

HParker (approved):

kddnewton:
@tenderlove The reason they're separated is because they're control flow operators, versus the others which are all calls. For example `foo += 1` is `foo = foo +1`. When we're compiling, we should be able to always look at the operator (which is already an interned symbol) and generate the same instructions.

For the `&&=` and `||=` operators, they are actually `foo && foo = 1` and `foo || foo = 1`, which is going to require truthy checks and jumps. These have to be compiled entirely differently then.

tenderlove:
> @tenderlove The reason they're separated is because they're control flow operators, versus the others which are all calls. For example `foo += 1` is `foo = foo +1`. When we're compiling, we should be able to always look at the operator (which is already an interned symbol) and generate the same instructions.
> 
> For the `&&=` and `||=` operators, they are actually `foo && foo = 1` and `foo || foo = 1`, which is going to require truthy checks and jumps. These have to be compiled entirely differently then.

Ya, I understand that the some are for control flow and others are not, but it seems that we're being inconsistent.  Method calls are grouped, but control flow is not?  Why do we group the one but not the other?

kddnewton:
@tenderlove I didn't group them together because they're different kinds of control flow. Sometimes it's just the difference between `branchif` vs `branchunless`, but for globals you have a defined check.

```
$ ruby --dump=insns -e '$foo &&= 1'
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,10)> (catch: false)
0000 getglobal                              :$foo                     (   1)[Li]
0002 dup
0003 branchunless                           10
0005 pop
0006 putobject_INT2FIX_1_
0007 dup
0008 setglobal                              :$foo
0010 leave
```

vs

```
$ ruby --dump=insns -e '$foo ||= 1'
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,10)> (catch: false)
0000 putnil                                                           (   1)[Li]
0001 defined                                global-variable, :$foo, true
0005 branchunless                           13
0007 getglobal                              :$foo
0009 dup
0010 branchif                               17
0012 pop
0013 putobject_INT2FIX_1_
0014 dup
0015 setglobal                              :$foo
0017 leave
```

I figured these were different enough that they merited their own node types.

enebo:
@kddnewton @tenderlove At the end of the day either way you would branch to different logic.  Keeping them separate makes that branch effectively free because it is just part of the type switch.

lib/yarp/desugar_visitor.rb
```diff
@@ -0,0 +1,267 @@
...
+    # Foo = Foo + bar
+    def visit_constant_operator_write_node(node)
+      desugar_operator_write_node(node, ConstantWriteNode, ConstantReadNode)
+    end
+
+    # Foo::Bar &&= baz
+    #
+    # becomes
+    #
+    # Foo::Bar && Foo::Bar = baz
```
eregon, line 83: I think this desugaring is incorrect, because `Foo` should only be executed once, similar to calls.
i.e. for
```ruby
module Foo; end

(p Foo)::Bar ||= 42
```
Foo is printed once and not twice by CRuby.

And I guess `(p Foo)::Bar ||= 42` is a `ConstantPathAndWriteNode` (can't check right now).
But even if it's not it would be observable e.g. if `Foo` triggers `const_missing`.

I think the simplest fix for now is to not desugar the `ConstantPath*WriteNode` since that's invalid semantically, but still desugar the rest where it's valid.

eregon:
I think this is a good compromise, thank you for writing out the reasoning in details (even though I prefer the state before, but I can see that has some disadvantages too, especially for C which doesn't have classes or interfaces to handle this kind of things generically).

I think it would be better if the desugaring happens in C (opt-in of course), that way it's also available for Java and it would likely be a lot more efficient too (can just mutate in place instead of reallocating the entire tree, or maybe even produce different nodes based on that flag) and I think most importantly it means the desugared tree can then be serialized.
The alternative would be to desugar manually in TruffleRuby, that's of course also possible although likely slower when loading from serialized form and that would be desugared already.

eregon:
> Sometimes it's just the difference between `branchif` vs `branchunless`, but for globals you have a defined check.

Could this defined check be part of the desugaring too? Otherwise it's probably invalid in some cases with globals.

kddnewton:
@eregon yeah https://github.com/ruby/yarp/pull/1348

lib/yarp/desugar_visitor.rb
eregon: I'll make a PR for this.

kddnewton: Thanks that would be great

eregon: As expected:
```ruby
module Foo; end
# original
(p Foo)::Bar ||= 42 # prints Foo once
# vs
# desugared
defined?((p Foo)::Bar) ? (p Foo)::Bar : (p Foo)::Bar = 42 # prints Foo twice
```

A more complicated edge case:
```ruby
module FooM; end
class Object
  def self.const_missing(name)
    p name
    name == :Foo ? FooM : super
  end
end

# original
Foo::Bar ||= 42
Foo::Bar ||= 43
p Foo::Bar # => 42

# vs
# desugared
defined?(Foo::Bar) ? Foo::Bar : Foo::Bar = 42
defined?(Foo::Bar) ? Foo::Bar : Foo::Bar = 43
p Foo::Bar # => 43, and warning: already initialized constant FooM::Bar
```
which I think is because in reality it's like:
```ruby
tmp = Foo
defined?(tmp::Bar) ? tmp::Bar : tmp::Bar = value
```
and `defined?(Foo::Bar)` is just `nil` instead, even if `FooM::Bar` is set.

andrykonchin:
It seems to me reasonable (at least for the sake of consistency) to have target node for a method call, e.g. `CallTargetNode`.

It would be helpful to differentiate (without checking the context) a simple method call from a call without/with missing last argument - in the `rescue` operator and multi-assignment. For both 1) attribute assignment (`a.foo, b = []`) and 2) reference assignment `a[:foo], b = []`.

Does it make sense?

PR #1333, opened 2023-08-25
Revert version templating

kddnewton:
This is wreaking havoc on the CRuby setup. Going to revert until I have more time to look at it.


eregon:
What's the issue?

kddnewton:
When we mirror into CRuby, all of the files are in different places, and the order that things are built is dependent on their build system. When I merged this everything broke, and it was really difficult to fix. I think if we want to rebuild this, it'll need to be inside CRuby and then reverse-sync it back into YARP.

eregon:
I can imagine the path from gemspec to lib/yarp/version.rb is different, is there anything else?
Templating must of course happen before any compilation, and then at that point all versions are "burned in" and it shouldn't matter.

One could of course also run templating on sync'ing, but I suppose there are advantages to run templating by CRuby's Makefile (although that causes duplication of the list of templates and is more brittle).

Issue #1134, opened 2023-07-08
Use the constant pool for strings not necessirely as-is in the source (e.g. for CallNode#name)


eregon:
For things like `a.foo(2, 3)` it seems great to use the constant pool.
But I guess it can't for `a.foo = 42` (there is no `foo=` in that source).

If the constant pool wasn't based on offsets in the source it wouldn't be a problem though, so maybe the constant pool should just own the strings it points to, and serialize them as well.
This could also result in a speedup while deserializing, due to all constant pool strings being next to each in memory, vs spread in the source.
And it would help to remove the need to access the source while deserializing (related-ish: #807)

Not a blocker or anything, but it could be a nice optimization for deserialization speed, at the cost of some serialized size though (but removing the need to keep the source as a byte[]/char* around).

enebo:
@eregon I think so but then I think a requirement for us should be that we should not need to hold onto original Ruby source to process serialized code (it is a big memory requirement right now).  That would mean names could be in the serialized node but then it makes more sense to just keep them in a constant pool since many names are repeated in a source file.

eregon:
One thought about this: currently after Loader is done, the constant pool is fully resolved, so we don't need to keep the source around after Loader is done. So the constant pool could be a hybrid of owned strings and (start, length) in source.

kddnewton:
We support this now. We don't support it for `CallNode#name`, but I'm going to close this in favor of #1296, which indicates that.

eregon:
OK, I moved that part of #1296 to #1530

PR #1338, opened 2023-08-27
Rename constant pool fields to name or operator

eregon:
* `constant_id` and `operator_id` are confusing.
* Solves points 3 and 4 of https://github.com/ruby/yarp/issues/1296

### DIFF
--- a/lib/yarp.rb
+++ b/lib/yarp.rb
@@ -462,10 +462,10 @@ module YARP
           # order here so that we can compare properly.
           if params
             sorted = [
-              *params.requireds.grep(RequiredParameterNode).map(&:constant_id),
-              *params.optionals.map(&:constant_id),
+              *params.requireds.grep(RequiredParameterNode).map(&:name),
+              *params.optionals.map(&:name),
               *((params.rest.name ? params.rest.name.to_sym : :*) if params.rest && params.rest.operator != ","),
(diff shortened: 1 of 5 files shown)


lib/yarp.rb
```diff
@@ -462,10 +462,10 @@ def self.yarp_locals(source)
...
             sorted = [
-              *params.requireds.grep(RequiredParameterNode).map(&:constant_id),
-              *params.optionals.map(&:constant_id),
+              *params.requireds.grep(RequiredParameterNode).map(&:name),
+              *params.optionals.map(&:name),
               *((params.rest.name ? params.rest.name.to_sym : :*) if params.rest && params.rest.operator != ","),
-              *params.posts.grep(RequiredParameterNode).map(&:constant_id),
+              *params.posts.grep(RequiredParameterNode).map(&:name),
               *params.keywords.reject(&:value).map { |param| param.name.chomp(":").to_sym },
               *params.keywords.select(&:value).map { |param| param.name.chomp(":").to_sym }
```
eregon, line 470: It'd be consistent if `KeywordParameterNode#name` returns the Symbol without this extra handling, but for that it should probably be added to the constant pool first (and the constant pool needs to support owned strings first, #1134).

kddnewton (approved):
Looks great

PR #1328, opened 2023-08-25
fix: regexes and strings with escaped newline around a heredoc

flavorjones:
For example:

```ruby
<<-A.gsub(/b\
a
A
c/, "")
```

should be parsed as a heredoc containing `"x\n"`, and a regex that is comprised of two string tokens, `"a"` and `"b"`:

```
CallNode(168...193)(
  InterpolatedStringNode(168...172)(
    (168...172),
    [StringNode(182...184)(nil, (182...184), nil, "a\n")],
    (184...186)
  ),
  (172...173),
  (173...177),
  (177...178),
  ArgumentsNode(178...192)(
    [InterpolatedRegularExpressionNode(178...188)(
       (178...179),
       [StringNode(179...182)(nil, (179...182), nil, "b"),
        StringNode(186...187)(nil, (186...187), nil, "c")],
       (187...188),
       0
     ),
     StringNode(190...192)(
       (190...191),
       (191...191),
       (191...192),
       ""
     )]
  ),
  (192...193),
  nil,
  0,
  "gsub"
)]
```

Similarly for the string case:

```ruby
<<-A + "e\
d
A
f"
```

should be parsed as:

```
CallNode(281...298)(
  InterpolatedStringNode(281...285)(
    (281...285),
    [StringNode(292...294)(nil, (292...294), nil, "d\n")],
    (294...296)
  ),
  nil,
  (286...287),
  nil,
  ArgumentsNode(288...298)(
    [InterpolatedStringNode(288...298)(
       (288...289),
       [StringNode(289...292)(nil, (289...292), nil, "e"),
        StringNode(296...297)(nil, (296...297), nil, "f")],
       (297...298)
     )]
  ),
  nil,
  nil,
  0,
  "+"
)]
```

Note that Ripper tokenizes incorrectly, combining the two string parts into a single token, so we skip the ripper tests for this new test file.

Previously these cases were failing the `yp_newline_list_append` assertion. The regexp case was not properly flushing the heredoc, lexing a string token, or jumping to `next_start`. The string case was not properly lexing a string token or flushing the heredoc.

Found via the fuzzer.

PR #1339, opened 2023-08-27
fix %i %I %w %W when spanning a heredoc

flavorjones:
## Problem being fixed

This PR addresses situations where %i, %I, %w, and %W lists span a heredoc by using an escaped newline (similar to the regex scenario from #1328), for example:

```ruby
<<-A, %w[b\
a
A
c]
```

Previously these scenarios all failed an assertion, either during newline tracking or during parsing.

## Reviewers should focus on ...

Please take a look at the %W and %I fix commits, I'm not sure whether that's the best approach.

## Ripper compatibility

Note that Ripper cannot successfully parse the %w, %W, or %i scenarios. I'll file a bug report upstream for that tomorrow.

Note that Ripper also mishandles some of these scenarios (see comments in "test/yarp/fixtures/spanning_heredoc.txt").

## Ruby compatibility

Please note that the behavior of %W and %I in this PR doesn't perfectly match Ruby. I can try to fix this in a future PR, but in the meantime didn't want to block this work on it, since it's an improvement over current behavior.

Ruby will run this script:

    pp <<-A, %W[l\
    k
    A
    l]

and output

    ["l\n" + "l"]

Note the newline; this string is "l\nl". YARP tokenizes correctly:

    [PERCENT_UPPER_W(697...700)("%W["), 1025],
    [STRING_CONTENT(700...703)("l\\\n"), 1025],
    [STRING_CONTENT(707...708)("l"), 1025],
    [STRING_END(708...709)("]"), 2],

however YARP's AST is missing this newline:

    ArrayNode(697...709)(
      [InterpolatedStringNode(700...708)(
         nil,
         [StringNode(700...703)(nil, (700...703), nil, "l"), StringNode(707...708)(nil, (707...708), nil, "l")],
         nil
       )],
      (697...700),
      (708...709)
    )


kddnewton:
This looks absolutely correct, and pretty much exactly how I would have tried to solve this, but better. :)

kddnewton (approved):

PR #1347, opened 2023-08-28
[rust] add `ConstantId` wrapper for `yp_constant_id_t`

froydnj:
The ultimate goal of this PR is to add something like the proposed `ConstantId::as_slice`.

The design proposed here threads `yp_parser_t` into the various Rust structures.  This is necessary because `ConstantId` needs access to the associated parser it was created from so it can index into the parser's constant pool.

I can imagine an alternate world where one provides `yarp::Parser` (or similar) to `ConstantId::as_slice` instead, and any other place that requires some global state from `yp_parser_t`.  (If you squint, this is the same sort of idea that `ParseResult::as_slice` implements, except the latter function does it backwards and is probably redundant with `Location::as_slice`.)  I think the design adopted in this PR is the better one, because it gives stronger guarantees about tying the parser and the constant ID together, ultimately making things easier to use and requiring less checking in the implementation.

I think it's possible that the `PhantomData` on `ConstantId` is acceptable to the compiler, but ultimately somewhat superfluous, and would welcome better ideas from somebody who knows more about Rust FFI than I do.

### DIFF
--- a/rust/yarp/build.rs
+++ b/rust/yarp/build.rs
@@ -556,6 +556,16 @@ impl<'pr> ConstantId<'pr> {{
     fn new(parser: NonNull<yp_parser_t>, id: yp_constant_id_t) -> Self {{
         ConstantId {{ parser, id, marker: PhantomData }}
     }}
+
+    /// Returns a byte slice for the constant ID.
+    #[must_use]
+    pub fn as_slice(&self) -> &'pr [u8] {{
+        unsafe {{
+            let pool = &(*self.parser.as_ptr()).constant_pool;
+            let constant = &(*pool.constants.offset(isize::try_from(self.id).expect("id should be in range")));
+            std::slice::from_raw_parts(constant.start.cast::<u8>(), constant.length)
+        }}
+    }}
 }}
 
 impl std::fmt::Debug for ConstantId<'_> {{
--- a/rust/yarp/build.rs
+++ b/rust/yarp/build.rs
@@ -547,13 +547,14 @@ impl std::fmt::Debug for NodeList<'_> {{
 
 /// A handle for a constant ID.
 pub struct ConstantId<'pr> {{
+    parser: NonNull<yp_parser_t>,
     id: yp_constant_id_t,
     marker: PhantomData<&'pr mut yp_constant_id_t>
 }}
 
 impl<'pr> ConstantId<'pr> {{
-    fn new(id: yp_constant_id_t) -> Self {{
-        ConstantId {{ id: id, marker: PhantomData }}
+    fn new(parser: NonNull<yp_parser_t>, id: yp_constant_id_t) -> Self {{
+        ConstantId {{ parser, id, marker: PhantomData }}
     }}
 }}
 
@@ -565,6 +566,7 @@ impl std::fmt::Debug for ConstantId<'_> {{
 
 /// An iterator over the constants in a list.
 pub struct ConstantListIter<'pr> {{
+    parser: NonNull<yp_parser_t>,
     pointer: NonNull<yp_constant_id_list_t>,
     index: usize,
     marker: PhantomData<&'pr mut yp_constant_id_list_t>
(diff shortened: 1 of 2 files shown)


froydnj:
(I can also understand the concern about making these nodes bigger and probably slower than the corresponding code in C would be, and could deal with passing a `&ParseResult` or `&Parser` to the relevant functions instead, even if -- I think -- that would be potentially unsafe, because you could mix `ConstantId`s with a different `ParseResult`.  I guess you could bounds check the constant pool reference and return `Option<&[u8]>`, but that forces a very verbose API on clients.)

kddnewton:
Hey @froydnj! I think this makes a lot of sense, but I do understand the tradeoff here of making them bigger by including this pointer. @lopopolo what do you think?

rust/yarp/build.rs
```diff
@@ -539,29 +545,62 @@ impl std::fmt::Debug for NodeList<'_> {{
...
+            let pool = &(*self.parser.as_ptr()).constant_pool;
+            let constant = &(*pool.constants.offset(isize::try_from(self.id).expect("id should be in range")));
+            std::slice::from_raw_parts(constant.start.cast::<u8>(), constant.length)
+        }}
+    }}
+}}
+
+impl std::fmt::Debug for ConstantId<'_> {{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {{
+        write!(f, "{{:?}}", self.id)
```
lopopolo, line 573: This should use "debug struct" and similar helpers on fmt::Formatter

lopopolo (commented):
This seems fine to me but I worry that the Rust crate is going through contortions because it is treating the parser struct as movable when the C API does not.

From what I gathered of the C examples, the parser is stack allocated, fed some bytes, manipulated to extract everything the calling routine requires, and discarded before returning.

I wonder if a closure/interior iteration style API would be more appropriate such that the parser doesn't need to move, be heap allocated, or threaded everywhere.

kddnewton:
> This seems fine to me but I worry that the Rust crate is going through contortions because it is treating the parser struct as movable when the C API does not.
> 
> From what I gathered of the C examples, the parser is stack allocated, fed some bytes, manipulated to extract everything the calling routine requires, and discarded before returning.
> 
> I wonder if a closure/interior iteration style API would be more appropriate such that the parser doesn't need to move, be heap allocated, or threaded everywhere.

Do you mean you would pass a closure to `parse` and then extract the information you needed within that closure?

lopopolo:
Yes. The closure would mutable borrow some Rust structs and stuff information into them, basically.

I'm basing this on my expectations that I'll want an "artichoke native" AST so I don't have to carry around a parser for code that's loaded into the VM.

kddnewton:
Right. The Java ASTs that are built from C YARP load the constants as soon as the nodes are created. Maybe we should do the same here instead of lazily loading them?

lopopolo:
You could facilitate this by having the closure accept some form of `Pin<&mut Parser>` and threading that into accessor APIs.

This is pretty similar to e.g. how mruby's C APIs work. The interpreter state is passed around (in contrast to MRI where it is a global).

Using a `Pin` means I think you need to be less concerned about mixing a particular parser with different accessor structs since you can disambiguate with invariant lifetime generics.

kddnewton:
I think that would be really valuable, but also clearly much larger than this PR itself. I think I'll merge this and then hopefully we can take a look at that as a wholesale change.

kddnewton (approved):

lopopolo:
> Right. The Java ASTs that are built from C YARP load the constants as soon as the nodes are created. Maybe we should do the same here instead of lazily loading them?

Yea these sorts of API questions are I think why @turboladen suggested starting with `yarp-sys` and letting things bake before iterating on `yarp` crate. We seem to be pushing forward here without the API design and stakeholder consideration y'all did with the parser itself.

kddnewton:
Ahh I'm sorry. I was too eager to get something out that we could iterate on.

PR #1348, opened 2023-08-29
Desugar ||= more accurately

kddnewton:
Class variables, global variables, constants, and constant paths should actually desugar to `defined?` instead of just reading the value.

### DIFF
--- a/lib/yarp/desugar_visitor.rb
+++ b/lib/yarp/desugar_visitor.rb
@@ -8,26 +8,16 @@ module YARP
     #
     # @@foo && @@foo = bar
     def visit_class_variable_and_write_node(node)
-      AndNode.new(
-        ClassVariableReadNode.new(node.name_loc),
-        ClassVariableWriteNode.new(node.name_loc, node.value, node.operator_loc, node.location),
-        node.operator_loc,
-        node.location
-      )
+      desugar_and_write_node(node, ClassVariableReadNode, ClassVariableWriteNode)
     end
 
     # @@foo ||= bar
     #
     # becomes
     #
-    # @@foo || @@foo = bar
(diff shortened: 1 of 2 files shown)


jemmaissroff (approved):

lib/yarp/desugar_visitor.rb
```diff
@@ -263,5 +233,35 @@ def desugar_operator_write_node(node, write_class, read_class, arguments: [])
...
+    def desugar_or_write_node(node, read_class, write_class, arguments: [])
+      OrNode.new(
+        read_class.new(*arguments, node.name_loc),
+        write_class.new(*arguments, node.name_loc, node.value, node.operator_loc, node.location),
+        node.operator_loc,
+        node.location
+      )
+    end
+
+    # Don't desugar `x ||= y` to `defined?(x) ? x : x = y`
```
eregon, line 247: `Don't desugar` -> `Desugar`, right?

lib/yarp/desugar_visitor.rb
```diff
@@ -149,12 +134,7 @@ def visit_global_variable_and_write_node(node)
     #
     # $foo || $foo = bar
```
eregon, line 135: `defined?($foo) ? $foo : $foo = bar`

lib/yarp/desugar_visitor.rb
```diff
@@ -59,12 +44,7 @@ def visit_constant_and_write_node(node)
     #
     # Foo || Foo = bar
```
eregon, line 45: `defined?(Foo) ? Foo : Foo = bar`

eregon (commented):
Thanks!

eregon:
^ I'll do these since I'm touching that file anyway.

PR #1352, opened 2023-08-29
Switch from const char * to const uint8_t *

kddnewton:
`char` signedness is platform-dependent. As such there is sometimes some weirdness when we need to compare against non-ascii values. Furthermore you can't use `char` to index into lookup tables, so we end up always casting it to an `unsigned char`.

In general, since we're dealing with unknown data (the source), we should treat it all as just raw bytes. (Side note: this is already what the Java code does.)

This commit translates all of our `char` handling to `uint8_t`. You can see that there are many fewer casts now, since everything is always unsigned.

Believe it or not, this means we can't run on as many platforms as we previously could - `uint8_t` isn't guaranteed to be defined on platforms where 1 byte != 8 bits, but it looks like Ruby isn't support supported on such platforms any (ruby/ruby uses `uint8_t`) so we should be okay.

---

This commit is one I've wanted to make for a while, but kept getting delayed. Unfortunately it's very large now since the project is so big. Hopefully once we get this merged it will be minimally painful to maintain.

### DIFF
--- a/docs/encoding.md
+++ b/docs/encoding.md
@@ -61,22 +61,22 @@ typedef struct {
     // Return the number of bytes that the next character takes if it is valid
     // in the encoding. Does not read more than n bytes. It is assumed that n is
     // at least 1.
-    size_t (*char_width)(const char *c, ptrdiff_t n);
+    size_t (*char_width)(const uint8_t *b, ptrdiff_t n);
 
     // Return the number of bytes that the next character takes if it is valid
     // in the encoding and is alphabetical. Does not read more than n bytes. It
     // is assumed that n is at least 1.
-    size_t (*alpha_char)(const char *c, ptrdiff_t n);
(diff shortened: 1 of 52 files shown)


jemmaissroff:
Can we (/ did you) test this on ruby/ruby CI before merging it in here? Contents of it look straightforward enough to me, but I'm a little concerned about unknown unknowns when syncing this to Ruby. Probably worth creating a branch and running ruby/ruby CI to ascertain that's all good rather than deal with sync failures / breaking CI over there.

kddnewton:
@jemmaissroff good call. I just opened https://github.com/ruby/ruby/pull/8324.

eregon (approved):
Makes sense. All green (yarp and ruby CIs)

jemmaissroff (approved):

PR #1354, opened 2023-08-29
Add location tests

andyw8:
There has been some discussion of how columns behave slightly different to how it does in [Syntax Tree](https://github.com/ruby-syntax-tree/syntax_tree). Before proposing to change this, I want to document the current state.

(cc @vinistock)

### DIFF
--- a/test/yarp/location_test.rb
+++ b/test/yarp/location_test.rb
@@ -801,6 +801,17 @@ module YARP
       assert_equal expected, actual
     end
 
+    def test_attributes
+      result = YARP.parse("class Foo\nend")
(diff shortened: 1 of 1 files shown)


vinistock:
Have you confirmed that the only thing not matching is the `end_column`? I suppose the `end_offset` would've been caught already since it's likely used by Ruby itself.

kddnewton:
@andyw8 can you add this to `location_test.rb` instead?

andyw8:
> Have you confirmed that the only thing not matching is the end_column?

I'll confirm as part of the proposed change.

kddnewton:
Superseded by #1415

PR #1361, opened 2023-08-30
[rust] add `Location::join`

froydnj:
This function helps eliminate the need for exposing raw pointers in the public API, and is just plain easier to read than dealing with a bunch of offset arithmetic.

### DIFF
--- a/rust/yarp/build.rs
+++ b/rust/yarp/build.rs
@@ -419,6 +419,17 @@ impl<'pr> Location<'pr> {{
     pub const fn new(loc: &'pr yp_location_t) -> Location<'pr> {{
         Location {{ start: loc.start, end: loc.end, marker: PhantomData }}
     }}
+
+    /// Return a Location starting at self and ending at the end of other.
+    /// Returns None if self starts after other.
+    #[must_use]
+    pub fn join(&self, other: &Location<'pr>) -> Option<Location<'pr>> {{
+        if self.start > other.start {{
+            None
+        }} else {{
+            Some(Location {{ start: self.start, end: other.end, marker: PhantomData }})
+        }}
+    }}
 }}
 
 impl std::fmt::Debug for Location<'_> {{
--- a/rust/yarp/build.rs
+++ b/rust/yarp/build.rs
@@ -169,7 +169,7 @@ fn write_node(file: &mut File, node: &Node) -> Result<(), Box<dyn std::error::Er
     writeln!(file, "    #[must_use]")?;
     writeln!(file, "    pub fn location(&self) -> Location<'pr> {{")?;
     writeln!(file, "        let pointer: *mut yp_location_t = unsafe {{ &mut (*self.pointer).base.location }};")?;
-    writeln!(file, "        Location::new(unsafe {{ &(*pointer) }})")?;
+    writeln!(file, "        Location::new(self.parser, unsafe {{ &(*pointer) }})")?;
     writeln!(file, "    }}")?;
 
     for field in &node.fields {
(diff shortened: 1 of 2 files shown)


kddnewton:
I understand the impetus behind this change, but I don't think we want every location struct in the project to get wider on account of the new parser struct pointer. I can see how it makes it much easier to work with though. Is there a way we could do this without attaching the parser pointer?

froydnj:
Given that `Location` objects are created on-demand and they are generally expected to be transient objects, I think an increase from 16 to 24 bytes matters less than if `yp_location_t` was looking at the same increase.  I believe the same sort of argument applies to passing the parser around on `Node` and company in #1347, and I think if we've accepted the hit on mostly-transient `Node` objects, accepting it on `Location` objects is not significantly different.  

(Aside: if we're going to talk about sizes, can we talk about making `yp_location_t` objects not huge?  😅)

One not-so-great option would be dropping the parser provenance commit entirely, but I think the prudent thing to do in that case would be to make `join` `unsafe`, which is not something I'm excited about.  Dropping parser provenance -- and assuming that `start()` and `end()` also wind up being deleted, which I think is the right end state -- also means you can't provide any sort of offset API, which turns out to be pretty important.  (The specific thing I care about for the experiment I'm doing is finding out where an expression "really" ends, e.g. with whitequark's parser, something like:

```ruby
heredoc_end = expr.each_descendant(:dstr, :str).select(&:heredoc?).map { |n| n.loc.heredoc_end }.max_by(&:end_pos)
if heredoc_end && heredoc_end.end_pos > expr.loc.expression.end_pos
  ...
end
```

but I'm sure there are other examples)

Another not-so-great option would be to change the Rust `Location` to:

```rust
struct Location<'pr> {
    source: *const c_char, // or the parser, either way
    start: u32,
    end: u32,
    marker: PhantomData<&'pr [c_char]>
}
```

and make constructing them fallible, which would imply that things would blow up on > 4GB files.

I can't think of other options -- at least ones that don't involve restructuring `yp_location_t` in some way, which would be a rather large change.

froydnj:
After thinking for a little bit, a third alternative is to adopt the design very briefly sketched in #1347, where we would explicitly require a `ParseResult` (better would be a `Parser` struct and even better would be a `ParserContext` that just had information required to understand the AST, but not data only required during parsing) in `Location::join`, and then we could bounds-check the `Location` pointers against the `ParseResult`'s source.

I don't really like the asymmetry between requiring extra data in `Location::join` and (currently) not requiring it in `ConstantId::as_slice`.  I think whichever way we pick should be consistent between the two.

Also, I have been assuming the transience of `Location` and `Node` objects due to just using them during a single walk over the tree that doesn't really store `Location` or `Node` objects over the long term, but perhaps your concerns about the space required come from a different use-case?

froydnj:
Ping on reviewing this patch.  I think the open options are:

1. Keep `Location` the same size by not adding a `parser` field.  This option would changing to `join(&self, result: &ParseResult, other: &Location) -> Option<Location<'pr>>`,  so we can verify that the `Location`s being joined come from the same parser.  A hypothetical `{start,end}_offset` API, which is cleaner than the raw pointers currently exposed, would also require a `ParseResult` argument, which would make them much less ergonomic.
2. Change `Location` to store `{start,end}_offset` instead of pointers.  This option would not change the size of `Location`, but would still require the altered signature for `join` mentioned above.  We would also need to change `as_slice` to take a `ParseResult`, as `Location` no longer has direct access to the source text, and `as_slice` would need to return `Option<&'pr [u8]>`.
3. Do nothing and merge this PR as-is.  I won't rehash the arguments for or against the size of `Location` objects.

If option 1 or 2 are chosen, then for consistency's sake, `ConstantId::as_slice` probably should change to take a `ParseResult`.

WDYT @kddnewton ?

kddnewton:
Sorry I haven't gotten a chance to look at this this week, but I will first thing next week!

froydnj:
Ping on reviewing this patch.

kddnewton:
I'm good with option 3. I don't know why in my head I thought the locations were created eagerly.

kddnewton (approved):

PR #1363, opened 2023-08-30
[rust] fix accessing optional locations

froydnj:
Per the use of `OptionalLocation` pieces from `config.yml` elsewhere, e.g.:

https://github.com/ruby/yarp/blob/4d2f7afe1fd880f98b68d851eb9049af64b35fc5/templates/ext/yarp/api_node.c.erb#L170-L171

we should be looking at the `start` of the `yp_location_t` we're accessing to determine if the location is present, rather than the address of the `yp_location_t` itself (which will never be null).

### DIFF
--- a/rust/yarp/build.rs
+++ b/rust/yarp/build.rs
@@ -246,7 +246,8 @@ fn write_node(file: &mut File, node: &Node) -> Result<(), Box<dyn std::error::Er
             NodeFieldType::OptionalLocation => {
                 writeln!(file, "    pub fn {}(&self) -> Option<Location<'pr>> {{", field.name)?;
                 writeln!(file, "        let pointer: *mut yp_location_t = unsafe {{ &mut (*self.pointer).{} }};", field.name)?;
-                writeln!(file, "        if pointer.is_null() {{")?;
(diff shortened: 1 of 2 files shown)


kddnewton (approved):

rust/yarp/build.rs
```diff
@@ -243,7 +243,8 @@ fn write_node(file: &mut File, node: &Node) -> Result<(), Box<dyn std::error::Er
             NodeParamType::OptionalLocation => {
                 writeln!(file, "    pub fn {}(&self) -> Option<Location<'pr>> {{", node_param.name)?;
                 writeln!(file, "        let pointer: *mut yp_location_t = unsafe {{ &mut (*self.pointer).{} }};", node_param.name)?;
-                writeln!(file, "        if pointer.is_null() {{")?;
+                writeln!(file, "        let start: *const u8 = unsafe {{ (*pointer).start }};")?;
```
kddnewton, line 246: Can you get rid of the extra type here? In case it changes in the future, I don't think we need to be quite this explicit

kddnewton (changes requested):

rust/yarp/build.rs
froydnj: Sure, done!

PR #1364, opened 2023-08-30
Fixing start/end location confusion found by the fuzzer

flavorjones:
The fuzzed snippets added in this PR are not very obvious, but they result in nodes with start locations after their end locations. This causes assertions to fail during serialization.

Primarily this is happening because locations during error recovery are "best efforts" and nodes may appear out of order in a list; but there is also an implicit assumption in nodes _are_ in location order.

To address this, this PR introduces checks on both start and end locations when a node is being appended or created.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -3908,12 +3908,14 @@ yp_statements_node_location_set(yp_statements_node_t *node, const uint8_t *start
 // Append a new node to the given StatementsNode node's body.
 static void
 yp_statements_node_body_append(yp_statements_node_t *node, yp_node_t *statement) {
-    if (yp_statements_node_body_length(node) == 0) {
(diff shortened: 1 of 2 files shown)


src/yarp.c
```diff
@@ -1,4 +1,5 @@
 #include "yarp.h"
+#include "sys/param.h"
```
kddnewton, line 2: I think this is a POSIX header. I think for now I would just inline those functions.

kddnewton (changes requested):

src/yarp.c
flavorjones: oof.

flavorjones: Done.

kddnewton (approved):

PR #1365, opened 2023-08-30
Fix instance variable constants, and class variable constants

kddnewton:


jemmaissroff (approved):
Thanks for doing this one!

Issue #1268, opened 2023-08-16
Overloading `child_nodes` in `config.yml`


jemmaissroff:
I’m looking at adding a flag to nodes, and realizing how much we’re overloading the term child_nodes in config.yml . We’re using child_nodes currently to mean params. A small, simple improvement would be renaming this. But overloading the term still leads to quite a bit of filtering and casing when actually generating the code. 

I think it would be cleaner to separate out actual child nodes from other fields on a node (eg have fields on a node for child_nodes, locations, flags) to match how the templates are being used. We could define a params method which returns all of these when needed, but in most cases just use the filtered down nodes.

For example, something like this

```
  - name: WhileNode
    child_nodes:
      - name: keyword_loc
        type: location
      - name: predicate
        type: node
      - name: statements
        type: node?
        kind: StatementsNode
      - name: flags
        type: flags
        kind: LoopFlags
```

could become:

```
  - name: WhileNode
    locations:
      - name: keyword_loc
        type: location
    child_nodes:
      - name: predicate
        type: node
      - name: statements
        type: node?
        kind: StatementsNode
    flags:
      - name: flags
        type: flags
        kind: LoopFlags
```    

depending on how we’d implement it, we could even get rid of the type on some of the less complicated fields.

eregon:
What about `string` and `integer` fields?

Also if it's separated by category it wouldn't be possible to choose the order of fields belonging to different categories.
For instance the current ordering for CallNode seems nice as it's in order of things appearing in the source:
https://github.com/ruby/yarp/blob/8a0a7a1d28b82df26a614378b2007076033ff5d0/config.yml#L585-L607

I think renaming `child_nodes` to `fields` would be an easy fix here with most of the advantages.

Regarding iterating child params, there could be a convenience method for that, that's indeed done a lot manually right now.

enebo:
@jemmaissroff in JRuby for the current AST we differentiate between child nodes and fields in that we have a childNodes() which only returns the child nodes of the current node.  Non-nodes are fields (and actually we also save the child nodes as fields) but childNodes() is necessary for implementing a proper visitor.

Ultimately though the only need to know child nodes from other fields (or types) is mostly to give the list of children to a visitor right?  That must be done by looking for a type of node{,?,[]}.   If we guaranteed child_nodes was only nodes I think that could allow type to be removed so long as cardinality specifier was still there somewhere.  So:

```text
- name: PostExecutionNode
    child_nodes:
      - name: statements
        type: node?
[...]

PR #1366, opened 2023-08-30
Rename `child_nodes` in `config.yml` to `fields`

kddnewton:
Partially addresses #1268.

This renames the config field and also nests everything under the `YARP` module (since this file gets included into the rake file, which means it gets loaded anytime the rake file gets loaded, which means it's possible for all of ruby's build system to get access to these constants).

It doesn't split up the field types, but it's a nice start to get away from `child_nodes` which was a confusing phraseology.


templates/java/org/yarp/Loader.java.erb
```diff
@@ -257,19 +257,19 @@ public class Loader {
             case <%= index + 1 %>:
             <%-
             params = node.needs_serialized_length? ? ["buffer.getInt()"] : []
```
jemmaissroff, line 259: `params` --> `fields`

jemmaissroff (changes requested):

templates/java/org/yarp/Loader.java.erb
kddnewton: For this one I left it in because it's being used as the params to the function being described.

jemmaissroff (approved):

PR #1369, opened 2023-08-30
Desugaring fixes

eregon:
Some nodes were incorrectly desugared, with some location field and the value nodes being swapped.
This is fixed by making the ordering consistent in this PR.
If that's not desired, an alternative would be to provide lambdas to helpers like `desugar_or_write_node` to get the correct order (that can also be done in addition to making the order consistent).


test/yarp/desugar_visitor_test.rb
```diff
@@ -54,6 +54,29 @@ def ast_inspect(node)
     def assert_desugars(expected, source)
       ast = YARP.parse(source).value.accept(DesugarVisitor.new)
       assert_equal expected, ast_inspect(ast.statements.body.last)
+
+      # Ensure every node is only present once in the AST.
+      # If the same node is present twice it would most likely indicate it is executed twice, which is invalid semantically.
+      # This also acts as a sanity check that Node#child_nodes returns only nodes or nil (which caught a couple bugs).
+      all_nodes = {}.compare_by_identity
+      visitor = Class.new(Visitor) do
```
kddnewton, line 62: Can you extract this out to a class in the test file? It's difficult to read inlined like this.

kddnewton (changes requested):

PR #1371, opened 2023-08-30
fix: octal, hex, and unicode strings at the end of a file

flavorjones:
These were causing invalid memory access.

Also:
- clean up handling of EOF after `"\\"` by moving the check from `parser_lex` into `yp_unescape_calculate_difference`
- report a syntax error for unterminated bracket-style unicode strings, like `\u{1234`

### DIFF
--- a/src/unescape.c
+++ b/src/unescape.c
@@ -69,17 +69,15 @@ char_is_ascii_printable(const uint8_t b) {
 // Scan the 1-3 digits of octal into the value. Returns the number of digits
 // scanned.
 static inline size_t
-unescape_octal(const uint8_t *backslash, uint8_t *value) {
+unescape_octal(const uint8_t *backslash, uint8_t *value, const uint8_t *end) {
     *value = (uint8_t) (backslash[1] - '0');
-    if (!yp_char_is_octal_digit(backslash[2])) {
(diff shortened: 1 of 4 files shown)


src/unescape.c
```diff
@@ -236,13 +237,14 @@ unescape(yp_parser_t *parser, uint8_t *dest, size_t *dest_length, const uint8_t
 
                 unicode_cursor += yp_strspn_whitespace(unicode_cursor, end - unicode_cursor);
 
-                while ((*unicode_cursor != '}') && (unicode_cursor < end)) {
+                while ((unicode_cursor < end) && (*unicode_cursor != '}')) {
```
jemmaissroff, line 240: Why this change?

jemmaissroff (commented):
Looks good, just a small question

src/unescape.c
flavorjones: We need to check if `unicode_cursor` is past `end` _before_ dereferencing it, otherwise we could be illegally accessing memory past the end of the buffer.

flavorjones: Hmm, looking at it more closely and this whole loop construct is more complicated than we need. Let me take another pass at cleaning it up.

jemmaissroff (approved):
LGTM, thanks!

PR #1377, opened 2023-09-01
fix: report syntax error for invalid hex escape, and avoid double-reporting

### LINKED ISSUE #1367: invalid hex escapes should result in a syntax error
Here's the failing test:

```ruby
  def test_invalid_hex_escape
    assert_errors expression('"\\xx"'), '"\\xx"', [
      ["Invalid hex escape.", 1..3],
    ]
  end
```

Currently this does not indicate a syntax error, and the tree is constructed using the value returned from `unescape_hexadecimal_digit` when passed `'x'`: `"\t"`

```
#<YARP::ParseResult:0x00007fa2fead0e28
 @comments=[],
 @errors=[],
 @source=#<YARP::Source:0x00007fa2fea99888 @offsets=[0], @source="\"\\xx\"">,
 @value=ProgramNode(0...5)([], StatementsNode(0...5)([StringNode(0...5)((0...1), (1...4), (4...5), "\t")])),
 @warnings=[]>
```

Ruby's behavior:

```
$ ruby -e '"\xx"'
-e:1: invalid hex escape
"\xx"
```

I'm not sure what the right solution should be to propagate errors back into parser_lex for this, so I'm punting and writing this bug report.

flavorjones:
Closes #1367

## Fix 1: report syntax error for invalid hex escape

This change passes in an error list to `escape.c:unescape_hexadecimal` and reports a syntax error if the first character isn't hexadecimal, e,g:

```ruby
  def test_invalid_hex_escape
    assert_errors expression('"\\xx"'), '"\\xx"', [
      ["Invalid hex escape.", 1..3],
    ]
  end
```

## Fix 2: avoid double-reporting of syntax errors during string unescaping

Essentially, this change updates `yp_unescape_calculate_difference` to not create syntax errors, and we rely entirely on `yp_unescape_manipulate_string` to report syntax errors.

To do that, this PR adds another (!) parameter to `unescape`: `yp_list_t *error_list`. When present, `unescape` reports syntax errors (and otherwise does not).

However, an edge case that needed to be addressed is reporting syntax errors in this case:

```ruby
?\u{1234 2345}
```

In a string context, it's possible to have multiple codepoints by doing something like `"\u{1234 2345}"`; however, in the character literal context, this is a syntax error -- only a single codepoint is allowed.

Unfortunately, when `yp_unescape_manipulate_string` is called, there's nothing to indicate that we are in a "character literal" context and that only a single codepoint is valid. `yp_unescape_calculate_difference` has this context, but no longer reports errors; `yp_unescape_manipulate_string` doesn't have this context and so no syntax error is reported.

To make this work, this PR:

- introduces a new static utility function, `yarp.c:yp_char_literal_node_create_and_unescape`, which is called when we're parsing `YP_TOKEN_CHARACTER_LITERAL`
- introduces a new (unexported) function, `yp_unescape_manipulate_char_literal` which does the same thing as `yp_unescape_manipulate_string` but tells `unescape` that only a single codepoint is expected

## Things reviewers should focus on ...

Well, I'm adding a new public function, which is a notable decision. Maybe this doesn't need to be a public function, since we only call it in one place, when parsing `YP_TOKEN_CHARACTER_LITERAL`?

It also feels funny to me that we're exercising the same code path twice, once during lexing and again during parsing. There may be something deeper we can do avoid unescaping twice, like saving the result on the token.

I'm open to other suggestions on how to fix this.

### DIFF
--- a/include/yarp/unescape.h
+++ b/include/yarp/unescape.h
@@ -29,9 +29,9 @@ typedef enum {
     YP_UNESCAPE_ALL
 } yp_unescape_type_t;
 
-// Unescape the contents of the given token into the given string using the
-// given unescape mode.
+// Unescape the contents of the given token into the given string using the given unescape mode.
 YP_EXPORTED_FUNCTION void yp_unescape_manipulate_string(yp_parser_t *parser, yp_string_t *string, yp_unescape_type_t unescape_type);
+void yp_unescape_manipulate_char_literal(yp_parser_t *parser, yp_string_t *string, yp_unescape_type_t unescape_type);
 
 // Accepts a source string and a type of unescaping and returns the unescaped version.
 // The caller must yp_string_free(result); after calling this function.
--- a/src/unescape.c
+++ b/src/unescape.c
@@ -91,9 +91,10 @@ unescape_hexadecimal_digit(const uint8_t value) {
 // Scan the 1-2 digits of hexadecimal into the value. Returns the number of
 // digits scanned.
 static inline size_t
-unescape_hexadecimal(const uint8_t *backslash, uint8_t *value, const uint8_t *end) {
+unescape_hexadecimal(const uint8_t *backslash, uint8_t *value, const uint8_t *end, yp_list_t *error_list) {
     *value = 0;
     if (backslash + 2 >= end || !yp_char_is_hexadecimal_digit(backslash[2])) {
+        if (error_list) yp_diagnostic_list_append(error_list, backslash, backslash + 2, "Invalid hex escape.");
         return 2;
     }
     *value = unescape_hexadecimal_digit(backslash[2]);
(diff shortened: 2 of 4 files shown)


kddnewton:
@flavorjones yeah it's been on my list for quite a while ... https://github.com/ruby/yarp/issues/579

kddnewton (changes requested):
This is good work and gets us to a better place. I don't think this is going to be the final solution, but I'm very happy to merge this now as it gets us further along. The only thing I would change is that I don't think we need to export this. At the moment the other one is exported so we can fuzz it, but I'm fine not adding a fuzz target for this.

flavorjones:
@kddnewton Updated to not export `yp_unescape_manipulate_char_literal`

kddnewton (approved):

PR #1378, opened 2023-09-01
remove reallocs preparing for custom memory management

HParker:
If we move to a bump/arena allocator like in https://github.com/HParker/yarp/tree/introduce-yp_malloc

we likely don't want to support realloc as it likely requires extra book keeping. Instead we can malloc and memset so our allocator has less cases to consider.

### DIFF
--- a/src/util/yp_newline_list.c
+++ b/src/util/yp_newline_list.c
@@ -25,7 +25,7 @@ yp_newline_list_init(yp_newline_list_t *list, const uint8_t *start, size_t capac
 bool
 yp_newline_list_append(yp_newline_list_t *list, const uint8_t *cursor) {
     if (list->size == list->capacity) {
-        size_t * original_offsets = list->offsets;
+        size_t *original_offsets = list->offsets;
 
         list->capacity = (list->capacity * 3) / 2;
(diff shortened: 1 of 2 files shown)


src/util/yp_newline_list.c
```diff
@@ -25,8 +25,12 @@ yp_newline_list_init(yp_newline_list_t *list, const uint8_t *start, size_t capac
 bool
 yp_newline_list_append(yp_newline_list_t *list, const uint8_t *cursor) {
     if (list->size == list->capacity) {
+        size_t * original_offsets = list->offsets;
```
kddnewton, line 28:
```suggestion
        size_t *original_offsets = list->offsets;
```

kddnewton (approved):

src/util/yp_string_list.c
```diff
@@ -12,8 +12,11 @@ yp_string_list_init(yp_string_list_t *string_list) {
 void
 yp_string_list_append(yp_string_list_t *string_list, yp_string_t *string) {
     if (string_list->length + 1 > string_list->capacity) {
+        yp_string_t * original_string = string_list->strings;
```
kddnewton, line 15:
```suggestion
        yp_string_t *original_string = string_list->strings;
```

PR #1387, opened 2023-09-04
Rename YP_NODE_*_NODE to YP_*_NODE

eregon:
As noticed by a few people the `YP_NODE_` prefix felt redundant given it also ends in `_NODE`.


kddnewton (approved):

jemmaissroff:
> Don't merge until we update CRuby, otherwise this will break their CI.

Can we do this as a more proper migration so we don't have to worry about timing? Can we add a commit to alias YP_*_NODE to YP_NODE_*_NODE temporarily, then make the change in CRuby and then revert that single commit from here?

kddnewton (changes requested):
Marking as "request changes"

eregon:
This feels very similar to the issue we had with TruffleRuby in YARP's CI.
Maybe importing YARP in CRuby shouldn't be done on every commit but manually or semi-automatically (example: a PR per week or so)?

> Can we do this as a more proper migration so we don't have to worry about timing? Can we add a commit to alias YP_NODE to YP_NODE_NODE temporarily, then make the change in CRuby and then revert that single commit from here?

Fine by me, but please push that to this branch then.

eregon:
I added the deprecated aliases in an extra commit, this should be safe to merge now.

I think it would be best to manually import yarp in CRuby for now, to avoid lots of ordering issues and tensions about what gets merged first, very similar to when TruffleRuby was in YARP's CI. It's just so much simpler and comfortable if adopting new APIs can be done when updating to a newer YARP commit (I have a fair amount of experience of that with YARP in TruffleRuby). I think it is likely there will be more breaking changes in YARP (breaking in terms of it would break CRuby if CRuby always pulls YARP master automatically).

PR #1392, opened 2023-09-04
Read template in UTF-8

nobu:
Explicitly specify the encoding, instead of `Encoding.default_external` temporarily.


nobu:
GitHub Actions are troubling?
Containers can't even set up.

PR #1393, opened 2023-09-05
Introduce YARP::Pattern

kddnewton:
This is a feature that I have in Syntax Tree that I've been missing in YARP, so I'm porting it over. This allows you to define a `YARP::Pattern` object that knows how to compare itself against YARP nodes and determine if it matches. Effectively you can think of this as an object that represents a pattern matching expression.

For example:

```ruby
case node
in InstanceVariableReadNode[name: /^@B/i]
  # matched an ivar read that starts with b or B
end
```

That works well when you have a specific call site where you want to compare against nodes, but doesn't work as well if you want it as a variable such that you can pass it around and compare programmatically. Instead, you can now do:

```ruby
pattern = YARP::Pattern.new("InstanceVariableReadNode[name: /^@B/i]")

if pattern.call(node)
  # matched an ivar read that starts with b or B
end
```

These patterns also know how to walk trees to find all of the nodes that will match, which you can do by calling the `#scan` method, as in:

```ruby
pattern = YARP::Pattern.new("InstanceVariableReadNode[name: /^@B/i]")
result = YARP.parse("@foo + @bar + @baz")

pattern.scan(result.value) do |node|
  # each yielded node will match the given pattern
end
```

`#scan` will also return an enumerable if a block is not given.

### DIFF
--- a/lib/yarp/pattern.rb
+++ b/lib/yarp/pattern.rb
@@ -1,0 +1,239 @@
+# frozen_string_literal: true
+
+module YARP
+  # A pattern is an object that wraps a Ruby pattern matching expression. The
+  # expression would normally be passed to an `in` clause within a `case`
+  # expression or a rightward assignment expression. For example, in the
+  # following snippet:
+  #
+  #     case node
+  #     in ConstantPathNode[ConstantReadNode[name: :YARP], ConstantReadNode[name: :Pattern]]
+  #     end
+  #
+  # the pattern is the `ConstantPathNode[...]` expression.
+  #
+  # The pattern gets compiled into an object that responds to #call by running
+  # the #compile method. This method itself will run back through YARP to
+  # parse the expression into a tree, then walk the tree to generate the
+  # necessary callable objects. For example, if you wanted to compile the
+  # expression above into a callable, you would:
+  #
+  #     callable = YARP::Pattern.new("ConstantPathNode[ConstantReadNode[name: :YARP], ConstantReadNode[name: :Pattern]]").compile
+  #     callable.call(node)
+  #
+  # The callable object returned by #compile is guaranteed to respond to #call
+  # with a single argument, which is the node to match against. It also is
+  # guaranteed to respond to #===, which means it itself can be used in a `case`
+  # expression, as in:
+  #
+  #     case node
+  #     when callable
+  #     end
+  #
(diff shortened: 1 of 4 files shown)


lib/yarp/pattern.rb
```diff
@@ -0,0 +1,239 @@
...
+  # following snippet:
+  #
+  #     case node
+  #     in ConstantPathNode[ConstantReadNode[name: :YARP], ConstantReadNode[name: :Pattern]]
+  #     end
+  #
+  # the pattern is the `ConstantPathNode[...]` expression.
+  #
+  # The pattern gets compiled into an object that responds to #call by running
+  # the #compile method. This method itself will run back through Linter to
```
eregon, line 16: Linter -> YARP? (same below)

eregon:
The `.call` calls in this PR will be megamorphic calls, in that is they will see many different blocks.
That's significantly slower than calling a given block (or a small number of them) and is likely a considerable overhead for this API.

Could the interface be:
```ruby
parse_result.scan(-> { |node| node in InstanceVariableReadNode[name: /^@B/i] }) { |node| p node }
# or maybe
parse_result.scan(InstanceVariableReadNode, -> { |node| node.name in /^@B/i }) { |node| p node }
```
Then it would just use Ruby's pattern matching, which is a lot more efficient by avoiding these megamorphic calls and the compilation overhead (e.g. all the extra lambdas, etc).

kddnewton:
@eregon That API is already possible today, but doesn't solve the use case of being able to pass around the patterns as objects. The point here is for us to be able to have a DSL that represents the whole match without having to call Ruby code.

I think we can definitely improve the efficiency of the procs as we iterate. For example we could inline a lot of the attribute checks directly into the hash patterns. We could also potentially cache a lot of the attribute checks so that nodes don't get scanned multiple times.

kddnewton:
@eregon sure, as a CLI:

```
$ yarp search 'InstanceVariableReadNode' test.rb
```

eregon:
That could be done by using `eval` like `pattern = eval "-> { |node| node in #{ARGV.first} }"`, no?

jemmaissroff (approved):

kddnewton:
Not if we don't want to trust the input implicitly

PR #1394, opened 2023-09-05
Use allocator in yp string free

HParker:
relies on: https://github.com/ruby/yarp/pull/1375

Moves frees from yp_string_free to the allocator. Review only the last commit for changes outside of creating the allocator.


HParker:
I am going to restart this one tracking a malloc source instead of a free. There are too many sources for this alloc

PR #1395, opened 2023-09-05
Use allocator in diagnostics

HParker:
relies on: https://github.com/ruby/yarp/pull/1375

review only last commit.


src/yarp.c
```diff
@@ -8170,7 +8170,7 @@ parse_write(yp_parser_t *parser, yp_node_t *target, yp_token_t *operator, yp_nod
                     target = (yp_node_t *) yp_local_variable_write_node_create(parser, constant_id, 0, value, &message, operator);
 
                     if (token_is_numbered_parameter(message.start, message.end)) {
-                        yp_diagnostic_list_append(&parser->error_list, message.start, message.end, "reserved for numbered parameter");
+                        yp_diagnostic_list_append(&parser->allocator, &parser->error_list, message.start, message.end, "reserved for numbered parameter");
```
kddnewton, line 8173: Since we're introducing a new parameter to every single call, I think it would be good to have `yp_parser_error(parser, start, end, message)` and `yp_parser_warning(parser, start, end, message)`. That can be a separate commit that we can merge today, before we add the allocator, and then the allocator is only a 2 line change.

HParker: Ah sure! I didn't want to add another method where the only difference is which pointer it grabs from the parser. 😅 happy to make that change.

PR #1402, opened 2023-09-05
Introduce a BlockLocalVariableNode

kddnewton:
This is a tradeoff that I think is worth it. Right now we have a location list that tracks the location of each of the block locals. Instead, I'd like to make that a node list that has a proper node in each spot in the list. In doing so, we eliminate the need to have a location list at all, making it simpler on all of the various consumers as we have one fewer field type. There should be minimal memory implications here since this syntax is exceedingly rare.

### DIFF
--- a/bin/memsize
+++ b/bin/memsize
@@ -16,7 +16,7 @@ results =
           8
         when "location", "location?"
           16
-        when "node[]", "string", "token", "token?", "location[]", "constant[]"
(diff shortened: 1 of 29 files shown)


eregon (approved):
Nice simplification, and agreed a node is far more convenient here and it's so rare it can't matter much in terms of memory.

PR #1404, opened 2023-09-05
Add constants on constants

### LINKED ISSUE #1391: Use the constant pool for Constant*Node
The other kind of "variables" already seem to all use the constant pool, but constants do not use it yet.
This also prevents skipping locations for serializations to Java, e.g. `ConstantWriteNode#name_loc` as that is the only way to get the name right now.
Related: #807 and #1133

kddnewton:
Fixes #1391

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -867,6 +867,8 @@ nodes:
           ^^^^^^^^^
   - name: ConstantAndWriteNode
     fields:
+      - name: name
+        type: constant
       - name: name_loc
         type: location
       - name: operator_loc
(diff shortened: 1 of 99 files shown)


config.yml
```diff
@@ -807,6 +807,8 @@ nodes:
           ^^^^^^^^^
   - name: ConstantAndWriteNode
     fields:
+      - name: name
+        type: constant
       - name: name_loc
```
jemmaissroff, line 812: Do we want a name and a `name_loc`?

jemmaissroff (changes requested):

jemmaissroff (approved):

Issue #1310, opened 2023-08-23
The block argument of a call node should consistently be in CallNode#block or ArgumentsNode#block


eregon:
Right now it's there if it's a literal block:
```
$ bin/parse -e 'foo(1) {}'      
#<YARP::ParseResult:0x00007f5b82d8f830
 @comments=[],
 @errors=[],
 @source=#<YARP::Source:0x00007f5b82c7c970 @offsets=[0], @source="foo(1) {}">,
 @value=
  ProgramNode(0...9)(
    [],
    StatementsNode(0...9)(
      [CallNode(0...9)(
         nil,
         nil,
         (0...3),
         (3...4),
         ArgumentsNode(4...5)([IntegerNode(4...5)()]),
         (5...6),
         BlockNode(7...9)([], nil, nil, (7...8), (8...9)),
         0,
         "foo"
       )]
    )
  ),
 @warnings=[]>
```

But it's inside ArgumentsNode as the last element otherwise:
```
$ bin/parse -e 'a = ->{}; foo(1, &a)'
#<YARP::ParseResult:0x00007fb3a766f230
 @comments=[],
 @errors=[],
 @source=#<YARP::Source:0x00007fb3a755c9b0 @offsets=[0], @source="a = ->{}; foo(1, &a)">,
 @value=
  ProgramNode(0...20)(
    [:a],
    StatementsNode(0...20)(
      [LocalVariableWriteNode(0...8)(:a, 0, LambdaNode(4...8)([], (4...6), nil, nil), (0...1), (2...3)),
       CallNode(10...20)(
         nil,
         nil,
         (10...13),
         (13...14),
         ArgumentsNode(14...19)(
           [IntegerNode(14...15)(), BlockArgumentNode(17...19)(LocalVariableReadNode(18...19)(:a, 0), (17...18))]
         ),
         (19...20),
         nil,
         0,
         "foo"
       )]
    )
  ),
 @warnings=[]>
```

Fixing this also means CallNode#block cannot be typed as BlockNode but must be BlockNode|BlockArgumentNode (so just `Node`).

From https://github.com/ruby/yarp/pull/1243#issuecomment-1686635150

eregon:
Alternatively, as discussed in https://github.com/ruby/yarp/pull/1243#issuecomment-1686598881, the block could be stored always in ArgumentsNode, as its own `node?` field.

enebo:
[...]

PR #1405, opened 2023-09-05
Call and operator write ergonomics

kddnewton:
Fixes https://github.com/ruby/yarp/issues/1236

This is definitely a tradeoff, because now we don't have a single call node anymore. However, I think this is acceptable considering the call node couldn't represent that actually 2 methods get called when you have these nodes. So it would have been confusing to client libraries anyway.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -582,6 +582,37 @@ nodes:
 
           break foo
           ^^^^^^^^^
+  - name: CallAndWriteNode
+    fields:
+      - name: receiver
+        type: node?
+      - name: call_operator_loc
+        type: location?
+      - name: message_loc
+        type: location?
+      - name: opening_loc
+        type: location?
+      - name: arguments
+        type: node?
+        kind: ArgumentsNode
+      - name: closing_loc
+        type: location?
+      - name: flags
+        type: flags
+        kind: CallNodeFlags
+      - name: read_name
+        type: string
+      - name: write_name
+        type: string
+      - name: operator_loc
+        type: location
+      - name: value
+        type: node
+    comment: |
+      Represents the use of the `&&=` operator on a call.
+
+          foo.bar &&= value
+          ^^^^^^^^^^^^^^^^^
   - name: CallNode
     fields:
       - name: receiver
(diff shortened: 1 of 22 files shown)


jemmaissroff (approved):
I like this change, and like that it mirrors the structure we already have in place here

config.yml
```diff
@@ -625,50 +656,70 @@ nodes:
...
+        type: location?
+      - name: flags
+        type: flags
+        kind: CallNodeFlags
+      - name: read_name
+        type: string
+      - name: write_name
+        type: string
+      - name: operator
+        type: constant
```
eregon, line 682: Duplicating the many fields of CallNode here feels rather like a code smell or anti-pattern.
And e.g. the logic to deal with visibility, arguments, block and various other aspects of calls now need to be duplicated for these 4 nodes, or all of these non-location parameters need to be passed to a common helper function.
Something to help with those concerns would be to track visibility in CallNodeFlags, and #1310. Then those operators calls are probably fairly simple to compile with little duplication.
I think before it would simply be enough to replace the receiver by a temporary variable and be done with it + a copy of the CallNode with/without the `=`, although we haven't written the code for that yet in TruffleRuby, so it might be slightly more complicated.

FWIW here is how the parser gem represents it:
```
$ CODE="foo.bar += baz" && ruby -rparser/current -e 'pp Parser::CurrentRuby.parse(ARGV[0])' "$CODE"     
s(:op_asgn,
  s(:send,
    s(:send, nil, :foo), :bar), :+,
  s(:send, nil, :baz))
```
Which is quite similar to current YARP/before this PR.

And this is how CRuby's current parser represents it:
```
$ ruby --dump=parsetree -e 'foo.bar += baz'                  
# @ NODE_SCOPE (id: 4, line: 1, location: (1,0)-(1,14))
# +- nd_tbl: (empty)
# +- nd_args:
# |   (null node)
# +- nd_body:
#     @ NODE_OP_ASGN2 (id: 3, line: 1, location: (1,0)-(1,14))*
#     +- nd_recv:
#     |   @ NODE_VCALL (id: 0, line: 1, location: (1,0)-(1,3))
#     |   +- nd_mid: :foo
#     +- nd_next->nd_vid: :bar
#     +- nd_next->nd_mid: :+
#     +- nd_value:
#         @ NODE_VCALL (id: 1, line: 1, location: (1,11)-(1,14))
#         +- nd_mid: :baz
```
Which seems closer to this change.

@enebo Does this change help?
What's the difference for https://gist.github.com/enebo/13cc0fcba7159b82a2be60828dd45e66?
What would `buildCallOperatorWrite()` look like with this change?

In summary I'm not really enthusiastic about this change but also we discussed it and I think it's fine to merge.
The addition of both `read_name` and `write_name` is certainly nice.

enebo: @eregon The main points for why I think it is fine:

  1. receiver/target is now no longer nested into another node
  2. The original node was not usable as it existed for read or write
  3. the names for read/write are not things we have to process (e.g chomp write to get read). we just use them (and this also be related to our symbol discussion in other bug since these should eventually be in constant pool).
  4. This is smaller due to one less node

My take on needing common methods is that I already need common methods because I will still be supporting the existing parser for years/ever (pure-Java mode).  I already have one aspect of that already.

So this is a big step forward at the cost of some repeated fields.

eregon: Right, those are good points, thanks for explaining them.
1. makes perfect sense to me (could also be independently but might be weird then).
2. indeed not usable as-is (and no copy/with method yet in Java but we could add some if useful)
3. agreed

PR #1406, opened 2023-09-05
Split warning and error reporting

HParker:
This has the side effect of using parser instead of the error list directly making arena allocation easier.

Alternative to make https://github.com/ruby/yarp/pull/1395 easier.


HParker:
Closing for now due to conflicts. Will reopen later with updated.

PR #1410, opened 2023-09-06
Constant on block parameter node

kddnewton:
This also introduces the `constant?` as a field type, since block parameters don't necessarily have a name associated with them (e.g., `def foo(&); end`).


eregon (approved):

eregon:
An alternative to optional constants would be to use a constant which corresponds to an empty string, but that would be error-prone (e.g. might added an empty string as a local var, that would be bad) so this seems much better.

PR #1412, opened 2023-09-06
Constants on def, kw param, kw rest param, rest param

### LINKED ISSUE #1133: Some nodes only provide the name through `name_loc` location
For example GlobalVariableWriteNode and ClassVariableWriteNode.

I think these should use the constant pool.

Related to https://github.com/ruby/yarp/issues/807: we shouldn't need locations fields for compilation and so wouldn't need to serialize them either for that purpose in JRuby & TruffleRuby.

kddnewton:
Fixes #1133


lib/yarp.rb
```diff
@@ -540,10 +540,10 @@ def self.yarp_locals(source)
             sorted = [
               *params.requireds.grep(RequiredParameterNode).map(&:name),
               *params.optionals.map(&:name),
-              *((params.rest.name ? params.rest.name.to_sym : :*) if params.rest && params.rest.operator != ","),
+              *((params.rest.name || :*) if params.rest && params.rest.operator != ","),
               *params.posts.grep(RequiredParameterNode).map(&:name),
-              *params.keywords.reject(&:value).map { |param| param.name.chomp(":").to_sym },
-              *params.keywords.select(&:value).map { |param| param.name.chomp(":").to_sym }
+              *params.keywords.reject(&:value).map(&:name),
+              *params.keywords.select(&:value).map(&:name)
```
eregon, line 546: Nice!

eregon (approved):

PR #1413, opened 2023-09-06
Introduce owned constants

### LINKED ISSUE #1373: Serializing metadata scope variables causes assert_failure/crash
I am back to hooking up evals and there is an issue in processing constant pool entries which come from metadata in yp_parse_serialize.   Largely, the problem is that `yp_ptrdifft_to_u32` assumes there is a connection between `constant->start` and `parser->start` but the constant in question comes from a pointer inside the metadata parameter (and not parser).

Here is a reduced testcase:

```c
#include "yarp.h"

void main(char **argv, int argc) {
  char *metadata = "\0\0\0\0\1\0\0\0\1\0\0\0\3\0\0\0foo";
  char *source = "p foo";

  yp_buffer_t buf;
  yp_buffer_init(&buf);
  yp_parse_serialize(source, 5, &buf, metadata);
}
```

kddnewton:
Before this commit, constants in the constant pool were assumed to be slices of the source string. This works in _almost_ all cases.

There are times, however, when a string needs to be synthesized. This can occur when passing in locals that need to be scoped through eval, or when generating method names like `foo=`.

After this commit, there is a single bit `owned` boolean on constants in the pool that indicates whether or not it is a slice of the source string. If it is not, it is assumed to be allocated memory that should be freed by the constant pool when the constant pool is freed.

When serializing, constants now take up 9 bytes instead of 8. The first byte indicates whether or not it is an owned string. When it is, instead of 4 bytes for the source offset and 4 bytes for the length it is instead 4 bytes for the buffer offset and 4 bytes for the length. The contents of the owned constants are embedded into the buffer after the constant pool itself.

Fixes #1373

### DIFF
--- a/templates/java/org/yarp/Loader.java.erb
+++ b/templates/java/org/yarp/Loader.java.erb
@@ -30,15 +30,26 @@ public class Loader {
         byte[] get(ByteBuffer buffer, int oneBasedIndex) {
             int index = oneBasedIndex - 1;
             byte[] constant = cache[index];
+
             if (constant == null) {
                 int offset = bufferOffset + index * 8;
                 int start = buffer.getInt(offset);
                 int length = buffer.getInt(offset + 4);
 
                 constant = new byte[length];
-                System.arraycopy(source, start, constant, 0, length);
+
+                if (Integer.compareUnsigned(start, 1 << 31) <= 0) {
+                    System.arraycopy(source, start, constant, 0, length);
+                } else {
+                    int position = buffer.position();
+                    buffer.position(Math.abs(start));
+                    buffer.get(constant, 0, length);
+                    buffer.position(position);
+                }
+
                 cache[index] = constant;
             }
+
             return constant;
         }
 
@@ -169,18 +180,6 @@ public class Loader {
             buffer.position(buffer.position() + 1); // continue after the 0 byte
             return null;
         }
-    }
-
-    private Nodes.Location[] loadLocations() {
-        int length = loadVarInt();
-        if (length == 0) {
-            return Nodes.Location.EMPTY_ARRAY;
-        }
-        Nodes.Location[] locations = new Nodes.Location[length];
-        for (int i = 0; i < length; i++) {
-            locations[i] = loadLocation();
-        }
-        return locations;
     }
 
     private byte[] loadConstant() {
--- a/templates/lib/yarp/serialize.rb.erb
+++ b/templates/lib/yarp/serialize.rb.erb
@@ -157,11 +157,16 @@ module YARP
 
         unless constant
           offset = constant_pool_offset + index * 8
-
           start = serialized.unpack1("L", offset: offset)
           length = serialized.unpack1("L", offset: offset + 4)
 
-          constant = input.byteslice(start, length).to_sym
+          constant =
+            if start.nobits?(1 << 31)
+              input.byteslice(start, length).to_sym
+            else
+              serialized.byteslice(start & ((1 << 31) - 1), length).to_sym
+            end
+
           constant_pool[index] = constant
         end
 
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@
 /doc/
 /pkg/
 /spec/reports/
-/test/serialized/
+/test/yarp/serialized/
 /top-100-gems/
 /tmp/
 /vendor/bundle
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -24,6 +24,7 @@ GEM
 PLATFORMS
   ruby
   universal-java-17
+  universal-java-20
 
 DEPENDENCIES
   ffi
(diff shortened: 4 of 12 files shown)


enebo:
@kddnewton maybe this could be tagged with a bit by reducing the offset space of the pool by a bit?

kddnewton:
@enebo do you mean the overall offset of the constant pool?

I don't think that will quite work since the constant pool can mix owned and unowned constants together.

enebo:
@kddnewton I was thinking it would be behind a function which would strip off the bit and then use the result as the actual offset.  It means the constant pool entry must be accessible using 23 bits instead of 24.

kddnewton:
@enebo sure, just updated

templates/lib/yarp/serialize.rb.erb
```diff
@@ -157,11 +157,16 @@ module YARP
 
         unless constant
           offset = constant_pool_offset + index * 8
-
           start = serialized.unpack1("L", offset: offset)
           length = serialized.unpack1("L", offset: offset + 4)
 
-          constant = input.byteslice(start, length).to_sym
+          constant =
+            if (start & (1 << 31)) == 0
```
eregon, line 164: Maybe a good use-case for Integer#anybits?/nobits?

templates/java/org/yarp/Loader.java.erb
```diff
@@ -30,15 +30,26 @@ public class Loader {
...
+
             if (constant == null) {
                 int offset = bufferOffset + index * 8;
                 int start = buffer.getInt(offset);
                 int length = buffer.getInt(offset + 4);
 
                 constant = new byte[length];
-                System.arraycopy(source, start, constant, 0, length);
+
+                if ((start & (1 << 31)) == 0) {
```
eregon, line 41: Another way to check this in Java is `(start < 0)` since `int` is signed 32-bit

templates/java/org/yarp/Loader.java.erb
```diff
@@ -30,15 +30,26 @@ public class Loader {
...
+                    System.arraycopy(source, start, constant, 0, length);
+                } else {
+                    int position = buffer.position();
+                    buffer.position(start - (1 << 31));
```
eregon, line 45:
```suggestion
                    buffer.position(start & 0x7fffffff);
```
since `1 << 31` doesn't fit in a signed 32-bit int so it feels rather hacky (also because `start` is never >= 1<<31).

eregon (approved):
Could you update `serialize.md` to document how it's serialized now?

templates/java/org/yarp/Loader.java.erb
eregon: But since this works and mirrors the Ruby code, feel free to keep it as-is (same below)

eregon:
Minor, the commit description still says `When serializing, constants now take up 9 bytes instead of 8` but it uses 8 with the updated PR.

templates/lib/yarp/serialize.rb.erb
kddnewton: Updated

templates/java/org/yarp/Loader.java.erb
kddnewton: Updated to use `Integer.compareUnsigned`

templates/java/org/yarp/Loader.java.erb
kddnewton: Updated to use `Math.abs(start)`

kddnewton:
@eregon updated the docs and the commit message

templates/java/org/yarp/Loader.java.erb
eregon: I don't think that's correct:
```
jshell> int start = -3
start ==> -3

jshell> start & 0x7fffffff
$2 ==> 2147483645

jshell> Math.abs(start)
$3 ==> 3
```

eregon: We should probably improve the Java toString() output, then the `build-java` CI job wouldn't catch it but it would be easy to check if it looks correct or not.
Alternatively when TruffleRuby starts using this we should notice as well in the daily CI job.

eregon: The only code that seems to trigger `if (constant->owned) {` seems ParseSerializeTest.
CallNode and operator nodes seem to use type `string` and not `constant` yet.

eregon: + that case doesn't actually use `foo` and adding `a = foo if rand < 0` at the top of the file considers foo as a call but it should be a LocalVariableReadNode. So this feels like it needs more testing and that's why this bug does not really happen yet.

kddnewton: It'll be fixed by https://github.com/ruby/yarp/pull/1466

eregon: Nice, I'm happy the better toString caught that :)

eregon: Ah actually just deserializing caught it, since https://github.com/ruby/yarp/pull/1454 isn't merged yet.

PR #1415, opened 2023-09-06
Fix `Location#end_column`

andyw8:
YARP's `end_column` is off by 1 compared to Syntax Tree, e.g.:

```
ruby-lsp(main):009> SyntaxTree.parse("class Foo\nend").child_nodes.first.body.first.location
=>
#<SyntaxTree::Location:0x00000001073f7a48
 @end_char=13,
 @end_column=3, # <-----
 @end_line=2,
 @start_char=0,
 @start_column=0,
 @start_line=1>
```

vs

```
ruby-lsp(main):034> location = YARP.parse("class Foo\nend").value.child_nodes.first.body.first.location
ruby-lsp(main):036> location.end_offset
=> 13
ruby-lsp(main):037> location.end_column
=> 2 # <-----
ruby-lsp(main):038> location.end_line
=> 2
ruby-lsp(main):039> location.start_offset
=> 0
ruby-lsp(main):040> location.start_column
=> 0
ruby-lsp(main):041> location.start_line
=> 1
```

This should probably be marked as a breaking change.

### DIFF
--- a/lib/yarp.rb
+++ b/lib/yarp.rb
@@ -95,7 +95,7 @@ module YARP
     # The column number in bytes where this location ends from the start of the
     # line.
     def end_column
-      source.column(end_offset - 1)
+      source.column(end_offset)
     end
 
     def deconstruct_keys(keys)
--- a/test/yarp/location_test.rb
+++ b/test/yarp/location_test.rb
@@ -817,6 +817,14 @@ module YARP
       node = result.value.statements.body.last
       node = yield node if block_given?
 
+      if expected.begin == 0
+        assert_equal 0, node.location.start_column
+      end
+
+      if expected.end == source.length
+        assert_equal source.split("\n").last.length, node.location.end_column
+      end
+
       assert_kind_of kind, node
       assert_equal expected.begin, node.location.start_offset
       assert_equal expected.end, node.location.end_offset


test/yarp/location_test.rb
```diff
@@ -817,6 +817,14 @@ def assert_location(kind, source, expected = 0...source.length)
       node = result.value.statements.body.last
       node = yield node if block_given?
 
+      if expected.begin == 0
```
andyw8, line 820: Test changes suggested by @kddnewton

kddnewton (approved):

lib/yarp.rb
```diff
@@ -95,7 +95,7 @@ def start_column
     # The column number in bytes where this location ends from the start of the
     # line.
     def end_column
-      source.column(end_offset - 1)
+      source.column(end_offset)
     end
```
eregon, line 99: It would be good to document the `end_column` is now exclusive (like `Range#exclude_end?`).

FWIW this seems consistent with `RubyVM::AbstractSyntaxTree::Node.last_column`:
```
irb(main):001:0> node=RubyVM::AbstractSyntaxTree.parse("class Foo\nend").children[2]
=> (CLASS@1:0-2:3 (COLON2@1:6-1:9 nil :Foo) nil (SCOPE@1:0-2:3 tbl: [] args: nil body: (BEGIN@1:9-1:9 nil)))
irb(main):002:0> node.last_column
=> 3
irb(main):003:0> node.last_lineno
=> 2
```

PR #1416, opened 2023-09-06
Consider source encoding for `slice`

vinistock:
Currently, when using `Node#slice` for a source that includes unicode characters, we're getting the wrong return. The reason is because we're not taking the encoding into consideration.

Example
```ruby
YARP.parse("á").value.slice
# => "\xC3\xA1"
# => Encoding #<Encoding:ASCII-8BIT>
```
We should take the encoding into consideration, so that `slice` returns the expected strings.

### DIFF
--- a/test/yarp/encoding_test.rb
+++ b/test/yarp/encoding_test.rb
@@ -96,5 +96,11 @@ module YARP
       encoding = YARP.lex("# encoding: ascii-8bit").value[0][0].value.encoding
       assert_equal Encoding.find("ascii-8bit"), encoding
     end
+
+    def test_slice_encoding
+      slice = YARP.parse("# encoding: Shift_JIS\nア").value.slice
+      assert_equal (+"ア").force_encoding(Encoding::SHIFT_JIS), slice
+      assert_equal Encoding::SHIFT_JIS, slice.encoding
+    end
   end
 end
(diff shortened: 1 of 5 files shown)


kddnewton (approved):

test/yarp/encoding_test.rb
```diff
@@ -96,5 +96,11 @@ def test_first_lexed_token
       encoding = YARP.lex("# encoding: ascii-8bit").value[0][0].value.encoding
       assert_equal Encoding.find("ascii-8bit"), encoding
     end
+
+    def test_slice_encoding
+      slice = YARP.parse("# encoding: Shift_JIS\nア").value.slice
+      assert_equal (+"ア").force_encoding(Encoding::SHIFT_JIS), slice
+      assert_equal Encoding::SHIFT_JIS, slice.encoding
```
eregon, line 103: It would be great to test this for YARP.dump + YARP.load too (which uses serialization)

eregon: Ah actually this is implicitly tested by the CI on JRuby & TruffleRuby which use serialization for `YARP.parse`, nevermind :)

kddnewton: It does, by virtue of testing with the FFI backend

Issue #941, opened 2023-06-02
Errors and warnings


kddnewton:
We've been relatively slapdash with the language used in our errors and warnings so far. Before we release v1 of this, I'd like to get the language, voice, and capitalization consistent.

Something I'd be interested in doing is having more structure with our errors. At the moment the strings are baked directly into the source and they're scattered all over the place. We don't have any kind of interpolation/formatting on them because we haven't had to yet, but it would be nice to have more details in them. For example, we have one error that reads `"Can't set variable"`. It would be nice to have at minimum `{ message: "Setting the {{name}} variable is invalid.", name: (0...5) }`.

Part of this work is also to enumerate all of the errors and warnings in the CRuby parser to ensure that we have all of them covered in YARP. Effectively the existing parser errors and warnings should be a strict subset of YARP's.

One possible solution would be to put all of the error messages into `config.yml` or another YAML file and template out a struct similar to our nodes. This would allow us to put names to the errors, which I like a lot because tools that consume these errors would no longer need to rely on the specific message which is pretty brittle. For example, in IRB (https://github.com/ruby/irb/blob/625f759ca8fbee6d8a248c6b924165b16453fa48/lib/irb/ruby-lex.rb#L306-L351) it looks specifically at the messages being received. With classes it would be much simpler.

Prior art in this area includes things like structured errors from TypeScript (https://github.com/yoavbls/pretty-ts-errors).

kddnewton:
Closing this for now since we now have `printf`-style messages when we want them and it's a lower priority.

PR #1418, opened 2023-09-07
Extract error messages into diagnostic.c and use canonical message IDs

### LINKED ISSUE #1379: Making syntax error message more consistent
There are some inconsistencies in syntax error messages. The benefit of consistently formatted messages would primarily be comprehensibility, particularly for developers who do not fluently read English.

I'm happy to offer a PR to make things more consistent, but that PR would reflect my personal choices.

One alternative first step might be to **adopt a lightweight style guide**. If you're open to it, I'm happy to draft something.

Some examples of inconsistencies ...

### Capitalization

Compare

```
Unexpected ...
^

unexpected ... when parent method is not forwarding.
^
```

### Contractions

Compare

```
Can't make alias for number variables.
^^^^^

Cannot use `&&=' on a multi-write.
^^^^^^
```

### Punctuation

Compare

```
Invalid Unicode escape sequence
                               ^

Invalid Unicode escape sequence.
                               ^
```

### Quote style

Compare

```
Expected predicate expression after `until`.
                                    ^^^^^^^

Expected a predicate after 'while'
                           ^^^^^^^

Expected a predicate after `if'.
                           ^^^^

Expected index after for.
                     ^^^
```

### Use of articles

Compare

```
Expected an expression after '('.
         ^^

Expected expression after `*'.
        ^
```

### Verbosity

Compare

```
Expected to be able to parse a pattern after `|'.
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Expected to find a class name after `class`.
         ^^^^^^^^^^^^^^^^^^^^

Expected an expression after the range operator.
         ^^^^^^^^^^^^^
```

And the error message that breaks the pattern entirely:

```
Missing exponent.
```

### Other

Oxford comma?

```
Expected a bare word, symbol or global variable argument.
```

When to give actionable advice?

```
ambiguity between regexp and two divisions: wrap regexp in parentheses or add a space after `/' operator
```

Copyedit for clarity?

```
Unexpected else without no when clauses in case statement.
```

Are there standard terms of art that we expect all developers to know?

```
Unterminated embdoc
```

flavorjones:
The parser now passes around `yp_diagnostic_id_t` for diagnostic messages instead of character strings, and we rely on the function `diagnostic_message()` to resolve that to a string.

In addition, many messages were edited so that the parser expresses coordinate ideas in similar form [1] using consistent voice and typographic conventions.

Also adds a `test:rust` rake task and updates `CONTRIBUTING.md`.

Closes #1379, and makes progress on #941.

  [1]: Strunk & White rule 19


kddnewton (changes requested):
Nice work! We need a rebase, the rust tests are failing because of the changed message, and I think we should prefix it with `YP_DIAG` instead of `YP_ERR` since it encompasses warnings as well.

flavorjones:
> I think we should prefix it with YP_DIAG instead of YP_ERR since it encompasses warnings as well

I had used `YP_ERR` for errors and `YP_WARN` for warnings ... happy to change it to `YP_DIAG` but wanted to make the convention this PR adopted clear first.

Will look at the rust failures, I hadn't run those locally before pushing.

kddnewton:
@flavorjones OH! Sorry I didn't see the warnings at the bottom. I think that's fine as it is.

flavorjones:
@kddnewton rebased, fixed the rust tests.

Also pushed a commit with a `test:rust` rake task and updated `CONTRIBUTING.md`.

kddnewton (approved):

src/diagnostic.c
```diff
@@ -1,12 +1,264 @@
...
+    [YP_ERR_UNARY_RECEIVER_MINUS]               = "Expected a receiver for unary `-`",
+    [YP_ERR_UNARY_RECEIVER_PLUS]                = "Expected a receiver for unary `+`",
+    [YP_ERR_UNARY_RECEIVER_TILDE]               = "Expected a receiver for unary `~`",
+    [YP_ERR_UNTIL_TERM]                         = "Expected an `end` to close the `until` statement",
+    [YP_ERR_WHILE_TERM]                         = "Expected an `end` to close the `while` statement",
+    [YP_ERR_WRITE_TARGET_READONLY]              = "Immutable variable as a write target",
+    [YP_ERR_WRITE_TARGET_UNEXPECTED]            = "Unexpected write target",
+    [YP_ERR_XSTRING_TERM]                       = "Expected a closing delimiter for the `%x` or backtick string",
+    [YP_WARN_AMBIGUOUS_FIRST_ARGUMENT_MINUS]    = "Ambiguous first argument; put parentheses or a space even after `-` operator",
+    [YP_WARN_AMBIGUOUS_FIRST_ARGUMENT_PLUS]     = "Ambiguous first argument; put parentheses or a space even after `+` operator",
```
paracycle, line 242: I know these have been extracted from the original source, but I cannot quite understand what `or a space even` part in these messages is supposed to mean. Are we trying to say "put parentheses or a space after `x` operator"?

flavorjones: These warnings are printed in cases like:

```
a +c
```

I agree they're not very clear; however this particular error message is nearly identical to the CRuby parser error message:

https://github.com/ruby/ruby/blob/094f336a274ce83bc9868c7f9be0962b54ec46a7/parse.y#L8581

In any case, suggestions on how to clarify the warning would probably be welcome.

PR #1420, opened 2023-09-07
Interpolated(X)StringNodes with 0/1 part are now (X)StringNodes

### LINKED ISSUE #1414: Heredocs with no interpolation should be StringNodes
A heredoc with no interpolation, like this:

```ruby
<<~HERE
doc
HERE
```

still has an InterpolatedStringNode wrapping a StringNode:

```
@ ProgramNode (location: (0...7))
├── locals: []
└── statements:
    @ StatementsNode (location: (0...7))
    └── body: (length: 1)
        └── @ InterpolatedStringNode (location: (0...7))
            ├── opening_loc: (0...7) = "<<~HERE"
            ├── parts: (length: 1)
            │   └── @ StringNode (location: (8...12))
            │       ├── opening_loc: ∅
            │       ├── content_loc: (8...12) = "doc\n"
            │       ├── closing_loc: ∅
            │       └── unescaped: "doc\n"
            └── closing_loc: (12...17) = "HERE\n"
```

We should instead have just a StringNode on its own in the case of a heredoc with no interpolation. Note that this would also apply to an empty heredoc.

jemmaissroff:
Prior to this commit, heredocs were always InterpolatedNodes, regardless of whether they had interpolation.

After this commit, if a heredoc InterpolatedNode has zero or one parts, it is instead converted into a StringNode or an XStringNode

Closes #1414


enebo:
@jemmaissroff I was wondering if this will affect syntactical libraries.  Is there a way to know this came from a heredoc originally?  Or is this even a real concern?  (note: I am happy with this change for compilation and JRuby/MRI does this in existing parsers)

jemmaissroff:
@enebo interesting. Off the top of my head, I can't think of how it would (we still preserve the heredoc opening and closing locs as the `<<-NAME`). But, if it becomes an issue I think we could definitely add a heredoc flag to the `StringNode`.

Also worth noting that this already would've been an issue, separate from the PR, with how `InterpolatedStringNodes` were working

jemmaissroff:
Going to try a different approach, closing this PR in the interim

eregon:
Replaced by https://github.com/ruby/yarp/pull/1427

PR #1422, opened 2023-09-07
Avoid an extra "stop" parameter to yp_strspn_whitespace_newlines

flavorjones:
and use yp_strspn_inline_whitespace instead.

Partially reverts implementation details from #1152

cc @tenderlove


flavorjones:
This PR came out of a side conversation with @kddnewton who suggested an approach like this instead. Would love feedback.

kddnewton (approved):

PR #1423, opened 2023-09-07
Add ParseResult#attach_comments! to tie comments to their locations

vinistock:
Development tools like RDoc or LSPs need to show comments in certain situations as they are considered documentation. The most prominent example is to implement hover in an LSP, where we want to show the documentation for a given node.

Currently, there's no easy way to figure out which comments are associated with each node. Let's bring in SyntaxTree's algorithm for attaching comments to locations so that we can easily ask a location what comments exist there.

### DIFF
--- a/lib/yarp.rb
+++ b/lib/yarp.rb
@@ -20,6 +20,10 @@ module YARP
       offsets.bsearch_index { |offset| offset > value } || offsets.length
     end
 
+    def line_offset(value)
+      offsets[line(value) - 1]
+    end
+
     def column(value)
       value - offsets[line(value) - 1]
     end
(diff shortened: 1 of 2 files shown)


lib/yarp.rb
```diff
@@ -230,6 +249,154 @@ def failure?
...
+      private
+
+      # Responsible for finding the nearest targets to the given comment within the context of the given encapsulating
+      # node.
+      def nearest_targets(node, comment)
+        comment_start = comment.location.start_offset
+        comment_end = comment.location.end_offset
+
+        targets = []
+        node.deconstruct_keys(nil).each do |key, value|
```
kddnewton, line 335: The only thing I don't like about this is relying on `deconstruct_keys`. I think this would be a really good place for either reflection or templating out a method. I'll do some thinking about this after this is merged.

kddnewton (approved):

PR #1427, opened 2023-09-08
Heredocs can create Interpolated(X)StringNodes or (X)StringNodes

### LINKED ISSUE #1414: Heredocs with no interpolation should be StringNodes

jemmaissroff:
Prior to this commit, heredocs were automatically InterpolatedNodes regardless of whether there was actually interpolation. With this commit, heredocs are only interpolate if there is actually interpolation

Closes #1414

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -11281,10 +11281,7 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
                             &((yp_token_t) { .type = YP_TOKEN_STRING_CONTENT, .start = parser->previous.end, .end = parser->current.start }),
                             &parser->current);
                 }
-                node->location = (yp_location_t) {
-                    .start = parser->previous.start,
-                        .end = parser->previous.end
-                };
+                node->location.end = parser->previous.end;
                 lex_state_set(parser, YP_LEX_STATE_END);
                 expect(parser, YP_TOKEN_HEREDOC_END, YP_ERR_HEREDOC_TERM);
                 return node;
@@ -11302,18 +11299,9 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
             if (YP_NODE_TYPE_P(part, YP_STRING_NODE) && match_any_type_p(parser, 2, YP_TOKEN_HEREDOC_END, YP_TOKEN_EOF)) {
                 // We only have a single string, so we can return it
                 yp_string_node_t *str_part = (yp_string_node_t *)part;
-                (str_part)->opening_loc = (yp_location_t) {
-                    .start = opening_token.start,
-                    .end = opening_token.end
-                };
-                (str_part)->closing_loc = (yp_location_t) {
-                    .start = parser->current.start,
-                    .end = parser->current.end
-                };
-                (str_part)->base.location = (yp_location_t) {
-                    .start = opening_token.start,
-                    .end = opening_token.end,
-                };
+                str_part->opening_loc = YP_LOCATION_TOKEN_VALUE(&opening_token);
+                str_part->closing_loc = YP_LOCATION_TOKEN_VALUE(&parser->current);
+                str_part->base.location = str_part->opening_loc;
                 if (quote == YP_HEREDOC_QUOTE_BACKTICK) {
                     part->type = YP_X_STRING_NODE;
                 }
@@ -11328,16 +11316,13 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
             else {
                 // We have multiple parts, continue parsing them
                 yp_node_list_t parts = YP_EMPTY_NODE_LIST;
+                yp_node_list_append(&parts, part);
 
                 while (!match_any_type_p(parser, 2, YP_TOKEN_HEREDOC_END, YP_TOKEN_EOF)) {
-                    if (part != NULL) {
+                    if ((part = parse_string_part(parser)) != NULL) {
                         yp_node_list_append(&parts, part);
                     }
-
-                    if ((part = parse_string_part(parser)) == NULL) continue;
                 }
-
-                yp_node_list_append(&parts, part);
 
                 if (quote == YP_HEREDOC_QUOTE_BACKTICK) {
                     node = (yp_node_t *) yp_interpolated_xstring_node_create(parser, &opening_token, &opening_token);
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -10037,6 +10037,53 @@ parse_method_definition_name(yp_parser_t *parser) {
     }
 }
 
+static int
+parse_heredoc_common_whitespace_for_single_node(yp_parser_t *parser, yp_node_t *node, int common_whitespace)
+{
+    const yp_location_t *content_loc = &((yp_string_node_t *) node)->content_loc;
+    int cur_whitespace;
+    const uint8_t *cur_char = content_loc->start;
+
+    while (cur_char && cur_char < content_loc->end) {
+        // Any empty newlines aren't included in the minimum whitespace
+        // calculation.
+        size_t eol_length;
+        while ((eol_length = match_eol_at(parser, cur_char))) {
+            cur_char += eol_length;
+        }
+
+        if (cur_char == content_loc->end) break;
+
+        cur_whitespace = 0;
+
+        while (yp_char_is_inline_whitespace(*cur_char) && cur_char < content_loc->end) {
+            if (cur_char[0] == '\t') {
+                cur_whitespace = (cur_whitespace / YP_TAB_WHITESPACE_SIZE + 1) * YP_TAB_WHITESPACE_SIZE;
+            } else {
+                cur_whitespace++;
+            }
+            cur_char++;
+        }
+
+        // If we hit a newline, then we have encountered a line that
+        // contains only whitespace, and it shouldn't be considered in
+        // the calculation of common leading whitespace.
+        eol_length = match_eol_at(parser, cur_char);
+        if (eol_length) {
+            cur_char += eol_length;
+            continue;
+        }
+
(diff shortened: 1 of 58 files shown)


kddnewton:
Sorry I haven't gotten a chance to look at this yet, but I will first thing next week!

src/yarp.c
```diff
@@ -10105,77 +10185,14 @@ parse_heredoc_dedent(yp_parser_t *parser, yp_node_t *node, yp_heredoc_quote_t qu
...
-            // At this point we have dedented all that we need to, so we need to find
-            // the next newline.
-            const uint8_t *breakpoint = next_newline(source_cursor, source_end - source_cursor);
-
-            if (breakpoint == NULL) {
-                // If there isn't another newline, then we can just move the rest of the
-                // string and break from the loop.
-                memmove(dest_cursor, source_cursor, (size_t) (source_end - source_cursor));
-                break;
+            yp_string_node_t *string_node = ((yp_string_node_t *) node);
```
kddnewton, line 10188: I think the indentation is off here

src/yarp.c
```diff
@@ -10105,77 +10185,14 @@ parse_heredoc_dedent(yp_parser_t *parser, yp_node_t *node, yp_heredoc_quote_t qu
...
+            string_node->unescaped = *(parse_heredoc_dedent_single_node(parser, &string_node->unescaped, dedent_next, common_whitespace, quote));
```
kddnewton, line 10189: Since you're passing a reference for the string into the function here, I don't think this assignment should be necessary.

src/yarp.c
```diff
@@ -11218,42 +11235,112 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
...
-            } else {
-                node = (yp_node_t *) yp_interpolated_string_node_create(parser, &parser->current, NULL, &parser->current);
-            }
 
             parser_lex(parser);
-            yp_node_t *part;
+
+            if (parser->current.type == YP_TOKEN_HEREDOC_END) {
+                if (quote == YP_HEREDOC_QUOTE_BACKTICK) {
+                    yp_x_string_node_t *x_str_node = yp_xstring_node_create(parser, &parser->previous, &parser->previous, &parser->current);
```
kddnewton, line 11243: Can you pass in a `yp_token_t` for the content token instead of updating everything below? Something like
```suggestion
                    yp_node_t *node = (yp_node_t *) yp_xstring_node_create(parser, &parser->previous, &((yp_token_t) { .type = YP_TOKEN_STRING_CONTENT, .start = parser->previous.end, .end = parser->current.start }, &parser->current);
                    node->location.end = parser->previous.end;
```

src/yarp.c
```diff
@@ -11218,42 +11235,112 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
...
+                }
+                lex_state_set(parser, YP_LEX_STATE_END);
+                expect(parser, YP_TOKEN_HEREDOC_END, YP_ERR_HEREDOC_TERM);
+                return node;
+            }
+
+            yp_token_t opening_token = parser->previous;
+
+            yp_node_t *part = NULL;
+            yp_node_list_t parts = YP_EMPTY_NODE_LIST;
```
kddnewton, line 11273: Here you're creating a list, then appending all of the parts, then checking the size of the list. I think this might be easier to read if you instead parsed the first part, checked its contents and the next token, and then only created a list if it was necessary.

For example:

```c
yp_node_t *part = parse_string_part(parser);
if (part == NULL) {
    // We couldn't parse anything, so just return a missing node.
    return (yp_node_t *) yp_missing_node_create(parser, parser->previous.start, parser->previous.end);
}

if (YP_NODE_TYPE(part, YP_STRING_NODE) && match_any_type_p(parser, 2, YP_TOKEN_HEREDOC_END, YP_TOKEN_EOF)) {
    // Here we only have a single string, so we can return just the one string part.
}

// Otherwise, we need to parse all of the parts of the string.
yp_node_list_t parts = YP_EMPTY_NODE_LIST;
yp_node_list_append(&parts, part);

while (!match_any_type_p(parser, 2, YP_TOKEN_HEREDOC_END, YP_TOKEN_EOF)) {
    if ((part = parse_string_part(parser)) == NULL) continue;
    yp_node_list_append(&parts, part);
}

// Now we have all the parts of the string, create the correct node and return it.
```

kddnewton (changes requested):

src/yarp.c
```diff
@@ -11218,42 +11235,109 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
...
+
                 } else {
-                    yp_interpolated_string_node_append((yp_interpolated_string_node_t *) node, part);
+                    node = (yp_node_t *)yp_string_node_create(
+                            parser,
+                            &parser->previous,
+                            &((yp_token_t) { .type = YP_TOKEN_STRING_CONTENT, .start = parser->previous.end, .end = parser->current.start }),
+                            &parser->current);
                 }
+                node->location = (yp_location_t) {
```
kddnewton, line 11256: From your calls to `yp_string_node_create`, `node->location.start` should already be set correctly. So I think this line can be:

```suggestion
                node->location.end = parser->previous.end;
```

src/yarp.c
```diff
@@ -11218,42 +11235,109 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
...
+                    int common_whitespace = parse_heredoc_common_whitespace_for_single_node(parser, node, -1);
+                    parse_heredoc_dedent_single_node(parser, &str_part->unescaped, true, common_whitespace, quote);
+                }
+            }
+            else {
+                // We have multiple parts, continue parsing them
+                yp_node_list_t parts = YP_EMPTY_NODE_LIST;
+
+                while (!match_any_type_p(parser, 2, YP_TOKEN_HEREDOC_END, YP_TOKEN_EOF)) {
+                    if (part != NULL) {
```
kddnewton, line 11305: The flow of this is kind of hard to read, because it looks like you're delaying appending parts onto the list based on the previous loop iteration. I think this might be easier if you append them right when you parse them.

```suggestion
                    if ((part = parse_string_part(parser)) != NULL) yp_node_list_append(&parts, part);
```

assuming before the loop that you `yp_node_list_append(&parts, part);`.

src/yarp.c
```diff
@@ -11218,42 +11235,109 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
...
+                yp_string_node_t *str_part = (yp_string_node_t *)part;
+                (str_part)->opening_loc = (yp_location_t) {
+                    .start = opening_token.start,
+                    .end = opening_token.end
+                };
+                (str_part)->closing_loc = (yp_location_t) {
+                    .start = parser->current.start,
+                    .end = parser->current.end
+                };
+                (str_part)->base.location = (yp_location_t) {
```
kddnewton, line 11285: I think you can simplify this to `str_part->base.location = str_part->opening_loc` since you just set it.

src/yarp.c
```diff
@@ -11218,42 +11235,109 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
...
-            if (indent == YP_HEREDOC_INDENT_TILDE) {
-                parse_heredoc_dedent(parser, node, quote);
+            if (YP_NODE_TYPE_P(part, YP_STRING_NODE) && match_any_type_p(parser, 2, YP_TOKEN_HEREDOC_END, YP_TOKEN_EOF)) {
+                // We only have a single string, so we can return it
+                yp_string_node_t *str_part = (yp_string_node_t *)part;
+                (str_part)->opening_loc = (yp_location_t) {
+                    .start = opening_token.start,
+                    .end = opening_token.end
+                };
+                (str_part)->closing_loc = (yp_location_t) {
```
kddnewton, line 11281:
```suggestion
                str_part->closing_loc = YP_LOCATION_TOKEN_VALUE(&parser->current);
```

src/yarp.c
```diff
@@ -11218,42 +11235,109 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
...
             }
 
-            // If this is a heredoc that is indented with a ~, then we need to dedent
-            // each line by the common leading whitespace.
-            if (indent == YP_HEREDOC_INDENT_TILDE) {
-                parse_heredoc_dedent(parser, node, quote);
+            if (YP_NODE_TYPE_P(part, YP_STRING_NODE) && match_any_type_p(parser, 2, YP_TOKEN_HEREDOC_END, YP_TOKEN_EOF)) {
+                // We only have a single string, so we can return it
+                yp_string_node_t *str_part = (yp_string_node_t *)part;
+                (str_part)->opening_loc = (yp_location_t) {
```
kddnewton, line 11277:
```suggestion
                str_part->opening_loc = YP_LOCATION_TOKEN_VALUE(&opening_token);
```

kddnewton (changes requested):

kddnewton (approved):

kddnewton:
@jemmaissroff just needs a rebase

PR #1437, opened 2023-09-11
Fix Regexp flag order

seven1m:
This fixes the flag values so they match MRI:

```ruby
irb(main):001:0> Regexp::IGNORECASE
=> 1
irb(main):002:0> Regexp::EXTENDED
=> 2
irb(main):003:0> Regexp::MULTILINE
=> 4
```


kddnewton (approved):
Beautiful!

PR #1438, opened 2023-09-11
Mark flags as private

kddnewton:
The flags integer is an implementation detail. We want people to use the query methods to access the individual fields so we are freed from having to maintain a specific order. As such, this commit changes the Ruby API to mark all flags fields as private attr_readers.

The only one that has a clear use case is returning the set of options given to regular expressions, to mirror the Regexp#options API. So, to support this use case, this commit introduces RegularExpressionNode#options and InterpolatedRegularExpressionNode#options. These APIs provide back the same integer so that they can be used interchangeably.

### DIFF
--- a/lib/yarp.rb
+++ b/lib/yarp.rb
@@ -389,30 +389,6 @@ module YARP
     end
   end
 
-  class FloatNode < Node
-    def value
-      Float(slice)
-    end
-  end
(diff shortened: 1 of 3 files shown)


seven1m (commented):
Nice!

PR #1439, opened 2023-09-11
Handle parsing local variable singleton method definition

### LINKED ISSUE #1435: Possible bug with `def obj.foo; end`?
The following Ruby code produces an unexpected (to me) AST:

```ruby
o = Object.new
def o.foo; end
```

...which gives:

```ruby
irb(main):001:0> YARP.parse('o = Object.new; def o.foo; end')
=> 
#<YARP::ParseResult:0x000055cdb0e77d58
 @comments=[],
 @errors=[],
 @source=#<YARP::Source:0x000055cdb0e7cd08 @offsets=[0], @source="o = Object.new; def o.foo; end">,
 @value=
  ProgramNode(0...30)(
    [:o],
    StatementsNode(0...30)(
      [LocalVariableWriteNode(0...14)(:o, 0, (0...1), CallNode(4...14)(ConstantReadNode(4...10)(:Object), (10...11), (11...14), nil, nil, nil, nil, 0, "new"), (2...3)),
       DefNode(16...30)(:foo, (22...25), CallNode(20...21)(nil, nil, (20...21), nil, nil, nil, nil, 2, "o"), nil, nil, [], (16...19), (21...22), nil, nil, nil, (27...30))]
    )
  ),
 @warnings=[]>
```

Specifically the receiver of the DefNode, e.g.

```ruby
irb(main):011:0> node.child_nodes.first.child_nodes[1].receiver
=> CallNode(20...21)(nil, nil, (20...21), nil, nil, nil, nil, 2, "o")
```

Is this a bug? I would expect a `LocalVariableReadNode` or maybe a `LocalVariableTargetNode` instead of a `CallNode`.

This is what Whitequark Parser produces:

```ruby
irb(main):002:0> Parser::CurrentRuby.parse('o = Object.new; def o.foo; end')
=> 
s(:begin,                                        
  s(:lvasgn, :o,                                 
    s(:send,                                     
      s(:const, nil, :Object), :new)),           
  s(:defs,                                       
    s(:lvar, :o), :foo,                          
    s(:args), nil))
```

...vs what it would produce if `o` wasn't a known local variable:

```ruby
irb(main):001:0> Parser::CurrentRuby.parse('def o.foo; end')
=> 
s(:defs,
  s(:send, nil, :o), :foo,
  s(:args), nil)
```

kddnewton:
Fixes #1435


jemmaissroff (approved):

seven1m (commented):
Nice!

PR #1450, opened 2023-09-12
Only keep semantic fields in Java, i.e. skip location fields

eregon:
* Add a include_location_fields argument for serialize functions.
* Add a byte in the header to indicate whether location fields are included.
* Fixes https://github.com/ruby/yarp/issues/807

cc @enebo

### DIFF
--- a/rakelib/serialization.rake
+++ b/rakelib/serialization.rake
@@ -1,39 +1,34 @@
 # frozen_string_literal: true
 
-fixtures = File.expand_path("../test/yarp/fixtures", __dir__)
-serialized_dir = File.expand_path("../serialized", fixtures)
+task "test:java_loader" do
+  # Recompile with YARP_SERIALIZE_ONLY_SEMANTICS_FIELDS=1
+  # Due to some JRuby bug this does not get propagated to the compile task, so require the caller to set the env var
+  # ENV["YARP_SERIALIZE_ONLY_SEMANTICS_FIELDS"] = "1"
+  raise "this task requires $SERIALIZE_ONLY_SEMANTICS_FIELDS to be set" unless ENV["YARP_SERIALIZE_ONLY_SEMANTICS_FIELDS"]
 
-desc "Serialize test fixtures and save it to .serialized files"
-task "test:serialize_fixtures" do
-  $:.unshift(File.expand_path("../lib", __dir__))
-  require "yarp"
-  require "fileutils"
-
-  Dir["**/*.txt", base: fixtures].each do |relative|
-    path = "#{fixtures}/#{relative}"
-    serialized_path = "#{serialized_dir}/#{relative}"
-
-    serialized = YARP.dump_file(path)
-    FileUtils.mkdir_p(File.dirname(serialized_path))
-    File.write(serialized_path, serialized)
-  end
+  Rake::Task["clobber"].invoke
+  Rake::Task["test:java_loader:internal"].invoke
 end
 
-task "test:java_loader" do
+task "test:java_loader:internal" => :compile do
+  fixtures = File.expand_path("../test/yarp/fixtures", __dir__)
+
+  $:.unshift(File.expand_path("../lib", __dir__))
+  require "yarp"
+  raise "this task requires the FFI backend" unless YARP::BACKEND == :FFI
+  require "fileutils"
   require 'java'
   require_relative '../tmp/yarp.jar'
   java_import 'org.yarp.Nodes$Source'
 
   Dir["**/*.txt", base: fixtures].each do |relative|
     path = "#{fixtures}/#{relative}"
-    serialized_path = "#{serialized_dir}/#{relative}"
-    serialized = File.binread(serialized_path).unpack('c*')
-
     puts
     puts path
+    serialized = YARP.dump_file(path)
     source_bytes = File.binread(path).unpack('c*')
     source = Source.new(source_bytes.to_java(:byte))
-    parse_result = org.yarp.Loader.load(serialized, source)
+    parse_result = org.yarp.Loader.load(serialized.unpack('c*'), source)
     puts parse_result.value
   end
 end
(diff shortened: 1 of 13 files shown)


enebo:
@eregon We cannot toggle this yet.  regular expressions do not unescape properly.  I also still use non-escaped string value via location in one place (which perhaps is another problem).  I will make an issue for one right now and perhaps a second will be forthcoming.

enebo:
I cannot link so this is first issue: https://github.com/ruby/yarp/issues/1452

eregon:
@kddnewton Do you think #1452 could be fixed soon?
Otherwise I could change the logic to have something like `YARP::Field#semantic_field?` and special-case for the few locations we still need until such issues are fixed.

rakelib/serialization.rake
```diff
@@ -8,12 +8,13 @@ task "test:serialize_fixtures" do
   $:.unshift(File.expand_path("../lib", __dir__))
```
eregon, line 8: This task can be simplified further by being merged into the one below, since we have the FFI backend now and can serialize from JRuby (instead of having to use CRuby for that).

eregon: Done

eregon:
> If you want to take a look, that would be great. Otherwise I'll add it to my list.

I don't think I'll have time for that one.

I updated the PR so we can mark individual location fields and still serialize those until they have a replacement.

eregon:
@enebo I preserved the `content_loc` fields you mentioned in https://github.com/ruby/yarp/issues/1452, see the changes to config.yml.

Do you also need `StringNode#opening_loc` ? I'm not sure what you meant in https://github.com/ruby/yarp/issues/1452#issuecomment-1716271367, that case has `opening_loc: ∅`.
In any case it's very easy to add back a given field (one-line change) if it turns out to be necessary for now.

enebo:
@eregon if you look at the other String nodes you will see the ones with delimeters have it but ones coming from interpolated strings do not.  So it is neccesary until unescaped bytes for regexp is fixed.   This is my current logic:

```java
boolean interpolated = node.opening_loc != null && source[node.opening_loc.startOffset] != '\'';
```

eregon:
@enebo OK I kept StringNode#opening_loc for now.
I think this could be handled in visitInterpolatedRegularExpressionNode by handling the parts and not using the generic visitor but handling StringNode there specially.
But it's not a big deal to keep this field for now, the `unescaped` needs to be fixed anyway.

eregon:

kddnewton (changes requested):
This PR has two names for the same thing: `semantic_fields` and `location_fields`. Can you call them all `semantic_fields`?

eregon:
@kddnewton The problem is how should the boolean be named?
Right now it's `include_location_fields`, `include_semantics_fields` would be wrong, maybe `only_semantics_fields` then?

kddnewton:
I would probably just call it `semantic`.

eregon:
OK, I'll update to that tomorrow

eregon:
I have changed the approach, now there is an env var (YARP_SERIALIZE_ONLY_SEMANTICS_FIELDS) which is used at templating time and decides whether to serialize non-semantic/location fields or not.
That way there is no runtime overhead and serialize.c does not have any runtime checks for this.
This is closer to what I had initially in mind, i.e., template-time configuration, which is used (& needed) in https://github.com/ruby/yarp/pull/1493 as well.

eregon:
See https://github.com/ruby/yarp/pull/1532#issuecomment-1722218391 for the savings on serialized size with this PR

src/yarp.c
```diff
@@ -14436,6 +14436,7 @@ yp_serialize(yp_parser_t *parser, yp_node_t *node, yp_buffer_t *buffer) {
     yp_buffer_append_u8(buffer, YP_VERSION_MAJOR);
     yp_buffer_append_u8(buffer, YP_VERSION_MINOR);
     yp_buffer_append_u8(buffer, YP_VERSION_PATCH);
+    yp_buffer_append_u8(buffer, YP_SERIALIZE_ONLY_SEMANTICS_FIELDS ? 1 : 0);
```
eregon, line 14439: `YP_SERIALIZE_ONLY_SEMANTICS_FIELDS` needs to be defined just for this usage.
I think we should `yp_serialize` or the part writing the header to `serialize.c.erb`, but that seemed out of scope of this PR.

kddnewton:
@eregon I'm nervous a bit about the implications of having this be an environment variable. It means everything is dependent on how the gem/project was templated, as opposed to the options specified. This now means there could be a mismatch between the `serialize.c` file serializing the AST and `serialize.rb` doing the deserializing. 

In general I think almost all of this PR is really good, I'm just very nervous about making the runtime behavior tied to a build-time specification, especially considering how many different consumers there are. Would you consider moving this back to a runtime specification?

eregon:
> @eregon I'm nervous a bit about the implications of having this be an environment variable. It means everything is dependent on how the gem/project was templated, as opposed to the options specified. This now means there could be a mismatch between the `serialize.c` file serializing the AST and `serialize.rb` doing the deserializing.

Yes the mismatch is possible but it is checked and errors properly in case of the mismatch with the added byte in the serialization header.

> In general I think almost all of this PR is really good, I'm just very nervous about making the runtime behavior tied to a build-time specification, especially considering how many different consumers there are. Would you consider moving this back to a runtime specification?

I believe we need build-time specification for https://github.com/ruby/yarp/pull/1493 anyway.
So in my view it's necessary at least for Java consumers of YARP.

I'm not very keen to make it a runtime specification because it makes the code more complex and might have some performance overhead, when there is AFAIK no value/use-case to specify this dynamically.

If it turns out to be needed to be a runtime parameter we can always change this fairly easily.
I pushed https://github.com/ruby/yarp/compare/main...eregon:yarp:semantic-fields-bak which was the approach using a runtime parameter in case we want to get inspiration from it later.

kddnewton (approved):
Okay. We can go with this for now. I'm still nervous about this change because we've tied runtime behavior to build-time config, but hopefully I'm wrong.

eregon:
Thank you for merging.

> we've tied runtime behavior to build-time config, but hopefully I'm wrong.

Only when YARP is used as the JRuby/TruffleRuby parser.

When YARP is used as a gem, location fields are always included and hence the CI still passes on JRuby/TruffleRuby.
`serialize.rb.erb` checks and expects that location fields are included, so that's clearly defined.
Conversely `Loader.java.erb` checks and expects that location fields are not included.
The only part needing to handle both cases is `serialize.c.erb` and that's just a couple lines of difference.

Issue #574, opened 2023-03-19
Parse other magic comments besides encoding


kddnewton:
To support compilation, we will need to store various other options enabled by magic comments. Particularly the `frozen_string_literal` pragma. We will need to both parse this and store it on the parser.

eregon:
We need this in the serialized output as well of course. IMO it should accept arbitrary keys and values, and can be stored like a list/array of strings in the serialized form

eregon:
One thought about when looking with @andrykonchin today is for serialization we could serialize this next to the encoding, as `varint number of magic comments; [magic comment name, magic comment value]*`.
We'd need `yp_parse()` to store the magic encodings somewhere though (unless it does already?)

PR #1451, opened 2023-09-12
Frozen string literal comments

### LINKED ISSUE #157: Magic comment syntax
We need to better handle magic comments. At the moment we only support very limited checking for `frozen_string_literal: true`. There is a formula for handling this written in CRuby that we should mimic - it needs to additionally handle the emacs format.

For known magic comments, we should store booleans so that consumers don't have to reparse the magic comments.

kddnewton:
Parse out frozen string literal comments, add a static_literal common flag, and a string-specific frozen flag.

Note that this doesn't close #157 or #574 yet, because we are not exposing this flag yet and we still aren't supporting all of the different magic comment syntax. But this is a stepping stone to get there.

### DIFF
--- a/include/yarp/parser.h
+++ b/include/yarp/parser.h
@@ -378,30 +378,30 @@ struct yp_parser {
     yp_node_flags_t integer_base;
 
     // Whether or not we're at the beginning of a command
-    unsigned int command_start : 1;
+    bool command_start;
 
     // Whether or not we're currently recovering from a syntax error
-    unsigned int recovering : 1;
+    bool recovering;
 
     // Whether or not the encoding has been changed by a magic comment. We use
     // this to provide a fast path for the lexer instead of going through the
     // function pointer.
-    unsigned int encoding_changed : 1;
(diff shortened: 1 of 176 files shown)


include/yarp/parser.h
```diff
@@ -390,6 +376,32 @@ struct yp_parser {
...
+    // Whether or not we're at the beginning of a command
+    unsigned int command_start : 1;
+
+    // Whether or not we're currently recovering from a syntax error
+    unsigned int recovering : 1;
+
+    // Whether or not the encoding has been changed by a magic comment. We use
+    // this to provide a fast path for the lexer instead of going through the
+    // function pointer.
+    unsigned int encoding_changed : 1;
```
eregon, line 389: Because this is performance-sensitive it seems worth to have its own 8-bits/byte.
Otherwise it means having to use an extra `& mask` whenever checking this.

kddnewton: TIL, that makes a lot of sense. Updated.

PR #1453, opened 2023-09-12
Fix behaviour of locations for comments

andyw8:
Fixes https://github.com/ruby/yarp/issues/1443

This is a breaking change: Comments will no longer include the line break. 

Co-authored with @kddnewton and @vinistock.

### DIFF
--- a/test/yarp/comments_test.rb
+++ b/test/yarp/comments_test.rb
@@ -7,7 +7,7 @@ module YARP
     def test_comment_inline
       source = "# comment"
 
-      assert_comment source, :inline, 0..9
+      assert_comment source, :inline, [0, 9, 1, 1, 0, 9]
       assert_equal [0], Debug.newlines(source)
     end
 
@@ -18,7 +18,7 @@ module YARP
       end
       RUBY
 
-      assert_comment source, :inline, 10..22
+      assert_comment source, :inline, [10, 21, 2, 2, 2, 13]
     end
 
     def test_comment___END__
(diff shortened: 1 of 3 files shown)


test/yarp/comments_test.rb
```diff
@@ -74,19 +74,27 @@ def bar
...
-    def assert_comment(source, type, location)
+    def assert_comment(source, type, locations)
+      start_offset, end_offset, start_line, end_line, start_column, end_column = locations
+
       result = YARP.parse(source)
       assert result.errors.empty?, result.errors.map(&:message).join("\n")
-      assert_equal result.comments.first.type, type
-      assert_equal result.comments.first.location.start_offset, location.begin
-      assert_equal result.comments.first.location.end_offset, location.end
+      assert_equal type, result.comments.first.type
```
andyw8, line 89: The assertions were backwards previously.

andyw8:
Looking into the failing CI jobs.

test/yarp/comments_test.rb
```diff
@@ -74,19 +75,40 @@ def bar
...
+      assert_equal("# bar method\n# bar end", method_node.location.comments.map { |c| c.location.slice }.join("\n"))
+      assert_equal("# baz invocation", call_node.location.comments.map { |c| c.location.slice }.join("\n"))
     end
 
     private
 
-    def assert_comment(source, type, location)
+    def assert_comment(source, type, locations)
+      start_offset, end_offset, start_line, end_line, start_column, end_column = locations
+      expected = {
```
andyw8, line 87: We changed to use a hash so that the failure would report the full set of differences in location, rather than stopping at the first difference.

kddnewton (approved):

PR #1465, opened 2023-09-13
Add implicit nodes for ommitted hash values

### LINKED ISSUE #1447: Request for way to know omitted hash value type
For hash value omission like `{ lvar: }`, `{ method: }`, `{ A: }`, I want to know the type of omitted value.

```ruby
YARP.parse("a=1;{a:}").value.statements.body[1]
# => HashNode(4...8)((4...5), [AssocNode(5...7)(SymbolNode(5...7)(nil, (5...6), (6...7), "a"), nil, nil)], (7...8))
YARP.parse("b=1;{a:}").value.statements.body[1]
# => HashNode(4...8)((4...5), [AssocNode(5...7)(SymbolNode(5...7)(nil, (5...6), (6...7), "a"), nil, nil)], (7...8))
```

For RubyVM::AST, it can be distinguished by `LVAR` `VCALL`.
```ruby
RubyVM::AbstractSyntaxTree.parse "a=1;{a:}"
# => 
(SCOPE@1:0-1:8
 tbl: [:a]
 args: nil
 body: (BLOCK@1:0-1:8 (LASGN@1:0-1:3 :a (LIT@1:2-1:3 1)) (HASH@1:4-1:8 (LIST@1:5-1:7 (LIT@1:5-1:7 :a) (LVAR@1:5-1:7 :a) nil))))

RubyVM::AbstractSyntaxTree.parse "b=1;{a:}"
# => 
(SCOPE@1:0-1:8
 tbl: [:b]
 args: nil
 body: (BLOCK@1:0-1:8 (LASGN@1:0-1:3 :b (LIT@1:2-1:3 1)) (HASH@1:4-1:8 (LIST@1:5-1:7 (LIT@1:5-1:7 :a) (VCALL@1:5-1:7 :a) nil))))

# Difference: (LVAR@1:5-1:7 :a) vs (VCALL@1:5-1:7 :a)
```

kddnewton:
Fixes #1447

We now always insert an implicit node into a hash value when the hash value is omitted. This should make it easier on compilers and static analysis tools to determine what is actually going on without having to look at the local tables themselves. Because we have different nodes types for these nodes, tools can still consistently handle/ignore them as they like. As a bonus, all of the implicit nodes are very small.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1413,6 +1413,19 @@ nodes:
 
           1.0i
           ^^^^
+  - name: ImplicitNode
+    fields:
+      - name: value
+        type: node
+    comment: |
+      Represents a node that is implicitly being added to the tree but doesn't
+      correspond directly to a node in the source.
+
+          { foo: }
+            ^^^^
+
+          { Foo: }
+            ^^^^
   - name: InNode
     fields:
       - name: pattern
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -2737,6 +2737,22 @@ yp_else_node_end_keyword_loc_set(yp_else_node_t *node, const yp_token_t *keyword
     node->end_keyword_loc = YP_LOCATION_TOKEN_VALUE(keyword);
 }
 
+// Allocate and initialize a new ImplicitNode node.
+static yp_implicit_node_t *
+yp_implicit_node_create(yp_parser_t *parser, yp_node_t *value) {
+    yp_implicit_node_t *node = YP_ALLOC_NODE(parser, yp_implicit_node_t);
+
+    *node = (yp_implicit_node_t) {
+        {
+            .type = YP_IMPLICIT_NODE,
+            .location = value->location
+        },
+        .value = value
+    };
+
+    return node;
+}
+
 // Allocate and initialize a new IntegerNode node.
 static yp_integer_node_t *
 yp_integer_node_create(yp_parser_t *parser, yp_node_flags_t base, const yp_token_t *token) {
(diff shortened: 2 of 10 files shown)


eregon (changes requested):
I would like at least a reply to my comment, I think it raises some valid concerns, and inf act the approach taken in this PR differs from both RubyVM::AST and parser gem, they both simply desugar as if the value was in the source and use the same node type to read a local var, etc.

kddnewton:
@eregon I'm in the middle of looking at options. It's not merged yet. I will reply once I've looked at the options sufficiently.

kddnewton:
We'll go with `ImplicitNode(CallNode)` etc.

config.yml
```diff
@@ -1413,6 +1413,19 @@ nodes:
...
+  - name: ImplicitNode
+    fields:
+      - name: value
+        type: node
+    comment: |
+      Represents a node that is implicitly being added to the tree but doesn't
+      correspond directly to a node in the source.
+
+          { foo: }
+            ^^^^
```
eregon, line 1425: This example seems not very clear, maybe:
```suggestion
          { foo: }
            ^^^^
gets translated to (assuming `foo` is a local variable):
HashNode([AssocNode(SymbolNode, ImplicitNode(LocalVariableReadNode))])
```

eregon (approved):
Thanks!
I noticed none of the snapshots contain a ImplicitNode(LocalVariableReadNode), is that example missing (would be good to add it), or maybe the logic to search for a local variable does not work?

PR #1466, opened 2023-09-13
Numbered parameters

### LINKED ISSUE #1158: Support numbered block parameters
Currently `_1`, `_2` etc in a block without explicit parameters are parsed as a method call:

```
bin/parse -e 'proc { _1.to_s }'
#<YARP::ParseResult:0x000000010cfd6e18
 @comments=[],
 @errors=[],
 @source=#<YARP::Source:0x000000010fabeab8 @offsets=[0], @source="proc { _1.to_s }">,
 @value=
  ProgramNode(0...16)(
    [],
    StatementsNode(0...16)(
      [CallNode(0...16)(
         nil,
         nil,
         (0...4),
         nil,
         nil,
         nil,
         BlockNode(5...16)(
           [],
           nil,
           StatementsNode(7...14)([CallNode(7...14)(CallNode(7...9)(nil, nil, (7...9), nil, nil, nil, nil, 0, "_1"), (9...10), (10...14), nil, nil, nil, nil, 0, "to_s")]),
           (5...6),
           (15...16)
         ),
         0,
         "proc"
       )]
    )
  ),
 @warnings=[]>
```

kddnewton:
Fixes #1158


src/yarp.c
```diff
@@ -10030,6 +10064,44 @@ parse_variable_call(yp_parser_t *parser) {
...
+            } else if (outer_scope_using_numbered_params_p(parser)) {
+                yp_diagnostic_list_append(&parser->error_list, parser->previous.start, parser->previous.end, YP_ERR_NUMBERED_PARAMETER_OUTER_SCOPE);
+            } else {
+                // Indicate that this scope is using numbered params so that
+                // child scopes cannot.
+                parser->current_scope->numbered_params = true;
+
+                // When you use a numbered parameter, it implies the existence
+                // of all of the locals that exist before it. For example,
+                // referencing _2 means that _1 must exist. Therefore here we
```
jemmaissroff, line 10083: Not sure I'm following this. Seems like this is valid Ruby, even though it outputs `nil`:

```
$ ruby -e "[1].each { p _2 }" 
nil
```

src/yarp.c
```diff
@@ -10030,6 +10064,44 @@ parse_variable_call(yp_parser_t *parser) {
             return (yp_node_t *) yp_local_variable_read_node_create(parser, &parser->previous, (uint32_t) depth);
         }
 
+        if (!parser->current_scope->closed && token_is_numbered_parameter(parser->previous.start, parser->previous.end)) {
+            // Now that we know we have a numbered parameter, we need to check
+            // if it's allowed in this context. If it is, then we will create a
+            // local variable read. If it's not, then we'll create a normal call
+            // node but add an error.
+            if (parser->current_scope->explicit_params) {
```
jemmaissroff, line 10072: Should we store information about the current scope not being able to have children with numbered params in all of these cases and not only the happy path? We already know that the current scope can't have children with scoped params. But any children will still have to look past this scope to see that, even though we know the information now. 

I know this will only happen in the error case, and am unsure how much we're trying to optimize that / error recovery.

jemmaissroff (changes requested):

src/yarp.c
kddnewton: Yup! When you reference `_2` in a block without explicit parameters, it adds `_1` and `_2` to the local table.

src/yarp.c
kddnewton: Sorry I'm not sure I'm following. There's two errors this thing is catching. One is:

```ruby
foo { || _1 }
```

is disallowed because of the combination of explicit and implicit parameters. The other is:

```ruby
foo { _1; bar { _1 } }
```

which is that you can't use implicit parameters in a block that is inside another scope that is using implicit parameters. We need both sets of information to stop both of these.

I'm not sure if that answered your question.

kddnewton: @jemmaissroff I see what you're saying now. We cache this information on the current scope and check it here: https://github.com/ruby/yarp/pull/1466/files#diff-71ae829a627932e547cf825850d9a6b4773b009b0b68ae8eb1d8ef02e6d98987R10126. So that function bails early.

kddnewton: Wait no, that isn't what you were saying. Fixing now.

jemmaissroff (approved):

PR #1477, opened 2023-09-13
Run `sudo apt-get update` before `sudo apt-get install`

eregon:
* As recommended in https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners
* Fixes https://github.com/ruby/yarp/issues/1476


kddnewton:
This is already addressed on `main`

PR #1478, opened 2023-09-13
free strings in the ruby extension

froydnj:
Noticed this while trying to valgrind a `YARP::Debug.profile_file` loop with a bunch of files.


kddnewton (approved):
Nice catch thanks

PR #1486, opened 2023-09-14
Attempting to fix valgrind caching

kddnewton:


.github/workflows/main.yml
```diff
@@ -174,12 +174,15 @@ jobs:
...
+    - uses: actions/checkout@v3
     - name: Cache valgrind
       uses: actions/cache@v3
+      id: cache-valgrind
       with:
         path: valgrind-3.20.0
         key: ${{ runner.os }}-valgrind-3.20.0
-    - name: Install valgrind from source
+    - name: Download valgrind
+      if: steps.cache-valgrind.outputs.cache-hit != 'true'
```
eregon, line 185: This will mean `libc6-dbg` might not be installed (on cache hit).
Is it only needed for `./configure` and `make`, or also at runtime?
If the latter, I think best to just move the apt-get stuff to another step.

eregon (approved):
Right, `actions/checkout@v3` wipes the CWD, so this LGTM

.github/workflows/main.yml
```diff
@@ -188,8 +191,10 @@ jobs:
         cd valgrind-3.20.0
```
eregon, line 191: Something that might speed up the job is remove the `v` from `tar xvf`, because that produces a lot of lines and I suspect it's not cheap to transfer thousands of lines.

.github/workflows/main.yml
kddnewton: Yeah I ended up having to. I thought it was possible the built artifact would contain the necessary functions, but apparently not.

.github/workflows/main.yml
kddnewton: Removed

PR #1493, opened 2023-09-14
Convert constant pool entries to a string type in Java Loader

eregon:
* Fixes https://github.com/ruby/yarp/issues/1389
* Fields of type `string` as kept as byte[].

@enebo Could you try if this works for JRuby with RubySymbol?
You'll need to subclass Loader to implement the byte[]->RubySymbol conversion.

I still need to check how this works with TruffleRuby but it looks pretty good.


templates/java/org/yarp/Loader.java.erb
```diff
@@ -15,38 +15,59 @@ public class Loader {
...
+
+    public Charset getEncodingCharset(String encodingName) {
+        encodingName = encodingName.toLowerCase(Locale.ROOT);
+        if (encodingName.equals("ascii-8bit")) {
+            return StandardCharsets.US_ASCII;
+        }
+        return Charset.forName(encodingName);
+    }
+
+    public <%= JAVA_STRING_TYPE %> bytesToString(byte[] bytes) {
```
eregon, line 28: Maybe it should be `bytesToIdentifier()` or so and `JAVA_IDENTIFIER_TYPE` but that's so long.

eregon:
This works nicely in TruffleRuby: https://github.com/eregon/truffleruby/compare/bd/adopt-yarp-20230914%5E...eregon:truffleruby:bd/adopt-yarp-20230914

eregon:
This is ready but needs https://github.com/ruby/yarp/pull/1450 to be merged first.

templates/java/org/yarp/Loader.java.erb
eregon: I renamtes to `bytesToName()`.
Maybe `JAVA_NAME_TYPE` would be good for the constant?

templates/template.rb
```diff
@@ -348,11 +351,10 @@ def locals
...
 
-          {
-            nodes: config.fetch("nodes").map { |node| NodeType.new(node) }.sort_by(&:name),
-            tokens: config.fetch("tokens").map { |token| Token.new(token) },
-            flags: config.fetch("flags").map { |flags| Flags.new(flags) }
-          }
+          nodes = config.fetch("nodes").map { |node| NodeType.new(node) }.sort_by(&:name)
+          tokens = config.fetch("tokens").map { |token| Token.new(token) }
+          flags = config.fetch("flags").map { |flags| Flags.new(flags) }
+          binding
```
kddnewton, line 357: I'm not keen to expose the entire binding to the template files, I'd much rather keep the scope of what they can access limited. Can you revert this change?

eregon: The only extra thing available in this binding is `config` which is not used anyway (and we could hide it if we care about that one), and having a constant scope of `YARP` which is quite a bit more convenient.
Is that the concern?

I can use `YARP::JAVA_STRING_TYPE` otherwise, but it is really verbose for every place I need the type.

kddnewton: Yeah we do that with the other constants, I'd rather keep it consistent.

eregon: Actually I'd be happy to remove all those `YARP::` prefixes in templates for field classes, they used to not be there and IMO it was more a little bit more readable that way. But I didn't want to do that big change in this PR.

I'll just undo this part, we can always do it later and should be its own PR anyway for consistency.

eregon: I reverted those changes.

kddnewton (approved):

PR #1497, opened 2023-09-14
force constant pool capacity to be a power of 2

froydnj:
...so that we can use cheap masks for resolving hash table insertions instead of expensive modulo.

This change improves the performance of:

```ruby
debug = YARP.const_get(:Debug)
files.each do |file|
  debug.profile_file(file)
end
```

by about 5% on a subset of Stripe's codebase.  Some of this might just be due to having slightly higher default constant pool sizes that don't requiring resizing.


kddnewton (approved):
Nice, this is great.

PR #1498, opened 2023-09-14
remove `ruby_memcheck` from the Gemfile

froydnj:
`ruby_memcheck` doesn't appear to be used and not requiring it avoids transitively dragging in `nokogiri` as well as a couple of other things.  But really it avoids `nokogiri`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,5 +7,4 @@ gemspec
 gem "rake"
 gem "rake-compiler"
 gem "test-unit"
-gem "ruby_memcheck", platform: %i[mri truffleruby mswin mingw x64_mingw]
 gem "ffi", platform: %i[mri mswin mingw x64_mingw]
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -7,17 +7,10 @@ GEM
   remote: https://rubygems.org/
   specs:
     ffi (1.15.5)
-    mini_portile2 (2.8.4)
-    nokogiri (1.15.3)
-      mini_portile2 (~> 2.8.2)
-      racc (~> 1.4)
     power_assert (2.0.3)
-    racc (1.7.1)
     rake (13.0.6)
     rake-compiler (1.2.5)
       rake
-    ruby_memcheck (2.0.1)
-      nokogiri
     test-unit (3.6.1)
       power_assert
 
@@ -30,7 +23,6 @@ DEPENDENCIES
   ffi
   rake
   rake-compiler
-  ruby_memcheck
   test-unit
   yarp!
 


froydnj:
Argh, ignore me, apparently my editor's `rg` wasn't getting invoked with the proper options. :(

seven1m:
Somewhat related, we're going to some decent lengths to avoid running `bundle install` for YARP in the Natalie project. We store the generated (templated) sources and overlay them [here](https://github.com/natalie-lang/natalie/blob/5278d83e9934eee95d8526aac706f01f45610201/Rakefile#L153-L171). We're also not using the `rake compile` command so we can avoid the need for rake-compiler. We do that [here](https://github.com/natalie-lang/natalie/blob/5278d83e9934eee95d8526aac706f01f45610201/Rakefile#L477-L478)

Nokogiri is a beast and not installing it is a high priority for me. 😄 

All that to say, the dependencies used for YARP are avoidable, and I wonder if there's a future where those dependencies could be optional: used by YARP developers but not required by library consumers.

eregon:
I think for Ruby impls it's kind of expected you vendor yarp and tweak things and don't build with `rake`.
FWIW https://github.com/ruby/yarp/blob/main/docs/build_system.md documents this fairly extensively (feel free to add Natalie there if you like).

What could be done I think is to move `ruby_memcheck` to a `:development` or `:memcheck` or `:extras` Bundler group or so.
Then it would be easy to to skip installing those with `bundle install --without memcheck`.

And rescue + return in LoadError from `require "ruby_memcheck"`: https://github.com/ruby/yarp/blob/7c38d6d465021526262fa9ef2dca7322103d703d/rakelib/test.rake#L20

seven1m:
Thanks for pointing out that build_system.md doc. I don't know why I didn't consider I can run templates.rb via Natalie to generate the templated sources. That's great. ❤️

seven1m:
Thanks @eregon, that makes sense. I had missed that build_system.md doc, so that's helpful.

eregon:
> I don't know why I didn't consider I can run templates.rb via Natalie to generate the templated sources.

@seven1m FWIW TruffleRuby doesn't do that, it runs `bundle exec rake templates` on CRuby.
Because well, you know, it would need a parser to run `rake` but in the future we would only have YARP as the parser for TruffleRuby.

seven1m:
Bootstrapping a Ruby kinda breaks my brain ... but I wonder how MRI will do it? Will the templates be vendored like we do with Natalie?

kddnewton:
@seven1m CRuby doesn't actually use the Ruby library, it only uses the C API. So it's effectively just another make target for them to template. That's why we have this weird CI check: https://github.com/ruby/yarp/blob/main/.github/workflows/main.yml#L26-L39 to ensure we don't use Ruby syntax > 2.5 in the templating code.

The other CRuby developers disagree with @eregon's take here because by having it integrated, it means they can change things within the CRuby codebase without having to reintegrate. This is going to be particularly important going forward because PRs will eventually reverse-sync from CRuby as people need to make changes to syntax.

Vendoring is totally an option, and I think makes a lot of sense for a project like Natalie that isn't going to be actively developing YARP itself.

As for the gem, I think it would be fine to do what @eregon is suggesting here and only install those gems optionally.

Issue #1501, opened 2023-09-15
`case 1; end` should be rejected


mame:
A `case` clause with no `when` clause must be rejected.

```
$ ./local/bin/irb -ryarp
irb(main):001> RubyVM::AbstractSyntaxTree.parse("case 1; end")
<internal:ast>:59:in `parse': syntax error, unexpected `end', expecting `when' (SyntaxError)
case 1; end
        ^~~

        from (irb):1:in `<main>'
        from <internal:kernel>:187:in `loop'
        from /home/mame/work/ruby/local/lib/ruby/gems/3.3.0+0/gems/irb-1.8.1/exe/irb:9:in `<top (required)>'
        from ./local/bin/irb:25:in `load'
        from ./local/bin/irb:25:in `<main>'
irb(main):002> YARP.parse("case 1; end")
=>
#<YARP::ParseResult:0x00007fbe4a1037a8
 @comments=[],
 @errors=[],
 @source=#<YARP::Source:0x00007fbe49f2d7d0 @offsets=[0], @source="case 1; end">,
 @value=
  @ ProgramNode (location: (0...11))
├── locals: []
└── statements:
    @ StatementsNode (location: (0...11))
    └── body: (length: 1)
        └── @ CaseNode (location: (0...11))
            ├── predicate:
            │   @ IntegerNode (location: (5...6))
            │   └── flags: decimal
            ├── conditions: (length: 0)
            ├── consequent: ∅
            ├── case_keyword_loc: (0...4) = "case"
            └── end_keyword_loc: (8...11) = "end",
 @warnings=[]>
```

seven1m:
Similarly, this is accepted by YARP but not by MRI:

```ruby
case 1
else
  2
end
```

PR #1518, opened 2023-09-15
forbid empty `case` conditions list

shaver:
Resolves issue #1501: a `case` clause with no `when` clause must be rejected

First yarp patch, please lmk what I messed up!


shaver:
Closing in favour of PR#1508!

PR #1519, opened 2023-09-15
fix computation of memsize for node lists

froydnj:
The first bit ensures that we don't double-count, the second bit ensures that we actually do count. 😅


templates/src/node.c.erb
```diff
@@ -14,7 +14,9 @@ yp_node_memsize_node(yp_node_t *node, yp_memsize_t *memsize);
 // Calculate the size of the node list in bytes.
 static size_t
 yp_node_list_memsize(yp_node_list_t *node_list, yp_memsize_t *memsize) {
-    size_t size = sizeof(yp_node_list_t) + (node_list->capacity * sizeof(yp_node_t *));
```
kddnewton, line 17: So the reasons these functions are here is so we can eventually use the typeddata API. In the event a node list isn't directly embedded on a node, this will now underestimate the size.

I think I'd rather us keep this in place, but then subtract it out from the node memsize calculation. Either that or provide some kind of _internal_memsize function that this calls.

(Also we're going to have the same problem with the constant list.

kddnewton (changes requested):

templates/src/node.c.erb
froydnj: > So the reasons these functions are here is so we can eventually use the typeddata API. In the event a node list isn't directly embedded on a node, this will now underestimate the size.

Do you mean that you want to move to a world where nodes merely have a pointer to the node list, for space reasons (nodes become much smaller!), and for Ruby API reasons?  Something like (at least at the C level, I didn't touch the Ruby API level):

https://github.com/ruby/yarp/compare/main...froydnj:froydnj-smaller-node-lists?expand=1

(This change appears to be slightly faster (~1%, if that) than the current setup and basically equal wrt memory usage.)

...and then the Ruby API exposes typeddata that borrows a pointer to the node list and a reference to the owning node?

In any event, I think subtracting out the node list size from the node upfront is probably the easiest way to handle this.

froydnj:
I don't love the commit I just pushed, but I think it's slightly better than the version where you have:

```erb
            memsize->memsize += sizeof(*cast) - sizeof(yp_node_list_t) * <%= node.fields.count { |f| f.is_a?(YARP::NodeListField) };
```

...and then you have a bunch of `- sizeof(yp_node_list_t) * 0` for nodes that don't have node lists.  I don't feel at all strongly about that, though, and happy to change it out, which would make the ERB simpler.

(I am also not an ERB expert, so there may be a better way to do this.)

froydnj:
Ahhh...I should just fix constant id lists while I'm here too, I'll push a patch for that in a bit.

templates/src/node.c.erb
kddnewton: It's an option I'd like to consider in the future. At the moment we're too far away from typeddata because we also support the FFI backend, but I think it could be a big win for initial performance and consumers that are looking at a subset of the tree.

PR #1523, opened 2023-09-15
Add node event dispatcher

vinistock:
This commit changes the node template to create a dispatcher class, which can be used to walk an AST an emit events to all registered listeners.

This is useful when performing several types of analysis on the same AST to avoid having to perform several rounds of visits. Two main examples are:

- LSPs: each different request can be a listener, which contains logic specific only to that request
- Linters: each linter rule can be a listener that also encapsulates logic only related to itself

For both of them, this allows the complete analysis to be completed in a single round of visits, saving numerous method dispatches and improving performance.

### DIFF
--- a/templates/lib/yarp/node.rb.erb
+++ b/templates/lib/yarp/node.rb.erb
@@ -48,6 +48,31 @@ module YARP
       }.compact.join(", ") %>]
     end
 
+    # def compact_child_nodes: () -> Array[Node]
+    def compact_child_nodes
+      <%- if node.fields.any? { |field| field.is_a?(YARP::OptionalNodeField) } -%>
+      compact = []
+      <%- node.fields.each do |field| -%>
+      <%- case field -%>
+      <%- when YARP::NodeField -%>
+      compact << <%= field.name %>
+      <%- when YARP::OptionalNodeField -%>
+      compact << <%= field.name %> if <%= field.name %>
+      <%- when YARP::NodeListField -%>
+      compact.concat(<%= field.name %>)
+      <%- end -%>
+      <%- end -%>
+      compact
+      <%- else -%>
+      [<%= node.fields.map { |field|
+        case field
+        when YARP::NodeField then field.name
+        when YARP::NodeListField then "*#{field.name}"
+        end
+      }.compact.join(", ") %>]
+      <%- end -%>
+    end
+
     # def comment_targets: () -> Array[Node | Location]
     def comment_targets
       [<%= node.fields.map { |field|
@@ -136,6 +161,13 @@ module YARP
       <%- end -%>
       inspector.to_str
     end
+
+    # Returns a symbol representation of the type of node.
+    #
+    # def human: () -> Symbol
+    def human
+      :<%= node.human %>
+    end
   end
 
   <%- end -%>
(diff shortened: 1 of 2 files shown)


test/yarp/dispatcher_test.rb
```diff
@@ -0,0 +1,40 @@
...
+        def foo
+          something(1, 2, 3)
+        end
+      RUBY
+
+      dispatcher.dispatch(root)
+
+      assert_equal([:call_node_enter, :call_node_leave], @listener.events_received)
+    end
+  end
```
kddnewton, line 39:
```suggestion
    class TestListener
      attr_reader :events_received

      def initialize
        @events_received = []
      end

      def call_node_enter(node)
        events_received << :call_node_enter
      end

      def call_node_leave(node)
        events_received << :call_node_leave
      end
    end

    def test_dispatching_events
      listener = TestListener.new

      dispatcher = Dispatcher.new
      dispatcher.register(listener, :call_node_enter, :call_node_leave)

      root = YARP.parse(<<~RUBY).value
        def foo
          something(1, 2, 3)
        end
      RUBY

      dispatcher.dispatch(root)
      assert_equal([:call_node_enter, :call_node_leave], listener.events_received)
    end
  end
```

kddnewton (approved):

templates/lib/yarp/node.rb.erb
```diff
@@ -48,6 +48,31 @@ module YARP
       }.compact.join(", ") %>]
     end
 
+    # def compact_child_nodes: () -> Array[Node]
+    def compact_child_nodes
+      <%- if node.fields.any? { |field| field.is_a?(YARP::OptionalNodeField) } -%>
```
eregon, line 53: Another possibility would be to use `[node_field, *optional_node_field, *node_list_field]` because there is no `#to_a` on Node and `*nil` is `[]`.
Not sure if faster or slower though, it would make the logic here a bit simpler.

templates/lib/yarp/node.rb.erb
```diff
@@ -157,6 +189,55 @@ module YARP
...
+
+      while (node = queue.shift)
+        case node.human
+        <%- nodes.each do |node| -%>
+        when :<%= node.human %>
+          listeners[:<%= node.human %>_enter]&.each { |listener| listener.<%= node.human %>_enter(node) }
+          queue = node.compact_child_nodes.concat(queue)
+          listeners[:<%= node.human %>_leave]&.each { |listener| listener.<%= node.human %>_leave(node) }
+        <%- end -%>
+        end
```
eregon, line 223: These big `case/when/end` statements (`dispatch` is 570 lines) are pretty bad for optimizing JITs like on TruffleRuby and JRuby, it's likely too much code in a single method to compile or optimize well, somewhat similar to https://github.com/whitequark/parser/issues/871.
Also this relies very heavily on `case expr; when :symbol; when :symbol2; ...; end` to use a Hash internally, otherwise it would be very slow as it would test each Symbol in order. I think that's the case on CRuby but on no other Ruby implementation.

I think something that would be fast for all Rubies and behave better for JIT would be to use a Hash of node.human to Procs containing the body of each `when` (one per node type to avoid needing `listerner.send` which would be slow).

templates/lib/yarp/node.rb.erb
```diff
@@ -157,6 +189,55 @@ module YARP
...
+    # def dispatch: (Node) -> void
+    def dispatch(root)
+      queue = [root]
+
+      while (node = queue.shift)
+        case node.human
+        <%- nodes.each do |node| -%>
+        when :<%= node.human %>
+          listeners[:<%= node.human %>_enter]&.each { |listener| listener.<%= node.human %>_enter(node) }
+          queue = node.compact_child_nodes.concat(queue)
```
eregon, line 220: Is this the correct order?
Maybe it should be `queue.concat(node.compact_child_nodes)`?

It looks like this adds child nodes before the current elements in the array (so on the "left side" i.e. near 0 index) and `queue.shift` above removes from the "left side" too.
So it looks like it behaves somewhat like a stack:
```ruby
q = [1]
q.shift

q = [2, 3].concat(q)
p q # => [2, 3]
p q.shift # => 2

q = [4].concat(q)
p q # => [4, 3]
p q.shift # => 4, the last added element is the first out in this case
# Not fully LIFO though because if it was `q = [4, 5].concat(q)` above, then 4 is out before 5
```

PR #1528, opened 2023-09-16
fix: handling escaped whitespace in a %w list

### LINKED ISSUE #1505: Different 'unescaped' result for %w[] vs regular string
```ruby
irb(main):001:0> YARP.parse('"foo\ bar"').value.statements.body.first.unescaped
=> "foo bar"
irb(main):002:0> YARP.parse('%w[foo\ bar]').value.statements.body.first.elements.first.unescaped
=> "foo\\ bar"
```

flavorjones:
Introduces a new flavor of unescaping, YP_UNESCAPE_WHITESPACE, which is the same as MINIMAL but also unescapes whitespace.

Note that a spanning_heredoc.txt fixture test is updated to be less wrong, but YARP's behavior doesn't yet fully match Ruby in this case.

Fixes #1505

### DIFF
--- a/ext/yarp/extension.c
+++ b/ext/yarp/extension.c
@@ -491,6 +491,12 @@ unescape_minimal(VALUE self, VALUE source) {
     return unescape(source, YP_UNESCAPE_MINIMAL);
 }
 
+// Escape the given string minimally plus whitespace. Returns the unescaped string.
+static VALUE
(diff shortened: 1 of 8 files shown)


flavorjones:
The spanning_heredoc issue is https://github.com/ruby/yarp/issues/1529

PR #1549, opened 2023-09-19
Print locations using line / col instead of offsets

jemmaissroff:
This commit changes printing of locations to use the format (start_line,start_column)-(end_line,end_column) instead of using offsets.


lib/yarp.rb
```diff
@@ -71,7 +71,7 @@ def copy(**options)
 
     # Returns a string representation of this location.
     def inspect
-      "#<YARP::Location @start_offset=#{@start_offset} @length=#{@length}>"
+      "#<YARP::Location @start_offset=#{@start_offset} @length=#{@length} start_line=#{start_line}>"
```
jemmaissroff, line 74: This change I'm slightly unsure of, since `start_line` is not an instance variable. However, this output seems helpful to me for cases like printing error messages.

kddnewton: It's only debug code, I think it's fine.

kddnewton (approved):
Just needs a rebase

eregon:
A more concise format that RubyVM::AST uses is `1:0-1:9`, but I think the format in this PR is easier to read :+1: 

Columns are currently in number of bytes, that seems the case for RubyVM::AST too.
I think it's fine internally but e.g. in an editor it might not match for a UTF-8 file with non-ASCII characters.

kddnewton:
@jemmaissroff could you look into the memcheck failure? Weird that it's failing on this, but I'm hoping it's surfacing something we can fix. I've found spin to be the best way to get access to valgrind.

PR #1551, opened 2023-09-19
Ignore parsing specific files on TruffleRuby to make CI green

jemmaissroff:
CI is currently failing due to an issue with Ripper on the latest TruffleRuby version. This commit removes the offending tests from running, to ensure CI is green again.

### DIFF
--- a/test/yarp/parse_test.rb
+++ b/test/yarp/parse_test.rb
@@ -58,6 +58,9 @@ module YARP
       # These fail on TruffleRuby due to a difference in Symbol#inspect: :测试 vs :"测试"
       next if RUBY_ENGINE == "truffleruby" and %w[seattlerb/bug202.txt seattlerb/magic_encoding_comment.txt].include?(relative)
 
+      # These fail on TruffleRuby due to a Ripper difference
+      next if RUBY_ENGINE == "truffleruby" and %w[symbols.txt unparser/corpus/literal/def.txt].include?(relative)
+
       filepath = File.join(base, relative)
       snapshot = File.expand_path(File.join("snapshots", relative), __dir__)
 


kddnewton (approved):

jemmaissroff:
cc @eregon

test/yarp/parse_test.rb
```diff
@@ -58,6 +58,9 @@ def test_parse_lex_file
       # These fail on TruffleRuby due to a difference in Symbol#inspect: :测试 vs :"测试"
       next if RUBY_ENGINE == "truffleruby" and %w[seattlerb/bug202.txt seattlerb/magic_encoding_comment.txt].include?(relative)
 
+      # These fail on TruffleRuby due to a Ripper difference
+      next if RUBY_ENGINE == "truffleruby" and %w[symbols.txt unparser/corpus/literal/def.txt].include?(relative)
```
eregon, line 62: This skips all other asserts for those files, but we already have `ripper_should_match` and `ripper_enabled` for finer-grained excludes.
Like I mentioned on some other PR, I think we should only use Ripper in tests on CRuby, consider it broken or incompatible on non-CRuby because it's too hacky.

eregon: Ah `ripper_enabled = RUBY_VERSION >= "3.2.0"` above, that used to be false before the truffleruby 23.1.0 release.
I suggest to change that to `ripper_enabled = RUBY_ENGINE == "ruby" && RUBY_VERSION >= "3.2.0"`

eregon: https://github.com/ruby/yarp/pull/1553

Issue #1537, opened 2023-09-16
Copy & embed any string/constant pool entry in the serialized output


eregon:
I think we should copy & embed any string/constant pool entry in the serialized output and therefore not need the source at all when deserializing.
That would mean larger serialized size but no need to read the source at all which could be a nice speedup in the case the serialized form is saved on disk and newer/equal the source file.
I think the serialized size wouldn't be worse than before + eager loading constant pool and strings, it would be about the same, but gaining to not have to read the source.
That might actually be fairly easy, and would be less temporal memory than serialized + source at the same time in memory (but that's not a big deal).

This would also remove the need for accessing `owned` when serializing the constant pool: https://github.com/ruby/yarp/pull/1449#issuecomment-1722241146

_Originally posted by @eregon in https://github.com/ruby/yarp/issues/1479#issuecomment-1719184286_

enebo:
There are two levels here to consider: loaded_tree and serialized_blob.

For loaded tree the only source we depend on at this point is Integer, Float, and Rational.  They need to be moved into the tree and not loaded from source.  We just haven't made a decision on how to do this.  Keeping source a round AND the tree is a huge memory cost over the legacy parser.  I assume we 100% will be doing this soon.  We are so close to not having source here.

For serialization I think there is a pretty big carrot to moving constants into serialization: pre-compilation.  It would be nice to eventually have a stable serialized format we can load and not bother to load from source at all (assuming it is trusted).  We have plans to replace our AOT serialization format with this since Loader requires no C.

As @eregon said we both do lazy method loading and whatever allows us to use as little memory as possible (also helps for startup time).    Ultimately, what we would have for methods would be a slice of the constant pool and a slice of the offsets representing the method body (whether we actually slice this or keep whole serialized blob is up to experimentation).  The time saved will be not loading huge portions of the tree until it is used.  As some Rails commands I have see up to 80% of all method definitions loaded are not called.

The last bit also mentioned by @eregon but he only mentions memory...it is faster to make this pool in C than have us create it in Java when Java is just starting up and still somewhat cold.  If we have to depend on C we may as well have C do as much heavy lifting for us as we can get.

@kddnewton I have said this many times now but there is a pretty large dual concern in prism and we keep trying to compromise to serve both.  I think it is still working but there will always be some tradeoff so long as we use a single format.  A larger serialized format is less memory for compilation but more memory for syntax consumers.  

[...]

PR #1554, opened 2023-09-19
make `yp_constant_id_t`s useful for constant pool indexing

froydnj:
This turns out to be really useful (#1537, #1533, #1449, etc.), and we can make everything smaller while doing so.


kddnewton:
Performance/memory impact?

kddnewton (commented):
Great work. I'll need to merge this into ruby/ruby first to ensure I don't break CI since this is a breaking change to the C API, then I'll merge this.

kddnewton:
Merged this in #1605

PR #1556, opened 2023-09-20
Do not allow trailing commas in calls without parenthesis

### LINKED ISSUE #1548: Ruby rejects `foo 1,` but YARP accepts it
```
$ ./local/bin/irb -ryarp
irb(main):001> RubyVM::AbstractSyntaxTree.parse("foo 1,")
<internal:ast>:59:in `parse': syntax error, unexpected end-of-input (SyntaxError)
foo 1,
      ^

        from (irb):1:in `<main>'
irb(main):002> YARP.parse("foo 1,")
=>
#<YARP::ParseResult:0x00007fe4d95a22b8
 @comments=[],
 @errors=[],
 @source=#<YARP::Source:0x00007fe4d93f9218 @offsets=[0], @source="foo 1,">,
 @value=
  @ ProgramNode (location: (0...5))
├── locals: []
└── statements:
    @ StatementsNode (location: (0...5))
    └── body: (length: 1)
        └── @ CallNode (location: (0...5))
            ├── receiver: ∅
            ├── call_operator_loc: ∅
            ├── message_loc: (0...3) = "foo"
            ├── opening_loc: ∅
            ├── arguments:
            │   @ ArgumentsNode (location: (4...5))
            │   └── arguments: (length: 1)
            │       └── @ IntegerNode (location: (4...5))
            │           └── flags: decimal
            ├── closing_loc: ∅
            ├── block: ∅
            ├── flags: ∅
            └── name: "foo",
 @warnings=[]>
```

haldun:
Closes #1548

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -11921,6 +11921,13 @@ parse_arguments_list(pm_parser_t *parser, pm_arguments_t *arguments, bool accept
         // argument to this method call.
         parse_arguments(parser, arguments, true, PM_TOKEN_EOF);
 
+        // If we have done with the arguments and still not consumed the comma,
+        // then we have a trailing comma where we need to check whether it is
+        // allowed or not.
+        if (parser->previous.type == PM_TOKEN_COMMA && !match1(parser, PM_TOKEN_SEMICOLON)) {
+            pm_parser_err_previous(parser, PM_ERR_EXPECT_ARGUMENT);
+        }
+
         pm_accepts_block_stack_pop(parser);
     }
 
--- a/test/prism/errors_test.rb
+++ b/test/prism/errors_test.rb
@@ -1676,6 +1676,12 @@ module Prism
       ], compare_ripper: false # Ripper does not check 'void value expression'.
     end
 
+    def test_trailing_comma_in_calls
+      assert_errors expression("foo 1,"), "foo 1,", [
+        ["Expected an argument", 5..6]
+      ]
+    end
+
     private
 
     def assert_errors(expected, source, errors, compare_ripper: RUBY_ENGINE == "ruby")


src/yarp.c
```diff
@@ -9771,7 +9776,7 @@ parse_arguments_list(yp_parser_t *parser, yp_arguments_t *arguments, bool accept
         // If we get here, then the subsequent token cannot be used as an infix
         // operator. In this case we assume the subsequent token is part of an
         // argument to this method call.
-        parse_arguments(parser, arguments, true, YP_TOKEN_EOF);
+        parse_arguments(parser, arguments, true, YP_TOKEN_EOF, false);
```
kddnewton, line 9779: Instead of passing this argument here, could we instead put the check for the trailing comma here? That way it wouldn't be inside `parse_arguments` at all and we could just put it where you're currently passing `false`

kddnewton (changes requested):

src/yarp.c
```diff
@@ -9773,6 +9773,12 @@ parse_arguments_list(yp_parser_t *parser, yp_arguments_t *arguments, bool accept
         // argument to this method call.
         parse_arguments(parser, arguments, true, YP_TOKEN_EOF);
 
+        // If we have done with the arguments and still not consumed the comma, then we have a trailing comma
+        // where we need to check whether it is allowed or not.
+        if (parser->previous.type == YP_TOKEN_COMMA && !match1(parser, YP_TOKEN_SEMICOLON)) {
```
kddnewton, line 9778: I think the check for semicolon here would allow `foo 1,;`. Is that intentional?

src/yarp.c
```diff
@@ -9773,6 +9773,12 @@ parse_arguments_list(yp_parser_t *parser, yp_arguments_t *arguments, bool accept
...
+            yp_diagnostic_list_append(&parser->error_list, parser->previous.start, parser->previous.end, YP_ERR_EXPECT_ARGUMENT);
+        }
```
kddnewton, line 9780:
```suggestion
        // If we have done with the arguments and still not consumed the comma,
        // then we have a trailing comma where we need to check whether it is
        // allowed or not.
        if (parser->previous.type == PM_TOKEN_COMMA && !match1(parser, PM_TOKEN_SEMICOLON)) {
            pm_parser_err_previous(parser, PM_ERR_EXPECT_ARGUMENT);
        }
```

kddnewton: Sorry this took so long, but yeah this looks good. I've added a suggestion here for how to update it since it has been so long.

kddnewton (changes requested):

kddnewton (approved):

PR #1560, opened 2023-09-20
Do not allow BEGIN except the toplevel

### LINKED ISSUE #1540: Ruby rejects `def foo; BEGIN {}; end` but YARP accepts it
`BEGIN {}` is permitted only at toplevel.

It might be a bit arguable if YARP should do this check because this seems to be semantic.  However, I think that it would be more convenient to have this check at YARP than to have each of the various tools that use YARP implement this check.

```
$ ./local/bin/irb -ryarp
irb(main):001> RubyVM::AbstractSyntaxTree.parse("def foo; BEGIN {}; end")
<internal:ast>:59:in `parse': BEGIN is permitted only at toplevel (SyntaxError)
def foo; BEGIN {}; end
         ^~~~~

        from (irb):1:in `<main>'
irb(main):002> YARP.parse("def foo; BEGIN {}; end")
=>
#<YARP::ParseResult:0x00007f7b471751f0
 @comments=[],
 @errors=[],
 @source=#<YARP::Source:0x00007f7b46e8f940 @offsets=[0], @source="def foo; BEGIN {}; end">,
 @value=
  @ ProgramNode (location: (0...22))
├── locals: []
└── statements:
    @ StatementsNode (location: (0...22))
    └── body: (length: 1)
        └── @ DefNode (location: (0...22))
            ├── name: :foo
            ├── name_loc: (4...7) = "foo"
            ├── receiver: ∅
            ├── parameters: ∅
            ├── body:
            │   @ StatementsNode (location: (9...17))
            │   └── body: (length: 1)
            │       └── @ PreExecutionNode (location: (9...17))
            │           ├── statements: ∅
            │           ├── keyword_loc: (9...14) = "BEGIN"
            │           ├── opening_loc: (15...16) = "{"
            │           └── closing_loc: (16...17) = "}"
            ├── locals: []
            ├── def_keyword_loc: (0...3) = "def"
            ├── operator_loc: ∅
            ├── lparen_loc: ∅
            ├── rparen_loc: ∅
            ├── equal_loc: ∅
            └── end_keyword_loc: (19...22) = "end",
 @warnings=[]>
```

haldun:
Closes #1540


src/yarp.c
```diff
@@ -12137,6 +12137,9 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
             yp_statements_node_t *statements = parse_statements(parser, YP_CONTEXT_PREEXE);
 
             expect1(parser, YP_TOKEN_BRACE_RIGHT, YP_ERR_BEGIN_UPCASE_TERM);
+            if ((parser->current_context->context != YP_CONTEXT_MAIN) && (parser->current_context->context != YP_CONTEXT_PREEXE)) {
```
kddnewton, line 12140: Can you extract out a variable for the `context` so we're not accessing it twice?

kddnewton (changes requested):

kddnewton (approved):

kddnewton:
This one needs a rebase as well

PR #1564, opened 2023-09-20
Check for bang and question mark in parameter name

### LINKED ISSUE #1563: Ruby rejects `def foo(x!); end` but YARP accepts it!
```
$ ./local/bin/irb -ryarp
irb(main):001> RubyVM::AbstractSyntaxTree.parse("def foo(x!); end")
<internal:ast>:59:in `parse': syntax error, unexpected method, expecting ')' (SyntaxError)
def foo(x!); end
        ^~

        from (irb):1:in `<main>'
irb(main):002> YARP.parse("def foo(x!); end")
=>
#<YARP::ParseResult:0x00007fc094bdc368
 @comments=[],
 @errors=[],
 @source=#<YARP::Source:0x00007fc094ba6b28 @offsets=[0], @source="def foo(x!); end">,
 @value=
  @ ProgramNode (location: (0...16))
  ├── locals: []
  └── statements:
      @ StatementsNode (location: (0...16))
      └── body: (length: 1)
          └── @ DefNode (location: (0...16))
              ├── name: :foo
              ├── name_loc: (4...7) = "foo"
              ├── receiver: ∅
              ├── parameters:
              │   @ ParametersNode (location: (8...10))
              │   ├── requireds: (length: 1)
              │   │   └── @ RequiredParameterNode (location: (8...10))
              │   │       └── name: :x!
              │   ├── optionals: (length: 0)
              │   ├── rest: ∅
              │   ├── posts: (length: 0)
              │   ├── keywords: (length: 0)
              │   ├── keyword_rest: ∅
              │   └── block: ∅
              ├── body: ∅
              ├── locals: [:x!]
              ├── def_keyword_loc: (0...3) = "def"
              ├── operator_loc: ∅
              ├── lparen_loc: (7...8) = "("
              ├── rparen_loc: (10...11) = ")"
              ├── equal_loc: ∅
              └── end_keyword_loc: (13...16) = "end",
 @warnings=[]>
```

haldun:
Closes #1563

@kddnewton I am not sure whether this test is sufficient or the approach is correct. May it be better to change the lexer so it won't parse the token as an identifier when we are inside the parameters?


kddnewton:
@haldun I'm also surprised by this. I wonder if we're missing some kind of state transition?

kddnewton:
@haldun ahh! It looks like we're missing a token then. I know this is quite a bigger fix, but would you be up for adding a `YP_TOKEN_METHOD_NAME` token? (That's basically what FID is.)

haldun:
@kddnewton let me try.

PR #1565, opened 2023-09-20
Check predicate of `while` and `until`

haldun:
Rejects where the predicate of `while` and `until` is not "closed".

```ruby
while 0 0; foo; end
until 0 0; foo; end
```

Ruby does not allow these:

```
ruby -c test.rb
test.rb: test.rb:1: syntax error, unexpected integer literal, expecting `do' for condition or ';' or '\n' (SyntaxError)
while 0 0; foo; end
        ^
test.rb:2: syntax error, unexpected integer literal, expecting `do' for condition or ';' or '\n'
until 0 0; foo; end
        ^
```

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -8267,6 +8267,18 @@ expect2(yp_parser_t *parser, yp_token_type_t type1, yp_token_type_t type2, yp_di
     parser->previous.type = YP_TOKEN_MISSING;
 }
 
+// This function is the same as expect2, but it expects one of three token types.
+static void
+expect3(yp_parser_t *parser, yp_token_type_t type1, yp_token_type_t type2, yp_token_type_t type3, yp_diagnostic_id_t diag_id) {
+    if (accept3(parser, type1, type2, type3)) return;
+
+    const uint8_t *location = parser->previous.end;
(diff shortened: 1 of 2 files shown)


src/yarp.c
```diff
@@ -12769,7 +12769,9 @@ parse_expression_prefix(yp_parser_t *parser, yp_binding_power_t binding_power) {
             yp_node_t *predicate = parse_expression(parser, YP_BINDING_POWER_COMPOSITION, YP_ERR_CONDITIONAL_UNTIL_PREDICATE);
             yp_do_loop_stack_pop(parser);
 
-            accept3(parser, YP_TOKEN_KEYWORD_DO_LOOP, YP_TOKEN_NEWLINE, YP_TOKEN_SEMICOLON);
+            if (!accept3(parser, YP_TOKEN_KEYWORD_DO_LOOP, YP_TOKEN_NEWLINE, YP_TOKEN_SEMICOLON)) {
```
kddnewton, line 12772: Can you write a `expect3` for this purpose?

kddnewton (changes requested):

kddnewton (approved):

PR #1569, opened 2023-09-20
use `yp_statements_node_body_length` a little bit more

froydnj:
Just a small cleanup.

### DIFF
--- a/src/yarp.c
+++ b/src/yarp.c
@@ -725,6 +725,10 @@ yp_statements_node_create(yp_parser_t *parser);
 static void
 yp_statements_node_body_append(yp_statements_node_t *node, yp_node_t *statement);
 
+// Get the length of the given StatementsNode node's body.
+static size_t
+yp_statements_node_body_length(yp_statements_node_t *node);
+
 // This function is here to allow us a place to extend in the future when we
 // implement our own arena allocation.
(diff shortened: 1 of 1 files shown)


eregon:
I restarted the failed truffleruby job, it's a race condition in RubyGems/Bundler: https://github.com/oracle/truffleruby/issues/3273

src/yarp.c
```diff
@@ -450,6 +450,8 @@ yp_parser_optional_constant_id_token(yp_parser_t *parser, const yp_token_t *toke
     return token->type == YP_TOKEN_NOT_PROVIDED ? 0 : yp_parser_constant_id_token(parser, token);
 }
 
+static size_t yp_statements_node_body_length(yp_statements_node_t *node);
```
kddnewton, line 453: A small nit, but would you mind moving this prototype up with `yp_statements_node_create` and `yp_statements_node_body_append`?

kddnewton (changes requested):

src/yarp.c
froydnj: Done.  Since the grouping meant that the prototype is now declared a little later in the file, I've reverted one of the changes.

kddnewton (approved):

PR #1599, opened 2023-09-25
Fix assertion failure for fwd params after rest

### LINKED ISSUE #1594: `YARP.parse 'def f(**,...)'` exits with assertion failed
Same for `a{|**,...` and `->**,...`

```
[tomoya.ishida:yarp]% ruby -Ilib -ryarp -e "YARP.parse 'def f(**,...)'"

Assertion failed: (params->keyword_rest == NULL), function yp_parameters_node_keyword_rest_set, file yarp.c, line 3707.
zsh: abort      ruby -Ilib -ryarp -e "YARP.parse 'def f(**,...)'"
```

haldun:
Closes #1594


src/yarp.c
```diff
@@ -9280,7 +9280,12 @@ parse_parameters(
 
                     yp_parser_local_add_token(parser, &parser->previous);
                     yp_forwarding_parameter_node_t *param = yp_forwarding_parameter_node_create(parser, &parser->previous);
-                    yp_parameters_node_keyword_rest_set(params, (yp_node_t *)param);
+                    if (params->keyword_rest == NULL) {
+                        yp_parameters_node_keyword_rest_set(params, (yp_node_t *)param);
```
kddnewton, line 9284: I love where you're going with this, but I actually want to favor the `...` here, and instead put the existing keyword rest into the `posts`. Could you move them around?

kddnewton (changes requested):

src/yarp.c
haldun: Done in https://github.com/ruby/yarp/pull/1599/commits/5a912de16792577e926d54282b42dcfd9e8de47e

kddnewton (approved):
Just needs a rebase

kddnewton (approved):

PR #1601, opened 2023-09-26
Statements versus expressions

### LINKED ISSUE #1547: Ruby rejects `alias x y.z` but YARP accepts it
Looks like YARP accepts it as `(alias x y).z`

```
$ ./local/bin/irb -ryarp
irb(main):001> RubyVM::AbstractSyntaxTree.parse("alias x y.z")
<internal:ast>:59:in `parse': syntax error, unexpected '.', expecting end-of-input (SyntaxError)
alias x y.z
         ^

        from (irb):1:in `<main>'
irb(main):002> YARP.parse("alias x y.z")
=>
#<YARP::ParseResult:0x00007fc4846bcc18
 @comments=[],
 @errors=[],
 @source=#<YARP::Source:0x00007fc484376158 @offsets=[0], @source="alias x y.z">,
 @value=
  @ ProgramNode (location: (0...11))
├── locals: []
└── statements:
    @ StatementsNode (location: (0...11))
    └── body: (length: 1)
        └── @ CallNode (location: (0...11))
            ├── receiver:
            │   @ AliasMethodNode (location: (0...9))
            │   ├── new_name:
            │   │   @ SymbolNode (location: (6...7))
            │   │   ├── opening_loc: ∅
            │   │   ├── value_loc: (6...7) = "x"
            │   │   ├── closing_loc: ∅
            │   │   └── unescaped: "x"
            │   ├── old_name:
            │   │   @ SymbolNode (location: (8...9))
            │   │   ├── opening_loc: ∅
            │   │   ├── value_loc: (8...9) = "y"
            │   │   ├── closing_loc: ∅
            │   │   └── unescaped: "y"
            │   └── keyword_loc: (0...5) = "alias"
            ├── call_operator_loc: (9...10) = "."
            ├── message_loc: (10...11) = "z"
            ├── opening_loc: ∅
            ├── arguments: ∅
            ├── closing_loc: ∅
            ├── block: ∅
            ├── flags: ∅
            └── name: "z",
 @warnings=[]>
 ```

### PR DESCRIPTION
**Pull request by kddnewton**
Currently, everything is structured as a pratt parser, which means everything is effectively being treated as an expression. This works for the majority of Ruby code. However, there are some things in the Ruby grammar that are considered "statements", that cannot be chained with operators.

There are two ways to solve this. (1) We could restructure the parser to call `parse_statement` when we're calling `parse_statements` instead of `parse_expression`. This would be the more classical approach to top-down recursive descent. (2) We could explicitly mark some expressions as "statements" and then add an error when they are found to be in a position that the user is attempting to chain.

I like option 2 because it leads to better error recovery. Every time you attempt to chain something with an operator, you're making a syntax error, but that might not be immediately obvious. Consider `+ foo 1`. In this case you have to put parentheses around the argument to `foo`, even though it's completely possible for us to tell that it should be parsed as `+(foo 1)` based on operator precedence. Instead, we can now mark that as a syntax error and tell you to please insert parentheses. But we still parse it as a unary because that's closer to what the developer intended in the first place.

This PR adds the ability to mark certain nodes as statements, and adds appropriate error messages in some places. It does not do all of the statements, which is going to require some more time. This is the first of what is probably going to be many PRs. This PR itself fixes #1547.

### CODE REVIEW

File: config.yml [hunk 1]
```diff
@@ -328,6 +318,21 @@ tokens:
     comment: "a separator between words in a list"
   - name: __END__
     comment: "marker for the point in the file at which the parser should stop"
+  - name: KEYWORD_IF_MODIFIER
+    value: 32768
```
jemmaissroff, line 322: Can we add a comment about the YP_MODIFIER_FLAG (or something to do with 2^15) here too so that they don't seem random?

templates/include/yarp/ast.h.erb
```diff
@@ -54,7 +59,7 @@ typedef uint16_t yp_node_flags_t;
 // all nodes (the ones listed below). Others are specific to certain node types.
 static const yp_node_flags_t YP_NODE_FLAG_STATEMENT = 0x1;
 static const yp_node_flags_t YP_NODE_FLAG_NEWLINE = 0x2;
-static const yp_node_flags_t YP_NODE_FLAG_STATIC_LITERAL = 0x3;
+static const yp_node_flags_t YP_NODE_FLAG_STATIC_LITERAL = 0x4;
```
jemmaissroff, line 62: Why this change?

config.yml
```diff
@@ -328,6 +318,21 @@ tokens:
...
+    value: 32769
+    comment: "unless in the modifier form"
+  - name: KEYWORD_WHILE_MODIFIER
+    value: 32770
+    comment: "while in the modifier form"
+  - name: KEYWORD_UNTIL_MODIFIER
+    value: 32771
+    comment: "until in the modifier form"
+  - name: KEYWORD_RESCUE_MODIFIER
+    value: 32772
```
eregon, line 334: Such high values means it's no longer possible to simply use an Array of token value -> token symbol or so anymore, notably this shows up in https://github.com/ruby/yarp/actions/runs/6314711880/job/17145622799?pr=1601.
A Hash would likely be much slower and use more memory.
So maybe some logic for the special cases >= 32768?

templates/include/yarp/ast.h.erb
kddnewton: The flags are a bit field, so it's important that they're all powers of 2, otherwise they won't work correctly. 0x3 was a bug

jemmaissroff: Can we make the flags being powers of 2 more explicit then, perhaps in the style of CRuby, eg `0x1`, `0x1 << 1`, `0x1 << 2`?

config.yml
eregon: Another idea is if the only flag we need (so far at least) for tokens is "statement vs expression" the check for "is it a statement token?" could be just "token >= KEYWORD_IF_MODIFIER", i.e., having the statement tokens at the end (and adding a comment in this file for where the limit is, or even an unused token to separate clearly).

kddnewton:
Closing this until it solves more problems

PR #1602, opened 2023-09-26
Rename YARP to prism

kddnewton:


jemmaissroff:
Looks good to me on this side! As discussed, let's implement on ruby/ruby first and then port back to ruby/yarp (or ruby/prism I guess 😂)

jemmaissroff (approved):

PR #1603, opened 2023-09-26
Remove visitor usage to not rely on overriding visit

kddnewton:


test/yarp/newline_test.rb
```diff
@@ -36,12 +36,7 @@ def assert_newlines(base, relative)
 
       result = YARP.parse_file(filepath)
       assert_empty result.errors
-
-      result.mark_newlines!
-      visitor = NewlineVisitor.new(result.source)
```
eregon, line 41: The NewlineVisitor class above can be removed then, right?

kddnewton: It was in a subsequent commit

PR #1608, opened 2023-09-29
Add a type method for quick comparison

kddnewton:


templates/lib/prism/node.rb.erb
```diff
@@ -193,6 +193,23 @@ module Prism
       <%- end -%>
       inspector.to_str
     end
+
+    # Sometimes you want to check an instance of a node against a list of
+    # classes to see what kind of behavior to perform. Usually this is done by
+    # calling `node.is_a?(cls)` or `node.class == cls` or
+    # `[cls1, cls2].include?(node.class)`. All of these approaches are
+    # relatively so because of the method calls and/or array allocations.
```
eregon, line 201:
```suggestion
    # relatively slow because of the method calls and/or array allocations.
```

templates/lib/prism/node.rb.erb
```diff
@@ -193,6 +193,23 @@ module Prism
       <%- end -%>
       inspector.to_str
     end
+
+    # Sometimes you want to check an instance of a node against a list of
+    # classes to see what kind of behavior to perform. Usually this is done by
+    # calling `node.is_a?(cls)` or `node.class == cls` or
```
eregon, line 199: I doubt `node.type == :some_symbol` is faster than `node.is_a?(cls)` (and same for `node.class == cls`), in fact that `.type` is a megamorphic call target (it calls one of N `type` methods) so it's likely slower than `is_a?` which is better optimized (details depend on the method lookup impl).

But indeed for `[cls1, cls2].include?(node.class)` and `case/when` then `.type` is a good fit.
Could you adjust the comment to clarify those two are the use cases, but `node.is_a?(cls)` is fine and there is no need to change it?

eregon: Actually `SomeClass === node` is likely the most efficient, because the method lookup there is trivial, there is no megamorphic lookup or call in there.
I think it's Ruby implementations' job to make the more common `is_a?` perform as good or as close as possible to it.

kddnewton: I updated to comment! I agree it's the ruby implementation's job to make it perform better. But this is also a convenient way to write things in general instead of fetching a bunch of constants.

eregon: Looks great, thanks for the update!

PR #1614, opened 2023-09-29
Run memcheck on CRuby 3.2

eregon:
* memcheck on ruby-head regularly fails due to regressions in ruby-head. This makes it clear whether the failure is due to ruby-head or to the changes in prism.

Alternatively we could skip running memcheck on ruby-head?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -169,6 +169,10 @@ jobs:
       run: bundle exec rake serialized_size:topgems
 
   memcheck:
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: ["3.2", "head"]
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


eregon:
Ah, 3.2 also has leaks in Ripper: https://github.com/ruby/prism/actions/runs/6358039241/job/17269878515?pr=1614
Ripper is really the worst.

eregon:
@peterzhu2118 Maybe it would make sense to ignore those on 3.2 if that's possible? Because it's not really useful for gems using `ruby_memcheck` to have those reported.

eregon:
Do you think we could use `ruby_memcheck` in CRuby's CI to catch these issues faster? Maybe just for Ripper tests since that's a very common source of leaks?

peterzhu2118:
It's possible to use ruby_memcheck on CRuby CI. We're also working on adding Valgrind support natively in Ruby. In that case, we'll be able to detect memory leaks everywhere in Ruby.

PR #1615, opened 2023-09-30
Consolidate diagnostic appendings

kddnewton:


src/prism.c
```diff
@@ -5447,7 +5485,7 @@ lex_optional_float_suffix(pm_parser_t *parser) {
             parser->current.end += pm_strspn_decimal_number_validate(parser, parser->current.end);
             type = PM_TOKEN_FLOAT;
         } else {
-            pm_diagnostic_list_append(&parser->error_list, parser->current.start, parser->current.end, PM_ERR_INVALID_FLOAT_EXPONENT);
+            pm_parser_err_token(parser, &parser->current, PM_ERR_INVALID_FLOAT_EXPONENT);
```
jemmaissroff, line 5488: Is it worth pulling out a `pm_parser_err_token_current` helper that takes just a parser and an error and implies the current? Seems like we have the pattern `pm_parser_err_token(parser, &parser->current, ...)` 28 times

jemmaissroff (changes requested):

src/prism.c
kddnewton: Great idea, done!

jemmaissroff (approved):

PR #1619, opened 2023-10-03
Add full_name to `ConstantPathNode` and `ConstantPathTargetNode`

vinistock:
Add `full_name` to both `ConstantPathNode` and `ConstantPathTargetNode`, so that one can easily figure out the complete name of the constant being referenced.

### DIFF
--- a/lib/prism/node_ext.rb
+++ b/lib/prism/node_ext.rb
@@ -52,4 +52,48 @@ module Prism
       o
     end
   end
+
+  class ConstantReadNode < Node
+    # Returns the list of parts for the full name of this constant. For example: [:Foo]
+    def full_name_parts
+      [name]
+    end
+
+    # Returns the full name of this constant. For example: "Foo"
+    def full_name
+      name.name
+    end
+  end
+
+  class ConstantPathNode < Node
+    # Returns the list of parts for the full name of this constant path. For example: [:Foo, :Bar]
+    def full_name_parts
+      parts = [child.name]
+      current = parent
+
+      while current.is_a?(ConstantPathNode)
+        parts.unshift(current.child.name)
+        current = current.parent
+      end
+
+      parts.unshift(current&.name || :"")
+    end
+
+    # Returns the full name of this constant path. For example: "Foo::Bar"
+    def full_name
+      full_name_parts.join("::")
+    end
+  end
+
+  class ConstantPathTargetNode < Node
+    # Returns the list of parts for the full name of this constant path. For example: [:Foo, :Bar]
+    def full_name_parts
+      (parent&.full_name_parts || [:""]).push(child.name)
+    end
+
+    # Returns the full name of this constant path. For example: "Foo::Bar"
+    def full_name
+      full_name_parts.join("::")
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


kddnewton:
Ahh yeah these methods should go in node_ext.rb

lib/prism/node_ext.rb
```diff
@@ -52,4 +52,32 @@ def options
...
+      parts = []
+      current = self
+
+      while current.is_a?(Prism::ConstantPathNode) || current.is_a?(Prism::ConstantPathTargetNode)
+        parts.unshift(current.child.name)
+        current = current.parent
+      end
+
+      parts.unshift(current&.name || :"").join("::")
+    end
```
kddnewton, line 67:
```suggestion
    def full_name_parts
      parts = [child.name]
      current = parent

      while current.is_a?(ConstantPathNode)
        parts.unshift(current.child.name)
        current = current.parent
      end

      parts.unshift(current&.name || :"")
    end
    
    def full_name
      full_name_parts.join("::")
    end
```

lib/prism/node_ext.rb
```diff
@@ -52,4 +52,32 @@ def options
...
```
kddnewton, line 81:
```suggestion
    def full_name_parts
      (parent&.full_name_parts || [:""]).push(child.name)
    end

    def full_name
      full_name_parts.join("::")
    end
```

lib/prism/node_ext.rb
```diff
@@ -52,4 +52,32 @@ def options
       o
     end
   end
+
+  class ConstantPathNode < Node
```
kddnewton, line 56:
```suggestion
  class ConstantPathNode < Node
    def full_name_parts
      [name]
    end

    def full_name
      name.name
    end
  end

  class ConstantPathNode < Node
```

kddnewton (changes requested):
I added a couple of suggestions that I think will make this easier to maintain.

Right now we're duplicating the logic to extract the full name between the target and the path nodes. If we share it, it will only be in one place, which is slightly nicer. 

I'd like to also split up the logic between extracting the names versus joining them with `::`. I've seen a couple of use cases where we want just the symbols, and it'll be easier if we provide them without having to split on `::` again.

Also, would you mind adding documentation for these methods?

lib/prism/node_ext.rb
```diff
@@ -52,4 +52,48 @@ def options
       o
     end
   end
+
+  class ConstantReadNode < Node
```
vinistock, line 56: In your original suggestion, both this class and the one below where `ConstantPathNode`. I figured this one was supposed to be `ConstantReadNode`. Let me know if that's not right.

kddnewton: Perfect

lib/prism/node_ext.rb
```diff
@@ -52,4 +52,48 @@ def options
...
+    # Returns the list of parts for the full name of this constant. For example: ["Foo"]
```
kddnewton, line 57: Last thing, these examples should have symbols in the arrays

kddnewton (approved):

PR #1620, opened 2023-10-03
[rust] add flag accessors on nodes

froydnj:
The raw `flags()` accessor that is currently exposed is functional, but seeing as how none of the relevant constants for the flag values are exported in `prism-sys`, `flags()` isn't very useful.  Let's write actual functions for each of the relevant flags, which will hopefully result in clearer code and make it so people don't have to worry about the bit-twiddliness of flag storage.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -374,7 +374,7 @@ flags:
   - name: StringFlags
     values:
       - name: FROZEN
-        comment: "frozen by virtue of a frozen_string_literal comment"
+        comment: "frozen by virtue of a `frozen_string_literal` comment"
 nodes:
   - name: AliasGlobalVariableNode
     fields:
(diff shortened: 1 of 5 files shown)


kddnewton (commented):
Love this. What do you think of prefacing with is_? In my limited exposure to rust I think I've seen that more often?

froydnj:
> Love this. What do you think of prefacing with is_? In my limited exposure to rust I think I've seen that more often?

I think that's a great idea.  After doing that change, I think the only ones worth calling out are:

* range.is_exclude_end() reads weirdly (`excludes_end` would be ideal, but I think `exclude_end` was OK-ish)
* regex.is_ignore_case() reads weirdly (`ignores_case` would be ideal, but I think `ignore_case` was fine)
* regex.is_once() is...OK? (I think the original `once` was fine here, but I don't have especially strong feelings)

Everything else is neutral or an improvement.

kddnewton (approved):
Yeah, we can rename in a subsequent PR. This is a good improvement

PR #1626, opened 2023-10-04
[rust] add pre- and post-traversal hooks to `Visit`

froydnj:
When writing visitors in Rust, I have found cases where I want to maintain a stack of nodes as I walk through the tree (e.g. to implement something like `rubocop-ast`'s `Node#parent`), and it's tedious to write visitor methods for all possible node types.  (It might also be inefficient, as we might be pushing and popping nodes when it doesn't matter -- you could maintain a list of such nodes yourself, but that seems to go against the spirit of visitors.)

This PR adds such functionality to the `Visit` trait: pre- and post-traversal hooks for child-ful and leaf nodes.

I'm reasonably sure the types for the hooks are correct; we don't want:

* references to the concrete node types, because that makes the hooks kind of useless.  (I did consider making the concrete node classes implement a common `ConcreteNode` trait and then make the hooks generic over `ConcreteNode`, but that seemed like a lot of work for very little gain.)
* concrete node values, for similar reasons.
* `&Node`, because such references can't be held once the hooks return.

so by process of elimination, we want actual `Node` values.  This happens to be exactly what I want for node stacks, but it's possible it's not the correct thing in general.  Rust lifetimes should work in our favor to make sure that we don't hold onto the `Node` values any longer than we are able.

### DIFF
--- a/rust/prism/build.rs
+++ b/rust/prism/build.rs
@@ -386,18 +386,35 @@ fn write_node(file: &mut File, flags: &[Flags], node: &Node) -> Result<(), Box<d
 fn write_visit(file: &mut File, config: &Config) -> Result<(), Box<dyn std::error::Error>> {
     writeln!(file, "/// A trait for visiting the AST.")?;
     writeln!(file, "pub trait Visit<'pr> {{")?;
+    writeln!(file, "   /// Called prior to visiting a node with potential child nodes.")?;
+    writeln!(file, "   fn visit_branch_node_enter(&mut self, _node: Node<'pr>) {{")?;
+    writeln!(file, "   }}")?;
+    writeln!(file)?;
+    writeln!(file, "   /// Called after visiting a node with potential child nodes.")?;
+    writeln!(file, "   fn visit_branch_node_leave(&mut self) {{")?;
+    writeln!(file, "   }}")?;
+    writeln!(file)?;
+    writeln!(file, "   /// Called prior to visiting a node that cannot have child nodes.")?;
+    writeln!(file, "   fn visit_leaf_node_enter(&mut self, _node: Node<'pr>) {{")?;
+    writeln!(file, "   }}")?;
+    writeln!(file)?;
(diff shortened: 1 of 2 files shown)


froydnj:
Ping on this @kddnewton .  I'm not super-attached to the proposed API -- one method with some indication of whether the node is a child or not would work too.

froydnj:
> Hey @froydnj - sorry about the delay.
> 
> I can definitely see the use case for this - I've had to keep stacks around for various stuff as well.

Thanks for looking at this!

> What would you think of creating another trait for this? I can see in some cases you would want a stack, but certainly not in all - and I imagine we're paying a penalty here for calling 2 functions on each visit function (unless rustc knows those functions are empty and doesn't bother calling them?).

I would assume that `rustc` inlines these functions consistently (maybe not in debug builds), so there's no runtime overhead from having them be no-ops.  (This doesn't apply with `dyn Visitor` or something like that, but people tend to not do that.)  I can see from running `cargo test --release` and examining the assembly for the generated binary that this appears to be the case for the visitor in `visitor_test`.

> As for the naming, can you do `visit_branch_node_enter`/`visit_branch_node_leave` and `visit_leaf_node_enter`/`visit_leaf_node_leave`? That mirrors what we have in the dispatcher class.

Can do!

kddnewton (changes requested):
Just marking this as "request changes" for filtering purposes

kddnewton:
Hey @froydnj - do you have time to come back to this?

froydnj:
@kddnewton Yes!  It won't happen this week, I suspect, but probably next week.

kddnewton:
Yeah no problem! Just wanted to ping to make sure it didn't get lost. I'm out tomorrow through next Monday anyway.

froydnj:
Ahhh, I rebased and everything, but forgot to make the naming changes you wanted.  My mistake, I will fix that up!

kddnewton (approved):
Nice work!

Issue #1582, opened 2023-09-22
YARP parses `def f(a,(a),(_1),(a!),(a?));end` without error


tompng:
Destructured parameter seems to be unchecked.
```ruby
def f(a,(a),(_1),(a!),(a?));end
tap{|a,(a),(_1),(a!),(a?)|}
->(a,(a),(_1),(a!),(a?)){}
```
In ruby
`(a)` : duplicated argument name (SyntaxError)
`(_1)` : _1 is reserved for numbered parameter (SyntaxError)
`(a!)` and `(a?)` : syntax error, unexpected method (SyntaxError)

tompng:
Looks like it is fixed in #1588 and #1627

PR #1627, opened 2023-10-05
Check for duplicate parameter names in destructured params

haldun:
Add checks for duplicated parameter name in codes like:

```ruby
def f(a, (b, (a))); end
```

Related #1582


kddnewton (approved):

kddnewton:
Just needs a rebase

PR #1631, opened 2023-10-06
Introduce transparent scopes.

eightbitraptor:
A transparent scope is a scope that cannot have local variables added to it's local table. When a local is added to it's table, it instead gets added to the first non-transparent parent scope.

This is used in for loops to ensure the correct depth for local variables inside the body


eregon (approved):
Nice, this sets the correct depth for locals inside `for`

PR #1634, opened 2023-10-06
Attach the ast node node to the scope

eightbitraptor:
So when building instruction sequences for a scope we can reference items from the parent. This is useful for for loops, where the local variable tables from the parent scope will need to be referenced.

### DIFF
--- a/include/prism/node.h
+++ b/include/prism/node.h
@@ -33,6 +33,7 @@ PRISM_EXPORTED_FUNCTION const char * pm_node_type_to_str(pm_node_type_t node_typ
 // declare them here to avoid generating them.
 typedef struct pm_scope_node {
     pm_node_t base;
+    pm_node_t *ast_node;
     struct pm_parameters_node *parameters;
     pm_node_t *body;
     pm_constant_id_list_t locals;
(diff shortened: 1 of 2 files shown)


eregon:
"parent" sounds weird since ScopeNode is not a node part of the AST so it has no "parent".
Maybe it should just be called "node" or "ast_node" or so?

include/prism/node.h
```diff
@@ -33,6 +33,7 @@ PRISM_EXPORTED_FUNCTION const char * pm_node_type_to_str(pm_node_type_t node_typ
 // declare them here to avoid generating them.
 typedef struct pm_scope_node {
     pm_node_t base;
+    pm_node_t parent;
```
eregon, line 36: Needs *

eightbitraptor: Thanks for this. And my apologies! I pushed this up in a rush before a meeting

eightbitraptor:
> "parent" sounds weird since ScopeNode is not a node part of the AST so it has no "parent". Maybe it should just be called "node" or "ast_node" or so?

Thanks for the feedback. I've used `ast_node` as I prefer the specificity.

eregon (approved):
LGTM

PR #1640, opened 2023-10-11
Increase the depth for the For node index variable

eightbitraptor:
In CRuby's parser locals in for nodes appear deeper in the ast than they do in Prism, but we can fix that with a transparent scope

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -8583,8 +8583,10 @@ parse_target(pm_parser_t *parser, pm_node_t *target) {
                     pm_parser_local_add_location(parser, message.start, message.end);
                     pm_node_destroy(parser, target);
 
+                    uint32_t depth = 0;
+                    for (pm_scope_t *scope = parser->current_scope; scope && scope->transparent; depth++, scope = scope->previous);
                     const pm_token_t name = { .type = PM_TOKEN_IDENTIFIER, .start = message.start, .end = message.end };
-                    target = (pm_node_t *) pm_local_variable_read_node_create(parser, &name, 0);
+                    target = (pm_node_t *) pm_local_variable_read_node_create(parser, &name, depth);
 
                     assert(sizeof(pm_local_variable_target_node_t) == sizeof(pm_local_variable_read_node_t));
                     target->type = PM_LOCAL_VARIABLE_TARGET_NODE;
@@ -12746,6 +12748,7 @@ parse_expression_prefix(pm_parser_t *parser, pm_binding_power_t binding_power) {
             parser_lex(parser);
             pm_token_t for_keyword = parser->previous;
             pm_node_t *index;
+            pm_parser_scope_push_transparent(parser);
 
             // First, parse out the first index expression.
             if (accept1(parser, PM_TOKEN_USTAR)) {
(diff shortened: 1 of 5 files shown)


eregon:
Right, this makes sense, taking the first example:
```ruby
for i in 1..10
  i
end
```
and the diff of it:
```diff
        ├── @ ForNode (location: (1,0)-(3,3))
        │   ├── index:
        │   │   @ LocalVariableTargetNode (location: (1,4)-(1,5))
        │   │   ├── name: :i
-       │   │   └── depth: 0
+       │   │   └── depth: 1
        │   ├── collection:
        │   │   @ RangeNode (location: (1,9)-(1,14))
        │   │   ├── left:
        │   │   │   @ IntegerNode (location: (1,9)-(1,10))
        │   │   │   └── flags: decimal
        │   │   ├── right:
        │   │   │   @ IntegerNode (location: (1,12)-(1,14))
        │   │   │   └── flags: decimal
        │   │   ├── operator_loc: (1,10)-(1,12) = ".."
        │   │   └── flags: ∅
        │   ├── statements:
        │   │   @ StatementsNode (location: (2,0)-(2,1))
        │   │   └── body: (length: 1)
        │   │       └── @ LocalVariableReadNode (location: (2,0)-(2,1))
        │   │           ├── name: :i
        │   │           └── depth: 1
        │   ├── for_keyword_loc: (1,0)-(1,3) = "for"
        │   ├── in_keyword_loc: (1,6)-(1,8) = "in"
        │   ├── do_keyword_loc: ∅
        │   └── end_keyword_loc: (3,0)-(3,3) = "end"
```
The `i` inside should be at the same depth because `i` is assigned and read "inside the loop", i.e. in `each`'s block when it's desugared to (it must be since the value depend on that block call/iteration)
```ruby
i = nil
(1..10).each do |i|
  i
end
```
or maybe clearer:
```ruby
i = nil
(1..10).each do |temp|
  i = temp
  i
end
```
Notice the `i = nil` outside, that's the correct desugaring because the index variable is also available after the for loop:
```
$ ruby -e 'for i in 1..10; end; p i'
10
```

eregon (approved):

src/prism.c
```diff
@@ -8583,8 +8583,10 @@ parse_target(pm_parser_t *parser, pm_node_t *target) {
                     pm_parser_local_add_location(parser, message.start, message.end);
                     pm_node_destroy(parser, target);
 
+                    uint32_t depth = 0;
+                    for (pm_scope_t *scope = parser->current_scope; scope && scope->transparent; depth++, scope = scope->previous);
```
eregon, line 8587: I think this might be easier to read as a `while` on 4 lines:
```c
pm_scope_t *scope = parser->current_scope;
while (scope && scope->transparent) {
    depth++;
    scope = scope->previous;
}
```

eregon:
BTW it might be a mistake that the for index variable(s) are visible after the loop, at least I don't think there is a good reason for it, and it has a performance cost. But nevertheless that is the current CRuby semantics so prism should match that.

kddnewton (approved):

PR #1641, opened 2023-10-12
Refactor escape sequences

### LINKED ISSUE #579: Revisit strings
Currently there are a couple of places that we attach strings to the tree. The first is `CallNode#name` which is necessary because some method names can be synthesized, e.g., `foo=`, `call`, etc. This is handled on its own and is fine.

The other places are `RegularExpressionNode#unescaped`, `StringNode#unescaped`, `SymbolNode#unescaped`, and `XStringNode#unescaped`. These are necessary in the case that we have escape sequences in the middle of these nodes. To make it easier to compile, we unescape those values as parse time.

To unescape them we make two passes. The first is while we're lexing. Whenever we encounter an escape (`\`), we call the unescape function to determine how many characters the escape takes (`\a` takes 2, `\C-\M-x` takes 7, etc.). Then, when we go to create one of these nodes that has the string attached, we make _another_ pass through the string and build the `yp_string_t` that will be attached.

Instead, it would be more efficient to build the escaped version while we lex. We could do this with a single `yp_string_t` that we keep on the parser (named something like `current_string`). Because the parser is only calling `parser_lex` one token at a time, we can always use the most recent value whenever we're about to handle a `YP_TOKEN_STRING_CONTENT` or `YP_TOKEN_CHARACTER_LITERAL` (the only tokens that actually generates this kind of content). The nodes can then copy the `yp_string_t` struct onto themselves and the next time one would be generated the parser can clear out the previous value with a `memset`. At this point ownership of any allocated memory on the string passes to the nodes that now own it, so we'll be okay from a memory perspective.

### LINKED ISSUE #1341: `lex_compat` failing to split up heredoc lines
Found in https://github.com/HParker/yarp/tree/failing-files in:

* `0a898a1dc14339c490e88c01dfab8a77-cleanup_spec.txt`
* `1e28abf226c8093332e2e03ce3474063-core.txt`
* `219a088a7306e1039162606a53b2e144-iec_bibliography.txt`
* `289809ba2786a8a68567c3de9cbb9e55-base_spec.txt`
* `35bf9f87e24717ac4cab441cb85320a4-base_spec.txt`
* `3b2d233a41744ecfa16fe857dd553e42-i18n_spec.txt`
* `3f7d740f1333a54b7346ce95313b7aa7-refs_spec.txt`
* `40e107d46753780696d19a79a8786738-blocks_spec.txt`
* `4b7cf11062f4ff5322101f0371fd4f4d-validator.txt`
* `5297933427c753437af28182d79cac24-processor_spec.txt`
* `54c79d90582310427c711fc0b00cb243-macros_spec.txt`
* `5ffd51d3c9d21faccb52d9f8d87bd4c1-cleanup_spec.txt`
* `6ebc8113b09dfd118530013843234abc-processor_spec.txt`
* `88b0e991e5f0349906957cb489f6c018-benchmark.txt`
* `89602dd2a7b55f1fe44fcea38ec40cec-refs_spec.txt`
* `8fba0a70015eb005cef21b5d9a9f808c-ref_spec.txt`
* `90f1c75ceca563defc0c7b4fd26d207b-section_spec.txt`
* `c0c5f9e8458ded165345e00bbea0cdfa-section_spec.txt`
* `cc23e31e862146088a333f9f36abeae3-inline_spec.txt`
* `cdc1055eae0ac163ada1c3e75fe430ec-named_entity.txt`
* `e1a8b2d7c3bf60f25f6a4921442697eb-cli.txt`

```ruby
<<~A
  #{B}
C
A
```

### LINKED ISSUE #1452: Regular expression nodes unescape improperly.
/\d+/ will make the unescaped bytes as 'd+'.  This needs different unescaping logic than string.

kddnewton:
Lots of things happening in this pull request. I wanted to make it smaller, but it's kind of a whole-sale change so required bundling it up.

There are a two reasons this is happening. First, we are currently incorrectly handling escape sequences in regular expressions. Second, we are currently scanning through string-like contents once to get the overall length, then another time to perform the actual unescaping. Third, we are going back over heredocs after they have been parsed in order to determine common whitespace, when we can determine it as lex time.

At the high-level, the changes in this PR are:

* We now use a `pm_buffer_t` to keep track of the contents of the current string, which lives on the parser. Whenever we return a `PM_TOKEN_STRING_CONTENT` token from `parser_lex`, it is expected that the parser will properly handle `parser.current_string`. If there are no escape sequences, this is a shared string that points to the start and end of the token. If there are escape sequences, this is an owned string that has the correct bytes in it.
* When string-like nodes are initially created, they already have their unescaped strings on them, so we don't go back and re-process them anymore.
* For heredocs, we now scan for common whitespace at lex time. For tilde heredocs, we now split up the string contents on actual newlines (not escaped newlines). This has the effect of making it significantly easier to dedent them in post processing since each string node only needs a single `memmove`.
* The unescaping test has been entirely revamped to brute force try every possible escape sequence in every possible context. With this in place, I'm much more confident we're now getting this right.

In terms of user-facing changes, there is really only one noticeable one which is that tilde heredocs will now be interpolated string nodes if they have more than one line, regardless of whether or not they have interpolation. This is because they will contain two string nodes with their own unescapes that have been properly dedented. I think this is probably fine, since it's possible to have interpolated string nodes for other weird cases that only contain string nodes as well, as in:

```ruby
[<<~FOO, "bar\
FOO
baz"]
```

In this case there will be two inner string nodes for the quoted string, one on line 1 and one on line 3 (this is the case today).

Fixes #579 
Fixes #1341
Fixes #1452


eregon:
Great!

> In terms of user-facing changes, there is really only one noticeable one which is that tilde heredocs will now be interpolated string nodes if they have more than one line, regardless of whether or not they have interpolation. This is because they will contain two string nodes with their own unescapes that have been properly dedented. I think this is probably fine, since it's possible to have interpolated string nodes for other weird cases that only contain string nodes as well, as in:

I think it would be valuable to clean this up, i.e. to only have interpolation when it's truly needed/present in the source.
Of course this could be done in the compiler/tool, but it feels suboptimal if every compiler/tool has to do that and the "interpolated/exploded" form is of little to no use for users of the gem.

I think for that edge case it's not too bad to have the InterpolatedStringNode (such code is cursed anyway :smile:), but for:
```ruby
s = <<~EOF
  ab
   cd
    ef
EOF
```
it seems unexpected and non-intuitive to get an InterpolatedStringNode there.
In fact the `StringNode#unescaped` value is I think very much expected to contain `"ab\n cd\n  ef"` and not InterpolatedStringNode + 3 StringNode.
It might also have a non-trivial performance and footprint impact to have one node per line for big tilde heredocs.

eregon:
IOW, what are the advantages to have an InterpolatedStringNode for multi-line tilde heredocs?
Could we simplify them to a single StringNode when we are about to create the InterpolatedStringNode for the tilde heredoc, or maybe slightly later (I think we already do that for e.g. `:"abc")?

kddnewton:
I hear what you're saying about there being a lot of nodes and us being able to combine them. I think we could potentially do a further refactor to combine sequential string nodes into one as another pass on the heredocs as a follow-up to this. But in general we've been talking about using interpolatedstringnodes as a generic list of string nodes for a while, especially as a replacement for stringconcatnode. So I think I'd like to roll with this approach for now just to get us closer to that reality.

eregon:
Yeah I think it's fine to merge this PR as-is and fix that later.
For cases like:
```ruby
a = "abc" \
  "def"
```
Two StringNode is probably better for tools dealing with syntax and not semantics.

OTOH for multiline tilde heredocs I think there is no value to have multiple StringNode, because it's contiguous in the source.

jemmaissroff (approved):

PR #1685, opened 2023-10-13
Parse magic comments

### LINKED ISSUE #574: Parse other magic comments besides encoding

kddnewton:
* Magic comments now get attached to the parse result object
* We now support the emacs and vim syntax, so to my knowledge we are now 1:1 with CRuby

We still don't properly support:

* The `warn_indent` magic comment
* The `shareable_constant_value` magic comment

But now that we are handling the syntax all in one place, adding these should be much easier.

Fixes #574

### DIFF
--- a/test/prism/magic_comment_test.rb
+++ b/test/prism/magic_comment_test.rb
@@ -16,7 +16,8 @@ module Prism
       "# -*- CoDiNg: ascii -*-",
       "# -*- \s\t\v encoding \s\t\v : \s\t\v ascii \s\t\v -*-",
       "# -*- foo: bar; encoding: ascii -*-",
-      "# coding \t \r  \v   :     \t \v    \r   ascii-8bit\n"
+      "# coding \t \r  \v   :     \t \v    \r   ascii-8bit\n",
+      "# vim: filetype=ruby, fileencoding=big5, tabsize=3, shiftwidth=3\n"
     ]
 
     examples.each do |example|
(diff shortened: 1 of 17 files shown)


jemmaissroff (approved):

test/prism/magic_comment_test.rb
```diff
@@ -0,0 +1,37 @@
...
+      "# CoDiNg: ascii",
+      "# \s\t\v encoding \s\t\v : \s\t\v ascii \s\t\v",
+      "# -*- encoding: ascii -*-",
+      "# -*- coding: ascii -*-",
+      "# -*- eNcOdInG: ascii -*-",
+      "# -*- CoDiNg: ascii -*-",
+      "# -*- \s\t\v encoding \s\t\v : \s\t\v ascii \s\t\v -*-",
+      "# -*- foo: bar; encoding: ascii -*-",
+      "# coding \t \r  \v   :     \t \v    \r   ascii-8bit\n",
+      "# vim: filetype=ruby, fileencoding=big5, tabsize=3, shiftwidth=3\n"
```
nobu, line 20: Why only these lines  have a newline?

```suggestion
      "# coding \t \r  \v   :     \t \v    \r   ascii-8bit",
      "# vim: filetype=ruby, fileencoding=big5, tabsize=3, shiftwidth=3",
```

kddnewton: @nobu removed in https://github.com/ruby/prism/commit/4a7be1fc507076fd67d4a003946736382953c079

PR #1696, opened 2023-10-16
Change ScopeNode to point to previous ScopeNode

jemmaissroff:
Amend ScopeNode to point to previous ScopeNode, and to have void* pointers to constants and index_lookup_table

### DIFF
--- a/include/prism/node.h
+++ b/include/prism/node.h
@@ -33,10 +33,17 @@ PRISM_EXPORTED_FUNCTION const char * pm_node_type_to_str(pm_node_type_t node_typ
 // declare them here to avoid generating them.
 typedef struct pm_scope_node {
     pm_node_t base;
+    struct pm_scope_node *previous;
     pm_node_t *ast_node;
     struct pm_parameters_node *parameters;
     pm_node_t *body;
     pm_constant_id_list_t locals;
+    pm_parser_t *parser;
+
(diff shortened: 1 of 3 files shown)


tenderlove (approved):

include/prism/node.h
```diff
@@ -33,10 +33,17 @@ PRISM_EXPORTED_FUNCTION const char * pm_node_type_to_str(pm_node_type_t node_typ
...
     pm_node_t *ast_node;
     struct pm_parameters_node *parameters;
     pm_node_t *body;
     pm_constant_id_list_t locals;
+    pm_parser_t *parser;
+
+    // We don't have the CRuby types ID and st_table within Prism
+    // so we use void *
+    void *constants; // ID *constants
+    void *index_lookup_table; // st_table *index_lookup_table
```
eregon, line 46: This looks a bit hacky, could it be done with a separate struct like:
```c
struct cruby_compiler_scope {
  pm_scope_node_t prism_scope;
  ID *constants;
  st_table *index_lookup_table;
};
// ...
pm_scope_node_init(node, &scope.prism_scope, &prev_scope.prism_scope, parser);
```
Then it would be properly type and also `pm_scope_node_init` would make sense besides the usage that CRuby does.

Alternatively a single `void *data` field and then storing a struct there for extra info would also be cleaner and more generally useful.

PR #1699, opened 2023-10-17
Fix lex compat when dedent should be 0

### LINKED ISSUE #1289: Common whitespace miscalculation with interpolation
Found in https://github.com/HParker/yarp/tree/failing-files in:

* `1c9c09519189d629c86e97de8730dc20-spec_helper.txt`
* `b5f6f4a7b52d555511747ee0053356c9-p5rb.txt`
* `d11f0b76fcb6661ccd4ae51da4924c5b-codetool.txt`

```ruby
<<~H
  a
#{b}
  c
H
```

kddnewton:
Fixes #1289

### DIFF
--- a/test/prism/heredoc_dedent_test.rb
+++ b/test/prism/heredoc_dedent_test.rb
@@ -7,8 +7,13 @@ module Prism
     filepath = File.expand_path("fixtures/tilde_heredocs.txt", __dir__)
 
     File.read(filepath).split(/(?=\n)\n(?=<)/).each_with_index do |heredoc, index|
+      # The first example in this file has incorrect dedent calculated by
+      # TruffleRuby so we skip it.
+      next if index == 0 && RUBY_ENGINE == "truffleruby"
+
       define_method "test_heredoc_#{index}" do
         node = Prism.parse(heredoc).value.statements.body.first
+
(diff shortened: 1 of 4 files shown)


test/prism/heredoc_dedent_test.rb
```diff
@@ -7,8 +7,13 @@ class HeredocDedentTest < TestCase
     filepath = File.expand_path("fixtures/tilde_heredocs.txt", __dir__)
 
     File.read(filepath).split(/(?=\n)\n(?=<)/).each_with_index do |heredoc, index|
+      # The first example in this file has incorrect dedent calculated by
+      # TruffleRuby so we skip it.
+      next if index == 0 && RUBY_ENGINE == "truffleruby"
```
jemmaissroff, line 12: I don't know how likely this is, but this means we'll fail on truffleruby if we move around the examples in this file. Should we mark that in the fixtures file itself too?

jemmaissroff (approved):
One comment, fine to merge without addressing

test/prism/heredoc_dedent_test.rb
kddnewton: I think it's alright for now - this test will fail CI so we will be able to catch it.

PR #1700, opened 2023-10-17
Expose options on match last line nodes

kddnewton:


lib/prism/node_ext.rb
```diff
@@ -3,6 +3,17 @@
 # Here we are reopening the prism module to provide methods on nodes that aren't
 # templated and are meant as convenience methods.
 module Prism
+  module RegularExpressionOptions
+  # Returns a numeric value that represents the flags that were used to create
```
eregon, line 7:
```suggestion
    # Returns a numeric value that represents the flags that were used to create
```

lib/prism/node_ext.rb
```diff
@@ -3,6 +3,17 @@
 # Here we are reopening the prism module to provide methods on nodes that aren't
 # templated and are meant as convenience methods.
 module Prism
+  module RegularExpressionOptions
```
eregon, line 6: Maybe this should be a private constant?

eregon (approved):

PR #1701, opened 2023-10-17
Ensure last encoding flag wins

eileencodes:
You can't encoding a string with more than one encoding so ensure that the last one wins.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -779,6 +779,7 @@ parse_decimal_number(pm_parser_t *parser, const uint8_t *start, const uint8_t *e
 static inline pm_node_flags_t
 pm_regular_expression_flags_create(const pm_token_t *closing) {
     pm_node_flags_t flags = 0;
+    pm_node_flags_t mask = (uint16_t) 0xFF0F;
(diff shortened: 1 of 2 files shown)


tenderlove (approved):
great job!

src/prism.c
```diff
@@ -779,18 +779,21 @@ parse_decimal_number(pm_parser_t *parser, const uint8_t *start, const uint8_t *e
 static inline pm_node_flags_t
 pm_regular_expression_flags_create(const pm_token_t *closing) {
     pm_node_flags_t flags = 0;
+    pm_node_flags_t mask = (uint16_t) 0xFF0F;
```
eregon, line 782: Could this be `~(PM_REGULAR_EXPRESSION_FLAGS_EUC_JP | PM_REGULAR_EXPRESSION_FLAGS_ASCII_8BIT | ...)` instead? It seems brittle to hardcode flag values here.

eregon: Fixed in https://github.com/ruby/prism/pull/1721

PR #1704, opened 2023-10-18
assure symbol body is identifier

HParker:
previously, sometimes symbol body was a METHOD_NAME type. this should be identifier.


test/prism/fixtures/symbols.txt
```diff
@@ -32,6 +32,10 @@
 
 :a
 
+:a?
+
+:a!
```
HParker, line 37: This test is useful, but doesn't completely catch this case since lex_compat hides this issue and the syntax tree doesn't actually specify the `value`'s token type.

I think it is good enough for now and something that is difficult to create a comparison test for since there is nothing existing to compare to.

PR #1706, opened 2023-10-18
Fixes for truffleruby

eregon:


src/prism.c
```diff
@@ -8213,7 +8222,8 @@ parser_lex(pm_parser_t *parser) {
 
             // If we haven't found an escape yet, then this buffer will be
             // unallocated since we can refer directly to the source string.
-            pm_token_buffer_t token_buffer = { 0 };
+            pm_token_buffer_t token_buffer;
+            pm_token_buffer_init(&token_buffer);
```
kddnewton, line 8226: What is the issue you're seeing here? We don't want to call `pm_buffer_init` because that's going to allocate, and we don't want to allocate until we need the memory because of an escape sequence.

It seems like you're trying to allocate everything to zero, but that's what `pm_token_buffer_t token_buffer = { 0 };` does.

kddnewton (changes requested):

src/prism.c
kddnewton: Ahh I missed your commit message. I found this link for the bug in clang: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80454.

If you're trying to get rid of the warning, initializing to `{ { 0 }, 0 }` should be fine.

eregon: I think we should have another `pm_buffer_init_*` function then. How should we call it?
`{ { 0 }, 0 }` seems rather messy and I already found a nasty bug in Prism due to this because C is way too lenient with this kind of initialization (with pm_string_t.type).

eregon: > I think we should have another `pm_buffer_init_*` function then.

Notably this makes it much clearer that the code in `pm_buffer.c` cannot assume that `buffer->value` is non-null or that `buffer->capacity > 0`.

eregon: BTW this function seems broken (it will hang) if capacity is 0:
```c
// Append the given amount of space to the buffer.
static inline void
pm_buffer_append_length(pm_buffer_t *buffer, size_t length) {
    size_t next_length = buffer->length + length;

    if (next_length > buffer->capacity) {
        do {
            buffer->capacity *= 2;
        } while (next_length > buffer->capacity);

        buffer->value = realloc(buffer->value, buffer->capacity);
    }

    buffer->length = next_length;
}
```

eregon: I see, what probably saves it is `pm_buffer_init_capacity(&token_buffer->buffer, 16);` in `pm_token_buffer_escape()`.
This feels very brittle, but it is how it is.

> It's wasteful, especially when there is built-in syntax for this.

If it's about performance this could easily be a `static inline` function, then it should have no cost.

eregon: I suppose no answer means no?
I'll push a commit to use `{ { 0 }, 0 }` and :crossed_fingers: it actually works on macOS Mojave because I can easily check there today.

eregon:
@kddnewton Updated to `{ { 0 }, 0 }`.

Please merge this when CI is green as this is blocking truffleruby to update to latest Prism (as it fails to build in CI due to that warning and `-Werror` from Prism's `Makefile`)

kddnewton (approved):

PR #1709, opened 2023-10-18
Some constant pool fixes and enhancements

kddnewton:
* The constant pool didn't have the concept of a constant string like `pm_string_t`, meaning it allocated memory even if the value was known at compile time. This has been fixed.
* When `...` is present in a method definition, you can use `*` and `&` to forward subsets of all of the parameters. This has been fixed.
* I changed the way the `Debug` module worked when comparing locals in the locals test. This makes it much more obvious where we need anonymous locals to handle specific values.

### DIFF
--- a/include/prism/util/pm_constant_pool.h
+++ b/include/prism/util/pm_constant_pool.h
@@ -40,8 +40,9 @@ size_t pm_constant_id_list_memsize(pm_constant_id_list_t *list);
 void pm_constant_id_list_free(pm_constant_id_list_t *list);
 
 typedef struct {
-    unsigned int id: 31;
+    unsigned int id: 30;
     bool owned: 1;
+    bool constant: 1;
     uint32_t hash;
 } pm_constant_pool_bucket_t;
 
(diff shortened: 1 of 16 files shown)


include/prism/util/pm_constant_pool.h
```diff
@@ -39,9 +39,21 @@ size_t pm_constant_id_list_memsize(pm_constant_id_list_t *list);
...
+typedef unsigned int pm_constant_pool_bucket_type_t;
+
+// By default, each constant is a slice of the source.
+static const pm_constant_pool_bucket_type_t PM_CONSTANT_POOL_BUCKET_DEFAULT = 0;
+
+// An owned constant is one for which memory has been allocated.
+static const pm_constant_pool_bucket_type_t PM_CONSTANT_POOL_BUCKET_OWNED = 1;
+
+// A constant constant is known at compile time.
+static const pm_constant_pool_bucket_type_t PM_CONSTANT_POOL_BUCKET_CONSTANT = 2;
```
eregon, line 52: Would an `enum` work here? Or C doesn't like the `pm_constant_pool_bucket_type_t type: 2;` below with an enum type?

eregon (approved):
Nice!

include/prism/util/pm_constant_pool.h
eregon: It seems to work fine: https://github.com/ruby/prism/pull/1713

PR #1713, opened 2023-10-18
Use an enum for pm_constant_pool_bucket_type_t

eregon:
From https://github.com/ruby/prism/pull/1709#discussion_r1363933976


eregon:
Mmh, but maybe it's not portable?
I found https://stackoverflow.com/a/11983430/388803
I suspect all non-ancient C compiler deal with it just fine.

eregon:
This is not very important, I'll close this.

PR #1716, opened 2023-10-19
Add type templates for RBS and RBI

vinistock:
This PR suggest adding templates to automatically generate RBS and RBI signatures. Having the right type annotations for the entire node structure is extremely helpful when developing - and having Prism itself export those annotations makes it super easy for anyone using typecheckers to get them.

This PR includes two commits:

1. Add the templates
2. Add a step to verify the validity of the generated type annotations on CI


.github/workflows/main.yml
```diff
@@ -30,11 +30,14 @@ jobs:
     - name: Set up Ruby
       uses: ruby/setup-ruby@v1
       with:
-        ruby-version: "2.5"
```
kddnewton, line 33: This needs to be 2.5 and cannot change because of CRuby's workflow for packaging binaries.

kddnewton (changes requested):
I am generally pro this PR but let's get CI passing first

kddnewton:
Nice, this is a lot closer. A couple more things:

* Could you move the static type declarations into their own files, such that only the generated types live in the template? I've found this much easier to maintain than having to change template files. For example the types for `ParseResult` and `Location` would move into a separate rbi/rbs file.
* Could you remove `rbs_to_rbi` and instead put that directly on the various `Field` classes in `template.rb`? That should make it a little easier to read/maintain.
* There's this implicit dependency on `rbs` for the tests that I'd prefer to make explicit. Could you instead add `rbs` to the gemfile and run it directly from Ruby instead of in a separate process?

vinistock:
I addressed the feedback. Concerning using RBS directly rather than spawning a new process, it's a bit odd because the CLI class handles a lot of the logic [here](https://github.com/ruby/rbs/blob/142f477047e5a2287690941d8c2880fccad60944/lib/rbs/cli.rb#L443). To avoid duplicating all of that, I just used the CLI class directly - which is odd as well.

Let me know how you'd like to proceed.

kddnewton (approved):
This is great, thanks!

PR #1724, opened 2023-10-23
Revisit multi target nodes

### LINKED ISSUE #1610: Provide `RequiredDestructuredParameterNode` child nodes not as Array but separated
Currently `RequiredDestructuredParameterNode` has only field `parameters` but it would be helpful (and convenient) to keep child nodes separately - pre parameters, rest and post parameters like it's done in `ParametersNode`.

### LINKED ISSUE #1635: SplatNode vs RestParameterNode
Is this a bug?

```ruby
irb(main):001:0> Prism.parse('def foo(*); end').value.statements.body.first.parameters.rest
=> 
@ RestParameterNode (location: (1,8)-(1,9))                                                                                     
├── name: nil                                                                                                                   
├── name_loc: ∅                                                                                                                 
└── operator_loc: (1,8)-(1,9) = "*"

irb(main):002:0> Prism.parse('def foo((*)); end').value.statements.body.first.parameters.requireds.first.parameters.first
=> 
@ SplatNode (location: (1,9)-(1,10))                                                                                            
├── operator_loc: (1,9)-(1,10) = "*"                                                                                            
└── expression: ∅  
```

I would have expected a `RestParameterNode` in both cases, but maybe there's a semantic reason that a `SplatNode` has to be used in the second case?

### LINKED ISSUE #1642: Splats on first element of MultiWrites make a nested MultiWrite and not a Splat
Code snippet:

```ruby
*a, b, c = 1
```

AST output:

```text
@ ProgramNode (location: (1,0)-(1,12))
├── locals: [:a, :b, :c]
└── statements:
    @ StatementsNode (location: (1,0)-(1,12))
    └── body: (length: 1)
        └── @ MultiWriteNode (location: (1,0)-(1,12))
            ├── targets: (length: 3)
            │   ├── @ MultiTargetNode (location: (1,0)-(1,2))
            │   │   ├── targets: (length: 1)
            │   │   │   └── @ SplatNode (location: (1,0)-(1,2))
            │   │   │       ├── operator_loc: (1,0)-(1,1) = "*"
            │   │   │       └── expression:
            │   │   │           @ LocalVariableTargetNode (location: (1,1)-(1,2))
            │   │   │           ├── name: :a
            │   │   │           └── depth: 0
            │   │   ├── lparen_loc: ∅
            │   │   └── rparen_loc: ∅
            │   ├── @ LocalVariableTargetNode (location: (1,4)-(1,5))
            │   │   ├── name: :b
            │   │   └── depth: 0
            │   └── @ LocalVariableTargetNode (location: (1,7)-(1,8))
            │       ├── name: :c
            │       └── depth: 0
            ├── lparen_loc: ∅
            ├── rparen_loc: ∅
            ├── operator_loc: (1,9)-(1,10) = "="
            └── value:
                @ IntegerNode (location: (1,11)-(1,12))
                └── flags: decimal
```

Moving the splat to any other position will show up as a SplatNode.

### LINKED ISSUE #1707: Nicer structure for MultiWriteNode
Currently MultiWriteNode only has a `Node[] targets` field and a value field.
It would be much easier to deal with it if it was like in the JRuby AST:
```
MultipleAsgnNode:
  Node[] pre;
  Node? rest;
  Node[] post;
```

For instance:
```
bin/parse -e 'a, a2, *b, c, c2 = ary'
@ ProgramNode (location: (1,0)-(1,22))
├── locals: [:a, :a2, :b, :c, :c2]
└── statements:
    @ StatementsNode (location: (1,0)-(1,22))
    └── body: (length: 1)
        └── @ MultiWriteNode (location: (1,0)-(1,22))
            ├── targets: (length: 5)
            │   ├── @ LocalVariableTargetNode (location: (1,0)-(1,1))
            │   │   ├── name: :a
            │   │   └── depth: 0
            │   ├── @ LocalVariableTargetNode (location: (1,3)-(1,5))
            │   │   ├── name: :a2
            │   │   └── depth: 0
            │   ├── @ SplatNode (location: (1,7)-(1,9))
            │   │   ├── operator_loc: (1,7)-(1,8) = "*"
            │   │   └── expression:
            │   │       @ LocalVariableTargetNode (location: (1,8)-(1,9))
            │   │       ├── name: :b
            │   │       └── depth: 0
            │   ├── @ LocalVariableTargetNode (location: (1,11)-(1,12))
            │   │   ├── name: :c
            │   │   └── depth: 0
            │   └── @ LocalVariableTargetNode (location: (1,14)-(1,16))
            │       ├── name: :c2
            │       └── depth: 0
            ├── lparen_loc: ∅
            ├── rparen_loc: ∅
            ├── operator_loc: (1,17)-(1,18) = "="
            └── value:
                @ CallNode (location: (1,19)-(1,22))
                ├── receiver: ∅
                ├── call_operator_loc: ∅
                ├── message_loc: (1,19)-(1,22) = "ary"
                ├── opening_loc: ∅
                ├── arguments: ∅
                ├── closing_loc: ∅
                ├── block: ∅
                ├── flags: variable_call
                └── name: :ary
```
This requires scanning the targets to find if there is a SplatNode or not and internally makes arrays of what's before & after.
If that information was part of the AST it would be much easier to deal with.

It would also be more consistent with `ParametersNode`.

### LINKED ISSUE #1717: Excessive `MultiTargetNode` in multi-assignent expression
Not sure whether it's intentional behaviour but it seems there is an excessive MultiTargetNode:

```
bin/parse -e 'a, (b, c), d = []'
@ ProgramNode (location: (1,0)-(1,17))
├── locals: [:a, :b, :c, :d]
└── statements:
    @ StatementsNode (location: (1,0)-(1,17))
    └── body: (length: 1)
        └── @ MultiWriteNode (location: (1,0)-(1,17))
            ├── targets: (length: 2)
            │   ├── @ LocalVariableTargetNode (location: (1,0)-(1,1))
            │   │   ├── name: :a
            │   │   └── depth: 0
            │   └── @ MultiTargetNode (location: (1,3)-(1,12))
            │       ├── targets: (length: 2)
            │       │   ├── @ MultiTargetNode (location: (1,3)-(1,9))
            │       │   │   ├── targets: (length: 2)
            │       │   │   │   ├── @ LocalVariableTargetNode (location: (1,4)-(1,5))
            │       │   │   │   │   ├── name: :b
            │       │   │   │   │   └── depth: 0
            │       │   │   │   └── @ LocalVariableTargetNode (location: (1,7)-(1,8))
            │       │   │   │       ├── name: :c
            │       │   │   │       └── depth: 0
            │       │   │   ├── lparen_loc: (1,3)-(1,4) = "("
            │       │   │   └── rparen_loc: (1,8)-(1,9) = ")"
            │       │   └── @ LocalVariableTargetNode (location: (1,11)-(1,12))
            │       │       ├── name: :d
            │       │       └── depth: 0
            │       ├── lparen_loc: ∅
            │       └── rparen_loc: ∅
            ├── lparen_loc: ∅
            ├── rparen_loc: ∅
            ├── operator_loc: (1,13)-(1,14) = "="
            └── value:
                @ ArrayNode (location: (1,15)-(1,17))
                ├── elements: (length: 0)
                ├── opening_loc: (1,15)-(1,16) = "["
                └── closing_loc: (1,16)-(1,17) = "]"
```

I assume that `MultiTargetNode` represents an array destructuring (but it isn't stated explicitly). If I am correct - current behaviour looks like a bug.

Might be related to https://github.com/ruby/prism/issues/1642.

kddnewton:
Fixes #1610
Fixes #1635
Fixes #1642
Fixes #1707
Fixes #1717

There were previously some bugs around multi target nodes because they weren't properly checking their nesting. These have been fixed. There were also some complaints about the API. These have been addressed. The big changes are:

* `MultiTargetNode`/`MultiWriteNode` now has its targets split into `requireds`, `rest`, and `posts`. This should make it easier to work with because you won't have to scan for a splat node to determine how to compile it.
* `RequiredDestructuredParameterNode` has been removed in favor of using `MultiTargetNode`, since they effectively the same thing. When they are compiled, they end up being the first statements compiled into the method bodies (to destructure the actual parameters).


config.yml
```diff
@@ -1915,7 +1915,11 @@ nodes:
           ^^^^^^^^^^^^^^
   - name: MultiTargetNode
     fields:
-      - name: targets
+      - name: requireds
```
eregon, line 1918:
```suggestion
      - name: pres
```
I would suggest `pres` or `pre` here (and for MultiWriteNode), to match with `post/posts`.
Both of them are not really required in the context of multiple assignment, they are just the things to assign before and after the `rest`.

Although I guess `requireds` was chosen to match https://github.com/ruby/prism/blob/87a9deb6bd8d8251d58977d394ab947c5600495d/config.yml#L2022-L2032 but the concept of required doesn't really exist for multiple assignment (everything is optional).

eregon (approved):
Great!

src/prism.c
```diff
@@ -5668,6 +5667,8 @@ context_terminator(pm_context_t context, pm_token_t *token) {
         case PM_CONTEXT_FOR:
         case PM_CONTEXT_ENSURE:
             return token->type == PM_TOKEN_KEYWORD_END;
+        case PM_CONTEXT_FOR_INDEX:
```
jemmaissroff, line 5670: What's this?

kddnewton: When you're parsing the index of a for loop (`for index in collection`) there's slightly different behavior for parsing a multiple target node. So we need to detect that context.

jemmaissroff (approved):

enebo:
@kddnewton Is this failing a test because MRI needs an update first for this to go green?

config.yml
kddnewton: Renamed to `lefts`/`rights`

PR #1730, opened 2023-10-24
Use Method#arity instead of Method#parameters

seven1m:
This has the benefit of creating fewer objects in memory just to check the method signature.

(Selfish reason for this PR is that Natalie doesn't have `Method#parameters` yet 😄)


seven1m:
Actually, I'm going to close this. I could imagine some future version of Ruby where `unpack1` takes more arguments, and this will break. I will patch the use of `Method#parameters` in our build step. ❤️

PR #1731, opened 2023-10-24
Add `vendored` feature to `prism-sys` crate

ianks:
This PR adds a new `vendored` feature to the rust crates. Typically, this is the preferred way to build self-contained deps like prism. This allow for useful functionality, like [building for `wasm32-wasi`](https://github.com/ruby/prism/compare/main...ianks:yarp:rust-prism-src-with-wasm?expand=1#diff-981e8c796781372ed9897ba184dbd976d6afd72c87f356a2a14366dc104a457f). You can run the example like so:

```sh
$ cd rust/prism
$ examples/wasm/run.sh
```


ianks:
@kddnewton I consolidated into prism-sys, so should be good now

kddnewton:
@ianks got some test failures

PR #1732, opened 2023-10-24
Add PostExecutionNode to pm_scope_node_init

jemmaissroff:
See https://github.com/ruby/ruby/pull/8752 for context

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -711,6 +711,11 @@ pm_scope_node_init(const pm_node_t *node, pm_scope_node_t *scope, pm_scope_node_
             scope->locals = cast->locals;
             break;
         }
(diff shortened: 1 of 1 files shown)


kddnewton:
Since we keep changing this in prism but it's only ever related to CRuby, could we move this function into CRuby?

kddnewton (approved):

eregon:
> Since we keep changing this in prism but it's only ever related to CRuby, could we move this function into CRuby?

+1, that would definitely be cleaner, especially since the struct includes CRuby-specific fields now.

jemmaissroff:
👍 I can move it over. Will do once we get this PostExecutionNode change fully merged

PR #1733, opened 2023-10-24
Revisit prettyprint

### LINKED ISSUE #1681: Fix prettyprinting in C API
We have done a lot of work to make the pretty printing look nice in the Ruby API. We should mirror this in the `pm_print_node` function in the C API so that C can also benefit from the nice printing

kddnewton:
Fixes #1681

This gets `pm_prettyprint` inline with the Ruby implementation of `Node#inspect` so that we can use it with `--dump=prism`.

## External changes:

* `pm_print_node` is removed in favor of using a single entrypoint to the prettyprint module. (It was never meant to be public anyway, hence why it wasn't marked with `PRISM_EXPORTED_FUNCTION`.) Instead it should be used like:

```c
pm_buffer_t output_buffer = { 0 };
pm_prettyprint(&output_buffer, parser, node);

printf("%.*s", (int) output_buffer.length, output_buffer.value);
pm_buffer_free(&output_buffer);
```

* The signature for `pm_prettyprint` is now `(pm_buffer_t *, const pm_parser_t *, const pm_node_t *)`.
* A long-standing bug has been fixed relating to locations that end on newlines. They now return the correct (line, column).
* Optional constants are now marked with `∅` instead of `nil` when they are not present to be consistent.

## Internal changes:

* I removed the scan functionality from the newline list. I found a bunch of edge cases, and honestly I think the optimization might not even be worth it depending on the use case. I'd rather remove it and potentially bring it back if it proves to be a bottleneck.
* `pm_buffer_append_u8` has been renamed to `pm_buffer_append_byte` for clarity.
* `pm_buffer_append_u32` has been renamed to `pm_buffer_append_varint` for clarity.
* `pm_buffer_append_format` has been introduced to make prettyprinting easier.


templates/src/prettyprint.c.erb
```diff
@@ -1,111 +1,186 @@
...
-    printf("%.*s\n", (int) buffer.length, buffer.value);
-
-    pm_buffer_free(&buffer);
-}
-
 // Pretty-prints the AST represented by the given node to the given buffer.
 PRISM_EXPORTED_FUNCTION void
-pm_prettyprint(pm_parser_t *parser, pm_node_t *node, pm_buffer_t *buffer) {
-    prettyprint_node(buffer, parser, node);
+pm_prettyprint(pm_buffer_t *output_buffer, const pm_parser_t *parser, const pm_node_t *node) {
```
jemmaissroff, line 182: Reminder that we need to merge this into CRuby before we merge it here, since CRuby uses the `pm_prettyprint` API already.

jemmaissroff (approved):
LGTM, just the node about that this is definitely breaking for CRuby

templates/src/prettyprint.c.erb
kddnewton: Yeah for sure

lib/prism/parse_result.rb
```diff
@@ -16,16 +16,32 @@ def slice(offset, length)
...
     end
 
+    # Binary search through the offsets to find the line number for the given
+    # offset.
     def line(value)
-      offsets.bsearch_index { |offset| offset > value } || offsets.length
+      left = 0
+      right = offsets.length - 1
+
+      while left <= right
```
eregon, line 25: What's the reason to implement our own binary search instead of using `bsearch_index` here?
`bsearch_index` wasn't correct for some edge case maybe?

lib/prism/parse_result.rb
```diff
@@ -86,7 +102,7 @@ def end_offset
 
     # The line number where this location starts.
     def start_line
-      source.line(start_offset)
+      source.line(start_offset) + 1
```
eregon, line 105: It seems better that Source#line returns a 1-starting line.
I think it's a breaking change if Source#line returns a 0-starting line.

kddnewton: `Location#start_line` and `Location#end_line` are the user-facing APIs and those are still 1-indexed, so nothing really changing that's user-facing here. This just avoids some extra computation internally. `Source#line` is not meant to be used on its own (which is why it's private on almost all of the APIs).

PR #1735, opened 2023-10-24
Remove pm_scope_node_init

jemmaissroff:
It is only used by CRuby, so is moved to ruby/ruby.

Do not merge until https://github.com/ruby/ruby/pull/8756 is merged


kddnewton (approved):

kddnewton:
We can remove `pm_scope_node_t` as well I think

PR #1736, opened 2023-10-25
Use `cargo fmt`

ianks:
Consistent formatting is nice, so let's enable it. I tried to configure `rustfmt.toml` to minimize big diffs to keep the original struct of the code as much as possible, since the default settings made a very noisy diff.


kddnewton:
Generally +1 on this, but it looks like it's missing from the dependencies and also maybe a config file is missing?

ianks:
I switched up the `rustfmt.toml` files somehow I guess? Anyway, fixed now.

kddnewton (approved):

PR #1740, opened 2023-10-26
{Compiler,Visitor}#visit_with

kddnewton:
There are many times when you need to pass data down the tree as you are compiling/visiting it. For example, in the CRuby compiler we need to pass `popped` and the current `iseq` that is being compiled, in parser-prism we need to pass down the local table, etc.

This new API allows you to do that without having to store stacks of values like we were previously doing. It would look something like:

```ruby
class MyCompiler < Prism::Compiler
  def visit_program_node(node)
    visit_with(node.statements, top_iseq)
  end

  def visit_statements_node_with(node, iseq)
    visit_all_with(node.body, iseq)
  end

  def visit_integer_node_with(node, iseq)
    iseq.putobject(node.value)
  end
end
```

`visit`, `visit_all`, and `visit_child_nodes` now all have `_with` equivalents, and all nodes have `accept_with` equivalents that forward along the data.

If you would like to forward multiple data points, it's recommended to create your own custom object or to use a data structure like an array or hash.

### DIFF
--- a/templates/lib/prism/compiler.rb.erb
+++ b/templates/lib/prism/compiler.rb.erb
@@ -22,9 +22,19 @@ module Prism
       node&.accept(self)
     end
 
+    # Visit an individual node with a value.
+    def visit_with(node, value)
+      node&.accept_with(self, value)
+    end
+
     # Visit a list of nodes.
     def visit_all(nodes)
       nodes.map { |node| node&.accept(self) }
+    end
+
+    # Visit a list of nodes with a value.
+    def visit_all_with(nodes, value)
+      nodes.map { |node| node&.accept_with(self, value) }
     end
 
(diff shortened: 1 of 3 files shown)


eregon:
FWIW another way to do this, and that's what TruffleRuby currently uses, is to create a new instance of the visitor/compiled/translator whenever "changing scopes". Then you can store whatever you want directly on the visitor and create a new visitor whenever you need to change a field. It's also convenient to switch from one visitor class to another which handle different/extra nodes for instance.

Not against this approach, just wanted to mention there is already a way to do this without an explicit stack.
This might be more elegant for the Ruby API.

One disadvantage of this approach in a statically-typed language is the passed value would be a generic Object/void* and would need lots of casts. That does not apply to Ruby though.

PR #1741, opened 2023-10-26
Add KeywordSplat flag to ArgumentsNode

jemmaissroff:
Method calls with keyword splat args compile differently than without since they merge the keyword arg hash with the keyword splat hash. We know this information at parse time, so can set a flag which the compiler can use.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -329,6 +329,10 @@ tokens:
   - name: __END__
     comment: "marker for the point in the file at which the parser should stop"
 flags:
+  - name: ArgumentsNodeFlags
+    values:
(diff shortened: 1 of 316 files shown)


kddnewton (approved):

jemmaissroff:
Merging - test-all is failing because the templating isn't done, but shows no indication that it will fail on this PR otherwise

eregon:
Quick thought and wondering, could the flag be on `KeywordHashNode` instead?
Although having the flag on ArgumentsNode might be best as then we could also add a flag for "are there are keyword arguments" on ArgumentsNode for free then.

PR #1742, opened 2023-10-26
Fix constant path full name when parent is not a constant

vinistock:
When we introduced `full_name`, we forgot that Ruby allows you to do `self::Foo::Bar` or even `var::Foo::Bar`. The code currently breaks in those scenarios when trying to invoke `name`, since the nodes that represent the beginning of the path don't implement it (like `SelfNode`).

There are a few options on how we might fix this. This PR is just using `node.slice` if the node is not a constant read. We can omit the first part when it's not a constant. Or we can add a `name` field to the nodes that don't have it, just returning the slice.

Let me know how you prefer to deal with this scenario.

### DIFF
--- a/lib/prism/node_ext.rb
+++ b/lib/prism/node_ext.rb
@@ -79,7 +79,7 @@ module Prism
     # Foo::Bar::Baz -> does not raise because all parts of the constant path are
     # simple constants
     # var::Bar::Baz -> raises because the first part of the constant path is a
-    # local variablV
+    # local variable
     class DynamicPartsInConstantPathError < StandardError; end
 
     # Returns the list of parts for the full name of this constant path.
--- a/lib/prism/node_ext.rb
+++ b/lib/prism/node_ext.rb
@@ -74,6 +74,14 @@ module Prism
   end
 
   class ConstantPathNode < Node
+    # An error class raised when dynamic parts are found while computing a
+    # constant path's full name. For example:
+    # Foo::Bar::Baz -> does not raise because all parts of the constant path are
+    # simple constants
+    # var::Bar::Baz -> raises because the first part of the constant path is a
+    # local variablV
+    class DynamicPartsInConstantPathError < StandardError; end
+
     # Returns the list of parts for the full name of this constant path.
     # For example: [:Foo, :Bar]
     def full_name_parts
@@ -83,6 +91,10 @@ module Prism
       while current.is_a?(ConstantPathNode)
         parts.unshift(current.child.name)
         current = current.parent
+      end
+
+      unless current.is_a?(ConstantReadNode)
+        raise DynamicPartsInConstantPathError, "Constant path contains dynamic parts. Cannot compute full name"
       end
 
       parts.unshift(current&.name || :"")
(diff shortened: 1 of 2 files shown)


lib/prism/node_ext.rb
```diff
@@ -83,7 +83,8 @@ def full_name_parts
         current = current.parent
       end
 
-      parts.unshift(current&.name || :"")
+      name = current.is_a?(ConstantReadNode) ? (current&.name || :"") : current.slice
+      parts.unshift(name)
```
kddnewton, line 87: I don't really want to use `slice` here since it could end up with a whole mess of stuff. For example:

```ruby
foo. # comment
  bar::Baz::Qux
```

You end up with `["foo. # comment\n  bar", :Baz, :Qux]` and `"foo. # comment\n  bar::Baz::Qux"`.

Honestly, I'm not sure of the best solution here, but I think I would prefer to keep the node around in the list and just say it's one of the "parts". That would be:

```suggestion
      part = current.is_a?(ConstantReadNode) ? (current&.name || :"") : current
      parts.unshift(part)
```

Then `full_name` below would need to change as well. I wouldn't want people to call that method and it drop data or be confusing. For example, if you have `var::Foo`, you wouldn't want `full_name` to return `::Foo` (which would be wrong, it's not a top-level constant) and you wouldn't want it to return `Foo` (which is wrong, it should be scoped). There's no winning in this scenario. I think honestly it should raise a special error that indicates that the full name could not be created if the first part is not a symbol.

kddnewton (changes requested):

lib/prism/node_ext.rb
enebo: @kddnewton Is this a special case of full_name or a general issue of emitting syntax from the tree?  `syntax` of SelfNode could print out `self`. constant read would be `Foo`.  Constant Path would be syntax of the children.   The nodes know the logic for putting `::` in too.

kddnewton: I think you're basically talking about round-tripping back to the original source. That's what I'm trying to solve with https://github.com/ruby-syntax-tree/syntax_tree-prism. This API in particular is being used for indexing code, so it kind of fits into a small niche of nice to haves.

kddnewton:
@vinistock do you have time to come back to this? If not that's okay, I can take over if you'd prefer.

kddnewton:
@vinistock ping

vinistock:
Sorry! Got back to this today. I changed it so that it raises if any of the parts can't be made into a string name.

I created a custom error class for this and put it under the node itself - since it's only relevant to that particular node. Let me know if you prefer to move that somewhere else.

kddnewton (approved):

kddnewton:
@vinistock documentation check is failing because the error class is undocumented.

lib/prism/node_ext.rb
```diff
@@ -74,6 +74,14 @@ def full_name
   end
 
   class ConstantPathNode < Node
+    # An error class raised when dynamic parts are found while computing a
+    # constant path's full name. For example:
+    # Foo::Bar::Baz -> does not raise because all parts of the constant path are
+    # simple constants
+    # var::Bar::Baz -> raises because the first part of the constant path is a
+    # local variablV
```
seven1m, line 82:
```suggestion
    # local variable
```

vinistock: Oops. Thanks, Tim!

kddnewton (approved):

PR #1744, opened 2023-10-27
Improve comment generation in templates

paracycle:
The existing comment generation was hard to read and was making a lot of string manipulation. However, ERB files are already designed to do string manipulation, so we can use that instead.

So, instead of doing a split and a map, I opted to use the `#each_line` method to iterate over the lines of the file.

Also, in order to add an optional space padding at the beginning of the line, I opted to unconditionally pad it with a space and then right trim it. This makes sure that no space is left behind if the line is empty, but a space is added if the line is not empty. It also ensures that the line endings that are kept by `each_lines` are also removed.

### DIFF
--- a/templates/java/org/prism/Nodes.java.erb
+++ b/templates/java/org/prism/Nodes.java.erb
@@ -174,9 +174,11 @@ public abstract class Nodes {
 <%# NODES -%>
     <%- nodes.each do |node| -%>
 
+    /**
     <%- node.comment.each_line do |line| -%>
-    //<%= line.prepend(" ").rstrip %>
+     *<%= line.prepend(" ").rstrip %>
     <%- end -%>
+     */
     public static final class <%= node.name -%> extends Node {
         <%- if node.needs_serialized_length? -%>
         public final int serializedLength;
--- a/templates/java/org/prism/Nodes.java.erb
+++ b/templates/java/org/prism/Nodes.java.erb
@@ -174,7 +174,9 @@ public abstract class Nodes {
 <%# NODES -%>
     <%- nodes.each do |node| -%>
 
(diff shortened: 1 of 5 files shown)


kddnewton (changes requested):
I'm generally fine with this. I think we can rely on `node.comment` always being there? If there isn't one we should error. I think you accidentally broke Java by using `#` instead of `//`.

templates/java/org/prism/Nodes.java.erb
```diff
@@ -174,7 +174,9 @@ public abstract class Nodes {
 <%# NODES -%>
     <%- nodes.each do |node| -%>
 
-    <%= "#{node.comment.split("\n").map { |line| "// #{line}" }.join("\n    ")}\n" if node.comment -%>
+    <%- node.comment.to_s.each_line do |line| -%>
+    #<%= line.rjust(line.size + 1).rstrip %>
+    <%- end -%>
```
eregon, line 179: The best in Java would be to use a JavaDoc comment `/** ... * ... */` like templates/javascript/src/nodes.js.erb below

paracycle: Thanks, done.

paracycle:
> I think we can rely on `node.comment` always being there? If there isn't one we should error. 
I wasn't sure if that was something we could assume. Since you say so, I've removed the redundant `.to_s` calls.

> I think you accidentally broke Java by using `#` instead of `//`.
Sorry, copy/paste error. I fixed it and also changed Java to use Javadoc syntax instead.

eregon (approved):

kddnewton (approved):

PR #1747, opened 2023-10-30
Share comment generation logic

kddnewton:


templates/java/org/prism/Nodes.java.erb
```diff
@@ -175,8 +175,8 @@ public abstract class Nodes {
     <%- nodes.each do |node| -%>
 
     /**
-    <%- node.comment.each_line do |line| -%>
-     *<%= line.prepend(" ").rstrip %>
+    <%- node.each_comment_line do |line| -%>
+     * <%= line %>
```
eregon, line 179: This approach will leave trailing spaces for empty lines in comments.
`*<%= line %>` + `yield line.prepend(" ").rstrip` would avoid that

kddnewton: Ah I was wondering why that was there. I'll fix this.

eregon (approved):

PR #1752, opened 2023-10-31
Fix debug method names in truffleruby workflow

eregon:
TruffleRuby doesn't intend to rename those, they are debug methods, so let's use their current name.
This should let the truffleruby workflow be green.

FYI @andrykonchin


kddnewton (approved):

eregon:
Indeed it's green now: https://github.com/ruby/prism/actions/runs/6709570377/job/18232909748

PR #1755, opened 2023-10-31
Split KeywordParameterNode into Optional and Required

jemmaissroff:
Prior to this commit, KeywordParameterNode included both optional and required keywords. With this commit, it is split in two, with `OptionalKeywordParameterNode`s no longer having a value field.

The main motivation behind this change is to remain consistent with Prism's verbose types, and indicate the information known at parsing time which distinguishes these two node types. It will also lead to a minor memory save as `OptionalKeywordParameterNode`s no longer have empty `value` fields.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1692,24 +1692,6 @@ nodes:
 
           foo(a: b)
               ^^^^
-  - name: KeywordParameterNode
-    fields:
-      - name: name
-        type: constant
-      - name: name_loc
-        type: location
-      - name: value
-        type: node?
-    comment: |
-      Represents a keyword parameter to a method, block, or lambda definition.
-
-          def a(b:)
-                ^^
-          end
-
-          def a(b: 1)
-                ^^^^
-          end
   - name: KeywordRestParameterNode
     fields:
       - name: name
@@ -1997,6 +1979,18 @@ nodes:
 
           $1
           ^^
+  - name: OptionalKeywordParameterNode
+    fields:
+      - name: name
+        type: constant
+      - name: name_loc
+        type: location
+    comment: |
+      Represents an optional keyword parameter to a method, block, or lambda definition.
+
+          def a(b: )
+                ^^
+          end
   - name: OptionalParameterNode
     fields:
       - name: name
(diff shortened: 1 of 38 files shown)


lib/prism/debug.rb
```diff
@@ -114,8 +114,8 @@ def self.prism_locals(source)
                   AnonymousLocal
                 end
               end,
-              *params.keywords.reject(&:value).map(&:name),
-              *params.keywords.select(&:value).map(&:name)
+              *params.keywords.select { |kw| kw.is_a? OptionalKeywordParameterNode }.map(&:name),
+              *params.keywords.select { |kw| kw.is_a? RequiredKeywordParameterNode }.map(&:name),
```
paracycle, line 118:
```suggestion
              *params.keywords.grep(OptionalKeywordParameterNode).map(&:name),
              *params.keywords.grep(RequiredKeywordParameterNode).map(&:name),
```

Since `Enumerable#grep` uses `===` to check for equality: https://ruby-doc.org/3.2.0/Enumerable.html#method-i-grep

config.yml
```diff
@@ -1997,6 +1979,18 @@ nodes:
...
+      - name: name
+        type: constant
+      - name: name_loc
+        type: location
+    comment: |
+      Represents an optional keyword parameter to a method, block, or lambda definition.
+
+          def a(b: )
+                ^^
+          end
```
andrykonchin, line 1993: I would expect an "optional" keyword parameter to have a default value and required parameter not to have it.

eregon: Right the names or so seem mixed up.

Ruby's terminology is `required` means has no default:
```
$ ruby -e 'p method(def m(k1:, k2: val); end).parameters'
[[:keyreq, :k1], [:key, :k2]]
```

Just like for positional required/optional arguments, the optional is the one with a default value.

eregon (changes requested):
The names are inverted

config.yml
jemmaissroff: You're totally right, I got the names flipped. Thanks for catching!

eregon (approved):
Looks good.
It seems best to squash the commits so if someone looks at individual commits they don't get confused.

kddnewton (approved):

PR #1765, opened 2023-11-03
Rename suppress warnings to verbose

kddnewton:


lib/prism/ffi.rb
```diff
@@ -299,7 +299,7 @@ def dump_options(options)
       values << (options.fetch(:frozen_string_literal, false) ? 1 : 0)
 
       template << "C"
-      values << (options.fetch(:suppress_warnings, false) ? 1 : 0)
+      values << (options[:verbose] ? 0 : 1)
```
eregon, line 302: If verbose is set to true it's serialized as 0 = false, is that a bug?

kddnewton: `verbose` is the name of the flag coming from the outside, `suppress_warnings` is the value of the flag on in the inside. So this is intentional.

PR #1767, opened 2023-11-03
Always fully qualify types for RBIs and avoid using attributes

vinistock:
There were a couple of things I missed in #1716.

Tapioca always uses compact class/module style (e.g.: `class Foo::Bar`) and never nests definitions inside a namespace. Because of that, all of the types we use must be fully qualified or else they will cause a typechecking error.

For example

```ruby
class Prism::CallNode
  sig { returns(T.nilable(Location)) } # => Location does not exist. We need to use Prism::Location
  def message_loc; end
end
```

Also, Tapioca won't properly mix the signatures if we use `attr_reader`. We should always use regular method definitions in RBIs.

I tested this on Ruby LSP and the RBI generated after these changes is fully valid.


rbi/prism_static.rbi
```diff
@@ -2,79 +2,79 @@
 
 module Prism
   class ParseResult
```
kddnewton, line 4: @vinistock do you want to un-nest this? i.e., make it `class `Prism::ParseResult`?

vinistock: Yes, good point. It'll make it easier to spot future issues.

I also removed the `# typed: true` comments from the top of the files, because Tapioca adds those automatically - and we were ending up with duplicates.

kddnewton (approved):

paracycle:
> Also, Tapioca won't properly mix the signatures if we use `attr_reader`. We should always use regular method definitions in RBIs.

I think this is actually a bug in RBI wrt its logic when doing a merge. It should treat `attr_xxx` defined methods and readers/writers on the same footing and merge their signatures properly.

vinistock:
That's a good point. I created https://github.com/Shopify/tapioca/issues/1709 to investigate.

PR #1787, opened 2023-11-08
Remove non-ASCII source characters

kddnewton:


templates/template.rb
```diff
@@ -374,6 +374,14 @@ def template(name, write_to: nil)
       write_to ||= File.expand_path("../#{name}", __dir__)
       contents = heading + erb.result_with_hash(locals)
 
+      if (extension == ".c" || extension == ".h") && !contents.ascii_only?
+        # Enforce that we only have ASCII characters here. This is necessary
+        # for some locales that only allow ASCII characters in C source files.
```
nobu, line 379: This comment is not exact.
Maybe "other than UTF-8 locales"?

kddnewton: That makes sense! I'll update it.

PR #1791, opened 2023-11-09
Reset do_loop_stack around a body of a endless method definition

### LINKED ISSUE #1772: Fails to parse `while def f = g do end; end`
Endless method definition with block inside while condition is parsed different from Ruby.
`while def f = g do end` (syntax error in ruby) is parsed without error.
`while def f = g do end; end` (syntax ok in ruby) fails to parse.

```ruby
eval('while def f = g do end')
# => syntax error, unexpected end-of-input, expecting `do' for condition or ';' or '\n' (SyntaxError)

Prism.parse('while def f = g do end').success?
# => true

eval('while def f = g do end; end')
# => infinite loop

Prism.parse('while def f = g do end; end').success?
# => false
```

makenowjust:
Fix #1772

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -14326,6 +14326,7 @@ parse_expression_prefix(pm_parser_t *parser, pm_binding_power_t binding_power) {
                 equal = parser->previous;
 
                 context_push(parser, PM_CONTEXT_DEF);
+                pm_do_loop_stack_push(parser, false);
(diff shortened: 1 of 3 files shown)


kddnewton (approved):
Nice work!

makenowjust:
Thank you!

PR #1796, opened 2023-11-10
Build the ability to format errors

kddnewton:
Previously, we only supported error messages that were constant strings. This works for the most part, but there are some times where we want to include some part of the source in the error message to make it better.

For example, instead of "Token is reserved" it's better to write "_1 is reserved".

To do this, we now support allocating error messages at runtime that are built around format strings.

I don't actually think this should be the real solution here. Ideally we would provide the ability for these fields to be described so that we could template out some nice Ruby syntax error classes. But for now, this gets us closer to passing Ruby spec so I'd like to build this intermediate step.

cc @enebo this should allow us to start fixing some specific error message assertions. Now that we have this, would you be able to open an issue with the error messages that are not lining up?

### DIFF
--- a/include/prism/diagnostic.h
+++ b/include/prism/diagnostic.h
@@ -30,6 +30,13 @@ typedef struct {
 
     /** The message associated with the diagnostic. */
     const char *message;
+
+    /**
+     * Whether or not the memory related to the message of this diagnostic is
+     * owned by this diagnostic. If it is, it needs to be freed when the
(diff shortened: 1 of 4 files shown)


enebo:
@kddnewton sure.  I will start opening issues next week.  In many cases it is not even the dynamic aspect you are fixing here but just more explicit english grammar (like adding an article in prism which was omitted in original warn/error).

PR #1802, opened 2023-11-10
Rename librubyparser to libprism

### LINKED ISSUE #1790: Rename librubyparser to libprism
Right now we generate build artifacts as `librubyparser.*`. This is legacy from when I was building the prototype and the project was called `ruby-parser`. We should rename this to `libprism` for consistency. Also because it looks like maybe CRuby will be generating librubyparser for the universal parser (maybe? I'm not sure if I'm reading it right). Either way it's non-obvious that `librubyparser` has to do with prism.

kddnewton:
librubyparser was an artifact of the prototype that was initially named ruby-parser. Instead, this renames it to libprism to be consistent with the actual name.

Fixes #1790


jemmaissroff (approved):

kddnewton:
cc @enebo @eregon @seven1m this is going to effect you

PR #1803, opened 2023-11-10
Add source code and changelog metadata to gemspec

### LINKED ISSUE #1738: RubyGems page `Source Code` link points to a different repository
I tried searching for this in the issues but didn't find it.

Was just wondering if you were aware that the URL for the Source Code link of this gem in Ruby Gems actually points to a different [prism gem](https://github.com/mwunsch/prism):

image

I tried looking at ruby gems gemspec but found no way to alter this so I'm assuming it's something done by the owners?

The Homepage link is fine, as can be seen, but still, since I usually go for the Source Code link, it threw me off a bit. Also the github badge up top links to the wrong repo.

mateusdeap:
This should fix #1738

It simply specifies the uri for the source code in the gemspec and also adds the uri for the changelog, as per @kddnewton's suggestion.

### DIFF
--- a/prism.gemspec
+++ b/prism.gemspec
@@ -118,4 +118,6 @@ Gem::Specification.new do |spec|
 
   spec.extensions = ["ext/prism/extconf.rb"]
   spec.metadata["allowed_push_host"] = "https://rubygems.org"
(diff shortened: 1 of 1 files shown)


kddnewton (approved):
Nice work!

PR #1806, opened 2023-11-10
Fix template generation of Loader.java for JRuby

enebo:
The original work nearly worked but 3 fields needed to be accessed by the extended type JRuby needs to create to properly have `getName()` return RubySymbol instead of j.l.String.

This PR additional removes an unused field when generating JRuby version of Loader.java because it is only needed for `getName()` when it returns j.l.String.

### DIFF
--- a/templates/java/org/prism/Loader.java.erb
+++ b/templates/java/org/prism/Loader.java.erb
@@ -79,16 +79,18 @@ public class Loader {
 
     private final ByteBuffer buffer;
     private final Nodes.Source source;
-    private String encodingName;
+    protected String encodingName;
+    <%- if string_type == "String" -%>
     private Charset encodingCharset;
+    <%- end -%>
     private ConstantPool constantPool;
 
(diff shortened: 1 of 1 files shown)


kddnewton (approved):

templates/java/org/prism/Loader.java.erb
```diff
@@ -79,16 +79,18 @@ public class Loader {
 
     private final ByteBuffer buffer;
     private final Nodes.Source source;
-    private String encodingName;
+    protected String encodingName;
+    <%- if string_type == "String" -%>
     private Charset encodingCharset;
+    <%- end -%>
```
eregon, line 85: Does the extra field cause some warning or so? Otherwise it might be simpler to just leave it in even if unused for simplicity.
But no big deal, fine as you changed it here.

Probably the binary->StandardCharsets.US_ASCII mapping is quite specific to j.l.String, for RubySymbol you'd want to keep it binary. So then it's nice to remove the field for non-j.l.String to remove any incorrect usage :+1:

PR #1808, opened 2023-11-11
Remove disable static option

### LINKED ISSUE #1719: undefined symbol `pm_encoding_utf_8`
I've opened an issue earlier in [Shopify/vscode-ruby-lsp](https://github.com/Shopify/vscode-ruby-lsp), but I feel like this is more appropriate here. I've described my problem in this issue over here → https://github.com/Shopify/vscode-ruby-lsp/issues/858

```plain
# nm /home/magiruuvelvet/.gem/ruby/3.1.0/gems/prism-0.15.1/lib/prism/prism.so | grep utf
                 U pm_encoding_utf_8
                 U rb_utf8_encindex
                 U rb_utf8_encoding

# nm /home/magiruuvelvet/.gem/ruby/3.1.0/gems/prism-0.15.1/build/librubyparser.so | grep utf
000000000004b7a8 d pm_encoding_utf8_mac
000000000004b778 d pm_encoding_utf_8
0000000000027360 t pm_encoding_utf_8_alnum_char
00000000000271b0 t pm_encoding_utf_8_alpha_char
0000000000027510 t pm_encoding_utf_8_char_width
00000000000275f0 t pm_encoding_utf_8_isupper_char
000000000000ff50 r pm_utf_8_dfa

# ldd /home/magiruuvelvet/.gem/ruby/3.1.0/gems/prism-0.15.1/lib/prism/prism.so
File: /home/magiruuvelvet/.gem/ruby/3.1.0/gems/prism-0.15.1/lib/prism/prism.so
Format: elf64-x86-64
Arch: x86_64
AddressSize: 64bit
OS/ABI: SystemV (0x0)
Needed Libraries:
  libc.so.6
  libjemalloc.so.2
  libruby31.so.3.1
  librubyparser.so
```

I've tried recompiling the native extension with the `-fvisibility=hidden` flag **removed**, but same issue. Is this library supposed to be shared?

**EDIT:**

I resolved this issue by recompiling `prism` with the `--enable-static` option.

`gem install prism -v 0.15.1 -- --enable-static`

kddnewton:
It's causing issues with Gentoo builds and we're not testing this path anyway. Instead, require that the C extension statically links libprism.

Fixes #1719

### DIFF
--- a/ext/prism/extconf.rb
+++ b/ext/prism/extconf.rb
@@ -1,137 +1,100 @@
 # frozen_string_literal: true
 
-require "rbconfig"
-
-module Prism
-  module ExtConf
-    class << self
-      def configure
-        unless RUBY_ENGINE == "ruby"
-          # On non-CRuby we only need the shared library, so build only that and
-          # not the C extension. We also avoid `require "mkmf"` as that prepends
-          # the LLVM toolchain to PATH on TruffleRuby, but we want to use the
-          # native toolchain here since libprism is run natively.
-          build_shared_prism
-          File.write("Makefile", "all install clean:\n\t@#{RbConfig::CONFIG["NULLCMD"]}\n")
-          return
-        end
-
-        require "mkmf"
-        configure_c_extension
-        configure_libprism
-
-        create_makefile("prism/prism")
-
-        if static_link?
-          File.open('Makefile', 'a') do |mf|
-            mf.puts
-            mf.puts '# Automatically rebuild the extension if libprism.a changed'
-            mf.puts '$(TARGET_SO): $(LOCAL_LIBS)'
-          end
-        end
-      end
-
-      def configure_c_extension
-        append_cflags("-DPRISM_DEBUG_MODE_BUILD") if debug_mode_build?
-        append_cflags("-fvisibility=hidden")
-      end
-
-      def configure_libprism
-        if static_link?
-          static_archive_path = File.join(build_dir, "libprism.a")
-          unless File.exist?(static_archive_path)
-            build_static_prism
-          end
-          $LOCAL_LIBS << " #{static_archive_path}"
-        else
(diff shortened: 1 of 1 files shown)


eregon (approved):
Looks good

ext/prism/extconf.rb
```diff
@@ -1,136 +1,101 @@
...
+unless find_header("prism.h", File.expand_path("../../include", __dir__))
+  raise "prism.h is required"
 end
 
-Prism::ExtConf.configure
+# Next, ensure we can find the header for the C extension. Explicitly look for
+# the extension header in the parent directory because we want to consistently
+# look for `prism/extension.h` in our source files to line up with our mirroring
+# in CRuby.
+unless find_header("prism/extension.h", File.join(__dir__, ".."))
```
eregon, line 65: FWIW File.join just adds `/` even on Windows, so it's the same as `"#{__dir__}/.."` (which IMO is clearer, but subjective).
Alternatively `File.expand_path("..", __dir__)` like above might be nicer as it resolves the `..`.

ext/prism/extconf.rb
```diff
@@ -1,136 +1,101 @@
...
+end
+
+# By default, all symbols are hidden in the shared library.
+append_cflags("-fvisibility=hidden")
+
+# We need to link against the librubyparser.a archive, which is built by the
+# project's `Makefile`. We'll build it if it doesn't exist yet, and then add it
+# to `mkmf`'s list of local libraries.
+archive_target = "build/librubyparser.a"
+archive_path = File.expand_path(File.join("../..", archive_target), __dir__)
```
eregon, line 85:
```suggestion
archive_path = File.expand_path("../../#{archive_target}", __dir__)
```

eregon (approved):

PR #1809, opened 2023-11-12
Add character APIs for locations

### LINKED ISSUE #1788: Location APIs for characters
We store everything as an offset of bytes from the start of the source. This is correct and not going to change.

However lots of our consumers are dealing with encoded source strings and would like to be able to use the `[]` method without having to convert into # of characters. (Debatably, this should be called "grapheme_clusters", but I'm going with characters for now because it's shorter.)

The APIs that would be nice to provide:

* `Location#start_character_offset`
* `Location#end_character_offset`
* `Location#start_character_column`
* `Location#end_character_column`

A simple implementation of these would be:

```ruby
class Location
  def start_character_offset = source.source.byteslice(0, start_offset).length
  def end_character_offset = source.source.byteslice(0, end_offset).length
  def start_character_column = source.slice(source.line_offset(start_offset), start_column).length
  def end_character_column = source.slice(source.line_offset(end_offset), end_column).length
end
```

However, this would mean a lot of things being recalculated any time any of these APIs were called. My hypothesis is that if you're calling these APIs once, it's likely you'll call them again. In this case I think we should implement some caching on the `Source` object for `byte offset` => `character offset` and use that cache in these computations.

kddnewton:
Fixes #1788 by introducing:

* `Location#start_character_offset`
* `Location#end_character_offset`
* `Location#start_character_column`
* `Location#end_character_column`

### DIFF
--- a/lib/prism/parse_result.rb
+++ b/lib/prism/parse_result.rb
@@ -25,40 +25,50 @@ module Prism
 
     # Perform a byteslice on the source code using the given byte offset and
     # byte length.
-    def slice(offset, length)
-      source.byteslice(offset, length)
+    def slice(byte_offset, length)
+      source.byteslice(byte_offset, length)
     end
 
     # Binary search through the offsets to find the line number for the given
     # byte offset.
-    def line(value)
-      start_line + find_line(value)
+    def line(byte_offset)
+      start_line + find_line(byte_offset)
     end
 
     # Return the byte offset of the start of the line corresponding to the given
     # byte offset.
-    def line_offset(value)
-      offsets[find_line(value)]
+    def line_start(byte_offset)
+      offsets[find_line(byte_offset)]
     end
 
     # Return the column number for the given byte offset.
-    def column(value)
-      value - offsets[find_line(value)]
+    def column(byte_offset)
+      byte_offset - line_start(byte_offset)
+    end
+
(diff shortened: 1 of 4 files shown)


lib/prism/parse_result.rb
```diff
@@ -21,44 +21,87 @@ def initialize(source, start_line = 1, offsets = compute_offsets(source))
...
+    def character_column(byte_offset)
+      character(byte_offset) - character(line_start(byte_offset))
     end
 
     private
 
+    # A lazily-computed cache of byte offset => character offset mappings.
+    def character_offsets
+      @character_offsets ||= { 0 => 0 }
+    end
```
eregon, line 65: I'm not sure this would pay off because it could become pretty large for a big file and consume N bytes for every byte in the source.
N seems to be about 60 for integer entries in a Hash:
```
ruby --disable-gems -e 'h={}; 1.times {|i| h[i]=i}'
12MB
ruby --disable-gems -e 'h={}; 10_000_000.times {|i| h[i]=i}'
588MB
irb(main):003:0> 576 * 1024**2 / 10_000_000.0
=> 60.3979776
```

I think it's better to leave this uncached for now, or do one cache entry per line.

kddnewton: Fair point!

eregon (approved):

PR #1812, opened 2023-11-13
Add "Unexpected void value expression" error

makenowjust:
TODO:

- [x] Add "Unexpected void value expression" error
- [x] Add a check function for void value expression
- [x] Add assertions to the parser
- [x] Add tests


kddnewton (approved):
Nice!

PR #1822, opened 2023-11-13
Ensure serialized file is little endian

kddnewton:
So that it's portable to other machines

### DIFF
--- a/templates/java/org/prism/Loader.java.erb
+++ b/templates/java/org/prism/Loader.java.erb
@@ -54,6 +54,7 @@ public class Loader {
 
             if (constant == null) {
                 int offset = bufferOffset + index * 8;
+
                 int start = buffer.getInt(offset);
                 int length = buffer.getInt(offset + 4);
 
@@ -86,7 +87,7 @@ public class Loader {
     private ConstantPool constantPool;
 
     protected Loader(byte[] serialized, Nodes.Source source) {
-        this.buffer = ByteBuffer.wrap(serialized).order(ByteOrder.nativeOrder());
+        this.buffer = ByteBuffer.wrap(serialized).order(ByteOrder.LITTLE_ENDIAN);
         this.source = source;
     }
 
--- a/templates/src/serialize.c.erb
+++ b/templates/src/serialize.c.erb
@@ -47,6 +47,21 @@ pm_serialize_string(pm_parser_t *parser, pm_string_t *string, pm_buffer_t *buffe
     }
 }
 
+/**
+ * Serialize a 32-bit integer to the given address always in little-endian.
+ */
+static void
+pm_serialize_32(char *address, uint32_t value) {
+#ifdef PRISM_WORDS_BIGENDIAN
+    address[0] = (char) ((value >> 24) & 0xFF);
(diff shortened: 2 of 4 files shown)


templates/src/serialize.c.erb
```diff
@@ -47,6 +47,31 @@ pm_serialize_string(pm_parser_t *parser, pm_string_t *string, pm_buffer_t *buffe
...
+static void
+pm_serialize_32(char *address, uint32_t value) {
+    static bool little_endian = false;
+    static bool endian_check = false;
+
+    if (!endian_check) {
+        uint32_t test = 0x1;
+        little_endian = *((uint8_t *) &test) == 0x1;
+        endian_check = true;
+    }
```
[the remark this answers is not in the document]
eregon: CRuby seems to use `WORDS_BIGENDIAN` which seems defined as:
```c
#ifdef AC_APPLE_UNIVERSAL_BUILD
# undef WORDS_BIGENDIAN
# ifdef __BIG_ENDIAN__
#  define WORDS_BIGENDIAN
# endif
#endif
```

templates/java/org/prism/Loader.java.erb
```diff
@@ -54,6 +54,8 @@ public class Loader {
 
             if (constant == null) {
                 int offset = bufferOffset + index * 8;
+
+                // Read an integer in forced little endian.
```
eregon, line 58: I would remove this, that should be clear from the `this.buffer = ByteBuffer.wrap(serialized).order(ByteOrder.LITTLE_ENDIAN);` line below.

eregon (commented):
Looks good.
One improvement which I think is important is to use the preprocessor like `#ifndef WORDS_BIGENDIAN` instead of a runtime check like `if (little_endian) {`.

templates/src/serialize.c.erb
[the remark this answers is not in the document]
eregon: From https://github.com/ruby/ruby/blob/e6b2cd15e34cbeaf07d8ced2db89d02272e9d5af/include/ruby/internal/config.h#L85-L90

seven1m (commented):
Nice!

templates/src/serialize.c.erb
```diff
@@ -47,6 +47,21 @@ pm_serialize_string(pm_parser_t *parser, pm_string_t *string, pm_buffer_t *buffe
     }
 }
 
+/**
+ * Serialize a 32-bit integer to the given address always in little-endian.
+ */
+static void
+pm_serialize_32(char *address, uint32_t value) {
+    if (PRISM_WORDS_BIGENDIAN) {
```
eregon, line 55:
```suggestion
#ifdef PRISM_WORDS_BIGENDIAN
```
So the C compiler doesn't even see the code for the wrong endian.

kddnewton: 👍

eregon:
FWIW it seems ruby/ruby has RB_IO_BUFFER_HOST_ENDIAN, so we could likely use similar logic as that:
https://bugs.ruby-lang.org/issues/20431#note-5

Issue #1814, opened 2023-11-13
Ruby rejects empty group name in regexp `/(?<>)/` but prism accepts it


tompng:
```ruby
eval "/(?<>)/" # group name is empty: /(?<>)/ (SyntaxError)
eval "/(?<\\\n>)/" # group name is empty: /(?<>)/ (SyntaxError)
eval "/(?<\n>)/" # syntax ok. group name is "\n"

Prism.parse "/(?<>)/" # success (should be error)
Prism.parse "/(?<\\\n>)/" # success (should be error)
Prism.parse "/(?<\n>)/" # success (ok)
```

makenowjust:
I think it should not be rejected for now because Prism does not parse a regex correctly (However, it is analyzed for computing locals.)

PR #1836, opened 2023-11-15
Don't add an invalid identifier capture to locals

### LINKED ISSUE #1815: MatchWriteNode `/(?<a b>)/ =~ s` has wrong locals
For `/regexp_with_named_capture/ =~ s`, group name which is not valid as a local variable name has no effect.
```ruby
/(?<a>.)(?<@a>.)(?<$a>.)(?<a b>.)/ =~ '12345'
local_variables #=> ["a"]
instance_variables #=> []
global_variables.include?(:$a) #=> false
```

Prism's MatchWriteNode has wrong locals
```ruby
match_write_node = Prism.parse("/(?<a>.)(?<@a>.)(?<$a>.)(?<a b>.)/ =~ '12345'").value.statements.body[0]
match_write_node.locals
# => [:a, :@a, :$a, :"a b"]
```

Also, for `Prism.parse("/(?<a>)(?<a>)/ =~ s")`, locals are `[:a, :a]`. (I don't know if this is correct or wrong)

makenowjust:
Fix #1815

I think #1814 is partially fixed because this change makes the empty capture name `:""` not to be added to locals.


src/prism.c
```diff
@@ -15383,6 +15402,10 @@ parse_regular_expression_named_captures(pm_parser_t *parser, const pm_string_t *
             const uint8_t *source = pm_string_source(name);
             size_t length = pm_string_length(name);
 
+            if (!name_is_identifier(parser, source, length)) {
```
kddnewton, line 15405: I think this is good, but in this case it's possible to have a `MatchWriteNode` when there aren't actually any matches. In this case I think we shouldn't replace the call node with a match write. Instead, we could lazily create the `match` node once we are sure we have a named capture group that is valid, and then only set `result = match` if one was created.

kddnewton (changes requested):

kddnewton (approved):

PR #1838, opened 2023-11-15
Fix associativity of binary range with begin-less range

### LINKED ISSUE #1828: Ruby rejects `..1..` but Prism accepts it
Prism seems to parse as a combination of beginless range and endless range
```ruby
Prism.parse("..1..").success? #=> true
eval('..1..') #=> syntax error, unexpected .. (SyntaxError)
```

makenowjust:
Fix #1828


src/prism.c
```diff
@@ -16757,6 +16760,14 @@ parse_expression(pm_parser_t *parser, pm_binding_power_t binding_power, pm_diagn
         return node;
     }
 
+    // Range operators are non-associative, so that it does not associate with
+    // other range operators (i.e. `..1..` should be rejected.)
+    // For this reason, we check such a case for unary ranges here, and if so,
+    // it returns the node immediately,
+    if (is_udot && pm_binding_powers[parser->current.type].left >= PM_BINDING_POWER_RANGE) {
```
kddnewton, line 16767: I'd like to combine this work with #1885 and make it part of the same switch statement. Once that's merged, let's check if the node type is a range or flipflop, and if it is if it doesn't have a left. In that case we could return the node explicitly here like this is doing. That makes it a little less work since we won't be building the `is_udot` for every single node that goes through `parse_expression`.

kddnewton:
Merged in with #1909

Issue #1829, opened 2023-11-14
Ruby rejects `a....b` but Prism accepts it.


tompng:
It's parsed as `(a...).b`
```ruby
Prism.parse('a....b').success? #=> true
eval('a....b') #=> syntax error, unexpected '.', expecting end-of-input (SyntaxError)
```

nevinera:
Well, yours definitely looks simpler, unless having specific error messages is worth some trouble here :-)

I'd love it if someone would pick through my PR even if it's not destined for a merge - I'm just getting started on the repo, and I'd like to improve my approaches, mergeable or not.

makenowjust:
@nevinera This bug is because the associativity of `..` and `...` is wrong. So, it would need to be fixed in my way.

Modifying my implementation to produce a specific error message is not hard. You may modify it after it has been merged. (Honestly, I'm not sure it is helpful for developers because `....` seems rare and terrible and developers can easily find this without the specific error.)

nevinera:
I agree, and would not have posted my pr if I had seen yours first. I left it up in draft because I wanted feedback on it, but certainly it should not be merged. :-)

PR #1840, opened 2023-11-15
Add a parsing error when attempting to call a method directly on an unterminated range

nevinera:
(Resolves https://github.com/ruby/prism/issues/1829)

* First, add a new range flag called `END_NOT_SUPPLIED`, indicating not an "unterminated" range, but a range that had no right-hand value _specified_.
  - update the snapshots to reflect this additional flag
  - My C is rusty - is there a better way to do handle this optional argument?
* Add a check when handing an infix `.` - if the `node` is a range and has the END_NOT_SUPPLIED flag set, add a parser error (but keep going).
  - test the variations I thought of, and confirm that ripper fails all of those cases as well.


enebo:
@nevinera this is to satisfy how things are bound when looking at the tree in Ruby right?  I am not saying we don't need flags for this but in C and Java the tree is unambiguous (it is null and not nil).  It seems that in the cases of (`b..` or `..b`) the nulls are translated to nil for Ruby bindings.  In this sense I half wonder what other tree structures have this problem?

In JRuby internally we mark these kinds of ranges as `beginless` and `endless`.

enebo:
Sorry I misread the intent.  Unless I didn't in which you can still answer :)

  I think this PR is to capture that no end was specified on a failing parse.  If that is the case then I think the convention is to put a MissingNode where the end should be placed.

nevinera:
I was trying to deal with the parsing ambiguity - it seemed like the cruby parser wasn't willing to accept quite a few things after `a..`/`a...` that work fine after `a..nil`/`a...nil`. Based on makenowjust's solution, I gather that was just an associativity issue, and not one that needed special handling; that probably should have been obvious even without the understanding, since I already knew that Ripper/Prism were _lexing_ these the same way.

We definitely don't need flags for this - I had added it in the PR fundamentally because I'm having trouble recognizing the boundary between lexing and parsing in the C code.

nevinera:
I'm just new, both to the project and to language-parsing in general. I tried to tackle the most specific-seeming issue I could find, expecting not to come up with a good solution, but perhaps to work one out with someone. But it looks like I stepped into a big swatch of associativity-related tickets makenowjust is mowing through in a row :-)

PR #1842, opened 2023-11-15
Add CP51932 encoding 

wildmaples:
https://www.iana.org/assignments/charset-reg/CP51932

Resolves: https://github.com/ruby/prism/issues/1844


kddnewton (approved):

kddnewton:
Nice job @wildmaples!

PR #1848, opened 2023-11-16
Silence clang analyzer warnings for the memory leaks

haldun:
clang analyzer reports two false positives about leaking memory. This PR silences these since we know that these are not actual memory leaks.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -10327,6 +10327,8 @@ parse_write_name(pm_parser_t *parser, pm_constant_id_t *name_field) {
     name[length] = '=';
 
     // Now switch the name to the new string.
+    // This silences clang analyzer warning about leak of memory pointed by `name`.
+    // NOLINTNEXTLINE(clang-analyzer-*)
     *name_field = pm_constant_pool_insert_owned(&parser->constant_pool, name, length + 1);
 }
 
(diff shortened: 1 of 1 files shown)


src/prism.c
```diff
@@ -10328,6 +10328,11 @@ parse_write_name(pm_parser_t *parser, pm_constant_id_t *name_field) {
 
     // Now switch the name to the new string.
     *name_field = pm_constant_pool_insert_owned(&parser->constant_pool, name, length + 1);
+
+    #ifdef __clang_analyzer__
```
kddnewton, line 10332: I'm not sure how I feel about this addition, because it would actually case a double free. It's definitely going to be freed by the constant pool, so we're introducing a bug here just to satisfy the analyzer. Is there a way to put this into a config file instead?

kddnewton (changes requested):

src/prism.c
haldun: I couldn't find another way. I see usages of the symbol `__clang_analyzer__` in other projects so I think maybe it is OK. But I definitely agree with you on this may not be an ideal solution.

eregon:
Is it a bug of the clang analyzer then? Or a known limitation or something?

Maybe we need to annotate pm_constant_pool_insert_owned/pm_parser_local_add_owned to clarify they take ownership of the allocation? That would be less invasive than annotating every usage of them.

src/prism.c
haldun: OK, it seems like adding `// NOLINT` to the reported line also silences these warnings.

kddnewton (approved):

PR #1849, opened 2023-11-16
Optimize `lex_keyword`

haldun:
- Instead of testing `parser->current.start + vlen <= parser->end` every time in `lex_keyword`, I moved the test to `lex_identifier`.
- Pass already cached value of `parser->current.start` to `lex_keyword` instead of accessing it directly within `lex_keyword`.
- Updated documentation comment for `lex_keyword`.

These changes seem to bring ~3% improvement.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -6725,17 +6725,21 @@ lex_global_variable(pm_parser_t *parser) {
 
 /**
  * This function checks if the current token matches a keyword. If it does, it
- * returns true. Otherwise, it returns false. The arguments are as follows:
+ * returns the token type. Otherwise, it returns PM_TOKEN_EOF. The arguments are as follows:
  *
+ * * `parser` - the parser object
+ * * `current_start` - pointer to the start of the current token
  * * `value` - the literal string that we're checking for
- * * `width` - the length of the token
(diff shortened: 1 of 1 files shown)


src/prism.c
```diff
@@ -6817,69 +6821,68 @@ lex_identifier(pm_parser_t *parser, bool previous_command_start) {
         }
     }
 
-    if (parser->lex_state != PM_LEX_STATE_DOT) {
+    if ((parser->lex_state != PM_LEX_STATE_DOT) && (current_start + width <= end)) {
```
kddnewton, line 6824: Shouldn't this always be the case? If we already got this far into the identifier, then `current_start + width` would necessarily have been before the end.

kddnewton (changes requested):

src/prism.c
haldun: Let me try.

haldun: OK, I think you are right, I removed that test in https://github.com/ruby/prism/pull/1849/commits/23a68dcda2493b95f9be43de9250d4f9e04e39e0

kddnewton (approved):

Issue #1843, opened 2023-11-15
Implement remaining encodings


kddnewton:
Encodings are sets of configuration that describe how bytes map to codepoints. Some of the more well-known encodings include `US-ASCII`, `UTF-8`, and `Shift_JIS`.

Ruby source files can be encoded in 90 different encodings, depending on the presence of a `# encoding: xxx` magic comment at the top of the source file. Encoding support is necessary to properly parse identifiers, constants, and string content. As such, we need to support all of the encodings that CRuby supports.

Fortunately, we don't need as much information as CRuby does for each encoding. We only need:

* whether or not a codepoint is alphabetical
* whether or not a codepoint is alphanumeric
* whether or not a codepoint is uppercase

For single byte encodings (i.e., encodings that represent every codepoint with at most one byte) we put all of their configuration into `src/enc/pm_tables.c`. These contains sets of lookup tables for each encoding. Because they're all limited to a single byte, we use arrays of `uint8_t` integers of length 256. Each integer is a bitmap containing the three bits of information listed above. For an example PR on how to add one of these encodings, see #1851.

For multi-byte encodings (i.e., encodings that can represent codepoints with more than one byte) we put them into their own files under `src/enc/` because they require more logic. For the most part it involves writing the `pm_encoding_*_char_width` function to decode the next codepoint from the given `uint8_t` slice. For an example PR on how to add one of these encodings, see #1844.

The tasks listed below are the encodings we still have to implement. Some of these will be easier than others, since some of them are single byte. I don't actually know yet which ones are since I'm not familiar with any of these - my personal process is to go straight to wikipedia to figure it out. If you want to take on a particular encoding, make a comment on this issue and we'll pull it out into its own issue to track your work.

* [x] `Big5-HKSCS`
* [x] `Big5-UAO`
* [x] `CESU-8`
* [x] #1844
* [x] `CP949`
* [x] `CP950`
* [x] `EUC-JIS-2004`
* [x] `EUC-KR`
* [x] `EUC-TW`
* [x] `Emacs-Mule`
* [x] `GB12345`
* [x] `GB18030`
* [x] `GB1988`
* [x] `GB2312`
* [x] #1856
* [x] `IBM864`
* [x] `IBM865`
* [x] #1866
* [x] `IBM869`
* [x] `KOI8-U`
* [x] `MacJapanese`
* [x] `SJIS-DoCoMo`
* [x] `SJIS-KDDI`
* [x] `SJIS-SoftBank`
[...]

PR #1854, opened 2023-11-17
add Windows-874 encoding

### LINKED ISSUE #1857: `Windows-874`

pcai:
aka cp874. 

874 is a derivative of ISO-8859-11 but functionally equivalent for the required use case.

https://github.com/ruby/prism/issues/1843

Fixes #1857

### DIFF
--- a/bin/encodings
+++ b/bin/encodings
@@ -96,5 +96,6 @@ when "windows-1256" then table(Encoding::Windows_1256)
 when "windows-1257" then table(Encoding::Windows_1257)
 when "windows-1258" then table(Encoding::Windows_1258)
 when "windows-31j"  then lists("windows-31j", 0...0x10000, Encoding::Windows_31J)
+when "windows-874"  then table(Encoding::Windows_874)
 else raise "Unknown encoding `#{ARGV[0]}'"
 end
--- a/test/prism/encoding_test.rb
+++ b/test/prism/encoding_test.rb
@@ -50,6 +50,7 @@ module Prism
       Encoding::Windows_1256 => 0x00...0x100,
       Encoding::Windows_1257 => 0x00...0x100,
       Encoding::Windows_1258 => 0x00...0x100,
+      Encoding::Windows_874 =>  0x00...0x100,
       Encoding::Big5 =>         0x00...0x10000,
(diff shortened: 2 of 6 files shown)


bin/encodings
```diff
@@ -52,6 +52,7 @@ when "big5"         then lists("big5", 0...0x10000, Encoding::Big5)
 when "cp850"        then table(Encoding::CP850)
 when "cp852"        then table(Encoding::CP852)
 when "cp855"        then table(Encoding::CP855)
+when "cp874"        then table(Encoding::CP874)
```
kddnewton, line 55: Typically we're only putting the main constants in here, not the aliases. `Encoding::Windows_874` is the main one, `Encoding::CP874` is the alias.

test/prism/encoding_test.rb
```diff
@@ -12,6 +12,7 @@ class EncodingTest < TestCase
       Encoding::CP850 =>        0x00...0x100,
       Encoding::CP852 =>        0x00...0x100,
       Encoding::CP855 =>        0x00...0x100,
+      Encoding::CP874 =>        0x00...0x100,
```
kddnewton, line 15: Since this is an alias, we don't need to test this one.

test/prism/encoding_test.rb
```diff
@@ -54,7 +55,8 @@ class EncodingTest < TestCase
       Encoding::CP51932 =>      0x00...0x10000,
       Encoding::GBK =>          0x00...0x10000,
       Encoding::Shift_JIS =>    0x00...0x10000,
-      Encoding::Windows_31J =>  0x00...0x10000
+      Encoding::Windows_31J =>  0x00...0x10000,
+      Encoding::Windows_874 =>  0x00...0x10000
```
kddnewton, line 59: The maximum codepoint for `Windows_874` is `0xFF`, so we should put this up in the hash with the other 1-byte encodings.

kddnewton (changes requested):
Just a couple of small changes, nice work!

Other thing missing is adding this encoding to the list in `encoding.md`

kddnewton (approved):
Nice work!

kddnewton:
Nice work @pcai!

Screenshot 2023-11-17 at 3 07 47 PM

PR #1867, opened 2023-11-18
Big5 HKSCS encoding

rgarver:
Attempting to make a dent on https://github.com/ruby/prism/issues/1843

This is the first time I've touched C code in over 10 years. Be gentle.

I attempted to match the overall end-point of the existing Ruby implementation for this. This is not much different from the Big5 encoding.

This compiles and passes tests for me, but I'm not particularly familiar with encodings or modern C so I'm sure I missed something.

### DIFF
--- a/include/prism/enc/pm_encoding.h
+++ b/include/prism/enc/pm_encoding.h
@@ -158,6 +158,7 @@ extern const uint8_t pm_encoding_unicode_table[256];
 extern pm_encoding_t pm_encoding_ascii;
 extern pm_encoding_t pm_encoding_ascii_8bit;
 extern pm_encoding_t pm_encoding_big5;
+extern pm_encoding_t pm_encoding_big5_hkscs;
 extern pm_encoding_t pm_encoding_cp51932;
 extern pm_encoding_t pm_encoding_cp850;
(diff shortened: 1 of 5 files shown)


kddnewton (approved):
Looks good to me! (Don't worry, CRuby is still C99 so definitely not modern C.)

This is going to fail the CRuby tests because it's a new file, so I'll merge it into there first and then merge this PR.

rgarver:
I was waiting for this PR to pass before picking up Big5-UAO to make sure I had the pattern for these updates correct. I'll try to get that out soon if no one else gets to it first.

Thanks!

kddnewton:
> I was waiting for this PR to pass before picking up Big5-UAO to make sure I had the pattern for these updates correct. I'll try to get that out soon if no one else gets to it first.

Please do!

rgarver:
@kddnewton I think this may be pretty much identical to Big5-UAO. How do you feel about me renaming some of this to `...big5_star...` and producing two `pm_encoding_t` from it?

kddnewton:
@rgarver that sounds good - I'll merge this one, and if you want feel free to make that refactor as part of another PR

PR #1870, opened 2023-11-18
Add macCentEuro encoding

thomasmarshall:
Related #1843. 

This PR implements the [`macCentEuro` encoding](https://en.wikipedia.org/wiki/Mac_OS_Central_European_encoding). The lookup table is identical to `macRoman`, upon which `macCentEuro` is based.


kddnewton (approved):
Looks great!

PR #1873, opened 2023-11-18
Fix `bin/lex` when encountering mismatch at EOF.

martinemde:
When encountering a situation where there is a mismatch at the very end of a file, and when `Prism.lex` produces a shorter output than `.lex_compat` and/or `.lex_ripper`, `bin/lex` itself fails with NoMethodError when calling `[0]` on `new` which is `nil`. This prevents outputting the actual mismatch.

This fix allows `new` to be `nil`, outputing `nil` in the 3rd column of `bin/lex` in this situation. (I can update this output approach if needed.)

### DIFF
--- a/bin/lex
+++ b/bin/lex
@@ -55,6 +55,7 @@ prism_new_value = prism_new.value
   color = left == right ? "38;5;102" : "1;31"
 
   if ENV["VERBOSE"] || (left != right)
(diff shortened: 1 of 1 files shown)


kddnewton (approved):

martinemde:
Thanks Kevin, and thanks for the talk at RubyConf that got me inspired to check out the project.

kddnewton:
No problem - thanks for the contributions!

PR #1877, opened 2023-11-19
Allow `&` forwarding in a method having `...` parameter

### LINKED ISSUE #1839: `...` enables forwarding of `&`
One more forwarding edge case :(

```ruby
# On CRuby 3.2.2
def bar; block_given?; end
def foo(...) bar(&) end
foo # => false
foo {} # => true
```

But 

```
bin/parse -e 'def foo(...) bar(&) end'
Errors:
[#<Prism::ParseError @message="Unexpected `&` when the parent method is not forwarding" @location=#<Prism::Location @start_offset=17 @length=1 start_line=1>>]

AST:
@ ProgramNode (location: (1,0)-(1,23))
├── locals: []
└── statements:
    @ StatementsNode (location: (1,0)-(1,23))
    └── body: (length: 1)
        └── @ DefNode (location: (1,0)-(1,23))
            ├── name: :foo
            ├── name_loc: (1,4)-(1,7) = "foo"
            ├── receiver: ∅
            ├── parameters:
            │   @ ParametersNode (location: (1,8)-(1,11))
            │   ├── requireds: (length: 0)
            │   ├── optionals: (length: 0)
            │   ├── rest: ∅
            │   ├── posts: (length: 0)
            │   ├── keywords: (length: 0)
            │   ├── keyword_rest:
            │   │   @ ForwardingParameterNode (location: (1,8)-(1,11))
            │   └── block: ∅
            ├── body:
            │   @ StatementsNode (location: (1,13)-(1,18))
            │   └── body: (length: 1)
            │       └── @ CallNode (location: (1,13)-(1,18))
            │           ├── receiver: ∅
            │           ├── call_operator_loc: ∅
            │           ├── message_loc: (1,13)-(1,16) = "bar"
            │           ├── opening_loc: (1,16)-(1,17) = "("
            │           ├── arguments: ∅
            │           ├── closing_loc: (1,18)-(1,19) = ")"
            │           ├── block:
            │           │   @ BlockArgumentNode (location: (1,17)-(1,18))
            │           │   ├── expression: ∅
            │           │   └── operator_loc: (1,17)-(1,18) = "&"
            │           ├── flags: ∅
            │           └── name: :bar
            ├── locals: [:"..."]
            ├── def_keyword_loc: (1,0)-(1,3) = "def"
            ├── operator_loc: ∅
            ├── lparen_loc: (1,7)-(1,8) = "("
            ├── rparen_loc: (1,11)-(1,12) = ")"
            ├── equal_loc: ∅
            └── end_keyword_loc: (1,20)-(1,23) = "end"
```

makenowjust:
Fix #1839


src/prism.c
```diff
@@ -5687,12 +5687,16 @@ pm_parser_scope_push_transparent(pm_parser_t *parser) {
 static int
 pm_parser_local_depth(pm_parser_t *parser, pm_token_t *token) {
     pm_constant_id_t constant_id = pm_parser_constant_id_token(parser, token);
+    pm_constant_id_t ellipsis_id = token->type == PM_TOKEN_UAMPERSAND ? pm_parser_constant_id_constant(parser, "...", 3) : 0;
```
haldun, line 5690: What do you think about passing this as a parameter so the logic of this function stays the same but the caller sites can change its behavior?

makenowjust: Indeed, that sounds good.

src/prism.c
```diff
@@ -11176,8 +11176,14 @@ parse_arguments(pm_parser_t *parser, pm_arguments_t *arguments, bool accepts_for
 
                 if (token_begins_expression_p(parser->current.type)) {
                     expression = parse_value_expression(parser, PM_BINDING_POWER_DEFINED, PM_ERR_EXPECT_ARGUMENT);
-                } else if (pm_parser_local_depth(parser, &operator) == -1) {
-                    pm_parser_err_token(parser, &operator, PM_ERR_ARGUMENT_NO_FORWARDING_AMP);
+                } else {
+                    if (pm_parser_local_depth(parser, &operator) == -1) {
```
kddnewton, line 11180: What do you think about adding `&` to the local table instead of checking it here? That would make it so that this first check would work.

kddnewton (changes requested):

src/prism.c
makenowjust: Please see https://github.com/ruby/prism/issues/1839#issuecomment-1822133511.

kddnewton (approved):
@makenowjust could you rebase?

kddnewton:
Merged in with #1909

PR #1880, opened 2023-11-19
feat: Adds macCroatian encoding

davidwessman:
- Based on:
  https://en.wikipedia.org/wiki/Mac_OS_Croatian_encoding
  https://www.unicode.org/Public/MAPPINGS/VENDORS/APPLE/CROATIAN.TXT

Questions:

- ~Would it make sense to add what the values mean in the bitfields for `src/enc/pm_tables.c`? Would make it easier for contributions atleast.~
- ~Should greek letters be included as well? Based on the unicode document it said `letter` so we considered it alphanumeric.~
- ~We were thrown off by some of the added values for `mac*` encodings because a lot of them have `0` for the codepoints 128-255 - which seems wrong for most of them. Should they be updated?~


kddnewton:
> Would it make sense to add what the values mean in the bitfields for src/enc/pm_tables.c? Would make it easier for contributions atleast.

They're defined in `pm_encoding.h` and in the C documentation. I'm hesitant to also add them to the source in case things change, as we would be duplicating documentation that would then need to be updated. In general the more verbose documentation goes in the headers in this project.

> Should greek letters be included as well? Based on the unicode document it said letter so we considered it alphanumeric.

I can see how you would think that – it makes sense! But it's not necessarily which characters we see as alphanumeric as much as which characters are alphanumeric according to ctype/locale. I'm trusting onigmo to be right on this for this case.

> We were thrown off by some of the added values for mac* encodings because a lot of them have 0 for the codepoints 128-255 - which seems wrong for most of them. Should they be updated?

Nope, that's correct! It means none of them are considered alpha/alnum/upper, which is fine.

kddnewton (changes requested):
Could you add `macCroation` to the list of encodings we support in `encoding.md`?

kddnewton (approved):

PR #1881, opened 2023-11-19
fix: Updates Mac Roman encoding table

davidwessman:
Based on:
https://en.wikipedia.org/wiki/Mac_OS_Roman

Co-authored-by: Josefine Rost <nijrost@gmail.com>


orhantoy:
I worked on https://github.com/ruby/prism/pull/1875 and I had a similar observation that the output from `bin/encodings` did not match with what is actually a letter or whether or not it is uppercase according to the encoding specs. The regex matches used for the bitmask

https://github.com/ruby/prism/blob/635f595a362c7ba800674409832fd22d85624f92/bin/encodings#L12-L15

seems to only match latin letters.

I still went with the output from calling `bin/encodings maccyrillic` because that's what was done for other similar encodings 🤷

davidwessman:
Hm, I did not get that was what `bin/encodings` could be used for 🙈 
@kddnewton what is your conclusion?

duerst:
@davidwessman @kddnewton As can be seen from https://github.com/ruby/ruby/blob/master/enc/ascii.c#L102, the MacRoman is just a copy of ASCII-8BIT, which means that when used for source code, only ASCII characters will work for identifiers. It has its own transcoding table (see https://github.com/ruby/ruby/blob/master/enc/trans/macroman-tbl.rb), so it can be used when reading/writing data. Given that MacRoman is quite old and most probably no longer used for new stuff, that feels about right. The same applies for the other Mac-related encodings.

kddnewton:
@davidwessman yeah this is fine. Once all of the encodings are pulled in, I'll create the alias functionality and I assume we'll be able to drop a lot of the lookup tables. For now while it's all being worked on in parallel though, this is fine.

kddnewton:
Ahh yeah I see, yeah we shouldn't merge this PR as we don't want this additional information. It should only be what's coming out of `bin/encodings`.

PR #1883, opened 2023-11-19
Fix heredocs with leading spaces

### LINKED ISSUE #753: Fix heredocs with leading spaces
```
puts <<-'    ABC'
marker with spaces
    ABC
```

```
puts <<~'    ABC'
marker with spaces
    ABC
```

Worth noting that this behavior recently changed upstream:
* [bug report](https://bugs.ruby-lang.org/issues/19539)
* [fix](https://github.com/ruby/ruby/commit/67dd52d59cde0d2f1ebb3e299b605ed239b59f5b)

haldun:
Closes #753

@kddnewton I am not 100% happy with it but it solves the issue.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -9770,6 +9770,17 @@ parser_lex(pm_parser_t *parser) {
             const uint8_t *ident_start = lex_mode->as.heredoc.ident_start;
             size_t ident_length = lex_mode->as.heredoc.ident_length;
 
+            // If we have a - or ~ heredoc, then we should adjust the start and the
+            // length of the ident so that we can match the lines.
+            if (
+                lex_mode->as.heredoc.indent == PM_HEREDOC_INDENT_TILDE ||
+                lex_mode->as.heredoc.indent == PM_HEREDOC_INDENT_DASH
+            ) {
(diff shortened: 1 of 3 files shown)


src/prism.c
```diff
@@ -9770,6 +9770,17 @@ parser_lex(pm_parser_t *parser) {
...
 
+            // If we have a - or ~ heredoc, then we should adjust the start and the
+            // length of the ident so that we can match the lines.
+            if (
+                lex_mode->as.heredoc.indent == PM_HEREDOC_INDENT_TILDE ||
+                lex_mode->as.heredoc.indent == PM_HEREDOC_INDENT_DASH
+            ) {
+                size_t ident_leading_whitespace = pm_strspn_inline_whitespace(ident_start, (ptrdiff_t)ident_length);
+                ident_start += ident_leading_whitespace;
+                ident_length -= ident_leading_whitespace;
```
kddnewton, line 9781: I don't think this is going to work quite as we want. This is going to shrink the heredoc declaration such that only the main content is present, but that ignores the requirement that whitespace be present in the prefix.

So for example, with this PR, this is going to parse correctly:

```ruby
<<~'  FOO'
  hi
FOO
```

when it shouldn't.

kddnewton (changes requested):

PR #1884, opened 2023-11-19
feat: add encoding for IBM865

derekcmoore:
For: https://github.com/ruby/prism/issues/1843

Fixes: (issue for encoding will be added once created)

Documentation: https://en.wikipedia.org/wiki/Code_page_865


kddnewton (approved):
Looks great!

PR #1886, opened 2023-11-20
feat: adds encodings for ibm869

lynne-ashminov:
Adds encoding for IBM869 per https://github.com/ruby/prism/issues/1843

Documentation: https://en.wikipedia.org/wiki/Code_page_869


kddnewton (approved):
Looks right!

PR #1891, opened 2023-11-20
Move CallNode#name field between receiver and arguments

eregon:
* The same order as in source code.
* CallOrWriteNode, CallOperatorWriteNode, CallAndWriteNode already have the correct order so it was also inconsistent with them.

Currently the CallNode#name is the last field, which feels pretty strange when creating a new CallNode because it's crucial information only at the very end, and of course in source order it's between receiver and arguments.

Before:
```java
    public static final class CallNode extends Node {
        public final Node receiver;
        public final ArgumentsNode arguments;
        public final Node block;
        public final short flags;
        public final String name;
```
After
```java
    public static final class CallNode extends Node {
        public final Node receiver;
        public final String name;
        public final ArgumentsNode arguments;
        public final Node block;
        public final short flags;
```


config.yml
```diff
@@ -671,6 +671,8 @@ nodes:
         type: location?
       - name: message_loc
         type: location?
+      - name: name
+        type: constant
```
eregon, line 675: It seems when there is both a `constant` and `location` field the `location` is second, so I'll swap these, to be consistent with many other nodes which have it in that order.

kddnewton:
Can you rebase? Then this is fine.

kddnewton (changes requested):
Just marking as "request changes" until rebase

eregon:
Rebased

eregon:
I tried to repro the failure of the `test-all` job locally.
But I have a different error:
```
$ make -s test-all TESTS="prism --no-retry"
/home/eregon/code/ruby/test/prism/compiler_test.rb: linked to incompatible /home/eregon/.rubies/ruby-3.2.2/lib/libruby.so.3.2 - /home/eregon/prefix/ruby-master/lib/ruby/3.3.0+0/prism/prism.so (Test::Unit::ProxyError)
(also with other test files, it's just the first one failing I guess)

$ ldd /home/eregon/prefix/ruby-master/lib/ruby/3.3.0+0/prism/prism.so
ldd: warning: you do not have execution permission for `/home/eregon/prefix/ruby-master/lib/ruby/3.3.0+0/prism/prism.so'
	linux-vdso.so.1 (0x00007fff65b61000)
	libruby.so.3.2 => /home/eregon/.rubies/ruby-3.2.2/lib/libruby.so.3.2 (0x00007f66bfa00000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f66c0121000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f66bf823000)
	libz.so.1 => /lib64/libz.so.1 (0x00007f66c0107000)
	libcrypt.so.2 => /lib64/libcrypt.so.2 (0x00007f66c00cd000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f66c00ab000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f66c0291000)
```
It's pretty strange it's linked to 3.2 given I even `chruby ruby-master` in that shell.
I'll try to clean and rebuild.

eregon:
OK I can repro it locally now.
I guess the issue is the Makefile doesn't regenerate some generated file correctly when config.yml changes.

eregon:
Indeed, if I do:
```
rm -f prism/**/*.o
time make -j8 install-nodoc
make -s test-all TESTS="prism --no-retry"
```
Then the tests pass.

Maybe these dependencies in common.mk don't work:
```
prism/prism.$(OBJEXT): {$(VPATH)}prism/ast.h
```
This VPATH thing probably doesn't work.
Or maybe make doesn't understand `{$(VPATH)}prism/ast.h` and `prism/ast.h` are the same thing.

eregon:
To repro this it seems necessary to:
```
cd ruby
git checkout .
make clean
cd ../prism
bundle exec rake clobber
cd ../ruby
make -j8 install-nodoc
make -s test-all TESTS="prism --no-retry"
# ^ works fine
tool/sync_default_gems.rb prism
make -s test-all TESTS="prism --no-retry"
# ^ fails
```

eregon:
I can't spend more time debugging the complex CRuby build system so I'd like some help here.

Alternatively I could change the workflow to only build CRuby after sync_default_gems, that would most likely fix it.

kddnewton:
In that case can you just revert your changes to the CRuby build?

eregon:
Feel free to just push to this branch, you should have permission.
I'd like to at least see if that change fixes the CI job here.

eregon:
Ah there is the same problem in https://github.com/ruby/prism/pull/1903#issuecomment-1821628491, at least it's not just this PR.

My last commit does fix that workflow: https://github.com/ruby/prism/actions/runs/6949427614/job/18907514454?pr=1891
There is still an issue with incremental rebuild, but OTOH incremental rebuild on CRuby is frequently broken, and that can be fixed separately, this PR is not making it any worse.

So let's merge this then?

jemmaissroff:
Just noting that I rebased this branch, as referenced in https://github.com/ruby/prism/pull/1903#issuecomment-1821803356

kddnewton (approved):

PR #1898, opened 2023-11-20
Rename the big5-hkscs stuff to something more generic and add Big5-UAO sharing common code.

rgarver:
This adds support for Big5-UAO. See: https://github.com/ruby/prism/issues/1843

Rename some of the Big5-HKSCS files and definitions to reuse them for both cases since the logic appears to be the same.


kddnewton:
@rgarver this looks good but do you mind just putting everything into `big5.c`? That'll just simplify this whole thing.

rgarver:
No problem. That makes sense.

kddnewton (approved):

PR #1899, opened 2023-11-20
Warning for ENDs in methods

### LINKED ISSUE #1684: END in a method should warn
This warning happens in the legacy parsers for:

```ruby
      def foo
        END { }
      end
```

```text
../snippets/END1.rb:2: warning: END in method; use at_exit
```

This could be done during compile but it involves scope walking.   Also with runtimes using lazy method compilation we will see this warning only if the method is ever called.

haldun:
Closes #1684

### DIFF
--- a/include/prism/diagnostic.h
+++ b/include/prism/diagnostic.h
@@ -242,6 +242,7 @@ typedef enum {
     PM_WARN_AMBIGUOUS_FIRST_ARGUMENT_PLUS,
     PM_WARN_AMBIGUOUS_PREFIX_STAR,
     PM_WARN_AMBIGUOUS_SLASH,
+    PM_WARN_END_IN_METHOD,
 
     /* This must be the last member. */
     PM_DIAGNOSTIC_ID_LEN,
--- a/lib/prism/ffi.rb
+++ b/lib/prism/ffi.rb
@@ -299,7 +299,7 @@ module Prism
       values << (options.fetch(:frozen_string_literal, false) ? 1 : 0)
 
       template << "C"
-      values << (options[:verbose] ? 0 : 1)
+      values << (options.fetch(:verbose, true) ? 0 : 1)
 
       template << "L"
       if (scopes = options[:scopes])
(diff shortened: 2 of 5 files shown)


kddnewton:
Looks like you may need to update the FFI backend as well.

haldun:
@kddnewton I haven't touched FFI backend, do you have any pointers about what needs to be done?

eregon:
Warnings should already work fine in the FFI backend, but would be good to check for another existing warning.
I wonder if the issue is the `context_def_p` differs when using serialization.
You could check with Prism.dump + Prism.load to see if that has the warning.

haldun:
I don't see any tests for the existing warning messages. Do I miss something? Btw can you provide a Ruby snippet to trigger and existing warning to see whether we get the warnings in FFI?

eregon:
Mmh, it seems indeed warnings don't work with FFI currently.
`bundle exec bin/parse -e 'p //'` has 1 warning.
`PRISM_FFI_BACKEND=1 bundle exec bin/parse -e 'p //'` has no warnings.

eregon:
Ah that's likely the bug then.
I recall https://github.com/ruby/prism/pull/1765#discussion_r1384991034 maybe related to that?

Serialization itself is not the problem because `be ruby -Ilib -rprism -e 'p Prism.load "-e", Prism.dump("p //")'` has the warning (that's non-FFI mode).

haldun:
@eregon Yes that seems to be the issue. The tests are passing when I revert that line back. I am not sure what should be the ideal solution here though.

kddnewton:
I don't see how `context_def_p` could be different in the serialization environment - that would be very concerning seeing as the parser is invoked the exact same way.

This diff ought to fix it

```diff
diff --git a/lib/prism/ffi.rb b/lib/prism/ffi.rb
index c910fd3aae..847990ed9a 100644
--- a/lib/prism/ffi.rb
+++ b/lib/prism/ffi.rb
@@ -299,7 +299,7 @@ module Prism
       values << (options.fetch(:frozen_string_literal, false) ? 1 : 0)
 
       template << "C"
-      values << (options[:verbose] ? 0 : 1)
+      values << (options.fetch(:verbose, true) ? 0 : 1)
 
       template << "L"
       if (scopes = options[:scopes])
```

eregon:
> I don't see how `context_def_p` could be different in the serialization environment - that would be very concerning seeing as the parser is invoked the exact same way.

FWIW I was thinking when going through pm_serialize_parse maybe if given data it thinks it's an eval vs top-level parse. But that's not the problem here, and that part seems fine.

kddnewton (approved):

PR #1901, opened 2023-11-21
feat: adds encoding for TIS-620

thdaraujo:
[related to #1843]

Adds encoding for TIS-620 (Thai Industrial Standard 620-2533)

Documentation: [wikipedia](https://en.wikipedia.org/wiki/Thai_Industrial_Standard_620-2533)

According to [this paragraph](https://en.wikipedia.org/wiki/Thai_Industrial_Standard_620-2533#Variants), this encoding should be equivalent to [ISO-8859-11](https://en.wikipedia.org/wiki/ISO/IEC_8859-11), except for hex A0:
```
A nearly identical version of TIS-620 has been adopted as ISO/IEC 8859-11 in 2001,
the sole difference being that ISO/IEC 8859-11 defines hex A0 as a non-breaking space,
while TIS-620 leaves it undefined but reserved. (In practice, this small distinction is usually ignored.) 
```

~Should this just point to the ISO-8859-11 table, or is it worth it to have a new table defined, as done in this PR?~

Thanks!


docs/encoding.md
```diff
@@ -81,7 +82,7 @@ For each of these encodings, prism provides a function for checking if the subse
 
 ## Support for other encodings
 
-If an encoding is encountered that is not supported by prism, prism will call a user-provided callback function with the name of the encoding if one is provided. That function can be registered with `pm_parser_register_encoding_decode_callback`. The user-provided callback function can then provide a pointer to an encoding struct that contains the requisite functions that prism will use those to parse identifiers going forward.
+If an encoding is encountered that is not supported by prism, prism will call an user-provided callback function with the name of the encoding if one is provided. That function can be registered with `pm_parser_register_encoding_decode_callback`. The user-provided callback function can then provide a pointer to an encoding struct that contains the requisite functions that prism will use to parse identifiers going forward.
```
kddnewton, line 85: Getting rid of `those` is fine, but it should be `a` not `an`.

kddnewton (changes requested):
One small change, otherwise looks good

kddnewton:
Worth having its own table, no problem.

docs/encoding.md
thdaraujo: oops you're right, fixed!

kddnewton (approved):

kddnewton:
Nice work!

PR #1903, opened 2023-11-21
Add SPLAT flag on ArrayNode indicating if it contains splat element(s)

### LINKED ISSUE #1680: Add a flag to Arrays which contain a single splat element
At parse time, we know if an array only has a single splat element. This compiles differently. Instead of having to recompute this information at compile time, we can add it as a flag on an ArrayNode so it's known earlier.

### LINKED ISSUE #1896: [COMPILER] Stack underflow using splat assignment
Minimal repro to get stack underflow:

```
a, *b = *c
1
```

This seems to be an issue even not in the "popped" case in that the instructions look wrong for even `*b = *c`

I found this running this code 

` RUBY_ISEQ_DUMP_DEBUG=prism make test-all TESTS=../test/-ext-/tracepoint/test_tracepoint.rb`

jemmaissroff:
This commit puts a SPLAT flag on any ArrayNodes which contain SplatNode elements

Closes #1680, also helps close #1896

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -331,9 +331,14 @@ tokens:
 flags:
   - name: ArgumentsNodeFlags
     values:
-      - name: KEYWORD_SPLAT
+      - name: CONTAINS_KEYWORD_SPLAT
         comment: "if arguments contain keyword splat"
     comment: Flags for arguments nodes.
+  - name: ArrayNodeFlags
+    values:
+      - name: CONTAINS_SPLAT
+        comment: "if array contains splat nodes"
+    comment: Flags for array nodes.
   - name: CallNodeFlags
     values:
       - name: SAFE_NAVIGATION
@@ -459,6 +464,9 @@ nodes:
         type: location?
       - name: closing_loc
         type: location?
+      - name: flags
+        type: flags
+        kind: ArrayNodeFlags
     comment: |
       Represents an array literal. This can be a regular array using brackets or
       a special array using % like %w or %i.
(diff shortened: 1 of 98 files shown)


kddnewton (approved):
Looks good

kddnewton:
Actually, could you call it `CONTAINS_SPLAT`? Thinking about it it looks like it might mean "this array is splatted" not necessarily that it has one inside

jemmaissroff:
> Actually, could you call it `CONTAINS_SPLAT`? Thinking about it it looks like it might mean "this array is splatted" not necessarily that it has one inside

Sure, do you want me to change the ArgumentsNodeFlag to `CONTAINS_KEYWORD_SPLAT` too then? I also liked `CONTAINS_SPLAT` better, but thought it was best to be consistent.

Also just confirming - that Ruby test-all failure is fine and the bug still exists there on not templating?

kddnewton:
Yeah `CONTAINS_KEYWORD_SPLAT` would be better now that you mention it and I see it with this context.

The test-all suite is fine, something isn't rebuilding properly, I'm not sure what it is.

kddnewton (approved):

jemmaissroff:
Waiting until https://github.com/ruby/ruby/pull/8987 merges before merging this PR

eregon:
Given I have already rebased https://github.com/ruby/prism/pull/1891 2-3 times, I'd appreciate if we'd merge that first.
Plus, it would fix the test-all CI check.

jemmaissroff:
> Given I have already rebased #1891 2-3 times, I'd appreciate if we'd merge that first. Plus, it would fix the test-all CI check.

I'm going to merge this and then will rebase your branch! I don't want to override a different review, and this is currently blocking for me. Seems like it's either that I'd wait and then rebase this, or can rebase yours directly

eregon:
That works, thanks!

PR #1905, opened 2023-11-21
Remove handling OptionalNodeField in set_newline_flag template

makenowjust:
~A `Node#set_newline_flag` template for `OptionalNodeField` seems buggy.
If an `OptionalNodeField` is marked as a newline field, `NoMethodError` is raised at runtime.
Since there is no such field so far, it is not a problem.
This PR is for preventing the potential of bugs in the future.~

Because they are buggy, and actually they are dead code.


eregon (approved):

eregon:
Could you do the same in https://github.com/ruby/prism/blob/6ca8896fc660592af958542cebda06737ec3d58b/templates/java/org/prism/Nodes.java.erb#L245-L246 ?
Alternatively, since this is dead code, it might be cleaner to simply remove `Prism::OptionalNodeField` there, and let it `raise` at templating time if it's ever needed.

makenowjust:
@eregon Indeed that sounds smart. Thank you.

PR #1906, opened 2023-11-22
Add KOI8-U encoding

heyogrady:
Adds single-byte [`KOI8-U`](https://en.wikipedia.org/wiki/KOI8-U) encoding (#1843).


src/enc/pm_tables.c
```diff
@@ -1418,7 +1442,7 @@ pm_encoding_single_char_width(PRISM_ATTRIBUTE_UNUSED const uint8_t *b, PRISM_ATT
  * checking if it's a valid codepoint in KOI-8 and if it is returning 1.
  */
 static size_t
-pm_encoding_koi8_r_char_width(const uint8_t *b, PRISM_ATTRIBUTE_UNUSED ptrdiff_t n) {
+pm_encoding_koi8_char_width(const uint8_t *b, PRISM_ATTRIBUTE_UNUSED ptrdiff_t n) {
```
heyogrady, line 1445: Renamed this function for use for both `KOI8-R` and `KOI8-U`, since the ranges are  the same.

kddnewton (changes requested):
Looks great, just need the update in encoding.md

kddnewton (approved):

PR #1908, opened 2023-11-22
Fix local variable depth in rescue clauses

camertron:
Local variables assigned in rescue clauses are accessible outside the rescue's scope:

```ruby
begin
  require "foo"
rescue LoadError
  my_var = "uh oh!"
end

my_var  # => "uh oh!"
```

For `begin` and `else` clauses, MRI does not push a stack frame and so locals can be assigned at depth 0. `rescue` clauses in contrast push a stack frame, so locals should be assigned at depth 1.

This PR pushes a transparent scope while parsing `rescue` clauses in order to increase the depth by 1.


kddnewton:
In general the approach is sound, I see that you're correctly pushing the transparent scope.

I'm curious why you need to change the handling of the depth though. Shouldn't that just work as in once the transparent scope is in place?

camertron:
@kddnewton you're absolutely right, the depth is already correctly calculated. I got rid of the other code 😄

eregon:
> `rescue` clauses in contrast [on MRI] push a stack frame, so locals should be assigned at depth 1.

This is actually only true on CRuby, and not on JRuby and TruffleRuby (also not on CRuby 1.8 FWIW):
```
~ ∴ ruby -ve 'begin; raise; rescue; puts caller_locations(0); end'
-e:1:in `rescue in <main>'
-e:1:in `<main>'

~ ∴ ruby -ve 'begin; raise; rescue; puts caller_locations(0); end'
truffleruby 24.0.0-dev-bb05e0bb, like ruby 3.2.2, GraalVM CE Native [x86_64-linux]
-e:1:in `<main>'

~ ∴ ruby -ve 'begin; raise; rescue; puts caller_locations(0); end'
jruby 9.4.4.0 (3.1.4) 2023-10-18 9c9bfd3dc7 OpenJDK 64-Bit Server VM 17.0.8+7 on 17.0.8+7 +jit [x86_64-linux]
-e:1:in `<main>'
```

I would really like to be able to trust the depth from Prism on TruffleRuby.

So I think either we need an option for this (but it probably doesn't make sense to set this except for using Prism as a Ruby engine parser) or a template-time env var to configure it (like `PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS`).
How about `PRISM_PUSH_SCOPE_FOR_RESCUE` (defaulting to true == CRuby behavior)?

In terms of implementation, because prism.c is not generated, you can use the trick to define in the header like https://github.com/ruby/prism/blob/e84c4662ed54f7be9528d8d9785836d6255b3ff5/templates/include/prism/ast.h.erb#L198

eregon (changes requested):
Marking as request changes because otherwise this will break JRuby and/or TruffleRuby due to incorrect depth there.

enebo:
Wow.  You learn something everyday and I didn't notice this worked this way for 20+ years.  This will indeed break JRuby and TR short-term since it is not just a parser change.  I guess the motivation for this was to increase backtrace visibility?  

I think we can come up with a way of subtracting the depth (which is what happens with `for` loop variables already) to maintain our current behavior.  This is somewhat icky since we have to recursively figure out how many rescues are between us and where the scope is really defined.   Since our system knows `for` is special we just count number of `for` scopes to substract.  We have no scopes for `rescue` (and I guess ensure?) so this will probably be some counter?  Not sure if other interwoven elements will make this more difficult.

As I see it there are two problems:

  1. Fixing the behavior so all impls have the same backtrace.  I assume this backtrace behavior is motivation?  If not someone should give more info.  I have only just seen this but it feels like a non-trivial task
  3. Having two depths means one serialization format will not be usable by all impls.  This is a longer term goal but being able to use the same blob means we can all share a universal pre-compiled source
  
I think @eregon workaround gives us time to figure out 1) but even if we decided not to impl 1) then we still need 2) in the future.  So I think any suggestion like this should only be temporary.

kddnewton:
Okay, so local variable depth is a whole confusing thing so I just added a doc for it. It's here: https://github.com/ruby/prism/blob/main/docs/local_variable_depth.md.

I think we have a shocking number of differences here that we will need to account for. I don't particularly want to deviate from CRuby, but it looks like we already do in quite a few places. In this case, I think we should clearly define our own concept of "depth" (since it's clearly not shared between implementations) and then make sure our documentation is super clear. CRuby is going to use our depth variable, but it's going to have to adjust it according to its own needs. JRuby/TruffleRuby will have to do the same. Hopefully we can converge around the most common cases and document the edge cases.

With this in mind, I think we shouldn't increase the scope depth for rescues, simply because it's easier to add to the number of scopes than it is to subtract. We'll handle this explicitly in the CRuby compiler.

enebo:
@kddnewton you also need to include BEGIN as well to that doc.

eregon:
That document looks great.
I commented on some clearer wording regarding `for` in https://github.com/ruby/prism/commit/ed8f5682cadc74486f1f6f2c87b54fa967ad6659#r133270393.
I think Prism, as it seems to be already the case, should only create new scopes when they need to be there semantically (i.e a separate set of variables only accessible in that scope and nested scopes).
IOW avoid artificial extra (semantically unnecessary) scopes which only exist on some Ruby implementations and are implementation details.

camertron:
Thanks for weighing in, everyone! Sounds like the compiler is the right place to make depth adjustments 👍 Thanks for that very nice doc too @kddnewton 😄

PR #1914, opened 2023-11-22
Optimize `lex_identifier`

haldun:
- Introduced `fast_char_is_identifier` which assumes `parser->encoding_changed` is `false`.
- Use `fast_char_is_identifier` inside `lex_identifier`.

I see a ~10% improvement, but it would be great if someone can confirm this too.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -5890,6 +5890,19 @@ char_is_identifier_start(pm_parser_t *parser, const uint8_t *b) {
 }
 
 /**
+ * Similar to char_is_identifier but this function assumes that the encoding
+ * has not been changed.
+ */
+static inline size_t
+char_is_identifier_utf8(const uint8_t *b, const uint8_t *end) {
+    if (*b < 0x80) {
+        return (*b == '_') || (pm_encoding_unicode_table[*b] & PRISM_ENCODING_ALPHANUMERIC_BIT ? 1 : 0);
+    } else {
+        return (size_t) (pm_encoding_utf_8_alnum_char(b, end - b) || 1u);
+    }
+}
+
+/**
  * Like the above, this function is also used extremely frequently to lex all of
  * the identifiers in a source file once the first character has been found. So
  * it's important that it be as fast as possible.
@@ -5907,11 +5920,8 @@ char_is_identifier(pm_parser_t *parser, const uint8_t *b) {
         } else {
             return 0;
         }
-    } else if (*b < 0x80) {
-        return (pm_encoding_unicode_table[*b] & PRISM_ENCODING_ALPHANUMERIC_BIT ? 1 : 0) || (*b == '_');
-    } else {
-        return (size_t) (pm_encoding_utf_8_alnum_char(b, parser->end - b) || 1u);
     }
+    return char_is_identifier_utf8(b, parser->end);
 }
 
 // Here we're defining a perfect hash for the characters that are allowed in
@@ -6985,9 +6995,16 @@ lex_identifier(pm_parser_t *parser, bool previous_command_start) {
     const uint8_t *end = parser->end;
     const uint8_t *current_start = parser->current.start;
     const uint8_t *current_end = parser->current.end;
+    bool encoding_changed = parser->encoding_changed;
 
-    while (current_end < end && (width = char_is_identifier(parser, current_end)) > 0) {
-        current_end += width;
+    if (encoding_changed) {
+        while (current_end < end && (width = char_is_identifier(parser, current_end)) > 0) {
+            current_end += width;
+        }
+    } else {
+        while (current_end < end && (width = char_is_identifier_utf8(current_end, end)) > 0) {
+            current_end += width;
+        }
     }
     parser->current.end = current_end;
 
@@ -7105,7 +7122,7 @@ lex_identifier(pm_parser_t *parser, bool previous_command_start) {
         }
     }
 
-    if (parser->encoding_changed) {
+    if (encoding_changed) {
         return parser->encoding.isupper_char(current_start, end - current_start) ? PM_TOKEN_CONSTANT : PM_TOKEN_IDENTIFIER;
     }
     return pm_encoding_utf_8_isupper_char(current_start, end - current_start) ? PM_TOKEN_CONSTANT : PM_TOKEN_IDENTIFIER;


src/prism.c
```diff
@@ -5889,6 +5889,19 @@ char_is_identifier_start(pm_parser_t *parser, const uint8_t *b) {
...
 }
 
+/**
+ * Similar to char_is_identifier but this function assumes that the encoding
+ * has not been changed.
+ */
+static inline size_t
+fast_char_is_identifier(const uint8_t *b, const uint8_t *end) {
+    if (*b < 0x80) {
+        return (*b == '_') || (pm_encoding_unicode_table[*b] & PRISM_ENCODING_ALPHANUMERIC_BIT ? 1 : 0);
```
eregon, line 5899: I wonder, isn't this part (`*b < 0x80`) valid for any ASCII-compatible encoding?
And only ASCII-compatible encodings are allowed for parsing so that's all Prism encodings.
If so changing `char_is_identifier` to check that first (regardless of encoding_changed) might make it as fast or nearly as the two separate functions, because most source bytes are < 0x80.
Could you try this maybe?

eregon: It does, because for an ASCII-compatible encodings, all codepoints < 0x80 must have the same meaning as in US-ASCII.

eregon: And FWIW `(0...128).each {|cp| ascii = cp.chr(Encoding::US_ASCII); p ascii; Encoding.list.select(&:ascii_compatible?).each { |enc| c = cp.chr(enc); raise c unless
c == ascii } }` proves it.

haldun: Is this close to your suggestion?

```c
static inline size_t
char_is_identifier(pm_parser_t *parser, const uint8_t *b) {
    if (*b < 0x80) {
        return (*b == '_') || (pm_encoding_unicode_table[*b] & PRISM_ENCODING_ALPHANUMERIC_BIT ? 1 : 0);
    }
    if (parser->encoding_changed) {
        size_t width;
        if ((width = parser->encoding.alnum_char(b, parser->end - b)) != 0) {
            return width;
        } else if (*b == '_') {
            return 1;
        } else if (*b >= 0x80) {
            return parser->encoding.char_width(b, parser->end - b);
        } else {
            return 0;
        }
    } else {
        return (size_t) (pm_encoding_utf_8_alnum_char(b, parser->end - b) || 1u);
    }
}
```

Interestingly it does not provide much in terms of performance comparing to the current two function version.

eregon: Yeah that's what I had in mind.

> Interestingly it does not provide much in terms of performance comparing to the current two function version.

OK, then the 2 functions is better, I'm fine with that.
Maybe it's about whether everything gets inlined or not, and that can't be the case when handling a non-UTF-8 encoding.

eregon (approved):
LGTM, would be good to have the review of @kddnewton too

src/prism.c
duerst: @haldun @eregon 
> It does, because for an ASCII-compatible encoding, all codepoints < 0x80 must have the same meaning as in US-ASCII.

Please be careful. Shift_JIS is a valid source encoding, and encodes some characters with two bytes, where the second byte can be a byte <0xFF. I have attached a file that shows this; unfortunately Github doesn't allow adding .rb files, so I had to change the extension to .txt. Please change it back when you download.
[ruby_shift_jis.txt](https://github.com/ruby/prism/files/13453742/ruby_shift_jis.txt)

A discussion of the structure of Shift_JIS can be found at https://en.wikipedia.org/wiki/Shift_JIS#Structure. Please note that the "first byte is even"/" first byte is odd" refers to the first byte in the original JIS code chart, not in Shift_JIS itself.

In general, the term "ASCII-compatible encoding" comes in various shades or degrees. In particular, there's a difference between (very simply worded) "ASCII characters are always encoded as ASCII bytes" and "ASCII bytes always stand for ASCII characters".

eregon: Right, but assuming we are at a codepoint boundary when calling `char_is_identifier()` then if the first byte is <0x80 it is correct to interpret it as the ASCII character. And it needs to be, otherwise it's really not ASCII-compatible in any sense.
Here I need "ASCII bytes *just after a codepoint boundary* always stand for ASCII characters" which holds for all `Encoding#ascii_compatible?`.

enebo: I question whether this will turn out to be reasonable or not.  I thought 0x80 would be but I see that in current lexer (both MRI and JRuby) it does not do this shortcut for identifiers but calls the more costly precise_mbclen.  This might be because if it is mbc it knows real length but since that is much more uncommon than ASCII I definitely wonder if there is some encoding which defies that 0x80 check (0x80 is used in other places in lexer but only when it knows it is utf-8 already -- or should be).

kddnewton: I'm not entirely comfortable with assuming things about other encodings, I'd rather stick to checking if the encoding changed and otherwise optimize only the UTF-8 encoding. It's going to be 99% of what is parsed anyway.

eregon: Just to be clear, this PR doesn't use my suggestion above, so anyway nothing changed semantically.

kddnewton:
I am also seeing about a 10% speed-up, great work @haldun

src/prism.c
eregon: Though I want to say, https://github.com/ruby/prism/pull/1914#discussion_r1403447919 is the hard proof I am correct that all codepoints < 0x80 mean the same in ASCII-compatible encodings, so I am not sure why people are doubting it here.
Yes, it assumes we are at a character boundary, but that is very much the expectation for any parser or lexer really, that it consumes one or multiple characters at a time and does not stop in the middle of a chracter.
But anyway, it turns out using that doesn't make it faster, so we don't need to use it.

src/prism.c
```diff
@@ -5889,6 +5889,19 @@ char_is_identifier_start(pm_parser_t *parser, const uint8_t *b) {
...
+    } else {
+        return (size_t) (pm_encoding_utf_8_alnum_char(b, end - b) || 1u);
```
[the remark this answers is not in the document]
haldun: Sorry I couldn't get this. Can you please elaborate more?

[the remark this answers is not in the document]
kddnewton: Actually ignore this, I'm not sure what I'm saying here.

kddnewton (approved):

PR #1916, opened 2023-11-23
Check circular references in default values of params

### LINKED ISSUE #1637: "circular argument reference" `SyntaxError`s are missing
In pre-Prism parsers this is detected during parse-time:

```ruby
def bar
  1
end

def foo(bar = bar)
  bar
end
```

Error:

```text
../snippets/def1.rb:4: circular argument reference - bar
```

I am thinking this is just missing since variable scoping is present.

makenowjust:
Fix #1637

The current parameter name is tracked by the new field `parser->current_param_name`. In CRuby impl., this value is tracked in the same way, so I think it is reasonable to add a new field to `pm_parser_t`.


src/prism.c
```diff
@@ -4037,19 +4037,30 @@ pm_local_variable_or_write_node_create(pm_parser_t *parser, pm_node_t *target, c
     return node;
 }
 
+static bool context_p(pm_parser_t *parser, pm_context_t context);
```
makenowjust, line 4040: This prototype declaration seems ugly ¯\_(ツ)_/¯

src/diagnostic.c
```diff
@@ -198,6 +198,7 @@ static const char* const diagnostic_messages[PM_DIAGNOSTIC_ID_LEN] = {
     [PM_ERR_OPERATOR_WRITE_BLOCK]               = "Unexpected operator after a call with a block",
     [PM_ERR_PARAMETER_ASSOC_SPLAT_MULTI]        = "Unexpected multiple `**` splat parameters",
     [PM_ERR_PARAMETER_BLOCK_MULTI]              = "Multiple block parameters; only one block is allowed",
+    [PM_ERR_PARAMETER_CIRCULAR]                 = "Invalid circular reference in a default parameter",
```
makenowjust, line 201: I am not sure here "Invalid" sounds like good wording.

src/prism.c
kddnewton: Can you move the whole context section above the constructors?

src/prism.c
```diff
@@ -4037,19 +4037,30 @@ pm_local_variable_or_write_node_create(pm_parser_t *parser, pm_node_t *target, c
...
+
 /**
  * Allocate a new LocalVariableReadNode node.
  */
 static pm_local_variable_read_node_t *
 pm_local_variable_read_node_create(pm_parser_t *parser, const pm_token_t *name, uint32_t depth) {
+    pm_constant_id_t name_id = pm_parser_constant_id_token(parser, name);
+
+    if (
+        context_p(parser, PM_CONTEXT_DEFAULT_PARAMS) &&
```
kddnewton, line 4050: This is going to be checked a lot, I'm wondering if we can make this more efficient by just checking `parser->current_param_name == name_id`, since it will always necessarily be `PM_CONTEXT_DEFAULT_PARAMS` if the value is set?

src/diagnostic.c
kddnewton:
```suggestion
    [PM_ERR_PARAMETER_CIRCULAR]                 = "Parameter default value references itself",
```

kddnewton (changes requested):

src/prism.c
makenowjust: 🆗

src/prism.c
kddnewton: Yes, `0` would work, constants have to be `>= 1`

makenowjust: Is a macro or a constant for this value defined?

makenowjust:
I found an edge case: `def foo(bar = (def baz(bar) = bar; 1)) = 2` is accepted by CRuby. [c789a83](https://github.com/ruby/prism/pull/1916/commits/c789a833c56babd742e1e1d416ffe99bbf2b5494) is a commit for supporting it.

kddnewton (approved):
@makenowjust nice work, this is quite complicated

PR #1919, opened 2023-11-24
Use let declaration in scanConstant

camertron:
The `startOffset` variable is reassigned on line 93 and should therefore be declared using `let` instead of `const`.


kddnewton (approved):
Thanks!

PR #1935, opened 2023-11-28
Add MacJapanese encoding

mattboldt:
MacJapanese appears to be an extension of Shift_JIS with some additions from Apple
https://en.wikipedia.org/wiki/Shift_JIS#MacJapanese

Apple's docs:
https://developer.apple.com/documentation/coreservices/1399915-encoding_variants_for_macjapanes

Not sure if these are a good source of truth, but I found them linked from a python library https://github.com/python/cpython/issues/45617

http://www.unicode.org/Public/MAPPINGS/VENDORS/APPLE/JAPANESE.TXT

http://www.unicode.org/Public/MAPPINGS/OBSOLETE/EASTASIA/JIS/SHIFTJIS.TXT

Relates to issue:
https://github.com/ruby/prism/issues/1843


eregon:
@duerst Do you know is there is any difference for the Ruby encodings `Encoding::Shift_JIS` and `Encoding::MacJapanese`?
If not, do you why they are not simply aliases?

mattboldt:
@eregon Oh, interesting. I assumed because it said Apple customized some chars, it couldn't simply be aliased. If that's not true, I'm happy to figure out how to alias it.

kddnewton:
@mattboldt could you add the file to the gemspec? If the tests are passing, this is good to go.

mattboldt:
@kddnewton Sorry, tests aren't passing, but I wanted to get some early feedback on it so I opened a draft. Sounds like we could potentially alias the existing Shift_JIS encoding instead?

kddnewton:
@mattboldt that'd be fine, why don't you try it. It would look like adding the encoding (as you have) and then defining it with the same callbacks in `pm_shift_jis.c`.

duerst:
> Sounds like we could potentially alias the existing Shift_JIS encoding instead?

As Ruby itself does that, I think prism also should just alias Shift_JIS.

It's correct that Shift_JIS and MacJapanese are not exactly the same, but then MacJapanese also comes in several variants. Many Japanese Ruby committers use Macs, and the fact that they didn't see the need for an implementation different from Shift_JIS probably means that this encoding isn't that important anymore. Also, this choice may have been made because other languages and conversion libraries make the same choice.

src/prism.c
```diff
@@ -6263,6 +6263,7 @@ parser_lex_magic_comment_encoding_value(pm_parser_t *parser, const uint8_t *star
                 ENCODING1("macCyrillic", pm_encoding_mac_cyrillic);
                 ENCODING1("macGreek", pm_encoding_mac_greek);
                 ENCODING1("macIceland", pm_encoding_mac_iceland);
+                ENCODING1("macJapanese", pm_encoding_mac_japanese);
```
kddnewton, line 6266:
```suggestion
                ENCODING1("MacJapanese", pm_encoding_mac_japanese);
                ENCODING1("MacJapan", pm_encoding_mac_japanese);
```

This one has another alias (`Encoding::MACJAPANESE.names`)

kddnewton (changes requested):

mattboldt:
Not sure what I missed here 🤔 

```
/usr/bin/ld: prism/prism.o: in function `parser_lex_magic_comment_encoding_value':
/home/runner/work/prism/prism/ruby/ruby/prism/prism.c:6305: undefined reference to `pm_encoding_mac_japanese'
/usr/bin/ld: /home/runner/work/prism/prism/ruby/ruby/prism/prism.c:6306: undefined reference to `pm_encoding_mac_japanese'
```

src/enc/pm_mac_japanese.c
```diff
@@ -0,0 +1,57 @@
...
+    if (pm_encoding_mac_japanese_char_width(b, n) == 1) {
+        return pm_encoding_ascii_isupper_char(b, n);
+    } else {
+        return 0;
+    }
+}
+
+/** mac_japanese encoding */
+pm_encoding_t pm_encoding_mac_japanese = {
+    .name = "mac_japanese",
```
kddnewton, line 51: The tests are failing on FFI because this isn't a valid identifier for `Encoding.find`. This should instead be:

```suggestion
    .name = "MacJapanese",
```

kddnewton (changes requested):
One more change!

src/enc/pm_mac_japanese.c
mattboldt: Fixed! Sorry for all the back and forth here, thanks for your help.

src/prism.c
kddnewton: No worries!

kddnewton (approved):

kddnewton:
Nice, just waiting on https://github.com/ruby/ruby/pull/9064

mattboldt:
Just realized I could manually test these changes via `bin/parse`

branch main
```
> bin/parse -e "# encoding: MacJapanese"
Comments:
[#<Prism::InlineComment @location=#<Prism::Location @start_offset=0 @length=23 start_line=1>>]

Magic comments:
[#<Prism::MagicComment @key="encoding" @value="MacJapanese">]

Errors:
[#<Prism::ParseError @message="unknown or invalid encoding in the magic comment" @location=#<Prism::Location @start_offset=12 @length=11 start_line=1>>]
```

branch enc_mac_japanese
```
> bin/parse -e "# encoding: MacJapanese"
Comments:
[#<Prism::InlineComment @location=#<Prism::Location @start_offset=0 @length=23 start_line=1>>]

Magic comments:
[#<Prism::MagicComment @key="encoding" @value="MacJapanese">]
```

@kddnewton  I assume something similar is done in the automated tests? Just curious how we can confirm it's working as expected once merged into CRuby

kddnewton:
Yup! @mattboldt you added it to `encoding_test.rb`, which does pretty much exactly this

Issue #1915, opened 2023-11-23
Could DefNode#parameters be never nil/null?


eregon:
Currently it's null for this case:
```
$ bin/parse -e 'def m; end'
@ ProgramNode (location: (1,0)-(1,10))
├── locals: []
└── statements:
    @ StatementsNode (location: (1,0)-(1,10))
    └── body: (length: 1)
        └── @ DefNode (location: (1,0)-(1,10))
            ├── name: :m
            ├── name_loc: (1,4)-(1,5) = "m"
            ├── receiver: ∅
            ├── parameters: ∅
            ├── body: ∅
            ├── locals: []
            ├── def_keyword_loc: (1,0)-(1,3) = "def"
            ├── operator_loc: ∅
            ├── lparen_loc: ∅
            ├── rparen_loc: ∅
            ├── equal_loc: ∅
            └── end_keyword_loc: (1,7)-(1,10) = "end"
```

But it could just be a ParametersNode with everything empty, and then be a non-optional field.
This leads to every use of a DefNode to have to consider it being nil/null and e.g. if it wants to know the number of required arguments then have this extra check.

I'm working around by creating it empty manually in TruffleRuby.

enebo:
@eregon this will increase the serialized size by creating a node just to have no contents in it.  I hot path this case to just be an arity check without going through all the other machinery of seeing how to bind stuff.  This is a rare case where I prefer a null field.

eregon:
@enebo Yeah I could see that.
OTOH it seems nicer to always have the node for e.g. Ruby users of the gem.

FWIW I'm using this in Java:
```java
public static final Nodes.ParametersNode ZERO_PARAMETERS_NODE = new Nodes.ParametersNode(EMPTY_NODE_ARRAY,
            EMPTY_NODE_ARRAY, null, EMPTY_NODE_ARRAY, EMPTY_NODE_ARRAY, null, null, 0, 0);
...
        var parametersNode = environment.parametersNode;
        if (parametersNode == null) {
            // parametersNode == null for a method means zero parameters: https://github.com/ruby/prism/issues/1915
            parametersNode = ZERO_PARAMETERS_NODE;
        }
```

enebo:
@eregon serialize+load and gem side of things do not need to behave the same way here.  config.yml says it can be null but ruby templates could do effectively what you are (construct an empty node).

For that matter since we both can stuff things via our Java engine env you could just make this zero param node in Loader for TR only.

enebo:
[...]

Issue #1762, opened 2023-11-02
Add information about proc's numbered parameters count


andrykonchin:
It would be helpful to have numbered parameters count (in proc) provided.

A proc with numbered parameters should be handled in a bit different way and we cannot rely on `ParametersNode` to have actual information about proc's arity and parameter names. We can rely only on `BlockNode#locals` and check if there are names like `_` + digit (1-9) and there are no `BlockParametersNode` or `ParametersNode`. So number of such locals (or the largest digit) would tell us how many numbered parameters are there.

It will be nice to have the numbered parameters count to be provided explicitly, e.g. as a `char`/`byte` fields. Also it might be simpler to have a separate node class (e.g. `NumberedParametersNode`) to use instead of `ParametersNode` in such a case. This way `NumberedParametersNode` would not contain numerous fields like `requireds`, `optionals` etc and have only one integer field.

enebo:
@andrykonchin this is even a little bit grosser on detection:

```ruby
proc { |_9a| p _9a }.call("a")
```

kddnewton:
Right, I think it's `^_\d$`.

enebo:
@kddnewton please prevent this tragedy :smiley: 

```java
   public int highestNumberedParameter() {
        if (type != Type.BLOCK) return 0;

        int biggest = 0;

        for (String name: variableNames) {
            if (name.matches("_\\d+$")) biggest = Integer.parseInt(name.substring(1, name.length()));
        }

        return biggest;
    }
```

kddnewton:
Yeah, I suppose I had to do that as well when translating to whitequark/parser. https://github.com/kddnewton/parser-prism/blob/cff12ee1955c651659b5230dc562aa5a2d8fbcc3/lib/parser/prism/compiler.rb#L911-L913. We can work on this.

kddnewton:
Nope! :) fortunately it only allows up to `_9` so you should be able to get rid of the `+`

eregon:
Currently the `block.parameters` is nil, but that seems not great given the block does have parameters, they are just numbered vs named.

How about  using a BlockParametersNode too with an array of `RequiredParameterNode` with names _1, _2, etc? That seems best to me.
One thing is those nodes wouldn't have a corresponding location in the source, but maybe we can just give them startOffset=0 length=0, or the whole block location maybe?

Or maybe a `NumberedParametersNode` or so, which would e.g. contain how many there are?

[...]

PR #1937, opened 2023-11-28
Add numbered_parameters field to BlockNode and LambdaNode

jemmaissroff:
We are aware at parse time how many numbered parameters we have on a BlockNode or LambdaNode, but prior to this commit, did not store that information anywhere in its own right.

The numbered parameters were stored as locals, but this does not distinguish them from other locals that have been set, for example in `a { b = 1; _1 }` there is nothing on the AST that distinguishes b from _1.

Consumers such as the compiler need to know information about how many numbered parameters exist to set up their own tables around parameters. Since we have this information at parse time, we should compute it here, instead of deferring the work later on.

I have confirmed this works with the compiler to help us fix existing bugs in compilation that rely on accurate parameters counts.


tenderlove (approved):
Looks good.  Just need to fix the compilation warnings

src/prism.c
```diff
@@ -12647,25 +12663,32 @@ parse_variable_call(pm_parser_t *parser) {
         }
 
         if (!parser->current_scope->closed && pm_token_is_numbered_parameter(parser->previous.start, parser->previous.end)) {
-            // Indicate that this scope is using numbered params so that child
-            // scopes cannot.
-            parser->current_scope->numbered_params = true;
```
kddnewton, line 12652: This optimization was here so that `outer_scope_using_numbered_params_p` didn't have to scan too many scopes, it would already be cached. Can we bring this back?

src/prism.c
```diff
@@ -5821,6 +5824,18 @@ pm_parser_local_add(pm_parser_t *parser, pm_constant_id_t constant_id) {
     }
 }
 
+/**
+ * Add a constant id to the local table of the current scope.
```
kddnewton, line 5828: This comment doesn't make a lot of sense in this context.

config.yml
```diff
@@ -596,6 +596,8 @@ nodes:
         type: location
       - name: closing_loc
         type: location
+      - name: numbered_parameters
```
kddnewton, line 599: With this configuration, consumers have to look at `numbered_parameters` every time they want to understand block/lambda nodes, and they have to implicitly know that if it's `> 0` then they should use the numbers and not check the regular parameters.

Another approach would be to create a `NumberedParametersNode`, which we put in place of parameters (which we already have the pointer for). We can do this because numbered parameters/other parameters are mutually exclusive. That way we wouldn't increase the size of every block/lambda node in the parser by 4 bytes and would instead only allocate for numbered parameters when necessary. Personally I think this is more clear to the consumer, because they don't have to know to check `numbered_parameters`. What do you think?

eregon: Agreed a node would be nicer, as already suggested in https://github.com/ruby/prism/issues/1762#issuecomment-1816204931
Wouldn't it just work without any special casing if it was a BlockParametersNode too with an array of `RequiredParameterNode` (or `NumberedParameterNode`) with names _1, _2, etc?

src/prism.c
jemmaissroff: I'm not sure I'm following - what part of this did we lose? My understanding is we're still setting `parser->current_scope->numbered_parameters` to a non-zero value, so all the same cases will still return true from `outer_scope_using_numbered_parameters_p`

config.yml
jemmaissroff: I initially thought about something like a `NumberedParametersNode`, but my hesitation was that it's not actually meaningful in terms of anything that's in the parameters. Usually the parameters node on something like a block or a lambda points concretely to the parameters actually being passed. Since numbered parameters are implied, they're never passed. A node like that would have, for instance, a null location, and somewhat arbitrary values (would we store, for instance, all of the numbered parameters that exist? the max one?). Is there prior art for this type of thing?

src/prism.c
jemmaissroff: Good catch! https://github.com/ruby/prism/pull/1945

config.yml
kddnewton: @eregon you're welcome to add that to the desugar visitor, but that's a ton of memory that I don't want to allocate every time someone uses numbered parameters.

enebo: My initial reaction to this PR is that it merges two concepts/types into one structure.  

As @kddnewton said it feels strange to check if numbered_parameters is checked to know whether you are dealing with an implicit numbered params block or it is actually just a block with no explicit parameters.  I feel this detail alone makes this as a possible point of confusion for new users.  I likely feel it could cause confusion for me if i haven't looked at this node in a year.

Beyond that it add 1-4 bytes to every block when serializing for what is still a somewhat uncommon syntax.  If it does become popular then I feel it also means why put those other fields (parameters+locals) onto a numbered parameters block.

eregon: It sounds like we agree on a NumberedParameterNode with a single integer field.
I think I'll still desugar it in TruffleRuby but that's no big deal and it makes sense to do it there.

Yeah having null parameters like currently is confusing, because it can mean no parameters at all or numbered parameters. I was thinking the same when touching code before filing https://github.com/ruby/prism/issues/1915. For methods null only means no parameters at all but for blocks it's ambiguous. So a NumberedParameterNode is definitely best to avoid such mistakes.

Issue #1783, opened 2023-11-07
Accept negative start line number


eregon:
The start line can be negative and this is used in ERB, other template engines, etc so they can inject extra lines at the start without affecting the line number of the user source:
https://github.com/ruby/erb/blob/8db8b8b5086713175523f7cab4bc8b779f069709/lib/erb.rb#L468

See https://github.com/ruby/prism/pull/1763#discussion_r1384983631

enebo:
:+1: 

```text
        // FIXME: I believe line number can be negative?
        // Line Number
        appendUnsignedInt(metadata, lineNumber);
```

eregon:
I splitted the previous label in two because it wasn't clear.
It's intended to signal priority yes.
The meaning is that this issue is blocking correct compilation (of such eval's) and blocking truffleruby's adoption of prism (which is just another case of compilation, so maybe I should remove the `/truffleruby` suffix).

eregon:
Yeah blocker was probably too strong without more context, I renamed to `needed for correct compilation`.
I'm trying to identify issues which do prevent switching from the current parser to Prism in TruffleRuby (by default, and removing the current parser).
I annotated the issues that I consider are blocking that, because it would be too incorrect or messy to work around (and since we are designing a new parser we don't want to make things difficult to use).
This specific issue I think is hard to work around so I think it qualifies for that (and it's likely easy to fix).

OTOH things like different warnings messages or missing SyntaxError I think are not very important for that switch, those could be fixed later from the POV of using Prism in TruffleRuby.

enebo:
Thanks for your perspective on it.  It is clearer now.

casperisfine:
Possible fix: https://github.com/ruby/prism/pull/1944

PR #1944, opened 2023-11-29
Implement negative line numbers

casperisfine:
Fix: https://github.com/ruby/prism/issues/1783

Ruby allows for 0 or negative line start, this is often used with `eval` calls to get a correct offset when prefixing a snippet.
    
 e.g.
    
```ruby
caller = caller_locations(1, 1).first
class_eval <<~RUBY, caller.path, caller.line - 2
  # frozen_string_literal: true
  def some_method
    #{caller_provided_code_snippet}
  end
RUBY
```

The first commit rename all `varint` stuff to `varuint` to avoid confusion.

The second commit introduce the `varsint` (signed) type, and change `line` to be signed.

### DIFF
--- a/docs/serialization.md
+++ b/docs/serialization.md
@@ -9,7 +9,7 @@ The syntax tree still requires a copy of the original source, as for the most pa
 
 Let us define some simple types for readability.
 
-### varint
+### varuint
 
 A variable-length integer with the value fitting in `uint32_t` using between 1 and 5 bytes, using the [LEB128](https://en.wikipedia.org/wiki/LEB128) encoding.
 This drastically cuts down on the size of the serialized string, especially when the source file is large.
@@ -18,15 +18,15 @@ This drastically cuts down on the size of the serialized string, especially when
 
 | # bytes | field |
 | --- | --- |
-| varint | the length of the string in bytes |
+| varuint | the length of the string in bytes |
 | ... | the string bytes |
 
 ### location
 
 | # bytes | field |
 | --- | --- |
-| varint | byte offset into the source string where this location begins |
-| varint | length of the location in bytes in the source string |
+| varuint | byte offset into the source string where this location begins |
+| varuint | length of the location in bytes in the source string |
 
 ### comment
 
@@ -71,18 +71,18 @@ The header is structured like the following table:
 | `1` | patch version number |
 | `1` | 1 indicates only semantics fields were serialized, 0 indicates all fields were serialized (including location fields) |
 | string | the encoding name |
-| varint | the start line |
-| varint | number of comments |
+| varuint | the start line |
+| varuint | number of comments |
 | comment* | comments |
-| varint | number of magic comments |
+| varuint | number of magic comments |
 | magic comment* | magic comments |
 | location? | the optional location of the `__END__` keyword and its contents |
-| varint | number of errors |
+| varuint | number of errors |
 | diagnostic* | errors |
-| varint | number of warnings |
+| varuint | number of warnings |
 | diagnostic* | warnings |
 | `4` | content pool offset |
-| varint | content pool size |
+| varuint | content pool size |
 
 After the header comes the body of the serialized string.
 The body consists of a sequence of nodes that is built using a prefix traversal order of the syntax tree.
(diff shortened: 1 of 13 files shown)


casperisfine:
Looks like I broke JRuby and Truffle. I suppose there's some Java version of varints? I'll have a look.

templates/lib/prism/serialize.rb.erb
```diff
@@ -147,6 +147,27 @@ module Prism
 
       # variable-length integer using https://en.wikipedia.org/wiki/LEB128
       # This is also what protobuf uses: https://protobuf.dev/programming-guides/encoding/#varints
+      def load_varsint
```
eregon, line 150: Just curious, did you write this code from the Wikipedia page?
I was thinking to use the zig-zag encoding like protobuf but this seems simpler so :+1:

eregon:
> Looks like I broke JRuby and Truffle. I suppose there's some Java version of varints? I'll have a look.

All the FFI backend CI jobs failed, including the one using CRuby.
Use `PRISM_FFI_BACKEND=true` to debug it locally on CRuby.
Probably an issue in encoding or decoding of varsint.

eregon:
In Loader.java.erb there is `source.setStartLine(loadVarInt());` so that needs to change too.

docs/serialization.md
```diff
@@ -11,7 +11,12 @@ Let us define some simple types for readability.
 
 ### varuint
 
-A variable-length integer with the value fitting in `uint32_t` using between 1 and 5 bytes, using the [LEB128](https://en.wikipedia.org/wiki/LEB128) encoding.
+A variable-length unsigned integer with the value fitting in `uint32_t` using between 1 and 5 bytes, using the [LEB128](https://en.wikipedia.org/wiki/LEB128) encoding.
+This drastically cuts down on the size of the serialized string, especially when the source file is large.
+
+### varsint
+
+A variable-length signed integer with the value fitting in `int32_t` using between 1 and 5 bytes, using the [LEB128](https://en.wikipedia.org/wiki/LEB128) encoding.
```
eregon, line 19: This should link to https://protobuf.dev/programming-guides/encoding/#signed-ints and mention ZigZag as it's using that.

src/prism.c
```diff
@@ -17238,7 +17236,7 @@ pm_serialize(pm_parser_t *parser, pm_node_t *node, pm_buffer_t *buffer) {
 PRISM_EXPORTED_FUNCTION void
 pm_serialize_parse(pm_buffer_t *buffer, const uint8_t *source, size_t size, const char *data) {
     pm_options_t options = { 0 };
-    if (data != NULL) pm_options_read(&options, data);
+    pm_options_read(&options, data);
```
eregon, line 17239: There are still 2 `if (data != NULL) pm_options_read(&options, data);` in `templates/src/serialize.c.erb`, those should be changed too to properly set the default start line.

eregon: Alternatively, `pm_options_t options = { 0 };` is kind of always wrong (it sets start line to 0), so maybe that should be replaced by
```c
pm_options_t options;
pm_options_init(&options);
```
or so (I would declare that function as static inline in `include/prism/options.h` just below the struct).

src/util/pm_buffer.c
```diff
@@ -151,6 +151,16 @@ pm_buffer_append_varuint(pm_buffer_t *buffer, uint32_t value) {
     }
 }
 
+/**
+ * Append a 32-bit signed integer to the buffer as a variable-length integer.
+ */
+void
+pm_buffer_append_varsint(pm_buffer_t *buffer, int32_t value) {
+    int64_t large_value = (int64_t)value;
+    uint32_t unsigned_int = (uint32_t)((2 * large_value) ^ (large_value >> (sizeof(int32_t) * 8 - 1)));
```
eregon, line 160: Is the conversion to `int64_t` needed?
The >> 31 shouldn't need it and is probably clearer without since it's for getting the sign bit.
The << 1 is probably fine too (https://protobuf.dev/programming-guides/encoding/#signed-ints).
So I'm thinking:
```c
uint32_t unsigned_int = ((uint32_t)(value) << 1) ^ (uint32_t)(value >> 31));
```
I found https://github.com/protocolbuffers/protobuf/blob/842f56b57a9e825d6cdd6c00947f6c19c18fe466/src/google/protobuf/io/coded_stream.h#L886-L887 so it seems correct.

templates/java/org/prism/Loader.java.erb
```diff
@@ -284,6 +284,11 @@ public class Loader {
         return x;
     }
 
+    private int loadVarSInt() {
+        int x = loadVarUInt();
+        return (x >> 1) ^ (-(x & 1));
```
eregon, line 289: Could you use `>>>` like https://github.com/protocolbuffers/protobuf/blob/main/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L508-L510?
I would also add a `// From URL` comment with that like above

eregon (commented):
Looks great!

src/util/pm_buffer.c
casperisfine: TIL about `>>>`.

Done both.

src/prism.c
casperisfine: Thanks, I fixed the missing callsite. `pm_options_init` would work but is a bit finicky to get right.

docs/serialization.md
casperisfine: Done.

kddnewton (approved):
Great work Jean!

We'll also need to update the deserializer for JS, but I'll handle that.

PR #1952, opened 2023-11-30
Finish encodings

### LINKED ISSUE #1843: Implement remaining encodings

kddnewton:
This PR does a couple of things related to encodings:

* Implements all of the remaining encodings that we didn't have support for
* Adds documentation for all of the remaining encodings
* Add tests for all of the remaining encodings
* Removes the old `pm_encoding_decode_callback_t` functionality, which was meant to mitigate our lack of encodings
* Consolidates all of the encoding files into one, since most of them refer to similar functions

Because it's now a fixed list, we can refer to encodings through an enum. With this, a couple of things will be opened up for future work that we should do:

* We're going to need an encoding field on strings, so this can just be a single byte integer since we have a handle for them now.
* We can now serialize just a single byte for the encoding, which we should do to standardize our serialization/make it smaller.

Fixes #1843


kddnewton:
Waiting on https://github.com/ruby/ruby/pull/9084

PR #1955, opened 2023-12-01
Fix comments after HEREDOCs again.

### LINKED ISSUE #1954: `<<HEREDOC #comment` at EOF is handled wrong
Unfortunately, #1948 did not fix the problem.

The general structure of the problem is the same, but unfortunately you can't look back a token to know when not to output on_nl.

```ruby
puts <<-HEREDOC.you.can.push.the.heredoc.end.very.far.back.to_s # darn
HEREDOC
```

No matter how far back the heredoc end is, you will still have the comment last. The heredoc's offset, which would push the eof further down, is lost way back in the token list.

I think the solution I offered wasn't quite right. I wonder if we need to hold onto the furthest end_offset reached at any point in the program (maybe just by a heredoc end), then use that as the eof start.

Here's the bin/lex tokens inspected of the above right now on main. I don't think going backwards until we find the heredoc is a viable solution.
<details>
  <summary>bin/lex output with inspect on tokens</summary>

```sh
❯ bin/lex
[[1, 0], :on_ident, "puts", CMDARG]
[[1, 5], :on_heredoc_beg, "<<-HEREDOC", CMDARG]
[[2, 0], :on_heredoc_end, "HEREDOC", CMDARG]
[[1, 15], :on_period, ".", DOT]
[[1, 16], :on_ident, "you", ARG]
[[1, 19], :on_period, ".", DOT]
[[1, 20], :on_ident, "can", ARG]
[[1, 23], :on_period, ".", DOT]
[[1, 24], :on_ident, "push", ARG]
[[1, 28], :on_period, ".", DOT]
[[1, 29], :on_ident, "the", ARG]
[[1, 32], :on_period, ".", DOT]
[[1, 33], :on_ident, "heredoc", ARG]
[[1, 40], :on_period, ".", DOT]
[[1, 41], :on_ident, "end", ARG]
[[1, 44], :on_period, ".", DOT]
[[1, 45], :on_ident, "very", ARG]
[[1, 49], :on_period, ".", DOT]
[[1, 50], :on_ident, "far", ARG]
[[1, 53], :on_period, ".", DOT]
[[1, 54], :on_ident, "back", ARG]
[[1, 58], :on_period, ".", DOT]
[[1, 59], :on_ident, "to_s", ARG]
[[1, 64], :on_comment, "# darn\n", ARG]
[[2, 7], :on_eof, "", BEG]
Ripper lex                                                             Prism compat lex                                                       Prism Lex                                                             
---------------------------------------------------------------------- ---------------------------------------------------------------------- ----------------------------------------------------------------------
[[2, 0], :on_heredoc_end, "HEREDOC", CMDARG]                           [[2, 0], :on_nl, "HEREDOC", BEG]                                       [:COMMENT, [64, 7]]                                                   
nil                                                                    [[2, 0], :on_heredoc_end, "HEREDOC", CMDARG]                           [:EOF, [78, 0]]  
```
</details>

martinemde:
The problem was deeper than just looking back a single token. You can push the heredoc_end token way back into the list. 

Unfortunately, we need to save the last location of a heredoc_end to see if it's the last token in the file. I feel like this is pretty inelegant, but it solve the problem more generally.

Fixes #1954 and all of the versions of this error that I brought up there.

### DIFF
--- a/lib/prism/lex_compat.rb
+++ b/lib/prism/lex_compat.rb
@@ -610,6 +610,7 @@ module Prism
       result = Prism.lex(source, **options)
       result_value = result.value
       previous_state = nil
+      last_heredoc_end = nil
 
       # In previous versions of Ruby, Ripper wouldn't flush the bom before the
       # first token, so we had to have a hack in place to account for that. This
@@ -664,6 +665,7 @@ module Prism
           when :on_heredoc_end
             # Heredoc end tokens can be emitted in an odd order, so we don't
             # want to bother comparing the state on them.
+            last_heredoc_end = token.location.end_offset
             IgnoreStateToken.new([[lineno, column], event, value, lex_state])
           when :on_ident
             if lex_state == Ripper::EXPR_END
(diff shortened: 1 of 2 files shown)


kddnewton (approved):
This makes sense!

PR #1962, opened 2023-12-01
Remove transparent scopes

eightbitraptor:

PR #1965, opened 2023-12-01
Update docs for For Nodes

eightbitraptor:
since transparent scopes have been removed, the local variables are now at the same level and we handle the mismatch in the CRuby compiler

cf. https://github.com/ruby/prism/pull/1962


eregon (approved):

eregon:
Thanks!

PR #1967, opened 2023-12-01
Prism.parse_success?(source)

kddnewton:
A lot of tools use Ripper/RubyVM::AbstractSyntaxTree to determine if a source is valid. These tools both create an AST instead of providing an API that will return a boolean only.

This new API only creates the C structs, but doesn't bother reifying them into Ruby/the serialization API. Instead it only returns true/false, which is significantly more efficient.

### DIFF
--- a/docs/ruby_api.md
+++ b/docs/ruby_api.md
@@ -25,6 +25,8 @@ The full API is documented below.
 * `Prism.load(source, serialized)` - load the serialized syntax tree using the source as a reference into a syntax tree
 * `Prism.parse_comments(source)` - parse the comments corresponding to the given source string and return them
 * `Prism.parse_file_comments(source)` - parse the comments corresponding to the given source file and return them
+* `Prism.parse_success?(source)` - parse the syntax tree corresponding to the given source string and return true if it was parsed without errors or warnings
+* `Prism.parse_file_success?(filepath)` - parse the syntax tree corresponding to the given source file and return true if it was parsed without errors or warnings
 
 ## Nodes
 
--- a/test/prism/ruby_api_test.rb
+++ b/test/prism/ruby_api_test.rb
@@ -20,6 +20,18 @@ module Prism
       assert_equal_nodes ast2, ast3
     end
 
+    def test_parse_success?
+      assert Prism.parse_success?("1")
+      refute Prism.parse_success?("<>")
+
+      assert Prism.parse_success?("m //", verbose: false)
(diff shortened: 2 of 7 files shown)


test/prism/ruby_api_test.rb
```diff
@@ -20,6 +20,18 @@ def test_ruby_api
       assert_equal_nodes ast2, ast3
     end
 
+    def test_parse_success?
+      assert Prism.parse_success?("1")
+      refute Prism.parse_success?("<>")
+
+      assert Prism.parse_success?("m //", verbose: false)
+      refute Prism.parse_success?("m //", verbose: true)
```
schneems, line 28: Not about this PR specifically but I find this a bit confusing/surprising. It seems like action at a distance. Why does changing verbosity affect the parse state? 

I checked the docs for that option https://rubydoc.info/gems/prism/Prism#parse-class_method it doesn't mention anything about affecting behavior (other than behavior of whether or not to warn).

schneems (commented):
My High-level initial thought on seeing the title was "why would that be needed since there's `Prism.parse().success?`". On reading the description it sounds like a performance improvement I would 100% use. 

My only suggestion would be to also add a `parse_failure?(source)`. I asked a question below.

test/prism/ruby_api_test.rb
kddnewton: So in this case `parse_success?` checks if there are no errors or warnings. Verbosity controls whether warnings are emitted. So by turning off verbosity, we can say `parse_success?` was successful because no warnings were emitted.

docs/ruby_api.md
```diff
@@ -25,6 +25,8 @@ The full API is documented below.
 * `Prism.load(source, serialized)` - load the serialized syntax tree using the source as a reference into a syntax tree
 * `Prism.parse_comments(source)` - parse the comments corresponding to the given source string and return them
 * `Prism.parse_file_comments(source)` - parse the comments corresponding to the given source file and return them
+* `Prism.parse_success?(source)` - parse the syntax tree corresponding to the given source string and return true if it was parsed without errors or warnings
```
eregon, line 28: It's quite unusual to consider warnings a failure.
Also https://rubydoc.info/gems/prism/Prism%2FParseResult:success%3F only mentions errors.

I think `Prism.parse_success?(source)` should be equivalent to `Prism.parse(source).success?` and the check should be no errors.
This method should not enable warnings by default, but I guess that's the default for `verbose` anyway?

PR #1968, opened 2023-12-01
Fix heredocs with leading spaces

### LINKED ISSUE #753: Fix heredocs with leading spaces

haldun:
Closes #753


test/prism/fixtures/tilde_heredocs.txt
```diff
@@ -95,3 +95,13 @@ EOF
   #{1}
 	b
 EOT
+
+<<~'  FOO'
```
kddnewton, line 99: The tests are failing because CRuby couldn't handle this case until Ruby 3.3. So it's going to fail for all cases < 3.3. In this case could you pull these examples into a different file (including the dash heredocs) and then explicitly ignore that fixture if < 3.3? We do something similar in `parse_test.rb`.

kddnewton (approved):

PR #1970, opened 2023-12-01
Add .DS_Store to .gitignore

schneems:
While I was poking around with prism locally, git was kind to point out that I had modified this (mac specific) file. It would be convenient if it was ignored by default.


kddnewton (approved):

schneems:
Thanks!

PR #1973, opened 2023-12-01
Diagnostic & Comment location

### LINKED ISSUE #1956: Rust Diagnostics & Comments do not provide access to their location
The only methods `Diagnostic` and `Comment` provide are a way to get their message:
https://github.com/ruby/prism/blob/602252373978b23b7b7fcd8c4881f5513138550f/rust/prism/src/lib.rs#L31-L45
https://github.com/ruby/prism/blob/602252373978b23b7b7fcd8c4881f5513138550f/rust/prism/src/lib.rs#L54-L69

`pm_diagnostic_t` and `pm_comment_t` do have pointers to the start and end of the source that generated them so it is possible to add a `location()` method to `Diagnostic` and `Comment`.

I'm also noticing that in the C source `pm_comment_t`and `pm_diagnostic_t` don't use `pm_location_t` either, just a pointer to the start and end of the source, which is identical to `pm_location_t`. Perhaps they should be replaced?
https://github.com/ruby/prism/blob/602252373978b23b7b7fcd8c4881f5513138550f/include/prism/parser.h#L385-L389

Speak2Erase:
This pull request refactors `pm_comment_t` and `pm)diagnostic_t` to use `pm_location_t` instead of a `start` and `end` field, as well as adding a `location` method to `Diagnostic` and `Comment` in the Rust API. 
The only downside to this is that `Comment` and `Diagnostic` store a pointer to the parser they came from now.

I'm pretty sure I didn't need to modify the Java or Javascript bindings at all, they seem to deserialize output from prism, and `pm_location_t` and two `start` and `end` fields are serialized identically. Please correct me if I'm wrong!

Fixes #1956


Speak2Erase:
I can't seem to recreate the memcheck failures. Are they an issue with upstream ruby?

kddnewton:
Yeah don't worry about that one, it's intermittent. @peterzhu2118 this is that same one that keeps occasionally cropping up.

PR #1976, opened 2023-12-02
Mention about `MAKEFLAGS` in `README.md`

haldun:
Added an environment variable `PRISM_MAKE_FLAGS` so we can pass arguments to `make` while running `rake compile`. I find this useful to speed up the build process by passing the `-j` flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,15 @@ to fetch the Ruby dependencies. Finally, run:
 rake compile
 ```
 
-to compile the shared library. It will be built in the `build` directory. To test that everything is working, run:
+to compile the shared library.
+
+One way to speed up the build process is to utilize `MAKEFLAGS` to pass options into `make`. For example, to use 10 jobs, run:
+
+```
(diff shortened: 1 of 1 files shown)


kddnewton:
Nice this is great! Can you document it somewhere?

eregon:
> I find this useful to speed up the build process by passing the `-j` flag.

There is already a standard env var that `make` uses for that: `MAKEFLAGS`.
I have `MAKEFLAGS=-j8` set globally.

eregon:
But of course not everyone will have that set, so maybe we should just pass `-j#{Etc.nprocessors}` in Rakefile?

kddnewton:
@haldun yeah if makeflags is the standard, let's go with that. Since you didn't know it and I didn't know it, could you add it to one of our docs so that other people can learn it as well? Effectively let's turn this PR into a docs PR.

kddnewton (approved):

PR #1983, opened 2023-12-04
Add locals_body_index to DefNode, BlockNode, LambdaNode

jemmaissroff:
The locals_body_index gives the index in the locals array where the locals from the body start. This allows compilers to easily index past the parameters in the locals array.


enebo:
@jemmaissroff Can I ask what the use-case is for this?  JRuby has never done anything like this but perhaps we want to.  The main thing which comes to mind is value initialization (parameters will always be initialized already).

jemmaissroff:
> @jemmaissroff Can I ask what the use-case is for this? JRuby has never done anything like this but perhaps we want to. The main thing which comes to mind is value initialization (parameters will always be initialized already).

This is for storing locals in the local table, and specifically to help with an edge case around MultiTargetNodes. CRuby pushes a placeholder internal variable onto the stack to represent multitargets in parameters, (or also keywords). For example:

```
$ ruby --dump=insns -e "def a((b), c:); d = 1; end"
....
local table (size: 5, argc: 1 [opts: 0, rest: -1, post: 0, block: -1, kw: 1@1, kwrest: -1])
[ 5] ?@0<Arg>   [ 4] c@1        [ 3] ?@2        [ 2] b@3        [ 1] d@4
```

We need to mimic that structure. CRuby actually puts these internal variables into the parse tree itself:

```
$ ruby --dump=parsetree -e "def a((b), c:); d = 1; end"
...
#         @ NODE_SCOPE (id: 16, line: 1, location: (1,0)-(1,26))
#         +- nd_tbl: (internal variable: 0x34359214079),:c,(internal variable: 0x34359214063),:b,:d
```

In prism's case, we have the `locals`, but it includes both body locals and parameter locals. We need to iterate over the parameter locals separately, to account for putting these internal variables into the local table correctly. Once we're done iterating over the parameters though, there's no marker for where the non-parameter locals start. We could track this as we were iterating over the parameters, but it adds a layer of unnecessary accounting since Prism already knows this information.

enebo:
@jemmaissroff thanks for the explanation.  We don't have this requirement so I am thinking about whether Java serialization removes this or whether we can leverage this for making a cheap use-def in our initial build (our first phase of building can yield unitialized lvars so we null guard all accesses).  Being able to trivially remove all parameters as used would mark many scopes outright.  We could do this now but it would be doing extra work (like this PR is trying to avoid).

[note: we do have a compiler pass for use-def but it happens only on the way to JIT compilation -- tmi]

kddnewton:
@jemmaissroff thanks for doing this, I think it looks really good. I think we can merge this as-is, but I had a thought.

Right now, we're adding 4 bytes to every method, block, and lambda. However this integer is only going to be incremented in the case that there are parameters. What do you think about putting this on the parameters node, and instead calling it `locals_count`? That way the compiler could assume 0, and if there _is_ a parameters node then use that as effectively the same thing. Then we would only be allocating those 4 bytes in the case that it was non-0, which would mean the default case would be much smaller.

Again, happy to merge as is if you want to just get this working, but I think this change would be worthwhile so I'm happy to do it after merge or now before.

jemmaissroff:
@kddnewton I'm on board with this change. If you have cycles to do it, will def be quicker for you than for me. If not though, I might merge as is, and keep going now that I'm unblocked to fix the related bugs in compilation, and then come back and change this (since it sounds like it shouldn't impact other implementations except by slightly increasing size)

kddnewton (approved):
For sure. Let's go ahead and merge, I'll do this after I finish the encoding stuff.

eregon:
> I am thinking about whether Java serialization removes this

It keeps it, it's just an uint32 field, but the impact on serialized size is negligible: [before](https://github.com/ruby/prism/actions/runs/7103757456/job/19337187230#step:8:1) vs [after](https://github.com/ruby/prism/actions/runs/7092914171/job/19305233669#step:8:1).

> What do you think about putting this on the parameters node, and instead calling it `locals_count`?

I misunderstood this as `locals_count` = `locals.size - number_of_parameter_local_vars`.
I'd suggest to name it more explicitly, like `number_of_parameter_locals` or `parameter_locals_count`.

What follows is before I realized this misunderstanding:
It's a bit unexpected to have the number of locals on the `ParametersNode`, basically it's keeping info about something which is outside the node (the `locals` field belongs to the DefNode/BlockNode/LambdaNode, not to the ParametersNode).

eregon:
FWIW I was wondering if there are some cases where parameter locals are not always at the start of the locals of where it might not be so well defined which locals are parameter locals or not.

```
$ bin/parse -e 'def m(a=(unset=true)); b = 3; end'
@ ProgramNode (location: (1,0)-(1,33))
├── locals: []
└── statements:
    @ StatementsNode (location: (1,0)-(1,33))
    └── body: (length: 1)
        └── @ DefNode (location: (1,0)-(1,33))
            ├── name: :m
            ├── name_loc: (1,4)-(1,5) = "m"
            ├── receiver: ∅
            ├── parameters:
            │   @ ParametersNode (location: (1,6)-(1,20))
            │   ├── requireds: (length: 0)
            │   ├── optionals: (length: 1)
            │   │   └── @ OptionalParameterNode (location: (1,6)-(1,20))
            │   │       ├── name: :a
            │   │       ├── name_loc: (1,6)-(1,7) = "a"
            │   │       ├── operator_loc: (1,7)-(1,8) = "="
            │   │       └── value:
            │   │           @ ParenthesesNode (location: (1,8)-(1,20))
            │   │           ├── body:
            │   │           │   @ StatementsNode (location: (1,9)-(1,19))
            │   │           │   └── body: (length: 1)
            │   │           │       └── @ LocalVariableWriteNode (location: (1,9)-(1,19))
            │   │           │           ├── name: :unset
            │   │           │           ├── depth: 0
            │   │           │           ├── name_loc: (1,9)-(1,14) = "unset"
            │   │           │           ├── value:
            │   │           │           │   @ TrueNode (location: (1,15)-(1,19))
            │   │           │           └── operator_loc: (1,14)-(1,15) = "="
            │   │           ├── opening_loc: (1,8)-(1,9) = "("
            │   │           └── closing_loc: (1,19)-(1,20) = ")"
            │   ├── rest: ∅
            │   ├── posts: (length: 0)
            │   ├── keywords: (length: 0)
            │   ├── keyword_rest: ∅
            │   └── block: ∅
            ├── body:
            │   @ StatementsNode (location: (1,23)-(1,28))
            │   └── body: (length: 1)
            │       └── @ LocalVariableWriteNode (location: (1,23)-(1,28))
            │           ├── name: :b
            │           ├── depth: 0
            │           ├── name_loc: (1,23)-(1,24) = "b"
            │           ├── value:
            │           │   @ IntegerNode (location: (1,27)-(1,28))
            │           │   └── flags: decimal
            │           └── operator_loc: (1,25)-(1,26) = "="
            ├── locals: [:a, :unset, :b]
            ├── locals_body_index: 2
            ├── def_keyword_loc: (1,0)-(1,3) = "def"
            ├── operator_loc: ∅
            ├── lparen_loc: (1,5)-(1,6) = "("
            ├── rparen_loc: (1,20)-(1,21) = ")"
            ├── equal_loc: ∅
            └── end_keyword_loc: (1,30)-(1,33) = "end"
```
So here `unset` is considered a parameter local, which seems fair, and the order is fine.

eregon:
However that example means that it's not enough to `nil`-initialize only body locals, `unset` would also need to be `nil`-initialized here (for the case an argument is passed).

enebo:
@eregon I think you are wrong.  Not because of what you are observing but because this PR needs to put `unset` after parameters.  If we tweak your example to:

```text
mri31 --dump=parsetree -e 'def m(a=(unset=true), c); b = 3; end'
```

You will see:

```text
#         @ NODE_SCOPE (id: 13, line: 1, location: (1,0)-(1,36))
#         +- nd_tbl: :a,:c,:unset,:b
```

@jemmaissroff you may need to make some additional changes for this as well?

kddnewton:
@enebo @eregon yeah they're working through it right now, but basically CRuby is using this index to say "everything after this index is in the right order" but "everything before is _likely_ in the right order". For destructured stuff or some keywords stuff, it has to be shuffled around.

PR #1989, opened 2023-12-05
Fix defined with new line

eileencodes:
It's possible to write the following and have it be valid Ruby:

```
defined?("foo"
)
```

But Prism wasn't taking the new line into account. This adds an `accept1` for a `PM_TOKEN_NEWLINE` to account for this. I've also updated the fixtures and snapshots to test this.


kddnewton (approved):
Looks good to me!

PR #2001, opened 2023-12-06
Simplify unterminated string

### LINKED ISSUE #1946: Could unclosed double quoted string without embed be StringNode?
Both single and double quoted string without embed is just a StringNode.
```ruby
'string' # StringNode
"string" # StringNode
"#{string}" # InterpolatedStringNode
```

If double quoted string is unclosed, it will be InterpolatedStringNode with empty parts.
```ruby
Prism.parse("'string") # StringNode(content is "")
Prism.parse('"string') # InterpolatedStringNode(parts is [])
```

For some code analysis like IRB's `require "path/to/complete` completion, it will be easy if it is a StringNode just like single quoted string.

kddnewton:
Fixes #1946


kddnewton:
@tompng now they are both `StringNode`

tompng:
Thank you 😃

PR #2007, opened 2023-12-06
Emit error for constant assignments in `def`s

### LINKED ISSUE #2004: Constant assignment in methods should be a syntax error
```ruby
def foo
  Bar = 1
end
```

haldun:
Closes #2004

I copied the error message from Ruby.


kddnewton (approved):
That was quick!

PR #2008, opened 2023-12-06
BlockParametersNode flag for ambiguous_param0

### LINKED ISSUE #1538: BlockParametersNode flag for ambiguous_param0
When you have a block like `foo { |a| }` the first parameter does autosplatting, and it's called `ambiguous_param0` in CRuby. You can tell it has this flag by running:

```
echo "foo { |a| }" | ruby -e 'pp RubyVM::InstructionSequence.compile($<.read).to_a'
```

which yields

```
["YARVInstructionSequence/SimpleDataFormat",
 3,
 2,
 1,
 {:arg_size=>0, :local_size=>0, :stack_max=>1, :node_id=>6, :code_location=>[1, 0, 1, 11], :node_ids=>[0, 0, -1]},
 "<compiled>",
 "<compiled>",
 "<compiled>",
 1,
 :top,
 [],
 {},
 [],
 [1,
  :RUBY_EVENT_LINE,
  [:putself],
  [:send,
   {:mid=>:foo, :flag=>4, :orig_argc=>0},
   ["YARVInstructionSequence/SimpleDataFormat",
    3,
    2,
    1,
    {:arg_size=>1, :local_size=>1, :stack_max=>1, :node_id=>4, :code_location=>[1, 4, 1, 11], :node_ids=>[-1, -1]},
    "block in <compiled>",
    "<compiled>",
    "<compiled>",
    1,
    :block,
    [:a],
    {:lead_num=>1, :ambiguous_param0=>true},
    [],
    [1, :RUBY_EVENT_B_CALL, [:putnil], :RUBY_EVENT_B_RETURN, [:leave]]]],
  [:leave]]]
```

you can see `ambiguous_param0` is in there. It _won't_ be in there if there are any other parameters or block locals. This should be implemented as a flag on `BlockParametersNode`.

haldun:
Closes #1538

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -342,7 +342,7 @@ flags:
   - name: BlockParametersNodeFlags
     values:
       - name: AMBIGUOUS_PARAM0
-        comment: "if there is only one required parameter"
+        comment: "if there is only one required parameter it changes the behavior of array iteration"
     comment: Flags for the block parameters nodes.
   - name: CallNodeFlags
     values:
@@ -634,6 +634,9 @@ nodes:
           end
   - name: BlockParametersNode
     fields:
+      - name: flags
+        type: flags
+        kind: BlockParametersNodeFlags
       - name: parameters
         type: node?
         kind: ParametersNode
@@ -643,9 +646,6 @@ nodes:
         type: location?
       - name: closing_loc
         type: location?
-      - name: flags
-        type: flags
-        kind: BlockParametersNodeFlags
     comment: |
       Represents a block's parameters declaration.
 
--- a/config.yml
+++ b/config.yml
@@ -339,6 +339,11 @@ flags:
       - name: CONTAINS_SPLAT
         comment: "if array contains splat nodes"
     comment: Flags for array nodes.
+  - name: BlockParametersNodeFlags
+    values:
+      - name: AMBIGUOUS_PARAM0
+        comment: "if there is only one required parameter"
+    comment: Flags for the block parameters nodes.
   - name: CallNodeFlags
     values:
       - name: SAFE_NAVIGATION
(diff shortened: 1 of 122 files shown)


config.yml
```diff
@@ -638,6 +643,9 @@ nodes:
         type: location?
       - name: closing_loc
         type: location?
+      - name: flags
```
kddnewton, line 646: Can you make this the first field in the list? We just did this yesterday but we're making all of them the first.

config.yml
```diff
@@ -339,6 +339,11 @@ flags:
       - name: CONTAINS_SPLAT
         comment: "if array contains splat nodes"
     comment: Flags for array nodes.
+  - name: BlockParametersNodeFlags
+    values:
+      - name: AMBIGUOUS_PARAM0
+        comment: "if there is only one required parameter"
```
kddnewton, line 345:
```suggestion
        comment: "if there is only one required parameter it changes the behavior of array iteration"
```

src/prism.c
```diff
@@ -11906,8 +11906,19 @@ parse_block_parameters(
             false
         );
     }
-
     pm_block_parameters_node_t *block_parameters = pm_block_parameters_node_create(parser, parameters, opening);
+
+    if (
```
kddnewton, line 11911: It's probably totally fine to do it here, but I'm wondering if you could do it inside of `parse_parameters` instead and take a pointer to a boolean.

The reason is, this is going to check every parameter type every time, whereas if we do it in the `parse_parameters` loop, then it's only going to check it when parameters are actually added.

Could you try that approach?

haldun: Sure I'll try. I've thought about it, but the checks here seemed quite cheap to me so that's why I implemented it in this way.

kddnewton: Hmm that's a good point. I'll do some benchmarks and get back to you, hold off for now.

kddnewton:
Thanks for your work on this @haldun but we're going to not go in this direction anymore.

Issue #1786, opened 2023-11-08
Rust crate packaging


turboladen:
It's been some months since I've been in this neck of the woods, and it seems lots has changed (which is exciting)! I just went to give a shot using the `prism` crate in a project and I get build failures with `prism-sys` (`vendor_dir` doesn't exist). Is this expected? After poking at it, it _looks_ like `prism-sys` expects you to build it from the prism repo/workspace (is that right?). ...which makes me wonder about the current thoughts on/state of crate packaging.

IIRC, there were some discussions about packaging the crates, but I didn't find anything in the Issues list, so I thought I'd ask before I started poking at it more:

* Is my experience the currently expected behavior?
* Have there been any more discussions on the topic? 
* Do you all have ideas floating around about how you'd like to approach this (or maybe approaches you want to avoid)? 

---

FWIW, here's the trace I get when trying to pull in `prism` to a project:

```
   Compiling prism v0.17.1 (https://github.com/ruby/prism#4a459b23)
   Compiling prism-sys v0.17.1 (https://github.com/ruby/prism#4a459b23)
error: failed to run custom build command for `prism-sys v0.17.1 (https://github.com/ruby/prism#4a459b23)`

Caused by:
  process didn't exit successfully: `/Users/steve.loveless/Development/projects/ruby_analyzer/target/debug/build/prism-sys-78c8b4886147ff3d/build-script-main` (exit status: 101)
  --- stderr
  thread 'main' panicked at /Users/steve.loveless/.cargo/git/checkouts/prism-964b59ab8481264d/4a459b2/rust/prism-sys/build/vendored.rs:8:5:
  Prism source directory does not exist, expected: /Users/steve.loveless/.cargo/git/checkouts/prism-964b59ab8481264d/4a459b2/rust/prism-sys/vendor/prism-0.17.1
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

kddnewton:
Hey @turboladen!

It's not so much the expected behavior as we just haven't figured it out yet. Ideally I'd like to release everything to cargo and have it be all packaged up correctly, I'm just not entirely sure how to do that yet.

I'm assuming you're pointing to the github directly in cargo? Maybe we need additional build options set up in order to properly handle that.

Overall the answer is we just don't know yet!

turboladen:
Ok, great! I'm interested in digging to this, so I might do that in the near future. Just wanted to make sure I wasn't gonna go off on some known undesirable path. 🙂 

[...]

PR #2009, opened 2023-12-07
Rust crate publishing (dry-run)

ngroman:
I think this is a possible fix for https://github.com/ruby/prism/issues/1786

This change consolidates the `prism` and `prism-sys` crates into a single workspace that shares one `Cargo.lock`.

The somewhat unfortunate part here is that for `prism-sys` I have to use `--allow-dirty` with `cargo publish` because the C files and headers that are copied over by `rake cargo:build` must also be published.

## Testing
Ran `git clean -xfd && RBENV_VERSION=3.1 bundle exec rake cargo:publish:dry`

Note this still fails with error because `prism-sys` does not exist on crates.io


kddnewton:
Sorry I didn't manage to get to this this week, I'll work on it next week. At first glance it looks good.

kddnewton:
@ngroman going to try to publish this today/tomorrow.

kddnewton:
This is now merged!

Issue #2015, opened 2023-12-07
[COMPILER] Compiling array within a hash


jemmaissroff:
Minimal repro:

```
{[] => 1}
```
leads to `method `hash' called on hidden T_ARRAY object`

Found from: `RUBY_ISEQ_DUMP_DEBUG=prism make test/coverage/test_coverage.rb`

paracycle:
There is a similar problem that exhibits itself with a similar message, which can be reproduced by:
```ruby
RubyVM::InstructionSequence.compile_prism(<<~RUBY).eval
  def foo a: [], b: []
    p a
    p b
  end
  foo(a: [Symbol])
RUBY
```
and it outputs: 
```
[Symbol]
<compiled>:2:in `p': method `inspect' called on hidden T_ARRAY object (0x00000001033d4de0 flags=0xa00002807) (NotImplementedError)
        from <compiled>:2:in `foo'
        from <compiled>:4:in `<compiled>'
        from ../test.rb:39:in `eval'
        from ../test.rb:39:in `<main>'
```
Note that since the second optional param is not passed, it causes an error on the `p b` line, so this version of the error looks like a default parameter bug.

I found this by isolating it from the failure of `RUBY_ISEQ_DUMP_DEBUG=prism make ../test/ostruct/test_ostruct.rb`

paracycle:
Oh I think I know what's wrong with #2016. The PR checks if `element` is any of the types `PM_NODE_ARRAY`, `PM_NODE_HASH` and `PM_NODE_RANGE`, but the element won't be any of those types for a hash, it will be `PM_NODE_ASSOC`.

Now, `PM_NODE_ASSOC` also has a static literal check of its own, but that only checks to see if the `value` (not the `key`) is an array, hash or range.

So, none of our code actually makes the `PM_HASH_NODE` in the above example to not be a static literal.

paracycle:
OK, I have what I think is a fix here: https://github.com/ruby/prism/pull/2055 and a test that passes on ruby/ruby with that change here: https://github.com/ruby/ruby/pull/9215

PR #2016, opened 2023-12-08
Hashes need to deoptimize based on their contents

tenderlove:
If a hash has children that are not "basic" types of objects, then we can't consider the hash to be a static literal.

Fixes: #2015

I didn't see how to add tests for this, so I didn't.  I _think_ this is the fix, but I don't understand how to test.


kddnewton (approved):
This definitely looks right.

We need to expose these flags to Ruby/serialization in order to properly test it. But for now this mirrors what we do with arrays, so it makes sense to do it here.

PR #2024, opened 2023-12-08
Split up CallNode in target position

### LINKED ISSUE #1723: Consider adding a Target counterpart for `CallNode`
It seems to me reasonable (at least for the sake of consistency) to have a target node for a method call, e.g. CallTargetNode.

It would be helpful to differentiate (without checking the context) a simple method call from a synthetic call node that is missing the last argument (that will be assigned in runtime). Such a target call node might be present in the `rescue` operator and multi-assignment and could be either attribute assignment (`a.foo, b = []`) or reference assignment (`a[:foo], b = []`).

Does it make sense?

kddnewton:
In this commit we're splitting up the call nodes that were in target positions (that is, for loop indices, rescue error captures, and multi assign targets).

Previously, we would simply leave the call nodes in place. This had the benefit of keeping the AST relatively simple, but had the downside of not being very explicit. If a static analysis tool wanted to only look at call nodes, it could easily be confused because the method would have 1 fewer argument than it would actually be called with.

This also brings some consistency to the AST. All of the nodes in a target position are now *TargetNode nodes. These should all be treated the same, and the call nodes can now be treated the same.

Finally, there is benefit to memory. Because being in a target position ensures we don't have some fields, we can strip down the number of fields on these nodes.

So this commit introduces two new nodes: CallTargetNode and IndexTargetNode. For CallTargetNode we get to drop the opening_loc, closing_loc, arguments, and block. Those can never be present. We also get to mark their fields as non-null, so they will always be seen as present.

The IndexTargetNode keeps around most of its fields but gets to drop both the name (because it will always be []=) and the message_loc (which was always super confusing because it included the arguments by virtue of being inside the []).

Overall, this adds complexity to the AST at the expense of memory savings and explicitness. I believe this tradeoff is worth it in this case, especially because these are very much not common nodes in the first place.

All of this should be visible in the snapshot changes, which you can check out in this PR. Fixes #1723.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -775,6 +775,32 @@ nodes:
 
           foo.bar ||= value
           ^^^^^^^^^^^^^^^^^
+  - name: CallTargetNode
+    fields:
+      - name: flags
+        type: flags
+        kind: CallNodeFlags
+      - name: receiver
+        type: node
+      - name: call_operator_loc
+        type: location
+      - name: name
+        type: constant
+      - name: message_loc
+        type: location
+    comment: |
+      Represents assigning to a method call.
+
+          foo.bar, = 1
+          ^^^^^^^
+
+          begin
+          rescue => foo.bar
+                    ^^^^^^^
+          end
+
+          for foo.bar in baz do end
+              ^^^^^^^
   - name: CapturePatternNode
     fields:
       - name: value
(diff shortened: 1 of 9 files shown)


config.yml
```diff
@@ -1581,6 +1607,35 @@ nodes:
...
+    fields:
+      - name: flags
+        type: flags
+        kind: CallNodeFlags
+      - name: receiver
+        type: node
+      - name: opening_loc
+        type: location
+      - name: arguments
+        type: node
```
tompng, line 1620: I think type is `node?` for `a[], = 1` and `a[&b], = 1`

jemmaissroff (approved):
This is much more explicit, thanks!

PR #2033, opened 2023-12-09
fix: escape newline

k0va1:
If you try to print error message using `printf` it'll add newline. So I think that would be better to escape `\n'


kddnewton (approved):
Thanks!

PR #2035, opened 2023-12-09
Fix git source build

tompng:
`gem 'prism', github: 'ruby/prism'` fails to bundle install with:
```
extconf.rb:57:in `<main>': prism.h is required (RuntimeError)
```
mkmf.log:
```
fatal error: prism/ast.h: No such file or directory
```

This pull request changes to generate templates before `find_header("prism.h", dir)`.


kddnewton (approved):
Thanks!

PR #2049, opened 2023-12-11
Add public macro and internal functions for working with node flags

paracycle:
The current way of setting/unsetting and testing flags on node was very verbose, and, as a result, was error prone.

This PR introduces the public macro `PM_NODE_FLAG_P(node, flag)` (exported from `ast.h`) to check if `flag` is set on `node`.

It also adds two methods in `prism.c` to set/unset flags on nodes: `pm_node_flag_set` and `pm_node_flag_unset`.

These make the code easier to read and less error-prone.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -865,6 +865,27 @@ pm_arguments_validate_block(pm_parser_t *parser, pm_arguments_t *arguments, pm_b
 }
 
 /******************************************************************************/
+/* Node flag handling functions                                               */
+/******************************************************************************/
+
+/**
+ * Set the given flag on the given node.
+ */
+static inline void
+pm_node_flag_set(pm_node_t *node, pm_node_flags_t flag) {
+    node->flags |= flag;
+}
+
+/**
+ * Remove the given flag from the given node.
+ */
+static inline void
+pm_node_flag_unset(pm_node_t *node, pm_node_flags_t flag) {
+    node->flags &= (pm_node_flags_t) ~flag;
+}
+
+
+/******************************************************************************/
 /* Node creation functions                                                    */
 /******************************************************************************/
 
--- a/src/prism.c
+++ b/src/prism.c
@@ -1173,12 +1173,12 @@ pm_array_node_elements_append(pm_array_node_t *node, pm_node_t *element) {
 
     // If the element is not a static literal, then the array is not a static
     // literal. Turn that flag off.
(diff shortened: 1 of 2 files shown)


templates/include/prism/ast.h.erb
```diff
@@ -116,6 +116,21 @@ static const pm_node_flags_t PM_NODE_FLAG_COMMON_MASK = (1 << (PM_NODE_FLAG_BITS
  */
 #define PM_NODE_TYPE_P(node, type) (PM_NODE_TYPE(node) == (type))
 
+/**
+ * Sets the given flag on the given node.
+ */
+#define PM_NODE_FLAG_SET(node, flag) (((pm_node_t *)(node))->flags |= (flag))
```
paracycle, line 122: Some of the `node`s passed to these macros are specific node types and others were of type `pm_node_t *`. That meant we need to do `->base.flags` for the former, but only `->flags` for the latter, which broke the symmetry.

Since casting a specific node type should always give the `base` pointer, this cast should do the right thing and allow us to access flags using just `->flags`.

kddnewton (approved):
This looks good, but can you rebase first with the keyword hash changes and use the macro?

paracycle:
> This looks good, but can you rebase first with the keyword hash changes and use the macro?

Yeap, that was what I was intending to do. Thanks for the review.

src/prism.c
```diff
@@ -3329,8 +3329,8 @@ pm_hash_node_elements_append(pm_hash_node_t *hash, pm_node_t *element) {
 
     // If the element is not a static literal, then the hash is not a static
     // literal. Turn that flag off.
-    if (PM_NODE_TYPE_P(element, PM_ARRAY_NODE) || PM_NODE_TYPE_P(element, PM_HASH_NODE) || PM_NODE_TYPE_P(element, PM_RANGE_NODE) || (element->flags & PM_NODE_FLAG_STATIC_LITERAL) == 0) {
-        hash->base.flags &= (pm_node_flags_t) ~PM_NODE_FLAG_STATIC_LITERAL;
+    if (PM_NODE_TYPE_P(element, PM_ARRAY_NODE) || PM_NODE_TYPE_P(element, PM_HASH_NODE) || PM_NODE_TYPE_P(element, PM_RANGE_NODE) || PM_NODE_FLAG_P(element, PM_NODE_FLAG_STATIC_LITERAL)) {
```
paracycle, line 3332: Ugh, this is not correct, since the previous code was checking for `== 0`, so the check should be `!PM_NODE_FLAG_P`. Fixing...

paracycle: Done.

kddnewton (approved):

PR #2069, opened 2023-12-13
Fix the implementation of the flag on keyword hash nodes

paracycle:
The previous implementation in #2046 was incorrect since it was just checking for all keys in assoc nodes to be static literals but the actual check is that all keys in assoc nodes must be symbol nodes (Ref: https://github.com/ruby/ruby/blob/2f1d6da8c45590bf3461ed4bf051a4e1009eaf85/compile.c#L4577-L4578).

This commit fixes that implementation, and, also, aliases the flag to `PM_KEYWORD_HASH_NODE_FLAGS_SYMBOL_KEYS` so that ruby/ruby can start using the new flag name.

I intend to later change the real flag name to `PM_KEYWORD_HASH_NODE_FLAGS_SYMBOL_KEYS` and remove the alias.

### DIFF
--- a/include/prism/parser.h
+++ b/include/prism/parser.h
@@ -17,6 +17,12 @@
 
 #include <stdbool.h>
 
+// TODO: remove this by renaming the original flag
+/**
+ * Temporary alias for the PM_NODE_FLAG_STATIC_KEYS flag.
+ */
+#define PM_KEYWORD_HASH_NODE_FLAGS_SYMBOL_KEYS PM_KEYWORD_HASH_NODE_FLAGS_STATIC_KEYS
+
 /**
  * This enum provides various bits that represent different kinds of states that
  * the lexer can track. This is used to determine which kind of token to return
--- a/src/prism.c
+++ b/src/prism.c
@@ -3949,7 +3949,7 @@ pm_keyword_hash_node_create(pm_parser_t *parser) {
         .base = {
             .type = PM_KEYWORD_HASH_NODE,
             .location = PM_OPTIONAL_LOCATION_NOT_PROVIDED_VALUE,
-            .flags = PM_KEYWORD_HASH_NODE_FLAGS_STATIC_KEYS
+            .flags = PM_KEYWORD_HASH_NODE_FLAGS_SYMBOL_KEYS
         },
         .elements = { 0 }
     };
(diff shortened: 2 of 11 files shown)


kddnewton (approved):
Do you need to check labels as well? Right now this will accept

```ruby
foo(:bar => 1)
```

I assume that's okay?

paracycle:
@kddnewton good question.

I checked with these changes on `ruby/ruby` with this code:
```ruby
def a(foo:) = p foo

a(:foo => 1)
```
and the output from the current and Prism compiler for this are:
```
== disasm: #<ISeq:<compiled>@<compiled>:0 (0,0)-(1,12)>
0000 definemethod                           :a, a
0003 putself                                                          (   1)[Li]
0004 putobject                              1
0006 send                                   <calldata!mid:a, argc:1, kw:[foo], FCALL|KWARG>, nil
0009 leave

== disasm: #<ISeq:a@<compiled>:0 (0,0)-(0,19)>
local table (size: 2, argc: 0 [opts: 0, rest: -1, post: 0, block: -1, kw: 1@1, kwrest: -1])
[ 2] foo@0      [ 1] ?@1
0000 putself                                [Ca]
0001 getlocal                               foo@0, 0
0004 send                                   <calldata!mid:p, argc:1, FCALL|ARGS_SIMPLE>, nil
0007 leave                                  [Re]
```

```
== disasm: #<ISeq:<compiled>@<compiled>:0 (0,0)-(1,12)>
0000 definemethod                           :a, a
0003 putself                                                          (   1)[Li]
0004 putobject                              1
0006 send                                   <calldata!mid:a, argc:1, kw:[foo], FCALL|KWARG>, nil
0009 leave

== disasm: #<ISeq:a@<compiled>:0 (0,0)-(0,19)>
local table (size: 2, argc: 0 [opts: 0, rest: -1, post: 0, block: -1, kw: 1@1, kwrest: -1])
[ 2] foo@0      [ 1] ?@1
0000 putself
0001 getlocal                               foo@0, 0
0004 send                                   <calldata!mid:p, argc:1, FCALL|ARGS_SIMPLE>, nil
0007 leave
```

and they eval to the same result of `1`.

I guess this is doing the right thing, right?

PR #2088, opened 2023-12-14
Update args of Prism.dump on document

riseshia:
Link: https://github.com/ruby/www.ruby-lang.org/pull/3156

function def: https://github.com/ruby/prism/blob/53ef1e8a193cc24d4ce3e0d2d2eade450125ce19/ext/prism/extension.c#L254-L258

From https://github.com/ruby/prism/pull/1763, filepath args of Prism.dump seems to moved in kwargs.

### DIFF
--- a/docs/ruby_api.md
+++ b/docs/ruby_api.md
@@ -14,7 +14,7 @@ The full API is documented below.
 
 ## API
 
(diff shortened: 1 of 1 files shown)


riseshia:
could you check this? @kddnewton

kddnewton (approved):

kddnewton:
Thanks @riseshia!

PR #2090, opened 2023-12-15
Add Node groups document

andrykonchin:
I found it useful to have such document because nodes in the `config.yml` file are sorted by name and the file is too big to scroll it and discover similar nodes.

Any suggestions are welcomed and appreciated.

### DIFF
--- a/docs/node_groups.md
+++ b/docs/node_groups.md
@@ -1,0 +1,272 @@
+# Node groups
+
+It might be helpful to have semantically close nodes grouped and categorised to easily navigate through the list of nodes in config.yml.
+
+## Generic containers
+
+- ProgramNode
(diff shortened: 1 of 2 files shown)


kddnewton:
I understand the motivation here, but I'm nervous about this thing getting out of sync/out of date pretty quickly. What do you think about putting a "category" or something similar in `config.yml` and then generating this document instead?

kddnewton (changes requested):
(Just marking this as "request changes")

kddnewton:
@andrykonchin do you have time to implement the changes here? Otherwise I'll close this and we can come back to it later.

kddnewton:
Let's come back to this later.

PR #2099, opened 2023-12-15
RBS type fixes for positional parameters

nixme:
Hello, I noticed there was mismatches between definitions of Node methods which used positional parameters and the RBS definitions which used keyword. This PR adjusts the templates to emit positional. I also updated prism_static.rbs based on the changes I saw in the other classes.

Question: I used steep to figure out what needed work. Do you want to add tests to prevent stuff from going out of sync, or is it too soon?

### DIFF
--- a/sig/prism_static.rbs
+++ b/sig/prism_static.rbs
@@ -1,32 +1,79 @@
 module Prism
-  class ParseResult
-    def value: () -> ProgramNode
-    def comments: () -> Array[Comment]
-    def errors: () -> Array[ParseError]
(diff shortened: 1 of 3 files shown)


kddnewton:
@nixme if you're up for adding steep in this PR, that would be great

kddnewton (approved):
Actually I'm good just merging this as is. But if you're up for adding Steep, that would be great.

nixme:
Yeah, I'm up for it. I'll try to throw a PR up this weekend.

Issue #2058, opened 2023-12-13
Adding kind for node[] fields where possible


eregon:
For at least some of the node[] we could give a kind and so that array's elements would have a proper type in documentation, Java code, etc.
Some example:
* MatchWriteNode#targets, (I think) always LocalVariableTargetNode

Some counter example:
* ParametersNode#requireds, not always RequiredParameterNode, can also be MultiTargetNode

In some cases it's not possible because multiple classes are possible, or a MissingNode is possible.

kddnewton:
Yes I agree. I think we should also use this opportunity to expand the syntax to support a set of nodes for languages where a union makes sense.

eregon:
Indeed, a union would be super helpful, even if it only ends up in documentation.
(Java doesn't have union types unfortunately)

eregon:
Probably we can consider this done when https://github.com/ruby/prism/pull/2101#discussion_r1445812790 is merged.

PR #2101, opened 2023-12-19
Type check RBS and Ruby files

nixme:
Type checking using Steep. (continued from #2099)

I wanted to get a gut check too see if what I'm doing in this PR is the right direction. I also have a few questions:

- ~I'm down to 80 errors, most of them nil-check related.~ **0 errors now**
	- I've avoiding touching `lib/prism/**/*.rb` except where I've thought the type checking found an actual bug (see inline comments), or to add steep type assertions via comments. 
	- ~To fix many of the rest, I'll need to touch the code a bit more. Mostly adding `or raise`, or extracting calls in a chain to a local variable and adding a type assertion. How do you feel about that?~ **I've added more runtime checks in the latest revisions**
	- ~Solving #2058 could fix 16 errors. Should I take a stab at that? I was thinking a `kinds:` list in config.yaml for `node[]`~ **Done. see config.yaml**
- I broke up `prism.rbs` and `prism_static.rbs` into RBS files that match their equivalent RB file in `lib/prism/**/*.rb` path. Is this okay? I can undo that if not.
- Since I'm using steep, I needed to add private methods to the RBS files so it can properly check everything. ~That means those members are in the RBS packaged with the gem. I can split that stuff up into a `_private` directory if you'd prefer.~ **Private RBS declarations are in _private directories not included in the gem**


lib/prism/desugar_compiler.rb
```diff
@@ -160,7 +160,7 @@ def desugar_operator_write_node(node, read_class, write_class, *arguments)
           0,
           read_class.new(*arguments, node.name_loc),
           nil,
-          node.operator_loc.slice.chomp("="),
+          node.operator_loc.slice.chomp("=").to_sym,
```
nixme, line 163: This was caught by steep, and I made this change because `CallNode` uses a Symbol for `name`

lib/prism/lex_compat.rb
```diff
@@ -240,10 +242,10 @@ class IgnoredNewlineToken < Token
       def ==(other) # :nodoc:
         return false unless self[0...-1] == other[0...-1]
 
-        if self[4] == Ripper::EXPR_ARG | Ripper::EXPR_LABELED
-          other[4] & Ripper::EXPR_ARG | Ripper::EXPR_LABELED > 0
+        if self[3] == Ripper::EXPR_ARG | Ripper::EXPR_LABELED
+          other[3] & Ripper::EXPR_ARG | Ripper::EXPR_LABELED > 0
         else
-          self[4] == other[4]
+          self[3] == other[3]
```
nixme, line 248: I couldn't quite figure out a good way to trigger these lines, but I believe these should be 3 not 4.

lib/prism/lex_compat.rb
```diff
@@ -841,7 +843,7 @@ def result
       # We sort by location to compare against Ripper's output
       tokens.sort_by!(&:location)
 
-      ParseResult.new(tokens, result.comments, result.magic_comments, result.data_loc, result.errors, result.warnings, [])
+      ParseResult.new(tokens, result.comments, result.magic_comments, result.data_loc, result.errors, result.warnings, Source.new(source))
```
nixme, line 846: steep caught this and I changed it to return a Source in the last position.

lib/prism/pack.rb
```diff
@@ -148,6 +148,8 @@ def describe
             end
           when LENGTH_MAX
             base + ", as many as possible"
+          else
+            raise
```
nixme, line 152: my read from pack.c is that INTEGER type would never return anything but `LENGTH_FIXED` or `LENGTH_MAX`, but I couldn't model that dependent enum structure in RBS. So added this `else` case to satisfy steep's flow analysis.

templates/sig/prism/node.rbs.erb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  interface _NodeImpl
+    def accept: (_Visitor) -> void
+    def comment_targets: () -> Array[node | Location]
+    def compact_child_nodes: () -> Array[node]
+    def inspect: (?NodeInspector) -> String
+  end
+
+  type node = Node & _NodeImpl
```
nixme, line 22: This class and interface structure ended up being necessary because methods taking a Node actually only can take a concrete subclass of Node, since they might call accept or inspect, etc.

If you're okay with making Node a module instead of a class, then I think we can simplify this since no instances of a Node will exist at runtime.

Another option if you want to keep Node as a class: If we split up the sigs into public (included in the gem) vs private, I'll move `class Node` definition into the private side. Then make a `_Node` interface that has all the methods currently in `class Node` and `interface _NodeImpl` and only the subclasses will `include _Node`. Then we can get rid of this type alias and use `_Node` everywhere.

sig/prism/debug.rbs
kddnewton, line 1: This module is meant to be private and used for local debugging only. I wouldn't want this showing up in type information for consumers.

sig/prism/lex_compat.rbs
kddnewton, line 1: This is meant to be a private implementation. The only thing the end-user should see is `Prism::lex_compat` giving them back an array of arrays.

Steepfile
```diff
@@ -0,0 +1,23 @@
...
+
+target :lib do
+  signature "sig"
+
+  check "lib"
+
+  ignore "lib/prism/serialize.rb"
+  ignore "lib/prism/ffi.rb"
+
+  library "ripper"
```
kddnewton, line 13: I hope this isn't actually necessary, as it should be hidden and only used for testing.

Steepfile
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+D = Steep::Diagnostic
```
kddnewton, line 3: Is this necessary?

Steepfile
```diff
@@ -0,0 +1,23 @@
...
+
+target :lib do
+  signature "sig"
+
+  check "lib"
+
+  ignore "lib/prism/serialize.rb"
+  ignore "lib/prism/ffi.rb"
```
kddnewton, line 11: Does this file need to be ignored? If it's just for making progress and we want to come back to it later that's fine, let's just notate it with a TODO.

Steepfile
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+D = Steep::Diagnostic
+
+target :lib do
+  signature "sig"
+
+  check "lib"
+
+  ignore "lib/prism/serialize.rb"
```
kddnewton, line 10: Does this file need to be ignored? If it's just for making progress and we want to come back to it later that's fine, let's just notate it with a TODO.

sig/manifest.yaml
```diff
@@ -0,0 +1,5 @@
+# For users of the Prism gem who use RBS, inform `rbs collection' of stdlib dependencies, since they do not appear in
+# the gemspec.
+dependencies:
+  - name: ripper
```
kddnewton, line 4: Is this necessary?

sig/prism/desugar_compiler.rbs
```diff
@@ -0,0 +1,49 @@
+module Prism
+  class DesugarCompiler < MutationCompiler
```
kddnewton, line 2: I'm not sure I actually want this entire compiler documented. The only thing you should know is that you do:

```ruby
node.accept(DesugarCompiler.new)
```

you shouldn't know about the inner workings of it. Especially not the private functions. Is there a way we can hide the implementation details here?

sig/prism/node_inspector.rbs
kddnewton, line 1: This should be a private implementation detail and should not be exposed to the consumer.

sig/prism/parse_result/comments.rbs
```diff
@@ -0,0 +1,39 @@
+module Prism
+  class ParseResult[T]
+    class Comments
```
kddnewton, line 3: This class is private and shouldn't be exposed to the user.

sig/prism/parse_result/newlines.rbs
```diff
@@ -0,0 +1,18 @@
+module Prism
+  class ParseResult[T]
+    class Newlines < Visitor
```
kddnewton, line 3: This class is private and shouldn't be exposed to the user.

sig/prism/pattern.rbs
```diff
@@ -0,0 +1,32 @@
...
+    attr_reader query: String
+
+    @compiled: Proc?
+
+    def initialize: (String query) -> void
+    def compile: () -> Proc
+    def scan: (node root) { (node) -> void } -> void
+            | (node root) -> ::Enumerator[node, void]
+
+    private
```
kddnewton, line 16: None of these private methods should be exposed to the user.

templates/sig/prism/dot_visitor.rbs.erb
```diff
@@ -0,0 +1,48 @@
...
+      def waypoint: (String value) -> void
+      def edge: (String value) -> void
+
+      def to_dot: () -> String
+    end
+
+    attr_reader digraph: Digraph
+
+    def initialize: () -> void
+    def to_dot: () -> String
```
kddnewton, line 37: This is the only part that should be exposed to the consumer.

kddnewton (changes requested):
Overall this is a really great start, thanks so much for doing this!

So far my only real feedback is that it looks like there are a lot of things that are typed that I wouldn't want the consumer to be aware of. In the Ruby code, I've marked them with `private_constant`. In general there are a lot of implementation details that I don't want people to know about because I want the freedom to change them later. This gets a little messy if we expose the types but not the constants.

If the issue is that we need the types to make steep work but don't want to expose them, then maybe we could use some kind of private types directory? I'm not sure, I haven't worked with RBS/steep in a while.

Thanks for doing this work!

lib/prism/lex_compat.rb
kddnewton: Agreed.

lib/prism/pack.rb
kddnewton: Agreed.

templates/sig/prism/node.rbs.erb
kddnewton: I think I would very much like a public/private split like you're saying here.

Gemfile
```diff
@@ -12,3 +12,4 @@ group :memcheck do
   gem "ruby_memcheck", platform: %i[mri mswin mingw x64_mingw]
 end
 gem "rbs", platform: %i[mri mswin mingw x64_mingw]
+gem "steep", platform: %i[mri mswin mingw x64_mingw]
```
eregon, line 15: `steep` seems to have many dependencies, how about making it part of a `group :types` or so, and also move `rbs` there? So it's possible to build prism without needing to install all these gems (some are C exts).

Steepfile
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+target :lib do
+  signature "sig"
+
+  # Tell Steep about stdlib require's
+  library "ripper" # RipperCompat
+  library "delegate"  # LexCompat
+  library "cgi"  # DotVisitor
```
nixme, line 9: Re: Ripper usage

* None of the public RBS files mention Ripper anymore
* `rake check_annotations` validates this because the `-r prism` call ignores dirs starting with `_` by convention.
* The gemspec only includes the public RBS files and no longer includes a manifest.yml

Now Steep is used within the project to statically type-check the implementations against the RBS files. The comment-based type annotations are still limited, so I've added RBS declarations as needed in the sig/prism/_private folder.

* Since there are `require`s to ripper, delegate, and cgi in various files in lib/prism/*.rb, I need to tell Steep about them so it can load the stdlib RBS files that are inside the `rbs` gem. Core Ruby RBS is loaded automatically.
* My hunch is steep doesn't automatically pull them in when it sees the `require`s because LOAD_PATH is dynamic, so it can't statically work that out.

config.yml
```diff
@@ -2266,16 +2281,36 @@ nodes:
...
         type: node[]
+        kind:
+          - RequiredKeywordParameterNode
+          - OptionalKeywordParameterNode
       - name: keyword_rest
         type: node?
+        kind:
+          - KeywordRestParameterNode
+          - ForwardingParameterNode
+          - NoKeywordsParameterNode
```
nixme, line 2313: A stab at #2058, especially where it helped get type errors down without adding assertions or checks.

I'm treating lists under `kind` as a union of node types. See associated changes in template.rb and build.rs

Steepfile
nixme: I've ignored additional files than the previous rev and added a TODO comment (but I've kept the WIP private RBS files for them).

This gets us down to 0 issues during `rake typecheck:steep`. I'll work on these files in future PRs if that's okay.

lib/prism/parse_result.rb
```diff
@@ -219,7 +215,8 @@ def join(other)
     # the beginning of the file. Useful for when you want a location object but
     # do not care where it points.
     def self.null
-      new(nil, 0, 0)
+      source = nil #: Source
+      new(source, 0, 0)
```
nixme, line 219: So this one's tricky. I chose to type `Location#source` as `Source`, not `Source | nil`. That let me avoid a lot of runtime nil checks, but on the flip side required unsound assertions here in `Location.null` and also in `DSL#Location`. Both places felt like edge-case usages, so I went with that.

Do you have a preference?

nixme (commented):
Here's another rev. I went ahead and did the public/private split. I've left a few comments inline.

Besides the split, I want to check that the changes to config.yml are okay. Thank you!

Steepfile
nixme: See comment in https://github.com/ruby/prism/pull/2101#discussion_r1445805825

lib/prism/ripper_compat.rb
```diff
@@ -108,7 +108,8 @@ def parse
 
     # Visit a CallNode node.
     def visit_call_node(node)
-      if !node.message.match?(/^[[:alpha:]_]/) && node.opening_loc.nil? && node.arguments&.arguments&.length == 1
+      message = node.message
+      if message && !message.match?(/^[[:alpha:]_]/) && node.opening_loc.nil? && node.arguments && node.arguments.arguments && node.arguments.arguments.length == 1
```
nixme, line 112: Steep's flow analysis doesn't follow `&.`, so I expanded it if that's okay. Otherwise I can wrangle a type assertion on a local variable.

nixme:
One more thing, if you pull down this branch, use vscode-steep with it, cause it's much more ergonomic to hover over local variables and see their types.

config.yml
eregon: Nice!

lib/prism/parse_result.rb
eregon: IMO the type should represent the reality, so Location#source can be nil and in fact it's not hard to see objects which have that when building new nodes with the DSL, which uses Location.null as default.

nixme: Addressed in ac0af0533b59e8ffe21f4a4cfb06c4447ec82174

lib/prism/parse_result.rb
```diff
@@ -113,12 +113,8 @@ def initialize(source, start_offset, length)
...
     # Create a new location object with the given options.
-    def copy(**options)
-      Location.new(
-        options.fetch(:source) { source },
-        options.fetch(:start_offset) { start_offset },
-        options.fetch(:length) { length }
-      )
+    def copy(source: self.source, start_offset: self.start_offset, length: self.length)
+      Location.new(source, start_offset, length)
     end
```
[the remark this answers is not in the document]
kddnewton: This is fine

lib/prism/parse_result.rb
```diff
@@ -221,6 +217,13 @@ def join(other)
     def self.null
       new(nil, 0, 0)
     end
+
+    private
+
+    # Access the associated source, raising if not present.
+    def source!
```
kddnewton, line 224: Why is this necessary? Shouldn't a source always be present?

nixme:
Sorry for the lag, I'll try picking this back up this weekend.

lib/prism/pattern.rb
```diff
@@ -174,7 +182,13 @@ def compile_hash_pattern_node(node)
 
       preprocessed =
         node.elements.to_h do |element|
-          [element.key.unescaped.to_sym, compile_node(element.value)]
+          key = element.key
+          if key.respond_to?(:unescaped)
```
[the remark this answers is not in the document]
nixme: `element` here is an `AssocNode`. Its `key` property is typed as just any node according to config.yml, and since not all nodes implement `unescaped`, I needed this check.

Should I revert and use a steep:ignore? Or if you know what all the possible key types could be, I could adjust config.yml.

lib/prism/pattern.rb
```diff
@@ -69,21 +69,29 @@ def initialize(query)
...
+      compile_node(in_node.pattern)
     end
 
     # Scan the given node and all of its children for nodes that match the
     # pattern. If a block is given, it will be called with each node that
     # matches the pattern. If no block is given, an enumerator will be returned
     # that will yield each node that matches the pattern.
     def scan(root)
-      return to_enum(__method__, root) unless block_given?
+      return to_enum(__method__ || raise, root) unless block_given?
```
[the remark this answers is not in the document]
nixme: RBS defines `Kernel#__method__` as returning `Symbol | nil`. Steep doesn't narrow the definition when inside a method. See https://github.com/ruby/rbs/blob/master/core/kernel.rbs#L729

I can revert and then steep:ignore this or shim kernel.rbs locally so it always returns a Symbol. What's your preference?

lib/prism/parse_result/comments.rb
```diff
@@ -171,7 +172,12 @@ def nearest_targets(node, comment)
 
     # Attach the list of comments to their respective locations in the tree.
     def attach_comments!
-      Comments.new(self).attach!
+      if ProgramNode === value
```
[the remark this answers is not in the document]
nixme: It's because `ParseResult` can hold non-node values like when using `Prism.lex`.

I've reverted this and used `steep:ignore` since it's so awkward.

lib/prism/parse_result.rb
nixme: See discussion with @eregon at https://github.com/ruby/prism/pull/2101#discussion_r1445829919

Because of the implementations of `Location.null` and `DSL#Location`, Location can have a nil source. We can switch back to saying Location always has a source and then ignoring type-check errors in the other places. What would you prefer?

nixme:
@kddnewton, @eregon I've fixed conflicts and updated the branch.

Notable changes: 
- Upgraded to Steep 1.7dev which adds support for `# steep:ignore` comments. Much easier to reduce the surface area of changes.
- I backed out a lot of RBS for files that I've ignored in Steepfile to cut down on this PR size. I'll reintroduce in future PRs when I work through that TODO list.

I've also responded to comment questions from before.

eregon (approved):
LGTM

lib/prism/pattern.rb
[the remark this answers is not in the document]
kddnewton: Ahh I see. Hash patterns can only have `SymbolNode` keys through their `AssocNode` nodes, but that's implicit. I think I'm fine with this line if you change it to:

```suggestion
          if key.is_a?(SymbolNode)
```

I'm assuming steep is smart enough to then not need the type annotation on the next line?

lib/prism/pattern.rb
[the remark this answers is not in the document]
kddnewton: Let's change this to:

```suggestion
      return to_enum(:scan, root) unless block_given?
```

that should be enough.

lib/prism/parse_result.rb
kddnewton: That's a great suggestion, let's ignore `Location.null` and `DSL#Location`, then type this as always being present.

kddnewton (changes requested):
Wow, great work. This is really, really nice. I have a couple (3) more very small changes, and then this is good to go.

lib/prism/parse_result.rb
eregon: I want to clarify the effects of marking as always non-nil:

Marking it as non-nil is unsound (as mentioned in the linked thread) and will mean NoMethodError instead of a clear(er) error when using Location methods that need a source.

For example here are two bugs on current master that correct typing would catch and improve:
```
$ bundle exec ruby -Ilib -rprism -e 'p Prism::Location.null'
lib/prism/parse_result.rb:207:in `start_line': undefined method `line' for nil:NilClass (NoMethodError)

      source.line(start_offset)
            ^^^^^
	from /home/eregon/code/yarp/lib/prism/parse_result.rb:170:in `inspect'
	from -e:1:in `p'
	from -e:1:in `<main>'

$ bundle exec ruby -Ilib -rprism -e 'p Prism::Location.null.start_character_offset'
lib/prism/parse_result.rb:181:in `start_character_offset': undefined method `character_offset' for nil:NilClass (NoMethodError)

      source.character_offset(start_offset)
            ^^^^^^^^^^^^^^^^^
	from -e:1:in `<main>'
```
compared to on this branch:
```
$ be ruby -Ilib -rprism -e 'p Prism::Location.null'                       
lib/prism/parse_result.rb:291:in `source!': Missing source (RuntimeError)
	from /home/eregon/code/yarp/lib/prism/parse_result.rb:203:in `start_line'
	from /home/eregon/code/yarp/lib/prism/parse_result.rb:166:in `inspect'
	from -e:1:in `p'
	from -e:1:in `<main>'
```

`Location.null` doesn't seem used anymore, so maybe that could be removed (OTOH that seems more efficient than many "empty" Location instances).
But `DSL#Location` is used.
One thought would be to use `nil` instead of a "location without source" but then that would make all location fields nil-able which seems way worse.
Another possibility would be to make `DSL#Location` use a "dummy/not-provided/empty" Source as default when not passed a Source. Maybe the best alternative?

Changing things more seems out of scope of this PR, so I think it's OK to merge with forcing the type as non-nil (or with the current state with `source!`, IMO that's better but @kddnewton decides).
It seems something worth solving properly though, as e.g. unsound types can be worse than no types in some aspects.

kddnewton: Maybe this is a documentation problem then. `Prism::Location.null` and `DSL::Location` are meant to provide a common interface, but not necessary to actually be used. I've used them both in testing, which is why they're in there in the first place. Honestly I've been thinking about dumping DSL for a while now anyway, and I know `Prism::Location.null` isn't actually used. Since these are lower-priority rarely used things, I don't want them to "infect" the types of other things.

eregon: Makes sense, let's solve this later then.

The DSL seems to mainly make things slightly shorter and an optional location:
```ruby
    def AliasMethodNode(new_name, old_name, keyword_loc, source = nil, location = Location())
      AliasMethodNode.new(source, new_name, old_name, keyword_loc, location)
    end
```
I think it's quite nice in `test/prism/errors_test.rb` though and would be useful for gems that transform the AST, e.g. to do their own custom desugaring or anything else.

I think the best and simplest solution there would be to use an empty source, like:
```ruby
    EMPTY_SOURCE = Source.new("")
    EMPTY_LOCATION = Location.new(EMPTY_SOURCE, 0, 0)

    def Location(source, start_offset, length)
      Location.new(source, start_offset, length)
    end

    def AliasMethodNode(new_name, old_name, keyword_loc, source = EMPTY_SOURCE, location = EMPTY_LOCATION)
      AliasMethodNode.new(source, new_name, old_name, keyword_loc, location)
    end
```

lib/prism/pattern.rb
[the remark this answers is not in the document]
nixme: Yep. Thanks for confirming. Changing to a `is_a?(SymbolNode)` and no longer need the type annotation as well.

lib/prism/pattern.rb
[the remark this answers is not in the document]
nixme: Addressed in 2d859af588500f4a708f6415feeab6c56b85bcf8

lib/prism/pattern.rb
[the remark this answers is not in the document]
nixme: Addressed in 2d859af588500f4a708f6415feeab6c56b85bcf8

lib/prism/parse_result.rb
nixme: I went with @kddnewton's preferred approach in 2d859af588500f4a708f6415feeab6c56b85bcf8, but I like @eregon's null object pattern approach and will throw up in a later PR to see what you guys think.

While I have you, how do you feel about committing templated codegen'ed output files into the repository? I'd add some sort of lint or test to ensure there's no drift.

It would ease browsing code on GithHub or comparing codegen'ed files between branches locally. But I understand if you don't prefer that.

kddnewton:
Great work @nixme! This is a really great contribution. Thank you!

PR #2113, opened 2023-12-27
Fix to check multiple block arguments for forwarding arg

### LINKED ISSUE #2111: Missing SyntaxError for `...` and explicit block argument
```ruby
def foo(&block)
  block.call
end

def bar(...)
  foo(...) { puts "custom block argument" }
end
```

This should lead to SyntaxError:

```
script.rb: script.rb:6: both block arg and actual block given (SyntaxError)
```

makenowjust:
Fix #2111


src/diagnostic.c
```diff
@@ -56,6 +56,7 @@ static const char* const diagnostic_messages[PM_DIAGNOSTIC_ID_LEN] = {
     [PM_ERR_ARGUMENT_AFTER_BLOCK]               = "unexpected argument after a block argument",
     [PM_ERR_ARGUMENT_AFTER_FORWARDING_ELLIPSES] = "unexpected argument after `...`",
     [PM_ERR_ARGUMENT_BARE_HASH]                 = "unexpected bare hash argument",
+    [PM_ERR_ARGUMENT_BLOCK_FORWARDING]          = "both a block argument and a forwarding argument; only one block is allowed",
     [PM_ERR_ARGUMENT_BLOCK_MULTI]               = "multiple block arguments; only one block is allowed",
```
andrykonchin, line 60: Does it make sense to use `PM_ERR_ARGUMENT_BLOCK_MULTI` message?

In all the other cases with two block arguments Prism returns `multiple block arguments; only one block is allowed`:

```ruby
foo(&a) {}

def bar(&)
  foo(&) { }
end


def bar()
  foo(&:to_s) { }
end
```

makenowjust: It makes sense of course. However, I believe the detailed error message helps users, so I introduce the new error message for this.

kddnewton (approved):

PR #2118, opened 2024-01-02
IndexTargetNode should always have ATTRIBUTE_WRITE

### LINKED ISSUE #2102: `IndexTargetNode` doesn't have `ATTRIBUTE_WRITE`
Not sure if it's expected behaviour, but it will be nice to have the flag set:

```
bin/parse -e 'a[b], = []'
@ ProgramNode (location: (1,0)-(1,10))
├── locals: []
└── statements:
    @ StatementsNode (location: (1,0)-(1,10))
    └── body: (length: 1)
        └── @ MultiWriteNode (location: (1,0)-(1,10))
            ├── lefts: (length: 1)
            │   └── @ IndexTargetNode (location: (1,0)-(1,4))
            │       ├── flags: ∅
            │       ├── receiver:
            │       │   @ CallNode (location: (1,0)-(1,1))
            │       │   ├── flags: variable_call
            │       │   ├── receiver: ∅
            │       │   ├── call_operator_loc: ∅
            │       │   ├── name: :a
            │       │   ├── message_loc: (1,0)-(1,1) = "a"
            │       │   ├── opening_loc: ∅
            │       │   ├── arguments: ∅
            │       │   ├── closing_loc: ∅
            │       │   └── block: ∅
            │       ├── opening_loc: (1,1)-(1,2) = "["
            │       ├── arguments:
            │       │   @ ArgumentsNode (location: (1,2)-(1,3))
            │       │   ├── flags: ∅
            │       │   └── arguments: (length: 1)
            │       │       └── @ CallNode (location: (1,2)-(1,3))
            │       │           ├── flags: variable_call
            │       │           ├── receiver: ∅
            │       │           ├── call_operator_loc: ∅
            │       │           ├── name: :b
            │       │           ├── message_loc: (1,2)-(1,3) = "b"
            │       │           ├── opening_loc: ∅
            │       │           ├── arguments: ∅
            │       │           ├── closing_loc: ∅
            │       │           └── block: ∅
            │       ├── closing_loc: (1,3)-(1,4) = "]"
            │       └── block: ∅
            ├── rest:
            │   @ ImplicitRestNode (location: (1,4)-(1,5))
            ├── rights: (length: 0)
            ├── lparen_loc: ∅
            ├── rparen_loc: ∅
            ├── operator_loc: (1,6)-(1,7) = "="
            └── value:
                @ ArrayNode (location: (1,8)-(1,10))
                ├── flags: ∅
                ├── elements: (length: 0)
                ├── opening_loc: (1,8)-(1,9) = "["
                └── closing_loc: (1,9)-(1,10) = "]"
```

kddnewton:
Fixes #2102

Because this is a user-facing change, we also need to deal with the fact that CRuby 3.3.0 was just released.

In order to support workflows that want to parse exactly as CRuby parses in a specific version, this PR introduces a new option to the options struct that is "version". This allows you to specify that you want "3.3.0" parsing.

I'm not sure if this is the correct solution. Another solution is to just fork and keep around the old branch for security patches. Or we could keep around a copy of the source files within this repository as another directory and only update when necessary. There are a lot of potential solutions here.

Because this change is so small and the check for it is so minimal, I've decided to go with this enum. If this ends up entirely cluttering the codebase with version checks, we'll come up with another solution. But for now this works, so we're going to go in this direction for a bit until we determine it's no longer working.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -2172,11 +2172,16 @@ pm_call_target_node_create(pm_parser_t *parser, pm_call_node_t *target) {
 static pm_index_target_node_t *
 pm_index_target_node_create(pm_parser_t *parser, pm_call_node_t *target) {
     pm_index_target_node_t *node = PM_ALLOC_NODE(parser, pm_index_target_node_t);
+    pm_node_flags_t flags = target->base.flags;
+
+    if (parser->version != PM_OPTIONS_VERSION_CRUBY_3_3_0) {
+        flags |= PM_CALL_NODE_FLAGS_ATTRIBUTE_WRITE;
+    }
 
     *node = (pm_index_target_node_t) {
         {
             .type = PM_INDEX_TARGET_NODE,
-            .flags = target->base.flags,
+            .flags = flags,
             .location = target->base.location
         },
         .receiver = target->receiver,
@@ -17319,6 +17324,9 @@ pm_parser_init(pm_parser_t *parser, const uint8_t *source, size_t size, const pm
         if (options->suppress_warnings) {
             parser->suppress_warnings = true;
         }
+
+        // version option
+        parser->version = options->version;
 
         // scopes option
         for (size_t scope_index = 0; scope_index < options->scopes_count; scope_index++) {
--- a/ext/prism/extension.c
+++ b/ext/prism/extension.c
@@ -22,6 +22,7 @@ ID rb_option_id_encoding;
 ID rb_option_id_line;
 ID rb_option_id_frozen_string_literal;
 ID rb_option_id_verbose;
+ID rb_option_id_version;
 ID rb_option_id_scopes;
 
 /******************************************************************************/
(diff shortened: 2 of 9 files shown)


kddnewton:
@enebo yeah great point. I just updated it

src/prism.c
```diff
@@ -2172,11 +2172,16 @@ pm_call_target_node_create(pm_parser_t *parser, pm_call_node_t *target) {
 static pm_index_target_node_t *
 pm_index_target_node_create(pm_parser_t *parser, pm_call_node_t *target) {
     pm_index_target_node_t *node = PM_ALLOC_NODE(parser, pm_index_target_node_t);
+    pm_node_flags_t flags = target->base.flags;
+
+    if (parser->version != PM_OPTIONS_VERSION_CRUBY_3_3_0) {
+        flags |= PM_CALL_NODE_FLAGS_ATTRIBUTE_WRITE;
+    }
```
eregon, line 2179: @andrykonchin and I are looking at which version we'd set.
We do want the flag to be set here.
But with the current that means we need to use PM_OPTIONS_VERSION_LATEST.

OTOH we do not want `it` handling from 3.4 (https://github.com/ruby/prism/pull/2124).
And using `PM_OPTIONS_VERSION_LATEST` would enable `it`.

So how to solve this?
Maybe we need `PM_OPTIONS_VERSION_CRUBY_3_3` (without _0) to mean 3.3 syntax with the latest fixes?
Alternatively, is it worth hiding this flag for Prism versions newer than the one shipped with 3.3.0? Maybe not?
It feels like this flag should not hurt anything, and if someone wants exactly the version of Prism as shipped in 3.3.0 then they can just specify that Prism release in their Gemfile.

eregon: @kddnewton WDYT?
Feel free to extract this to an issue/discussion if you prefer to discuss it somewhere else.

eregon: Maybe the fundamental question is whether this change would go in CRuby 3.3.1. I think it should as a bug fix.

PR #2120, opened 2024-01-02
Fix a comment about NumberedParametersNode

k0kubun:
I thought these arrows should point to `_1` or `_2` instead of the whole block.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -2126,7 +2126,7 @@ nodes:
       parameters within a block or lambda.
 
           -> { _1 + _2 }
-          ^^^^^^^^^^^^^^
+               ^^   ^^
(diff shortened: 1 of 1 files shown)


kddnewton:
I see why you got there, but there's some nuance here.

The node you're describing is the `NumberedParametersNode` (note the `s` after `Parameter`). Numbered parameters on their own are represented with a `LocalVariableReadNode`. The `NumberedParametersNode` location matches the location of the block/lambda that contains it because there isn't really a location for it, it's more like metadata.

k0kubun:
Ahh, my bad 🤦 Thanks for clarifying that!

Issue #2052, opened 2023-12-12
Support `it` default parameter


kddnewton:
Note that this is predicated on us figuring out how we want to support multiple versions.

PR #2124, opened 2024-01-02
Parse `it` default parameter

k0kubun:
close https://github.com/ruby/prism/issues/2052


src/prism.c
```diff
@@ -12825,6 +12833,14 @@ parse_variable_call(pm_parser_t *parser) {
                 return (pm_node_t *) pm_local_variable_read_node_create(parser, &parser->previous, 0);
             }
         }
+        else if (!parser->current_scope->closed && pm_token_is_it(parser->previous.start, parser->previous.end)) {
```
kddnewton, line 12836:
```suggestion
        } else if ((parser->version != PM_OPTIONS_VERSION_CRUBY_3_3_0) && !parser->current_scope->closed && pm_token_is_it(parser->previous.start, parser->previous.end)) {
```

We need to gate this with 3.3.0.

kddnewton (changes requested):
One small change and then great! Thanks!

kddnewton (approved):

kddnewton (dismissed):
I think this might be more complicated than this unfortunately.

If you get `it {}` I think it needs to change back into a method call. This is also the case for when it receivers arguments without parentheses.

k0kubun:
I tried some of such cases with local `Prism.parse`, and they seemed working. Though I guess I'm still missing something for passing lex-ruby/discourse/top-100, so I'll look into them.

src/prism.c
```diff
@@ -12824,6 +12855,8 @@ parse_variable_call(pm_parser_t *parser) {
                 pm_parser_local_add_token(parser, &parser->previous);
                 return (pm_node_t *) pm_local_variable_read_node_create(parser, &parser->previous, 0);
             }
+        } else if ((parser->version != PM_OPTIONS_VERSION_CRUBY_3_3_0) && !parser->current_scope->closed && pm_token_is_it(parser->previous.start, parser->previous.end)) {
```
eregon, line 12858: Maybe the test should be `parser->version >= PM_OPTIONS_VERSION_CRUBY_3_4_0`?
Because e.g. for 3.3.1 we wouldn't want to handle `it` specially.

k0kubun: For that, the enum would need to look like:

```c
typedef enum {
    PM_OPTIONS_VERSION_CRUBY_3_3_0 = 0,
    PM_OPTIONS_VERSION_CRUBY_3_4_0 = 1,
    PM_OPTIONS_VERSION_LATEST = 2
} pm_options_version_t;
```

However, even having the following enum on top of `main` (swapping 0 and 1 between 3_3_0 and LATEST) breaks a lot of stuff for some reason:

```c
typedef enum {
    PM_OPTIONS_VERSION_CRUBY_3_3_0 = 0,
    PM_OPTIONS_VERSION_LATEST = 1
} pm_options_version_t;
```

It doesn't seem like a trivial change, so I think it should be addressed in a separate PR.

eregon: Right.
Do you have a link to a failed job?
Maybe the issue is these constants are hardcoded in FFI too: https://github.com/ruby/prism/blob/829ac0ed3f449313584aae35db98fd7614eb9d63/lib/prism/ffi.rb#L318

It does seem better to avoid changing the values of the enum to avoid tricky missing recompilation issues, I was thinking `PM_OPTIONS_VERSION_CRUBY_3_3_0 = 330` would work nicely, and `PM_OPTIONS_VERSION_LATEST = 999`.

kddnewton (dismissed):
I think we should add `=> ^it` in this PR. Also we will need to document that the local variable read nodes for `it` will actually use `0it` for the constant name, since that's pretty weird.

src/prism.c
```diff
@@ -13549,8 +13607,13 @@ parse_pattern_primitive(pm_parser_t *parser, pm_diagnostic_id_t diag_id) {
                     parser_lex(parser);
                     pm_node_t *variable = (pm_node_t *) parse_variable(parser);
                     if (variable == NULL) {
-                        PM_PARSER_ERR_TOKEN_FORMAT(parser, parser->previous, PM_ERR_NO_LOCAL_VARIABLE, (int) (parser->previous.end - parser->previous.start), parser->previous.start);
-                        variable = (pm_node_t *) pm_local_variable_read_node_create(parser, &parser->previous, 0);
+                        if (pm_token_is_it(parser->previous.start, parser->previous.end)) {
```
kddnewton, line 13610: We need to gate this on version as well.

```suggestion
                        if (parser->version != PM_OPTIONS_VERSION_CRUBY_3_3_0 && pm_token_is_it(parser->previous.start, parser->previous.end)) {
```

kddnewton (dismissed):

eregon (commented):
Could you add a test ensuring that with version 3_3_0 `{ it }` is parsed as a method call?
That seems good to add and should help ensure the right checks are in place.

kddnewton (approved):
Great job @k0kubun!

PR #2131, opened 2024-01-03
Include field comments in Java source

eregon:
* Also use `field.each_comment_line` to simplify other usages of field comments.

### DIFF
--- a/templates/java/org/prism/Nodes.java.erb
+++ b/templates/java/org/prism/Nodes.java.erb
@@ -16,7 +16,6 @@ import java.util.Arrays;
 // @formatter:off
 public abstract class Nodes {
 
-    public static final byte[][] EMPTY_BYTE_ARRAY_ARRAY = {};
     public static final <%= string_type %>[] EMPTY_STRING_ARRAY = {};
 
     @Target(ElementType.FIELD)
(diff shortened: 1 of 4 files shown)


eregon:
Ready for review, it looks like this:
```java
    public static final class CallNode extends Node {
        public final short flags;
        /**
         * <pre>
         * The object that the method is being called on. This can be either
         * `nil` or a node representing any kind of expression that returns a
         * non-void value.
         *
         *     foo.bar
         *     ^^^
         *
         *     +foo
         *      ^^^
         *
         *     foo + bar
         *     ^^^
         * </pre>
         */
        @Nullable
        public final Node receiver;
```
FYI cc @enebo

templates/java/org/prism/Nodes.java.erb
```diff
@@ -4,6 +4,10 @@ package org.prism;
 import java.lang.Override;
 import java.lang.String;
 import java.lang.StringBuilder;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
```
kddnewton, line 10: I'm assuming all of these hit the minimum required Java for JRuby/TruffleRuby?

kddnewton (approved):
Looks good to me! Maybe @enebo should check first though

enebo:
@kddnewton @eregon looks good.

templates/java/org/prism/Nodes.java.erb
eregon: That's checked by the `build-java` CI job which uses Java 1.8 (specified in the Rakefile).

PR #2132, opened 2024-01-04
Document `ArrayNode`, `SourceFileNode` fields

sanjaii:


config.yml
```diff
@@ -496,12 +496,20 @@ nodes:
...
+        comment: |
+          Represents whether or not there is a splat operator in the array.
       - name: elements
         type: node[]
+        comment: |
+          Represent list of elements in the array.
       - name: opening_loc
         type: location?
+        comment: |
+          Represents optional source location for opening token.
```
kddnewton, line 508: So this comment is effectively tautological; `opening_loc, location?` conveys the same information as `Represents optional source location for opening token.` We want to answer some questions to make this a more descriptive comment:

* Why is it optional? When is it present and when is it not?
* What tokens can be in this slot? Is it always a `[` or are there other kinds?
* How does this location correspond with the closing location?

config.yml
```diff
@@ -496,12 +496,20 @@ nodes:
...
       - name: closing_loc
         type: location?
+        comment: |
+          Represents optional source location for closing token.
```
kddnewton, line 512: Same comment as above for this. We want to know more than just rephrasing the name and type.

config.yml
```diff
@@ -2494,6 +2502,8 @@ nodes:
     fields:
       - name: filepath
         type: string
+        comment: |
+          Represents the name of the file being parsed.
```
kddnewton, line 2506: We need more information than this.

* Is it the name or the path in the filesystem?
* What is the default value?
* How do you set this when you're calling `Prism.parse`?

config.yml
```diff
@@ -496,12 +496,20 @@ nodes:
       - name: flags
         type: flags
         kind: ArrayNodeFlags
+        comment: |
+          Represents whether or not there is a splat operator in the array.
       - name: elements
         type: node[]
+        comment: |
+          Represent list of elements in the array.
```
kddnewton, line 504: More information needed here:

* What kind of nodes can be present in this list?
* Is it every possible node or are there some restrictions?
* Will there always be at least one element or can this be empty?

kddnewton (changes requested):

config.yml
sanjaii: I don't really have much idea about the restrictions on the types of nodes that can be present in the list. Can you provide any source for reference?

config.yml
sanjaii:
```
require 'prism'

p __FILE__
p Prism.parse("__FILE__")
```
From this, what I understood is that the default value is `""`. If the current file being parsed, will the `filePath` value be the default value?

config.yml
kddnewton: I don't have one at the moment, unfortunately. That's why we want to do this documentation effort!

config.yml
kddnewton: There is documentation on `Prism.parse` that would be relevant here, namely passing the `filepath` keyword. Also `Prism.parse_file` will impact this.

kddnewton (changes requested):

sanjaii:
May I know why the builds are failing?

kddnewton:
If you click into "Details" it says `Non-ASCII character on line 1236 of /home/runner/work/prism/prism/include/prism/ast.h`. You've probably added a non-ascii character into the comments, which isn't allowed.

config.yml
sanjaii: I'm currently facing a challenge; I've manually tested a few scenarios where I had some uncertainties. Do you have any suggestions or ideas for testing which nodes can be contained within the elements? For instance, based on my understanding, a `WhenNode` cannot be directly placed in the array; it always needs to be nested under a `CaseNode`

PR #2135, opened 2024-01-07
Document `HashNode`, `AssocNode` and `AssocSplatNode` fields.

matthew-healy:
This PR adds documentation comments in `config.yml` for `HashNode`, `AssocNode` and `AssocSplatNode`.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1455,10 +1455,28 @@ nodes:
     fields:
       - name: opening_loc
         type: location
+        comment: |
+          The location of the opening brace.
+
+              { a => b }
+              ^
       - name: elements
         type: node[]
+        comment: |
+          The elements of the hash. These can be either `AssocNode`s or `AssocSplatNode`s.
+
+              { a: b }
+                ^^^^
+
+              { **foo }
+                ^^^^^
       - name: closing_loc
         type: location
+        comment: |
+          The location of the closing brace.
+
+              { a => b }
+                       ^
     comment: |
       Represents a hash literal.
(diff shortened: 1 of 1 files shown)


config.yml
```diff
@@ -543,10 +543,35 @@ nodes:
     fields:
       - name: key
         type: node
+        comment: |
+          The key of the association. This can be any node.
```
kddnewton, line 547: _Almost_ any node. It can't be a void statement, as in an `alias` statement. How about something like:

```suggestion
          The key of the association. This can be any node that represents a non-void expression.
```

config.yml
```diff
@@ -543,10 +543,35 @@ nodes:
...
+              { foo => bar }
+                ^^^
+
+              { def a; end => 1 }
+                ^^^^^^^^^^
       - name: value
         type: node?
+        comment: |
+          The value of the association, if present. This can be any node. Will
+          not be present if, e.g., this node is an element of a `HashPatternNode`.
```
kddnewton, line 561:
```suggestion
          The value of the association, if present. This can be any node that
          represents a non-void expression. It can be optionally omitted if this
          node is an element in a `HashPatternNode`.
```

kddnewton (changes requested):
Mostly really good, just a couple of notes of feedback

matthew-healy:
@kddnewton Thanks for the review! Have added your proposed changes.

kddnewton (approved):
Looks great

PR #2139, opened 2024-01-08
Document `*ReadNode` fields

matthew-healy:
This PR adds documentation comments in `config.yml` for:

- `InstanceVariableReadNode`
- `ClassVariableReadNode`
- `GlobalVariableReadNode`
- `BackReferencedReadNode`
- `NumberedReferenceReadNode`
- `LocalVariableReadNode`
- `ConstantReadNode`

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -2287,6 +2287,16 @@ nodes:
     fields:
       - name: number
         type: uint32
+        comment: |
+          The (1-indexed, from the left) number of the capture group. Numbered
+          references that would overflow a `uint32`  result in a `number` of
+          exactly `2**32 - 1`.
+
+              $1          # number `1`
+
+              $5432       # number `5432`
+
+              $4294967296 # number `4294967295`
     comment: |
       Represents reading a numbered reference to a capture in the previous match.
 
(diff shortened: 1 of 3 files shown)


config.yml
```diff
@@ -2013,8 +2070,33 @@ nodes:
...
+
+              foo = 1; foo # depth 0
+
+              bar = 2; tap { bar } # depth 1
+
+          The specific rules for calculating the depth may differ from
+          individual Ruby implementations, as they are not specified by the
+          language.
+
+          For more information, see [the Prism documentation](https://github.com/ruby/prism/blob/main/docs/local_variable_depth.md).
```
matthew-healy, line 2099: I'm not sure if this style of markdown link is supported by all of the codegen targets, but I think linking to the canonical documentation is more valuable than just including a cut-down facsimile. 

If there's a better way of doing this please let me know.

matthew-healy:
@kddnewton Thanks for the review! Have extracted the identifier lexing details into `lexing.md` as requested. Agree that it makes more sense to keep them in one place so they don't drift. :+1:

kddnewton (approved):
Nice work!

PR #2140, opened 2024-01-08
Document `ConstantPathNode` fields

matthew-healy:
This PR updates `config.yml` to add documentation for the fields of `ConstantPathNode`.


config.yml
```diff
@@ -1096,10 +1096,43 @@ nodes:
...
+
+              self::Test
+              ^^^^
+
+              a.b::C
+              ^^^
       - name: child
         type: node
+        comment: |
+          The right-hand node of the path. Always a `ConstantReadNode`.
```
matthew-healy, line 1116: The [Advent of Prism section](https://kddnewton.com/2023/12/03/advent-of-prism-part-3.html#constantpathnode) states that `child` is "almost always a constant read". 

Looking at the calls to `pm_constant_path_node_create` in `src/prism.c`, they're all passed the result of `pm_constant_read_node_create`, with the exception of [this one](https://github.com/ruby/prism/blob/a81b4ce9cbeda7118461beb78d10564c11c6319b/src/prism.c#L17020) which is passed a `MissingNode` & results in an error.

matthew-healy:
Not sure why the C documentation generation is failing - I can run `doxygen Doxyfile` fine locally. Is that check known to be flaky? :thinking:

kddnewton:
Nope, not flaky. Looks like there's a warning about `Foo` in the output of doxygen in the test run

matthew-healy:
The issue was that doxygen read the `::Foo` in the comment as a reference to part of the C codebase, which it then couldn't find. Have just removed that part of the comment. 

(I couldn't repro it locally initially because I'd forgotten to run `bundle exec rake templates` :sweat_smile:).

config.yml
```diff
@@ -1378,13 +1378,46 @@ nodes:
     fields:
       - name: parent
         type: node?
+        comment: |
+          The left-hand node of the path, if present. Can be `nil` or any kind
+          of a node returning a non-void value. Will be missing when constant
+          lookup is at the root of the module tree.
```
kddnewton, line 1384:
```suggestion
          The left-hand node of the path, if present. Can be `nil` or any kind of a node returning a non-void value. Will be`nil` when the constant lookup is at the root of the module tree.
```

kddnewton (changes requested):

kddnewton (approved):

config.yml
```diff
@@ -1378,13 +1378,44 @@ nodes:
     fields:
       - name: parent
         type: node?
+        comment: |
+          The left-hand node of the path, if present. Can be `nil` or any kind of a node returning a non-void value. Will be`nil` when the constant lookup is at the root of the module tree.
```
kddnewton, line 1382:
```suggestion
          The left-hand node of the path, if present. Can be `nil` or any kind of a node returning a non-void value. Will be `nil` when the constant lookup is at the root of the module tree.
```

kddnewton (changes requested):

PR #2162, opened 2024-01-10
Document `ClassVariableWriteNode` fields

matthew-healy:
This PR adds documentation comments to `config.yml` for the fields of `ClassVariableWriteNode`.

While doing this, I noticed that `operator_loc` had type `location?`, but I was unable to find any situation where the `=` operator wouldn't be present. The tests & the `parse:topgems` task still pass when using `PM_LOCATION_TOKEN_VALUE` instead of `PM_OPTIONAL_LOCATION_TOKEN_VALUE`, so I added that change in a separate commit, on the assumption that the optionality was in error. If that's not correct let me know & I'll drop that commit & amend the documentation with an example that lacks the `=` location.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1032,10 +1032,7 @@ nodes:
       - name: name
         type: constant
         comment: |
-          The name of the class variable, including the leading `@@`.
-
-          For more information on permitted class variable names, see
-          [the Prism documentation](https://github.com/ruby/prism/main/docs/lexing.md).
+          The name of the class variable, which is a `@@` followed by an [identifier](https://github.com/ruby/prism/blob/main/docs/parsing_rules.md#identifiers).
 
               @@abc = 123     # name `@@abc`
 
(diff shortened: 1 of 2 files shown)


config.yml
```diff
@@ -1031,12 +1031,40 @@ nodes:
     fields:
       - name: name
         type: constant
+        comment: |
+          The name of the class variable, including the leading `@@`.
+
+          For more information on permitted class variable names, see
+          [the Prism documentation](https://github.com/ruby/prism/main/docs/lexing.md).
```
kddnewton, line 1038:
```suggestion
          The name of the class variable, which is a `@@` followed by an [identifier](https://github.com/ruby/prism/blob/main/docs/parsing_rules.md#identifiers).
```

kddnewton (changes requested):

kddnewton (approved):

Issue #2169, opened 2024-01-11
Java generated code should generate an enum (or static final ints) for versions once versioning is settled


enebo:
I am manually writing 5 for 3.3 support in the request for serialized source but this ultimately should be passing a value that is referenced from code generated by prism.

There seems to still be some questions on how versioning will be handled and I will not need anything other than 3.3 support for at least a year so this is not all that important.  Just reminding us that we should add something eventually.

eregon:
Indeed.
Currently the versions constants are only part of the header at https://github.com/ruby/prism/blob/5f46d364ae814edd21be9bb2f7b766b7d44123ca/include/prism/options.h#L32-L38

and in lib/prism/ffi.rb.

> I am manually writing 5

5 doesn't seem to correspond to any of these constants, but maybe I'm missing something?

The version for serialization seems not yet documented in https://github.com/ruby/prism/blob/main/docs/serialization.md but it was added in https://github.com/ruby/prism/pull/2118/files#diff-3eb9e99cab36d75a322185ffbcfb68f5337241fd0ad897c79359f87d5f2d510d, I'll make a PR for that.

It seems JRuby wants the same as TruffleRuby for now, nice (https://github.com/ruby/prism/issues/2163).

enebo:
@eregon haha. I was not looking to closely I saw the 5 from the strncmp.  A case of moving too quickly to remove hs_err core dumps.

kddnewton:
I think this is done now, unless I'm misreading.

eregon:
Yes, added in https://github.com/ruby/prism/pull/2304

PR #2170, opened 2024-01-12
Document version in serialization.md

eregon:
From https://github.com/ruby/prism/issues/2169#issuecomment-1888863547

### DIFF
--- a/docs/serialization.md
+++ b/docs/serialization.md
@@ -175,6 +175,7 @@ The final argument to `pm_serialize_parse` is an optional string that controls t
 | ...     | the encoding bytes         |
 | `1`     | frozen string literal      |
 | `1`     | suppress warnings          |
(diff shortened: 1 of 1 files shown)


docs/serialization.md
```diff
@@ -175,6 +175,7 @@ The final argument to `pm_serialize_parse` is an optional string that controls t
 | ...     | the encoding bytes         |
 | `1`     | frozen string literal      |
 | `1`     | suppress warnings          |
+| `1`     | syntax version, see [pm_options_version_t](https://github.com/ruby/prism/blob/main/include/prism/options.h) for valid values |
```
eregon, line 178: "syntax version" is  my desired interpretation of it, feel free to suggest to something else.
Just "version" or "prism version" seems too confusing as it's not the same thing as the gem version.

PR #2172, opened 2024-01-14
Document AndNode and OrNode

nikhilbhatt:
Document And and OrNOde

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -475,10 +475,25 @@ nodes:
     fields:
       - name: left
         type: node
+        comment: |
+          Represents left side of expression.
+
+          left and right
+          ^^^^
       - name: right
         type: node
+        comment: |
+          Represents right side of expression.
+
+          left && right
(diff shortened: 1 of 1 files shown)


config.yml
```diff
@@ -475,10 +475,25 @@ nodes:
     fields:
       - name: left
         type: node
+        comment: |
+          Represents left side of expression.
```
kddnewton, line 479: We need a little more information here. We know it's the left side of the expression by virtue of the name and the type. We want to know more than just that. Specifically: what kind of nodes can be in this position?

This comment applies to `left`/`right` in all four cases in this PR

kddnewton (changes requested):

config.yml
nikhilbhatt: Updated

kddnewton (approved):
Nice, thanks!

PR #2174, opened 2024-01-14
Allow JavaScript visitor methods to accept arguments

camertron:
In the project I'm working on, I need to be able to pass arguments to the various visitor methods, eg:

```typescript
class MyVisitor extends Visitor {
  compile(ast: any) {
    this.visitProgramNode(ast, true);
  }

  visitProgramNode(node: ProgramNode, used: boolean) {
    // some code here
    this.visitAll(node.statements.body, used);
    // more code here
  }
}
```

This PR modifies the `Visitor` class and the various `Node` classes to forward the args they're passed.


eregon:
It sounds similar to the discussion in https://github.com/ruby/prism/pull/1740.
One disadvantage of adding this is it makes the case which does not need to pass arguments probably quite a bit slower.
And keeping the data as a field of the visitor is likely a lot faster too.

camertron:
Ok, that makes sense 👍 I hand-wrote a visitor before the official JavaScript one was available in Prism. I was hoping to not have to rewrite all my code 😅 But no problem, I'll close this.

Issue #2123, opened 2024-01-02
Document fields


kddnewton:
At the moment, our `config.yml` file has documentation for every node type in the syntax tree. However, there is little to no documentation for the individual fields on the nodes. This is a crucial piece of information for understanding how the nodes should be used and what they represent. Therefore, we would like to document each field on each node, and surface that documentation to the end user.

This issue serves as a meta-issue for that work. Each node is listed below, and should be checked off when all of its fields are completely documented.

In terms of style we are very much _not_ looking for "this is the receiver of the method call". Instead we are looking for "this can be `nil` or a node that represents a non-void expression". Generally we want to avoid tautological statements, and instead favor more useful comments. Examples are encouraged — where applicable — and should begin with 4 spaces. See https://github.com/ruby/prism/blob/c790abd3bd288f2dbcba985ba2b41b5e84d68619/config.yml#L717-L729 for an example.

- [x] AliasGlobalVariableNode
- [x] AliasMethodNode
- [x] AlternationPatternNode
- [x] AndNode
- [x] ArgumentsNode
- [x] ArrayNode
- [x] ArrayPatternNode
- [x] AssocNode
- [x] AssocSplatNode
- [x] BackReferenceReadNode
- [x] BeginNode
- [x] BlockArgumentNode
- [x] BlockLocalVariableNode
- [x] BlockNode
- [x] BlockParameterNode
- [x] BlockParametersNode
- [x] BreakNode
- [x] CallAndWriteNode
- [x] CallNode
- [x] CallOperatorWriteNode
- [x] CallOrWriteNode
- [x] CallTargetNode
- [x] CapturePatternNode
- [x] CaseMatchNode
- [x] CaseNode
- [ ] ClassNode
- [x] ClassVariableAndWriteNode
- [ ] ClassVariableOperatorWriteNode
- [ ] ClassVariableOrWriteNode
- [x] ClassVariableReadNode
- [ ] ClassVariableTargetNode
- [x] ClassVariableWriteNode
- [ ] ConstantAndWriteNode
- [ ] ConstantOperatorWriteNode
- [ ] ConstantOrWriteNode
- [ ] ConstantPathAndWriteNode
- [x] ConstantPathNode
- [ ] ConstantPathOperatorWriteNode
- [ ] ConstantPathOrWriteNode
- [ ] ConstantPathTargetNode
- [x] ConstantPathWriteNode
- [x] ConstantReadNode
- [ ] ConstantTargetNode
[...]

PR #2175, opened 2024-01-14
Document `IfNode` fields

dgollahon:
- Adds documentation for the fields of the `IfNode`. Also updates the overall description to mention ternary expressions.
- Part of #2123

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1589,26 +1589,87 @@ nodes:
     fields:
       - name: if_keyword_loc
         type: location?
+        comment: |
+          The location of the `if` keyword if present.
+
+              bar if foo
+                  ^^
+
+          The `if_keyword_loc` field will be `nil` when the `IfNode` represents a ternary expression.
       - name: predicate
         type: node
+        comment: |
+          The node for the condition the `IfNode` is testing.
+
+              if foo
+                 ^^^
+                bar
+              end
+
+              bar if foo
+                     ^^^
+
+              foo ? bar : baz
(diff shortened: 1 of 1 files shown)


dgollahon:
Open to feedback on level of detail or any stylistic changes. I just opened this one node to see if it was approximately what you were looking for but happy to open others as a follow-up.

I also wasn't exactly sure how to indicate multiple lines for something like: 

```ruby
              if foo
                bar
              elsif baz
              ^^^^^^^^^
                qux
                ^^^
              end
              ^^^
```

so let me know if you have an opinion on that. Are the contents of this file used to generate documentation somewhere that I can see? I wasn't sure how to preview my changes outside of the YAML.

config.yml
```diff
@@ -1554,26 +1554,84 @@ nodes:
...
+
+              bar if foo
+                     ^^^
+
+              foo ? bar : baz
+              ^^^
       - name: then_keyword_loc
         type: location?
+        comment: |
+          The location of the `then` keyword if present, `nil` otherwise.
```
kddnewton, line 1582: This can also be `?` in a ternary.

kddnewton (changes requested):
Great work, one small comment

config.yml
dgollahon: Updated, ready for another review @kddnewton. Sorry for the delay, I forgot I hadn't updated this.

kddnewton (approved):

PR #2180, opened 2024-01-15
Add ability to pass options to JavaScript's parsePrism function

camertron:
This PR makes it possible to pass a set of options to `parsePrism`. It replicates the behavior present in the Ruby `dump_options` method in ffi.rb by packing the options into a binary string and passing a pointer as the last argument to `pm_serialize_parse`.

### DIFF
--- a/lib/prism/ffi.rb
+++ b/lib/prism/ffi.rb
@@ -296,7 +296,7 @@ module Prism
         values << 0
       end
 
-      template << "L"
+      template << "l"
       values << options.fetch(:line, 1)
 
       template << "L"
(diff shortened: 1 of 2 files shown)


lib/prism/ffi.rb
```diff
@@ -296,7 +296,7 @@ def dump_options(options)
         values << 0
       end
 
-      template << "L"
+      template << "l"
```
camertron, line 299: The c code expects this value to be signed, so `"l"` is appropriate here.

kddnewton (approved):
Nice work!

PR #2185, opened 2024-01-16
Provide abstract methods in Prism::Node

kddnewton:
@st0012 I agree with your assessment here, I don't think we should be adding signatures for methods that don't exist. That feels pretty wrong.

I assume this was added because people wanted to work with multiple nodes and treat them the same, but sorbet's typing system makes interfaces very difficult without including a module. For that reason this PR introduces `Prism::AnyNode`, a type alias for all of the existing node types. Hopefully that should solve the root of the issue, because people can use that signature instead.

@vinistock sorry I didn't catch this in the initial code review when you first introduced the RBI. Can this solution work for your needs? I would really prefer to not provide signatures for methods that don't exist.

### DIFF
--- a/rbi/prism_static.rbi
+++ b/rbi/prism_static.rbi
@@ -32,9 +32,6 @@ class Prism::ParseWarning
 end
 
 class Prism::Node
-  sig { returns(T::Array[T.nilable(Prism::Node)]) }
-  def child_nodes; end
-
   sig { returns(Prism::Location) }
   def location; end
 
@@ -43,6 +40,21 @@ class Prism::Node
 
   sig { returns(String) }
   def to_dot; end
+
+  sig { params(visitor: Prism::Visitor).void }
+  def accept(visitor); end
+
+  sig { returns(T::Array[T.nilable(Prism::Node)]) }
+  def child_nodes; end
+
+  sig { returns(T::Array[Prism::Node]) }
+  def compact_child_nodes; end
+
+  sig { returns(T::Array[T.nilable(Prism::Node)]) }
+  def deconstruct; end
+
+  sig { returns(Symbol) }
+  def type; end
 end
 
 class Prism::Comment
--- a/templates/lib/prism/node.rb.erb
+++ b/templates/lib/prism/node.rb.erb
@@ -36,6 +36,44 @@ module Prism
     def to_dot
       DotVisitor.new.tap { |visitor| accept(visitor) }.to_dot
     end
+
+    # --------------------------------------------------------------------------
+    # :section: Node interface
+    # These methods are effectively abstract methods that must be implemented by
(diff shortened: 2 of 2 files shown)


vinistock:
I believe this would work, but it would require some updates on the Ruby LSP since everything that is currently typed `Node` needs to become `AnyNode`. Also, if I'm not mistaken, the team identified a performance bottleneck when using type aliases for unions - and this is a very large union.

Is there any reason why we couldn't do something like this?

```ruby
class Node
  def child_nodes
    raise NotImplementedError, "Classes inheriting from Node must implement this method"
  end
end
```

kddnewton:
Ahh I didn't realize it would be such a performance issue. Is there any way you could check out if it's bad? I think this is the "ideal" solution but if it's not possible due to performance then I think we can go with your suggestion.

I just really don't like these kind of abstract methods — they always feel like we're trying to work in a different language. But if necessary I can concede.

eregon:
IMO having `Node` as the common "super type/interface" is worth it for simplicity, because it just works vs requiring extra effort to type expressions/things as AnyNode (I'd imagine).
Whether methods are actually defined on Node or subclasses feels like implementation details (as an example maybe it's more efficient to `attr_reader :location` or so per subclass (maybe not)), but typing wise it's clearly better if declared on Node than on every subclass.

Another example is `accept`, this is defined per class (and needs to be) but typing-wise it should be declared on Node since every Node supports `accept` (BTW I'm not quite sure how type checking currently work since `accept` is not declared on Node but is [used as such](https://github.com/ruby/prism/blob/d18b8711ff6512940b7395fa166a2325ca465e6c/templates/lib/prism/visitor.rb.erb#L10)).
Or maybe Sorbet is smart enough that if every subclass defines `accept` then any Node, given that Node is abstract (how does it know?) must support `accept`?

eregon:
I also think of typing as similar to documentation.
It feels better if the common methods that every Node has are documented at https://ruby.github.io/prism/rb/Prism/Node.html and I think the simplest way for that is "abstract/NotImplementedError" methods defined in `Node`.

froydnj:
> Or maybe Sorbet is smart enough that if every subclass defines `accept` then any Node, given that Node is abstract (how does it know?) must support `accept`?

Sorbet is not quite that smart; it requires you to explicitly declare `accept` as `abstract`, see https://sorbet.org/docs/abstract .

eregon:
@froydnj Thanks. Given that `accept` doesn't seem declared on Node currently and `abstract` isn't used, does it mean Prism doesn't typecheck currently? (maybe that's simply not checked in CI?)
Ah probably that's the case, I thought https://github.com/ruby/prism/pull/2101 is merged already but not yet (although that's RBS not RBI).

vinistock:
Yeah, that should do the trick. Thanks.

PR #2187, opened 2024-01-16
Fix prism brace association for constant-like method/local calls

HParker:


kddnewton (approved):
Can you add a fixture for the local variable one as well? I think:

```ruby
foo = 1
foo {}
```

would do it

kddnewton (changes requested):
Whoops meant to mark as "request changes"

kddnewton (approved):

PR #2190, opened 2024-01-17
Update checkout action version

m-nakamura145:
This PR updates to latest checkout action version


kddnewton (approved):
Thanks!

If you feel like sending a PR for creating a dependabot update to do this, I would merge that as well! Something like this: https://github.com/ruby-syntax-tree/syntax_tree/blob/main/.github/dependabot.yml#L7-L10.

PR #2200, opened 2024-01-17
Remove flag gating for 3.3.0 bug fixes

### LINKED ISSUE #2163: No way to get bug fixes + set version to 3.3 (so exclude 3.4 syntax)
From https://github.com/ruby/prism/pull/2118#discussion_r1445987020

@andrykonchin and I are looking at which version we'd set for truffleruby 24.0 (the next feature release where we'll switch to Prism).
We do want/need the flag from https://github.com/ruby/prism/pull/2118 to be set (as well as other fixes/improvements).
But with the current code in https://github.com/ruby/prism/pull/2118 that means we'd need to use PM_OPTIONS_VERSION_LATEST.

OTOH we do not want `it` handling from 3.4 (https://github.com/ruby/prism/pull/2124) for truffleruby (if at all possible).
And using `PM_OPTIONS_VERSION_LATEST` would enable `it` (when that PR is merged).

So how to solve this?
Maybe we need `PM_OPTIONS_VERSION_CRUBY_3_3` (without _0) to mean 3.3 syntax with the latest fixes?
Alternatively, is it worth hiding the flag from https://github.com/ruby/prism/pull/2118 for Prism versions newer than the one shipped with 3.3.0? Maybe not?
It feels like this flag should not hurt anything, and if someone wants exactly the version of Prism as shipped in 3.3.0 then they can just specify that Prism release in their Gemfile.
In short adding PM_CALL_NODE_FLAGS_ATTRIBUTE_WRITE is not really a syntax change or different parsing or anything, it's just extra information or from a POV a bug fix for a missing flag.

@kddnewton WDYT?

_Originally posted by @eregon in https://github.com/ruby/prism/pull/2118#discussion_r1445987020_

kddnewton:
Fixes #2163

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -2203,14 +2203,10 @@ pm_index_target_node_create(pm_parser_t *parser, pm_call_node_t *target) {
     pm_index_target_node_t *node = PM_ALLOC_NODE(parser, pm_index_target_node_t);
     pm_node_flags_t flags = target->base.flags;
 
-    if (parser->version != PM_OPTIONS_VERSION_CRUBY_3_3_0) {
-        flags |= PM_CALL_NODE_FLAGS_ATTRIBUTE_WRITE;
-    }
-
     *node = (pm_index_target_node_t) {
         {
             .type = PM_INDEX_TARGET_NODE,
-            .flags = flags,
+            .flags = flags | PM_CALL_NODE_FLAGS_ATTRIBUTE_WRITE,
             .location = target->base.location
         },
         .receiver = target->receiver,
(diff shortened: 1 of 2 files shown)


src/prism.c
```diff
@@ -2203,14 +2203,10 @@ pm_index_target_node_create(pm_parser_t *parser, pm_call_node_t *target) {
     pm_index_target_node_t *node = PM_ALLOC_NODE(parser, pm_index_target_node_t);
     pm_node_flags_t flags = target->base.flags;
 
-    if (parser->version != PM_OPTIONS_VERSION_CRUBY_3_3_0) {
```
enebo, line 2206: I know this gating was removed but I wonder if we want inequality for this by default with specific equality when it is locked to something specific.

`parser->version >= PR_OPTIONS_VERSION_CRUBY_3_3`

It is pretty likely we will see ranges doing `<=` and `>=` if something appears then is removed later.  Just a drive-by comment.

kddnewton: That makes sense - I'll keep that in mind going forward.

PR #2202, opened 2024-01-17
Add Java class `Parser.ParsingOptions` to serialise parsing options into byte array

andrykonchin:
It's needed to implement `Kernel#eval` correctly and make local variables in outer scopes visible.

### DIFF
--- a/java/org/prism/Parser.java
+++ b/java/org/prism/Parser.java
@@ -6,6 +6,10 @@ public class Parser {
         System.load(path);
     }
 
-    public static native byte[] parseAndSerialize(byte[] source);
+    public static native byte[] parseAndSerialize(byte[] source, byte[] options);
+
+    public static byte[] parseAndSerialize(byte[] source) {
+        return parseAndSerialize(source, null);
+    }
 
 }
--- a/java/org/prism/Parser.java
+++ b/java/org/prism/Parser.java
@@ -1,6 +1,6 @@
 package org.prism;
 
-public class Parser {
+public abstract class Parser {
 
     public static void loadLibrary(String path) {
         System.load(path);
--- a/java/org/prism/ParsingOptions.java
+++ b/java/org/prism/ParsingOptions.java
@@ -5,7 +5,7 @@ import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 
 // @formatter:off
-public class ParsingOptions {
+public abstract class ParsingOptions {
     /** Serialize parsing options into byte array.
      *
      * @param filepath the name of the file that is currently being parsed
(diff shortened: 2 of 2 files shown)


kddnewton (changes requested):
This is great! I'm happy to have this merged in. The only change I would like to be made: all of the other places where we refer to parsing options, we have consistent naming, but this PR introduces a bunch of names that don't line up. Can you use the same name as the options used in other places?

They are:
* filepath
* encoding
* line
* frozen_string_literal (I'm fine with this being `frozenStringLiteral`)
* verbose
* version
* scopes

kddnewton:
@enebo are you okay with these changes? It's a breaking change for the Java API just want to make sure you're aware

kddnewton:
@andrykonchin not sure if you saw Tom's comment — could this be a function or do you need an object for this?

eregon:
Right, it could be a static method, and then the documentation for each field would instead go as `@param` in JavaDoc on the method.

`verbose` should be renamed to `suppress warnings` to align with https://github.com/ruby/prism/blob/main/docs/serialization.md#apis

andrykonchin:
@kddnewton I've addressed the comment. Could you please take a look?

java/org/prism/Parser.java
```diff
@@ -1,11 +1,96 @@
 package org.prism;
 
+import java.io.ByteArrayOutputStream;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+// @formatter:off
 public class Parser {
 
+    public static class ParsingOptions {
```
eregon, line 10: @andrykonchin Could you move this to a separate file?
It seems better because (I think) the Parser class around only class-loads successfully if there is the corresponding JNI method implemented.
If it's in another file that's no longer a concern.

andrykonchin: Done

eregon (approved):
LGTM

kddnewton:
Looks good to me, @enebo?

java/org/prism/ParsingOptions.java
```diff
@@ -0,0 +1,82 @@
+package org.prism;
+
+import java.io.ByteArrayOutputStream;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+// @formatter:off
+public class ParsingOptions {
```
eregon, line 8: There is a check in TruffleRuby that every class that can be marked as `final` is marked as such.
Could you make this class and Parser too `abstract` since they are not meant to be instantiated (at least currently)?
That will satisfy that check.

eregon: Correction, the check currently only looks at `src/main` and so not `src/yarp`.
But it seems good to mark as `abstract` for clarity anyway.

andrykonchin: Done

kddnewton (approved):
Looks good

PR #2206, opened 2024-01-18
Raise error from bin/parse for missing source code and file

eileencodes:
Previously, if you called `bin/parse` without `-e` or a file, you would get the following confusing error:

```
$ bin/parse
open: No such file or directory
bin/parse:20:in `<main>': undefined method `value' for nil:NilClass (NoMethodError)

value = result.value
              ^^^^^^
```

Now you get an `ArgumentError`:

```
$ bin/parse
open: No such file or directory
bin/parse:17:in `<main>': There was nothing to parse. You must pass source code with `-e` or a Ruby file. (ArgumentError)
```

### DIFF
--- a/bin/parse
+++ b/bin/parse
@@ -13,6 +13,8 @@ if ARGV[0] == "-e"
   result = Prism.parse(ARGV[1])
 else
   result = Prism.parse_file(ARGV[0] || "test.rb")
+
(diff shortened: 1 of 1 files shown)


kddnewton:
Ooo this is one of the things I have wanted to fix for a long time.

I'm down to merge this as is, because it fixes the problem, but we should make this raise going forward. I'll open a new issue.

kddnewton (approved):

PR #2211, opened 2024-01-18
Handle stovetop start in constant path full_name

HParker:
this prevents prism from raising `Constant path contains dynamic parts. Cannot compute full name` when a constant path starts with `::`.

### DIFF
--- a/lib/prism/node_ext.rb
+++ b/lib/prism/node_ext.rb
@@ -118,7 +118,7 @@ module Prism
         current = current.parent
       end
 
-      unless current.is_a?(ConstantReadNode)
+      unless current.is_a?(ConstantReadNode) || current == nil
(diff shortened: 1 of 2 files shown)


lib/prism/node_ext.rb
```diff
@@ -118,7 +118,7 @@ def full_name_parts
         current = current.parent
       end
 
-      unless current.is_a?(ConstantReadNode)
+      unless current.is_a?(ConstantReadNode) || current == nil
```
kddnewton, line 121:
```suggestion
      if !current.is_a?(ConstantReadNode) && !current.nil?
```

I never understand `unless` without significant headaches...

kddnewton (changes requested):

kddnewton (approved):

PR #2213, opened 2024-01-19
Unescape method name for unary methods.

### LINKED ISSUE #2226: Method defining !@
```ruby
class C; def !@; true; end; end
!C.new
```

eightbitraptor:
This PR removes the trailing `@` from the Method name in the prism AST for `DefNode`s that define unary methods.

CRuby's parser removes the `@` from the `mid` of `def !@; true; end` as seen here

```
#             |   @ NODE_DEFN (id: 3, line: 2, location: (2,0)-(4,3))*
#             |   +- nd_mid: :!
#             |   +- nd_defn:
```

I've opted to handle it inside Prism so that method names in the AST match what CRubys parser generates, rather than handling it in the compiler.

Fixes #2226

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -12719,20 +12719,24 @@ parse_string_part(pm_parser_t *parser) {
  * automatically drop trailing `@` characters. This happens at the parser level,
  * such that `~@` is parsed as `~` and `!@` is parsed as `!`. We do that here.
  */
+static const uint8_t *
+parse_operator_symbol_name(const pm_token_t *name) {
+    switch (name->type) {
+        case PM_TOKEN_TILDE:
+        case PM_TOKEN_BANG:
+            if (name->end[-1] == '@') return name->end - 1;
+        /* fallthrough */
+        default:
+            return name->end;
+    }
+}
+
 static pm_node_t *
 parse_operator_symbol(pm_parser_t *parser, const pm_token_t *opening, pm_lex_state_t next_state) {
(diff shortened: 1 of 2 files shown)


src/prism.c
```diff
@@ -15379,6 +15379,15 @@ parse_expression_prefix(pm_parser_t *parser, pm_binding_power_t binding_power, b
             pm_parser_scope_pop(parser);
             pm_parser_current_param_name_restore(parser, saved_param_name);
 
+            /**
+             * If the final character is @. As is the case when defining
+             * methods to override the unary operators, we should ignore
+             * the @ in the same way we do for symbols.
+             */
+            if (name.type == PM_TOKEN_BANG && name.end[-1] == '@') {
```
kddnewton, line 15387: It looks like we're missing `PM_TOKEN_TILDE` from this check.

Can you reuse the code in `parse_operator_symbol`? Maybe by adding another function since this one is specific to creating symbols.

Something like:

```c
static const uint8_t *
parse_def_node_name(const pm_token_t *name) {
    switch (name->type) {
        case PM_TOKEN_TILDE:
        case PM_TOKEN_BANG:
            if (name->end[-1] == '@') return name->end - 1;
        /* fallthrough */
        default:
            return name->end;
    }
}
```

(Untested)

kddnewton (changes requested):
Marking as "request changes"

kddnewton (approved):

PR #2214, opened 2024-01-19
Document `UnlessNode` fields

MatheusRich:
Adds documentation for the fields of `UnlessNode`.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -2757,18 +2757,56 @@ nodes:
     fields:
       - name: keyword_loc
         type: location
+        comment: |
+          The location of the `unless` keyword.
+
+              unless cond then bar end
+              ^^^^^^
+
+              bar unless cond
+                  ^^^^^^
       - name: predicate
         type: node
+        comment: |
+          The condition to be evaluated for the unless expression. Can be any
+          kind of node that represents a non-void expression.
+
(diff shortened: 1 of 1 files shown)


config.yml
```diff
@@ -2757,18 +2757,56 @@ nodes:
...
+          The location of the `then` keyword, if present.
+
+              unless cond then bar end
+                          ^^^^
       - name: statements
         type: node?
         kind: StatementsNode
+        comment: |
+          The body of the unless expression, if present. It's a list of non-void
+          expressions.
```
MatheusRich, line 2791: Open to suggestions for a better way to describe the expression body

config.yml
```diff
@@ -2757,18 +2757,56 @@ nodes:
...
+              unless cond then bar end
+              ^^^^^^
+
+              bar unless cond
+                  ^^^^^^
       - name: predicate
         type: node
+        comment: |
+          The condition to be evaluated for the unless expression. Can be any
+          kind of node that represents a non-void expression.
```
MatheusRich, line 2772: I couldn't find where/how we define "non-void expressions". Documenting that (if it's not present) would be useful too.

kddnewton (approved):

PR #2246, opened 2024-01-23
Add description RangeNode flags field

sanjaii:


kddnewton:
We're going to leave off comments on the flags field because we are probably going to change the way these are documented soon. Thanks for the contribution though!

PR #2261, opened 2024-01-23
Fix multiple assigns with newlines

haldun:
Fixes parsing code like:

```ruby
(
  a,
  b
) = c
```


src/prism.c
```diff
@@ -11140,6 +11140,8 @@ static pm_node_t *
 parse_targets_validate(pm_parser_t *parser, pm_node_t *first_target, pm_binding_power_t binding_power) {
     pm_node_t *result = parse_targets(parser, first_target, binding_power);
 
+    while (accept1(parser, PM_TOKEN_NEWLINE));
```
kddnewton, line 11143:
```suggestion
    accept1(parser, PM_TOKEN_NEWLINE);
```

I think we only need one since subsequent ones should be ignored

kddnewton (approved):

PR #2264, opened 2024-01-24
Use char_is_identifier_start() to check for valid method name and fix serialization

maxprokopiev:
I've noticed an interesting case where snippets like `foo.🌊 = 1` are not accepted and failing with:

```
[#<Prism::ParseError @message="unexpected write target" @location=#<Prism::Location @start_offset=10 @length=1 start_line=1>>]
```

The reason for that is because the current check is only for `_` and any alphanumeric char, so it makes sense to change that to use `char_is_identifier_start()` which will check for any valid identifier (while also making the condition cleaner ✨).

After fixing that I've noticed that test cases started to fail with:

```
...
prism/test/prism/test_helper.rb:60:in `assert_equal_nodes'
prism/test/prism/parse_test.rb:230:in `block (3 levels) in <class:ParseTest>'
prism/test/prism/parse_test.rb:225:in `each'
prism/test/prism/parse_test.rb:225:in `block (2 levels) in <class:ParseTest>'
<:🌊=>(UTF-8) expected but was
<:"\xF0\x9F\x8C\x8A=">(ASCII-8BIT)

diff:
? :🌊                =
?  "\xF0\x9F\x8C\x8A "
?  ? +

? Encoding: UTF  -8
?           ASCII  BIT
?           ???  +++
```

That seems to be an issue with not forcing encoding during the deserialization, which is also fixed in https://github.com/ruby/prism/commit/932324356961572ac8f3cd3b2743cbb0c81c5bf1


kddnewton (approved):
Yeah this looks good, thanks!

kddnewton:
It looks like there's a failure on truffleruby because of `Symbol#inspect`, where on TruffleRuby they're making it `:"🌊"` whereas on CRuby it's `:🌊`.

To fix this, can you move this fixture into its own file and then skip it on TruffleRuby in `parse_test.rb`? We have something like this already there because of this exact problem (which I forgot about...).

kddnewton:
And then we just need to regenerate the snapshots by rerunning the tests

maxprokopiev:
was running the tests with truffleruby locally 😅 pushed the snapshots in https://github.com/ruby/prism/pull/2264/commits/459a9f544ec4b04419bb85850ca31691dae3dcf4

kddnewton (approved):

eregon:
I filed https://github.com/oracle/truffleruby/issues/3407 to fix this in TruffleRuby

Issue #2082, opened 2023-12-14
The verbose option for parsing should have 3 levels, like $VERBOSE


eregon:
There are 3 states for `$VERBOSE` (true, false, nil) and some parser warnings warn at `false` while some only warn at `true` (and of course none warn at `nil`):
```
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-linux]
$ ruby -e 'def foo; END {}; end' 
-e:1: warning: END in method; use at_exit
$ ruby -e '(1; 2)'            
$ ruby -w -e '(1; 2)'
-e:1: warning: unused literal ignored
-e:1: warning: possibly useless use of a literal in void context
```

But the current API only has true/false, and nil means default and not a separate value (AFAIK):
https://ruby.github.io/prism/rb/Prism.html#method-c-parse
And in fact `pm_options_suppress_warnings_set` takes a boolean.

Also the serialization has a reversed option called suppress warnings:
https://github.com/ruby/prism/blob/main/docs/serialization.md#apis
And that caused quite a bit of confusion (e.g. #2080 but not only), so a good occasion to change that to a 3-states verbose.

_Originally posted by @eregon in https://github.com/ruby/prism/issues/2080#issuecomment-1856348153_

eregon:
A typical encoding of $VERBOSE (nil, false, true) values is (0, 1, 2), which also reflects the ruby `-W` flag:
```
     -W[level=2]    Turns on verbose mode at the specified level without printing the version message at the beginning.
                    The level can be;

                          0       Verbose mode is "silence". It sets the $VERBOSE to nil.

                          1       Verbose mode is "medium". It sets the $VERBOSE to false.

                          2 (default) Verbose mode is "verbose". It sets the $VERBOSE to true.  -W2 is the same as -w
```

kddnewton:
Yeah, makes sense

eregon:
One more very-related concern about this, that I forgot when filing this issue but remembered now:
I think we need to have a boolean on `‎Prism::ParseWarning`, for whether this is a $VERBOSE==true warning or $VERBOSE==false warning.

The reason is when caching the serialized output of a file, it would be best to reuse the serialized output if nothing changes but the `$VERBOSE` level (which is a runtime value).

[...]

PR #2265, opened 2024-01-24
Add level to warnings and errors to categorize them

eregon:
* Fixes https://github.com/ruby/prism/issues/2082


bin/parse
```diff
@@ -36,6 +37,6 @@ else
   parts.each_with_index do |(key, value), index|
     puts if index > 0
     puts "#{key}:"
-    puts value.inspect
+    pp value
```
eregon, line 40: This prints the warnings much more nicely, one per line.

OTOH it shows DATA as `(6,0)-(11,0)` instead of `#<Prism::Location @start_offset=90 @length=215 start_line=6>`.
That seems an OK trade-off to keep things simple.

kddnewton (changes requested):
This generally looks good. My only feedback is could you change this to be an integer? I'd like to reuse this field for error level as well (https://bugs.ruby-lang.org/issues/20024).

eregon:
> This generally looks good. My only feedback is could you change this to be an integer? I'd like to reuse this field for error level as well.

That makes a lot of sense, and then also makes it clear this new field is useful for both warnings and errors alike.

eregon:
Done.
I also modified `templates/javascript/src/deserialize.js.erb`.
That doesn't seem tested in CI BTW because it was not failing on the commit before even though I forgot to update it.
(OTOH Loader.java was failing CI before being edited as deserialization failed without the changes)

test/prism/errors_test.rb
```diff
@@ -1871,6 +1871,16 @@ def test_upcase_end_in_def
       ]
     end
 
+    def test_warnings_verbosity
+      warning = Prism.parse("def foo; END { }; end").warnings[0]
+      assert_equal "END in method; use at_exit", warning.message
+      assert_equal 1, warning.level
```
eregon, line 1877: Do we want fancier values here for the level or integers are fine?
If yes, which Symbol should they be for both warnings levels and the error level?
(in Java they could be an `enum`, basically a copy of the C enum)

eregon: I switched to Symbols, it seems much better compatibility-wise.

eregon:
Ah one more thing to do related to this is to remove the suppress warnings field in serialization.
What do you want to do for the verbose option to `Prism.parse`? It's no longer necessary but maybe we should keep it for compatibility?

kddnewton (approved):

PR #2268, opened 2024-01-24
Track whether a Symbol should have its encoding changed from the source encoding.

### LINKED ISSUE #1998: Symbols need encoding flags
Symbols can change encoding based on their internal bytes. Also, sometimes they will fall back to US-ASCII. For example:

```ruby
:foo.encoding
# => #<Encoding:US-ASCII>
```

```ruby
:"\xC2\x80".encoding
# => #<Encoding:UTF-8>
```

```ruby
# encoding: ascii
:"\xC2\x80".encoding
# => #<Encoding:ASCII-8BIT>
```

Note that this also includes symbols in `%I` lists (but not `%i`).

nirvdrum:
Ruby sets a `Symbol` literal's encoding to US-ASCII if the symbols consists only of US ASCII code points. Character escapes can also lead a `Symbol` to have a different encoding than its source's encoding. We need to track this for proper Ruby semantics.

Determining if a `Symbol` is ASCII-only requires performing a linear scan through its associated string. Ideally, the attached string would have an indication of whether it is ASCII-only, but we don't have that in place today. I think having the code range calculated by the lexer would be beneficial in general, but I'll leave that for future work. We currently determine a string's bounds in `pm_strpbrk`, so that strikes me as the most natural place to track if a string is ASCII-only. However, that would mean visiting each byte ourselves in Prism rather than delegating work to `strchr`. I don't know if supported platforms instrinsify `strchr` and didn't want to inadvertently harm performance.

In speaking with @kddnewton, we decided to perform an extra string walk for the time being. To keep the cost down, the scan only happens with strings associated with a `Symbol` (see: `parse_symbol_encoding`). The `SymbolNode` will track if it's ASCII-only via its flags.

Unfortunately, this approach means we need to hook in at various points because there are several different ways in which we create `SymbolNode` instances. In addition to `Symbol` literals, they come up in keyword arguments, `case` statements, and there's special handling of `Symbol` literals of operators. In some cases we eagerly populate the string. Other times the string data is filled in after the node is created. I think I've hit all of the cases we need to and I've added some tests to _test/prism/encoding_test.rb_ to verify.

Fixes #1998.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -5468,6 +5468,53 @@ pm_super_node_create(pm_parser_t *parser, const pm_token_t *keyword, pm_argument
 }
 
 /**
+ * Read through the contents of a string and check if it consists solely of US ASCII code points.
+ */
+static bool
+ascii_only_p( const pm_string_t *contents) {
+    const size_t length = contents->length;
+
+    for (size_t i = 0; i < length; i++) {
+        if (contents->source[i] & 0x80) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+/**
+ * Ruby "downgrades" the encoding of Symbols to US-ASCII if the associated encoding is ASCII-compatible and
+ * the Symbol consists only of US-ASCII code points. Otherwise, the encoding may be explicitly set with an
+ * escape sequence.
+ */
+static inline pm_node_flags_t
+parse_symbol_encoding(const pm_parser_t *parser, const pm_string_t *contents) {
+    // Don't set any flags on the Symbol if it hasn't been populated yet.
+    if (contents->source == NULL) {
+        return 0;
+    }
+
+    // Ruby stipulates that all source files must use an ASCII-compatible encoding. Thus, all symbols appearing in
+    // source are eligible for "downgrading" to US-ASCII.
+    if (ascii_only_p(contents)) {
+        return PM_SYMBOL_FLAGS_FORCED_US_ASCII_ENCODING;
+    } else {
+        // A Symbol may optionally have its encoding explicitly set.
+        //
+        // NB: an explicitly set encoding is ignored by Ruby if the Symbol consists of only US ASCII code points.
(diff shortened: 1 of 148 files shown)


nirvdrum:
@enebo I think data validation is orthogonal to what I was looking to do here. This was mostly about making sure encodings for `Symbol` gets set to what CRuby would. I think what you're looking for would be best solved by having the code range information on-hand, but obviously we don't have that yet.

For completeness, this is the error condition Tom is talking about:

```ruby
:"\xFF\xFF" #=> invalid symbol in encoding UTF-8 :"\xFF\xFF" (EncodingError)
```

kddnewton:
Yeah I'm fine with having that come later, there's a ticket open for it already.

kddnewton:
(Just doing a final review, this looks good)

enebo:
"I think what you're looking for would be best solved by having the code range information on-hand"

@nirvdrum indeed!  This is so obvious I am wondering why I didn't do that for my workaround.  Once this PR lands I will still need this check so until CR is handy so I will calc CR look for a BROKEN CR.

src/prism.c
```diff
@@ -5467,6 +5467,53 @@ pm_super_node_create(pm_parser_t *parser, const pm_token_t *keyword, pm_argument
...
+
+    return true;
+}
+
+/**
+ * Ruby "downgrades" the encoding of Symbols to US-ASCII if the associated encoding is ASCII-compatible and
+ * the Symbol consists only of US-ASCII code points. Otherwise, the encoding may be explicitly set with an
+ * escape sequence.
+ */
+static inline pm_node_flags_t
```
nirvdrum, line 5491: I mostly followed what we were doing with `parse_unescaped_encoding`. I don't know if `inline` is gaining us much here. I did not mark `ascii_only_p` as inline.

kddnewton (approved):

PR #2276, opened 2024-01-25
Mark forwarding parameters in the lexical scope

tenderlove:
@kddnewton I need some help to push this over the line 🙇


kddnewton (approved):
Waiting for this to be merged into CRuby, then this is good.

test/prism/errors_test.rb
```diff
@@ -781,8 +781,8 @@ def test_method_parameters_after_arguments_forwarding
           nil
         ),
         nil,
-        [:"...", :a],
```
eregon, line 784: If we remove these variables, then it might be good to have the new forwarding flags on the various "scope" nodes (DefNode, etc).
Although I'm unsure if needed or not, @andrykonchin probably you know better.
I guess it's not needed because we'd still see these special names e.g. under ParametersNode.
And it probably makes sense to not have them in the local tables since Prism local tables seems to be basically "user-visible variables" (e.g. through Binding#local_variables).

(in any case, not something that needs to be solved in this PR)

eregon:
FWIW I *think* this change is causing this failure:
https://github.com/ruby/prism/actions/runs/7718734483/job/21040616236
(given that the build in https://github.com/oracle/truffleruby/actions/runs/7717093994/job/21035403709 is fine)

I guess it's possible to workaround it (I haven't investigated yet), but just wanted to note it is kind of a breaking change (in case that part was not intended).

Before:
```
$ bin/parse -e 'def m(...); n(...); end'|grep locals
├── locals: []
            ├── locals: [:"..."]
            ├── locals_body_index: 1
```
After:
```
$ bin/parse -e 'def m(...); n(...); end'|grep locals
├── locals: []
            ├── locals: []
            ├── locals_body_index: 0
```

PR #2283, opened 2024-01-26
Error follow-up

kddnewton:
Split up the diagnostic levels so that error and warning levels aren't mixed. Also fix up deconstruct_keys implementation.


java/org/prism/ParseResult.java
```diff
@@ -13,35 +13,40 @@ public MagicComment(Nodes.Location keyLocation, Nodes.Location valueLocation) {
         }
     }
 
-    public enum DiagnosticLevel {
+    public enum ErrorLevel {
         /** The default level for errors. */
```
eregon, line 17:
```suggestion
        /** For errors that cannot be recovered from. */
```

lib/prism/parse_result.rb
```diff
@@ -353,7 +353,7 @@ def initialize(message, location, level)
 
     # Implement the hash pattern matching interface for ParseWarning.
     def deconstruct_keys(keys)
-      { message: message, location: location, verbose_only: verbose_only }
```
eregon, line 356: Oops, I missed that one, thank you for the fix.

src/diagnostic.c
```diff
@@ -66,227 +66,228 @@ struct diagnostic_data {
  * PM_WARNING_VERBOSE_NOT_NIL if they appear for `ruby -c -e 'code'`,
  * or PM_WARNING_VERBOSE_TRUE if they appear only with -w: `ruby -w -c -e 'code'`.
```
eregon, line 67:
```suggestion
 * PM_WARNING_LEVEL_DEFAULT if they appear for `ruby -c -e 'code'`,
 * or PM_WARNING_LEVEL_VERBOSE if they appear only with -w: `ruby -w -c -e 'code'`.
```

eregon: Also there are a few mentions of `pm_diagnostic_level_t` left, but that doesn't exist anymore.

eregon (approved):

PR #2284, opened 2024-01-27
Tiny modification to find_line method.

khezam:
A slight modification to the method named ```find_line(byte_offset)```

### DIFF
--- a/lib/prism/parse_result.rb
+++ b/lib/prism/parse_result.rb
@@ -65,10 +65,11 @@ module Prism
       right = offsets.length - 1
 
       while left <= right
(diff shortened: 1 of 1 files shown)


kddnewton:
Thanks for the contribution! What's the reason for the change - I'm assuming performance? Do you have a benchmark result you could share?

khezam:
Hi! After conducting a benchmark performance comparison, I initially thought my changes had a slight edge. However, upon reevaluating the benchmarks, it appears that this is not the case.

I believe this PR is not beneficial, so I will go ahead and close it.

Sorry for any confusion.

PR #2303, opened 2024-01-30
Check literals for receiver in `def`

### LINKED ISSUE #1921: Ruby rejects `def (1).f; end` but Prism accepts it.
Ruby and Prism parses `def 1.f; end` to be SyntaxError.
For `def (1).f; end`, Ruby and Prism is different.
```ruby
def (1).f; end
# => can't define singleton method for literals (SyntaxError)

Prism.parse('def (1).f; end').success?
# => true
```
Same for `def ((1; 1)).foo; end`.

Ruby 3.3dev also rejects these. (Ruby 3.2 accepts it. https://bugs.ruby-lang.org/issues/19877)
```ruby
def ((a; 1)).foo; end
def ((return; 1)).foo; end
```

haldun:
Closes #1921


src/prism.c
```diff
@@ -2771,6 +2807,10 @@ pm_def_node_create(
         end = end_keyword->end;
     }
 
+    if ((receiver != NULL) && (receiver->type == PM_PARENTHESES_NODE)) {
```
kddnewton, line 2810:
```suggestion
    if ((receiver != NULL) && PM_NODE_TYPE_P(receiver, PM_PARENTHESES_NODE))) {
```

src/prism.c
```diff
@@ -2743,6 +2743,42 @@ pm_constant_write_node_create(pm_parser_t *parser, pm_constant_read_node_t *targ
     return node;
 }
 
+static void
```
kddnewton, line 2746: Can we document this function?

src/diagnostic.c
```diff
@@ -260,6 +260,7 @@ static const pm_diagnostic_data_t diagnostic_messages[PM_DIAGNOSTIC_ID_LEN] = {
     [PM_ERR_RESCUE_TERM]                        = { "expected a closing delimiter for the `rescue` clause", PM_ERROR_LEVEL_FATAL },
     [PM_ERR_RESCUE_VARIABLE]                    = { "expected an exception variable after `=>` in a rescue statement", PM_ERROR_LEVEL_FATAL },
     [PM_ERR_RETURN_INVALID]                     = { "invalid `return` in a class or module body", PM_ERROR_LEVEL_FATAL },
+    [PM_ERR_SINGLETON_FOR_LITERALS]             = { "can't define singleton method for literals", PM_ERROR_LEVEL_FATAL },
```
kddnewton, line 263:
```suggestion
    [PM_ERR_SINGLETON_FOR_LITERALS]             = { "cannot define singleton method for literals", PM_ERROR_LEVEL_FATAL },
```

I know you're just matching the error message from CRuby, but let's try to go with `cannot` to match our own style. If someone complains we can change it back.

kddnewton (changes requested):

kddnewton (approved):
Great work!

PR #2317, opened 2024-01-31
Optimize also java

andreaTP:
Thanks to the work done by @mariofusco on perfecting the Control Flow we are now able to run the optimized version of `prism.wasm` in Chicory.


enebo (approved):
@kddnewton This will only affect prism.wasm for Java.  It also simplifies config.

kddnewton (approved):

Issue #2309, opened 2024-01-30
`test/ruby/test_pattern_matching.rb` - Unexpected node type in pattern matching expression


kddnewton:

etiennebarrie:
I've reduced it to:
```ruby
case nil
in 0,;
  true
end
```

etiennebarrie:
Or even
```ruby
foo in 0,
```

etiennebarrie:
I think there is a problem in the parser, not just in the compiler.
```ruby
foo in 0,;
```
can be parsed but not
```ruby
foo in 0,
```

PR #2330, opened 2024-02-01
Fix Array pattern with command at end-of-file

etiennebarrie:
This fixes an issue found in https://github.com/ruby/prism/issues/2309#issuecomment-1921624040

~~I can't just add the fixture in test/prism/fixtures/patterns-eof.txt because it requires an EOF to show up. Is it OK to start adding new fixture files?~~

~~I've run the rake task with an added semicolon which makes parsing work, then manually edited the snapshot to remove the last character from the various locations.~~

For `test_individual_snippets_` tests, the fixtures are split on two consecutive newlines and the newline is removed to create the snippets, so we can add a snippet for the issue.

https://github.com/etiennebarrie/prism/blob/3f6d57fd98f9a28a38c30bb43ba98efe56766a72/test/prism/parse_test.rb#L225-L226

But `test_filepath_patterns.txt` parses the whole file at once, so I had to add a separate snippet each time to end the previous pattern in the context of a file that continues.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -10717,14 +10717,6 @@ match4(const pm_parser_t *parser, pm_token_type_t type1, pm_token_type_t type2, 
 }
 
 /**
- * Returns true if the current token is any of the five given types.
- */
-static inline bool
-match5(const pm_parser_t *parser, pm_token_type_t type1, pm_token_type_t type2, pm_token_type_t type3, pm_token_type_t type4, pm_token_type_t type5) {
-    return match1(parser, type1) || match1(parser, type2) || match1(parser, type3) || match1(parser, type4) || match1(parser, type5);
-}
-
-/**
  * Returns true if the current token is any of the six given types.
  */
 static inline bool
@@ -11360,7 +11352,7 @@ parse_statements(pm_parser_t *parser, pm_context_t context) {
             break;
         }
 
-        // If we have a terminator, then we will parse all consequtive terminators
+        // If we have a terminator, then we will parse all consecutive terminators
         // and then continue parsing the statements list.
         if (accept2(parser, PM_TOKEN_NEWLINE, PM_TOKEN_SEMICOLON)) {
             // If we have a terminator, then we will continue parsing the statements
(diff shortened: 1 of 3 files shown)


haldun:
IMHO the fix can be 

```diff
diff --git a/src/prism.c b/src/prism.c
index 43aafeb8..017f694c 100644
--- a/src/prism.c
+++ b/src/prism.c
@@ -10716,14 +10716,6 @@ match4(const pm_parser_t *parser, pm_token_type_t type1, pm_token_type_t type2,
     return match1(parser, type1) || match1(parser, type2) || match1(parser, type3) || match1(parser, type4);
 }
 
-/**
- * Returns true if the current token is any of the five given types.
- */
-static inline bool
-match5(const pm_parser_t *parser, pm_token_type_t type1, pm_token_type_t type2, pm_token_type_t type3, pm_token_type_t type4, pm_token_type_t type5) {
-    return match1(parser, type1) || match1(parser, type2) || match1(parser, type3) || match1(parser, type4) || match1(parser, type5);
-}
-
 /**
  * Returns true if the current token is any of the six given types.
  */
@@ -14011,7 +14003,7 @@ parse_pattern(pm_parser_t *parser, bool top_pattern, pm_diagnostic_id_t diag_id)
         // Gather up all of the patterns into the list.
         while (accept1(parser, PM_TOKEN_COMMA)) {
             // Break early here in case we have a trailing comma.
-            if (match5(parser, PM_TOKEN_KEYWORD_THEN, PM_TOKEN_BRACE_RIGHT, PM_TOKEN_BRACKET_RIGHT, PM_TOKEN_NEWLINE, PM_TOKEN_SEMICOLON)) {
+            if (match6(parser, PM_TOKEN_KEYWORD_THEN, PM_TOKEN_BRACE_RIGHT, PM_TOKEN_BRACKET_RIGHT, PM_TOKEN_NEWLINE, PM_TOKEN_SEMICOLON, PM_TOKEN_EOF)) {
                 node = (pm_node_t *) pm_implicit_rest_node_create(parser, &parser->previous);
                 pm_node_list_append(&nodes, node);
                 break;
```

kddnewton:
@haldun I think that's probably correct. @etiennebarrie could you see if that fixes the problem?

kddnewton:
Whoops clicked the wrong button

kddnewton (approved):

PR #2359, opened 2024-02-03
add benchmark script

exterm:
Similar to what @kddnewton did in  https://github.com/kddnewton/parser-prism/blob/325959896668cf24e57d668c48689156be330950/bin/bench.

I needed an easy way to compare performance of the different parser options. I think it would make sense for it to live in this repo for now.

These are the results from a run on my local machine:

```
 Parser::CurrentRuby     1.000 i/100ms
       Parser::Prism    25.000 i/100ms
Prism::Translation::Parser
                         5.000 i/100ms
 Parser::CurrentRuby     15.994 (± 0.0%) i/s -     80.000 in   5.004759s
       Parser::Prism    235.991 (± 2.5%) i/s -      1.200k in   5.087801s
Prism::Translation::Parser
                         56.258 (± 1.8%) i/s -    285.000 in   5.068509s

Comparison:
       Parser::Prism:      236.0 i/s
Prism::Translation::Parser:       56.3 i/s - 4.19x  slower
 Parser::CurrentRuby:       16.0 i/s - 14.76x  slower
 ```

### CODE REVIEW

**Review by kddnewton — CHANGES_REQUESTED** [review]
Just marking as "request changes"

**Comment by exterm** [comment]
Done and I also added `ruby_parser` and its translator to the mix.

**Review by kddnewton — APPROVED** [review]
Thanks!

File: bin/prism [hunk 1]
```diff
@@ -246,6 +248,24 @@ module Prism
...
+    def benchmark(argv)
+      require "benchmark/ips"
+      require "parser/current"
+      require "ruby_parser"
+
+      filepath = argv.fetch(0) { File.expand_path("../lib/prism/translation/parser/compiler.rb", __dir__) }
+
+      Benchmark.ips do |x|
+        x.report("Parser::CurrentRuby") { Parser::CurrentRuby.parse_file(filepath) }
+        x.report("Parser::Prism") { Prism.parse_file(filepath) }
```
eregon, line 261: This should be labeled just `Prism` as there is no `Parser::Prism` (it confused me for the results in the PR description).
I was about to make a PR but @kddnewton already fixed it in 40059d3400dc5f6c7ef0360993276637dfe70903 :)

exterm: Ah dang, that's copy-pasta. Thanks for the note, Benoit

PR #2364, opened 2024-02-05
Support for Ruby 2.7

kddnewton:
Importantly this does _not_ support the syntax version of Ruby 2.7, it merely allows prism to run with Ruby 2.7. The syntax version being parsed will still be 3.3 or 3.4, depending on the given options.


exterm:
❤️

Issue #2207, opened 2024-01-18
Prism::parse_file* APIs should raise appropriate errors


kddnewton:
Right now, when you call any of these APIs:

* `Prism::dump_file`
* `Prism::lex_file`
* `Prism::parse_file`
* `Prism::parse_file_comments`
* `Prism::parse_lex_file`
* `Prism::parse_file_success?`

It will do a few things:

* Validate that the filepath is a string (or raise a `TypeError`)
* Try to open the file (or fail, call `perror` to print out the error for `errno`, and return `nil`)
* Try to stat the file (or fail, call `perror` to print out the error for `errno`, and return `nil`)
* Try to `mmap` the file (or fail, call `perror` and print out the error for `errno`, and return `nil`)
* Map the file, parse it, and return a parse result

I think we should change this behavior to raise an appropriate error. Fortunately, `errno` is set for us by `open`/`fstat`/`mmap`, so that is already in place. I think we should be able to call `rb_syserr_fail` with `errno` and the filepath and it should take care of everything for us. At that point `file_options` can return `void` because it will only ever exit through a Ruby error.

eregon:
> * Validate that the filepath is a string (or raise a `TypeError`)

This seems to only happen for Prism.parse but not Prism.parse_file:
```
$ ruby -Ilib -rprism -e 'pp Prism.parse(nil)'
-e:1:in `parse': wrong argument type NilClass (expected String) (TypeError)

pp Prism.parse(nil)
               ^^^
	from -e:1:in `<main>'
zsh: exit 1     ruby -Ilib -rprism -e 'pp Prism.parse(nil)'

$ ruby -Ilib -rprism -e 'pp Prism.parse_file(nil)'
open: Bad address
nil
$ echo $?
0
```

kddnewton:
Yes, `nil` is special-cased at the moment, that should be handled as well.

duerst:
> Right now, when you call any of these APIs:
> 
> * `Prism::dump_file`
> * `Prism::lex_file`
> * `Prism::parse_file`
[snip]
> It will do a few things:
> 
> * Validate that the filepath is a string (or raise a `TypeError`)
[snip]

Shouldn't this be a bit more "duck-typed", so that e.g. a file can be passed directly or some other object can be the source of the code to be parsed?

eregon:
These APIs are designed to be passed a filesystem path.
[...]

PR #2365, opened 2024-02-05
Implement file parsing error handling

eileencodes:
Implement file parsing error handling

This PR implements proper file parsing error handling. Previously
`file_options` would call `pm_string_mapped_init` which would print an
error from `perror`. However this wouldn't raise a proper Ruby error so
it was just a string output. I've done the following:

- Raise an error from `rb_syserr_fail` with the filepath in
`file_options`.
- No longer return `Qnil` if `file_options` returns false (because now 
it will raise)
- Update `file_options` to return `static void` instead of `static
bool`.
- Update `file_options` and `profile_file` to check the type so when
passing `nil` we see a `TypeError`.
- Delete `perror` from `pm_string_mapped_init`
- Update `FFI` backend to raise appropriate errors when calling
`pm_string_mapped_init`.
- Add tests for `dump_file`, `lex_file`, `parse_file`,
`parse_file_comments`, `parse_lex_file`, and `parse_file_success?`,
when a file doesn't exist and for `nil`.
- Updates the `bin/parse` script to no longer raise it's own 
`ArgumentError` now that we raise a proper error.

Fixes: ruby/prism#2207

### DIFF
--- a/test/prism/parse_test.rb
+++ b/test/prism/parse_test.rb
@@ -69,11 +69,96 @@ module Prism
       assert_equal 5, tokens.length
     end
 
+    def test_dump_file
+      assert_nothing_raised do
+        Prism.dump_file(__FILE__)
+      end
+
+      error = assert_raise Errno::ENOENT do
+        Prism.dump_file("idontexist.rb")
+      end
+
+      assert_equal "No such file or directory - idontexist.rb", error.message
+
+      assert_raise TypeError do
+        Prism.dump_file(nil)
+      end
+    end
+
+    def test_lex_file
+      assert_nothing_raised do
+        Prism.lex_file(__FILE__)
+      end
+
+      error = assert_raise Errno::ENOENT do
+        Prism.lex_file("idontexist.rb")
+      end
+
+      assert_equal "No such file or directory - idontexist.rb", error.message
+
+      assert_raise TypeError do
+        Prism.lex_file(nil)
+      end
+    end
+
     def test_parse_lex_file
       node, tokens = Prism.parse_lex_file(__FILE__).value
 
       assert_kind_of ProgramNode, node
       refute_empty tokens
+
+      error = assert_raise Errno::ENOENT do
+        Prism.parse_lex_file("idontexist.rb")
+      end
+
+      assert_equal "No such file or directory - idontexist.rb", error.message
+
+      assert_raise TypeError do
+        Prism.parse_lex_file(nil)
+      end
+    end
+
+    def test_parse_file
+      node = Prism.parse_file(__FILE__).value
+      assert_kind_of ProgramNode, node
+
+      error = assert_raise Errno::ENOENT do
+        Prism.parse_file("idontexist.rb")
+      end
+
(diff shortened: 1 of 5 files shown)


eileencodes:
Oops, forgot to run all the tests. Will update.

kddnewton (approved):
Great work!

kddnewton:
Looks like the windows tests are failing because it's a different errno error (I think that's expected). Can you check for a superclass?

kddnewton:
Ahh it looks like FFI is failing because it needs to be updated. You can check that with:

PRISM_FFI_BACKEND=1 bundle exec rake

eileencodes:
I'm a little bit stuck. FFI tests are failing because we check `LibRubyParser.pm_string_mapped_init` and raise if it's `false`. 

https://github.com/ruby/prism/blob/837af48f7176c1b67ae3a247ce1dc4221f386d82/lib/prism/ffi.rb#L163

I can't figure out how (or even if) I can get the `errno` here in order to raise a `SystemCallError.new(filepath, ????)` with the right error code. The errno is different for `nil` or doesn't exist, so I can't pass in the `Errno::ENOENT` constant.

eileencodes:
After pairing with Kevin (thank you!) we've changed `pm_string_mapped_init` to return the error code instead of a boolean so we can raise an appropriate `SystemCallError` from FFI.

eregon:
FWIW there is `FFI.errno` to get the errno.
That's fully reliable if the previous call was a call done with FFI (i.e. it's properly saved&restored around FFI calls and not just accessing the native `errno`).

test/prism/parse_test.rb
```diff
@@ -69,11 +69,108 @@ def test_parse_lex
...
+        Prism.dump_file("idontexist.rb")
+      end
+
+      assert_equal "No such file or directory - idontexist.rb", error.message
+
+      error = assert_raise Errno::EFAULT do
+        Prism.dump_file(nil)
+      end
+
+      assert_equal "Bad address", error.message
```
eregon, line 87: It feels a bit strange that nil would raise EFAULT.
I would expect a coercion error, just like `File.read` does:
```
$ ruby -e 'File.read nil'
-e:1:in `read': no implicit conversion of nil into String (TypeError)
```

test/prism/parse_test.rb
```diff
@@ -69,11 +69,112 @@ def test_parse_lex
...
+      end
+
+      assert_equal "No such file or directory - idontexist.rb", error.message
+
+      assert_raise TypeError do
+        Prism.parse_file_comments(nil)
+      end
+    end
+
+    def test_profile_file
```
kddnewton, line 164: This one you should remove, because it's not implemented in the FFI. We'll trust that it'll work since it's not supposed to be exposed anyway. But thank you for being thorough!

eileencodes:
Ok I think this is ready now! 😄

test/prism/parse_test.rb
eileencodes: It's now raising a `TypeError`.

kddnewton (approved):
Nice job!

Issue #2354, opened 2024-02-02
Ripper compatibility layer


kddnewton:
We want to replace Ripper with prism within the community as the default choice when a developer needs access to a parser. It's possible that given enough time we could replace all of the public uses ourselves, but we won't be able to migrate private repositories/applications that are using it. Even though Ripper has been marked as "early alpha" since it was first introduced, it has been around a long enough time that people have been relying on it for years.

We want to provide a compatibility layer such that people can continue to use Ripper for as long as they need to, but have the parsing be backed by prism. This means replacing all of Ripper's surface area, which is no small task. That surface area can loosely be broken up into two different groups: Ripper core and the Ripper API.

Ripper core is a SAX-style Ruby parser. This means it calls out to Ruby methods whenever a rule is reduced and uses the results of calling those methods as arguments to more methods higher in the tree. For example:

```ruby
class Foo < Ripper
  def on_binary(left, op, right) = [:binary, left, op, right]
  def on_int(value) = [:@int, value]
  def on_stmts_new = []
  def on_stmts_add(stmts, stmt) = stmts << stmt
  def on_program(stmts) = stmts
end
```

Using `Ripper.new("1 + 2").parse` you will get back `[[:binary, [:@int, "1"], :+, [:@int, "2"]]]`. The results of our `on_int` method get passed to `on_binary`, which get passed to `on_stmts_add`, which get passed up to `on_program`.

We're not going to put hooks into prism's parsing for every rule reduction. We would not be able to maintain it, and it would clutter the entire codebase. Instead, we're going to replicate Ripper by parsing the entire tree with prism and then walking the tree in the same order as Ripper to generate the same output. Note that we're going to be calling out to methods that are user-defined, not providing the implementation of those methods ourselves. The order of the methods that are called therefore very much matters. You can see the beginnings of this approach here: https://github.com/ruby/prism/blob/main/lib/prism/ripper_compat.rb.

Unfortunately, because Ripper exposes so much of the internals of the Ruby parser, you can also access various state on the parser. This includes the current line number, column, and parser state. We're not going to be able to replicate the parser state very accurately, so I don't think we should try. The lines and columns we can though. This will involve changing the line and column numbers to be in a specific state before calling out to the user-defined hooks.

[...]

PR #2372, opened 2024-02-06
More visitors and tests for RipperCompat

noahgibbs:
Part of issue #2354

More support for operators. Add basic support for ranges, parentheses and negative integers. Tests for each of these.

Full Ripper compatibility will require a lot more, but it seems good to start simple.

### DIFF
--- a/lib/prism/ripper_compat.rb
+++ b/lib/prism/ripper_compat.rb
@@ -108,15 +108,49 @@ module Prism
 
     # Visit a CallNode node.
     def visit_call_node(node)
-      if !node.message.match?(/^[[:alpha:]_]/) && node.opening_loc.nil? && node.arguments&.arguments&.length == 1
-        left = visit(node.receiver)
-        right = visit(node.arguments.arguments.first)
+      bounds(node.location)
 
-        bounds(node.location)
-        on_binary(left, node.name, right)
-      else
-        raise NotImplementedError
+      if node.message.match?(/^[[:alpha:]_]/)
+        val = on_ident(node.message)
+
+        return on_vcall(val)
       end
+
+      if node.opening_loc.nil?
+        left = visit(node.receiver)
+        if node.arguments&.arguments&.length == 1
+          right = visit(node.arguments.arguments.first)
+
+          on_binary(left, node.name, right)
+        elsif !node.arguments || node.arguments.empty?
+          on_unary(node.name, left)
+        else
+          raise NotImplementedError, "More than two arguments for operator"
+        end
+      else
+        # This is a method call, not an operator
+        raise NotImplementedError, "Non-nil opening_loc"
+      end
+    end
+
+    # Visit a RangeNode
+    def visit_range_node(node)
+      bounds(node.location)
+      left = visit(node.left)
+      right = visit(node.right)
+
+      if node.operator_loc.slice == "..."
+        on_dot3(left, right)
+      else
+        on_dot2(left, right)
+      end
+    end
+
+    # Visit a ParenthesesNode
+    def visit_parentheses_node(node)
+      bounds(node.location)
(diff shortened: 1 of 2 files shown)


noahgibbs:
Ooh, interesting. JRuby seems to get a different unary minus symbol here, :- rather than :-@.

enebo:
@noahgibbs can you tell me which case is wrong in JRuby?   I assume some incompatability in what we currently ship?

noahgibbs:
@enebo The change I had to make was with Ripper's unary minus. Specifically, in CRuby:

```ruby
noah@Noahs-MBP-3 prism % irb
irb(main):001> require "ripper"
=> false
irb(main):002> Ripper.sexp_raw("-7")
=> [:program, [:stmts_add, [:stmts_new], [:unary, :-@, [:@int, "7", [1, 1]]]]]
```

For JRuby, it looks like the same expression is getting the symbol colon-minus rather than colon-minus-at just after :unary. Trying this locally on JRuby 9.4.3.0 (it's what I already have installed), that's indeed what I see:

```ruby
noah@Noahs-MBP-3 prism % chruby jruby-9.4.3.0
noah@Noahs-MBP-3 prism % irb
irb(main):001:0> require "ripper"
=> false
irb(main):002:0>  Ripper.sexp_raw("-7")
=> [:program, [:stmts_add, [:stmts_new], [:unary, :-, [:@int, "7", [1, 1]]]]]
```

enebo:
Thanks @noahgibbs.  I filed an issue for this: https://github.com/jruby/jruby/issues/8091

lib/prism/ripper_compat.rb
```diff
@@ -133,8 +167,22 @@ def visit_imaginary_node(node)
...
+      text = node.slice
+      loc = node.location
+      if text[0] == "-"
+        # TODO: This is ugly. We test that a newline between - and 7 doesn't give an error, but this could still be bad somehow.
+        bounds_line_col(loc.start_line, loc.start_column + 1)
+        on_int_val = on_int(text[1..-1])
+        bounds(node.location)
+        if RUBY_ENGINE == "jruby"
+          on_unary(:-, on_int_val)
+        else
```
eregon, line 179: I would suggest to run `test/prism/ripper_compat_test.rb` only on CRuby (`return unless RUBY_ENGINE == "ruby"`) and match only CRuby's behavior here.
Any difference of Ripper on JRuby/TruffleRuby is most likely an unintended incompatibility with CRuby.

enebo:
@noahgibbs Already fixed and will be in JRuby 9.4.6.0 which will be out this month.  I appreciate you running JRuby for these tests as you will help me find some bugs :smile:

eregon:
@enebo Ah maybe I spoke too son then :sweat_smile: 
But do you think it's good for RipperCompat to mimic JRuby's differences/incompatibilities in Ripper?
Or rather the incompatible test(s) should be skipped on JRuby then?

FWIW in TruffleRuby we plan to replace Ripper with Prism::RipperCompat once it's complete enough.
For JRuby that would avoid having to maintain Ripper too.

noahgibbs:
What I'm doing short-term is to take Tom at his word. The next update to RipperCompat will check and only do this if JRUBY_VERSION is less than 9.4.6.0. The same approach will work if you integrate RipperCompat: it can stop bothering with the quirks for any version of JRuby or TruffleRuby that defaults to using RipperCompat :-)

If there turn out to be a lot of quirks, or quirks that are inconvenient to work around, I may revisit the approach. But differences of this size aren't a big deal. We *do* want this to be standard for Ruby tooling, which means supporting currently-in-existence versions of JRuby, TruffleRuby and pre-Prism CRuby. Presumably we figure out guidelines for how long to support these things when we're further along.

PR #2375, opened 2024-02-07
RipperCompat: support method calls with blocks

noahgibbs:
There are a few cases we don't correctly support yet ("foo bar 1") and the argument handling is messy here. I'm working on it. But we're supporting a few fixture files correctly now.

This is part of fixing issue #2354.

### DIFF
--- a/lib/prism/ripper_compat.rb
+++ b/lib/prism/ripper_compat.rb
@@ -116,30 +116,78 @@ module Prism
     end
 
     # Visit a CallNode node.
+    # Ripper distinguishes between many different method-call
+    # nodes -- unary and binary operators, "command" calls with
+    # no parentheses, and call/fcall/vcall.
     def visit_call_node(node)
       if node.variable_call?
-        if node.message.match?(/^[[:alpha:]_]/)
-          bounds(node.message_loc)
-          return on_vcall(on_ident(node.message))
-        end
+        raise NotImplementedError unless node.receiver.nil?
 
-        raise NotImplementedError, "Non-alpha variable call"
+        bounds(node.message_loc)
+        return on_vcall(on_ident(node.message))
       end
 
       if node.opening_loc.nil?
-        left = visit(node.receiver)
-        if node.arguments&.arguments&.length == 1
-          right = visit(node.arguments.arguments.first)
+        # No opening_loc can mean an operator. It can also mean a
+        # method call with no parentheses.
+        if node.message.match?(/^[[:punct:]]/)
+          left = visit(node.receiver)
+          if node.arguments&.arguments&.length == 1
+            right = visit(node.arguments.arguments.first)
 
-          on_binary(left, node.name, right)
-        elsif !node.arguments || node.arguments.empty?
-          on_unary(node.name, left)
+            return on_binary(left, node.name, right)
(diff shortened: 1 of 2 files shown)


lib/prism/ripper_compat.rb
```diff
@@ -238,13 +333,20 @@ def visit_number(node)
         value = yield slice[1..-1]
 
         bounds(node.location)
-        on_unary(RUBY_ENGINE == "jruby" ? :- : :-@, value)
+        on_unary(RUBY_ENGINE == "jruby" && Gem::Version.new(JRUBY_VERSION) < Gem::Version.new("9.4.6.0") ? :- : :-@, value)
```
kddnewton, line 336: Could you make this something like:

```ruby
if RUBY_ENGINE == "jruby" && Gem::Version.new(JRUBY_VERSION) < Gem::Version.new("9.4.6.0")
  def visit_unary_operator(value)
    value == :-@ ? :- : value
  end
else
  def visit_unary_operator(value)
    value
  end
end
```

That way the check is only performed once.

noahgibbs: Sure. Pushed.

lib/prism/ripper_compat.rb
```diff
@@ -238,13 +333,32 @@ def visit_number(node)
...
+      def visit_unary_operator(value)
+        value
+      end
+    end
+
+    # Visit a binary operator node like an AndNode or OrNode
+    def visit_binary_operator(node)
+      left_val = visit(node.left)
+      right_val = visit(node.right)
+      on_binary(left_val, node.operator_loc.slice.to_sym, right_val)
```
kddnewton, line 359:
```suggestion
      on_binary(left_val, node.operator, right_val)
```

lib/prism/ripper_compat.rb
```diff
@@ -203,6 +228,21 @@ def visit_statements_node(node)
       end
     end
 
+    private
+
+    # Ripper generates an interesting format of argument list.
+    # It seems to be very location-specific. We should get rid of
+    # this method and make it clearer how it's done in each place.
+    def args_node_to_arguments(args_node)
```
kddnewton, line 236: Just for my own paranoia could you move this down to the rest of the `private` methods? I know it's pedantic, but I'm always fearful of accidentally exposing methods we don't mean to.

kddnewton (changes requested):

lib/prism/ripper_compat.rb
noahgibbs: I've actually removed it entirely in my next change.

noahgibbs: Okay, moved it in this PR. But either way, disappearing soon.

lib/prism/ripper_compat.rb
noahgibbs: Oh hey - thought I had committed this because the button was greyed out, but that was because it was outdated. Added this.

kddnewton (approved):
Looking good!

kddnewton:
Oh shoot, sorry. That should have been `.operator.to_sym`

kddnewton (approved):

PR #2378, opened 2024-02-07
Dev-only CLI

kddnewton:
We keep adding more scripts to /bin that are doing the same kinds of processing. Instead, this commit consolidates them all into a single CLI that shares the same logic so that we can consistently read files in the same way.

It keeps around 2 binstubs for bin/lex and bin/parse since those are the most used and I'm sure people have built up muscle memory for those. Those scripts are now just wrappers for forwarding to bin/prism.


noahgibbs (approved):

noahgibbs:
Seems like a good idea.

enebo:
@kddnewton Is there a possibility of us making a non-dev script so we can run and generate AST, etc... from the installed gem?

kddnewton:
Yeah we could do that. I don't think CRuby will want us to ship that with it so we'll need to take a little care to make sure it doesn't get copied over, but that's doable. Could you open an issue for it?

enebo:
@noahgibbs I added https://github.com/jruby/jruby/issues/8094 for that newly discovered ripper issue.

kddnewton:
@enebo if you have CRuby 3.3, you can run `ruby --dump=prism -e 'whatever'` and you'll get the AST.

enebo:
@kddnewton Yeah that works for MRI for sure.  I have thought about adding similar thing to JRuby but have a bin/ast (which only currently works for legacy parser).

eregon:
It'd be quite convenient to have an executable, because that would then use the installed gem vs the version used in CRuby at CRuby build time.
I think it would make sense to add a `prism` or so executable to CRuby too, because it does ship Prism as a default gem so it should be consistent (it's same for e.g. rdoc).

PR #2379, opened 2024-02-08
RipperCompat: support for more features.

noahgibbs:
* add bin/prism ripper to compare Ripper output
* block arg handling is quirky, do it per-call-site
* required params for blocks
* boolean values
* various assign-operator support
* breaks
* early fragile begin/rescue/end
* more fixtures being checked

Part of #2354.

### DIFF
--- a/lib/prism/ripper_compat.rb
+++ b/lib/prism/ripper_compat.rb
@@ -132,7 +132,12 @@ module Prism
       end
 
       # A non-operator method call with parentheses
-      args = on_arg_paren(node.arguments.nil? ? nil : args_node_to_arguments(node.arguments))
+
+      args = if node.arguments.nil?
+        on_arg_paren(nil)
+      else
+        on_arg_paren(on_args_add_block(visit_elements(node.arguments.arguments), false))
+      end
 
       bounds(node.message_loc)
       ident_val = on_ident(node.message)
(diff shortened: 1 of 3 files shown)


noahgibbs:
I've removed the break-based code fixtures. It turns out that 3.2.2 parses a free-standing "break" statement just fine, but latest Ruby does not. For my unit tests I just put the break inside a method call. But we'll need to figure out what to do with fixtures that become unparseable by later versions of CRuby.

lib/prism/ripper_compat.rb
```diff
@@ -142,31 +147,92 @@ def visit_call_node(node)
...
+    end
+
+    # Visit a LocalVariableOrWriteNode.
+    def visit_local_variable_or_write_node(node)
+      visit_binary_op_assign(node)
+    end
+
+    # Visit nodes for +=, *=, -=, etc., called LocalVariableOperatorWriteNodes.
+    def visit_local_variable_operator_write_node(node)
+      visit_binary_op_assign(node, operator: node.operator.to_s + "=")
```
kddnewton, line 168:
```suggestion
      visit_binary_op_assign(node, operator: "#{node.operator}=")
```

lib/prism/ripper_compat.rb
```diff
@@ -142,31 +147,92 @@ def visit_call_node(node)
...
+    # Visit a BlockParametersNode.
+    def visit_block_parameters_node(node)
+      on_block_var(visit(node.parameters), no_block_value)
+    end
+
+    # Visit a ParametersNode.
+    # This will require expanding as we support more kinds of parameters.
+    def visit_parameters_node(node)
+      #on_params(required, optional, nil, nil, nil, nil, nil)
+      on_params(node.requireds.map { |n| visit(n) }, nil, nil, nil, nil, nil, nil)
```
kddnewton, line 197: Ahh this is the first place where we need `visit_all`. We should change the parent class to `Prism::Compiler` instead of `Prism::Visitor` so that we can use it.

```suggestion
      on_params(visit_all(node.requireds), nil, nil, nil, nil, nil, nil)
```

lib/prism/ripper_compat.rb
```diff
@@ -142,31 +147,92 @@ def visit_call_node(node)
...
     end
 
-    # Visit an AndNode
+    # Visit a RequiredParameterNode.
+    def visit_required_parameter_node(node)
+      bounds(node.location)
+      on_ident(node.name.to_s)
```
kddnewton, line 203:
```suggestion
      on_ident(node.name.name)
```

lib/prism/ripper_compat.rb
```diff
@@ -142,31 +147,92 @@ def visit_call_node(node)
...
+    end
+
+    # Visit a BreakNode.
+    def visit_break_node(node)
+      return on_break(on_args_new) if node.arguments.nil?
+
+      args_val = visit_elements(node.arguments.arguments)
+      on_break(on_args_add_block(args_val,false))
```
kddnewton, line 211:
```suggestion
      on_break(on_args_add_block(args_val, false))
```

lib/prism/ripper_compat.rb
```diff
@@ -142,31 +147,92 @@ def visit_call_node(node)
...
-    # Visit an OrNode
+    # Visit an OrNode.
     def visit_or_node(node)
       visit_binary_operator(node)
     end
 
+    # Visit a TrueNode.
+    def visit_true_node(node)
+      bounds(node.location)
+      on_var_ref(on_kw(node.slice))
```
kddnewton, line 227:
```suggestion
      on_var_ref(on_kw("true"))
```

lib/prism/ripper_compat.rb
```diff
@@ -142,31 +147,92 @@ def visit_call_node(node)
...
+    end
+
+    # Visit a FalseNode.
+    def visit_false_node(node)
+      bounds(node.location)
+      on_var_ref(on_kw(node.slice))
```
kddnewton, line 233:
```suggestion
      on_var_ref(on_kw("false"))
```

kddnewton (changes requested):
A couple of tiny things, but really good overall!

lib/prism/ripper_compat.rb
noahgibbs: Okay. I did this and changed the parent class.

kddnewton (approved):
Looks great!

noahgibbs:
Looks like Symbol#name isn't in Ruby 2.7, so need to switch back to .to_s on that line.

Issue #2380, opened 2024-02-08
Include the line offsets in the serialized output


eregon:
I'm profiling translation (which is the conversion Prism AST -> Truffle AST) time in TruffleRuby and the bottleneck seems pretty clear:
```
$ jt ruby --vm.agentpath:/home/eregon/Downloads/async-profiler-3.0-linux-x64/lib/libasyncProfiler.so=start,event=cpu,interval=1000000,flat=1000 -e 'Truffle::Debug.profile_translator File.read("src/main/ruby/truffleruby/core/io.rb"), 50_000'

          ns  percent  samples  top
  ----------  -------  -------  ---
  1835122755   13.74%     1835  org.prism.Nodes$Source.computeLineOffsets
   231075955    1.73%      231  org.truffleruby.parser.TranslatorEnvironment.<init>

          ns  percent  samples  top
  ----------  -------  -------  ---
  2850012917   20.39%      285  org.prism.Nodes$Source.computeLineOffsets
   220023193    1.57%       22  org.truffleruby.parser.TranslatorEnvironment.<init>

          ns  percent  samples  top
  ----------  -------  -------  ---
  1649999671   16.57%      165  org.prism.Nodes$Source.computeLineOffsets
   180003810    1.81%       18  IndexSetIterator::advance_and_next
```
That's computing the line offsets in Java.

Computing the line offsets in Ruby is also a bottleneck, as found in https://github.com/oracle/truffleruby/issues/3293.

I looked at what actually uses the line offsets in TruffleRuby:
* Used in setNewLineFlag(). Potentially C code could do that instead but then we'd need to pass these extra flags from C to Java.
* Used for coverage, when coverage is enabled to mark the line has "having code". It seems hard to avoid.
* For translating `__LINE__`. Maybe `SourceLineNode` could include a `varsint` field with the line number for convenience to avoid that one.
* Getting the line for warnings & errors.

eregon:
I'll implement this.

PR #2381, opened 2024-02-08
Serialize the newline_list to avoid recomputing it again later

eregon:
* Fixes https://github.com/ruby/prism/issues/2380

Related: https://github.com/ruby/prism/issues/2402

### DIFF
--- a/lib/prism/parse_result.rb
+++ b/lib/prism/parse_result.rb
@@ -15,10 +15,10 @@ module Prism
     attr_reader :offsets
 
     # Create a new source object with the given source code.
-    def initialize(source)
+    def initialize(source, start_line = 1, offsets = [])
       @source = source
-      @start_line = 1 # set after parsing is done
-      @offsets = [] # set after parsing is done
+      @start_line = start_line # set after parsing is done
+      @offsets = offsets # set after parsing is done
     end
 
     # Perform a byteslice on the source code using the given byte offset and
--- a/lib/prism/parse_result.rb
+++ b/lib/prism/parse_result.rb
@@ -12,15 +12,13 @@ module Prism
     attr_accessor :start_line
 
     # The list of newline byte offsets in the source code.
-    attr_reader :offsets
+    attr_accessor :offsets
 
-    # Create a new source object with the given source code and newline byte
-    # offsets. If no newline byte offsets are given, they will be computed from
-    # the source code.
-    def initialize(source, start_line = 1, offsets = compute_offsets(source))
+    # Create a new source object with the given source code.
+    def initialize(source)
       @source = source
-      @start_line = start_line
-      @offsets = offsets
+      @start_line = 1 # set after parsing is done
+      @offsets = [] # set after parsing is done
     end
 
     # Perform a byteslice on the source code using the given byte offset and
@@ -76,14 +74,6 @@ module Prism
       end
 
       left - 1
-    end
-
-    # Find all of the newlines in the source code and return their byte offsets
-    # from the start of the string an array.
-    def compute_offsets(code)
-      offsets = [0]
-      code.b.scan("\n") { offsets << $~.end(0) }
-      offsets
     end
   end
 
--- a/lib/prism/parse_result.rb
+++ b/lib/prism/parse_result.rb
@@ -9,10 +9,10 @@ module Prism
     attr_reader :source
 
     # The line number where this source starts.
-    attr_accessor :start_line
+    attr_reader :start_line
 
     # The list of newline byte offsets in the source code.
-    attr_accessor :offsets
+    attr_reader :offsets
 
     # Create a new source object with the given source code.
     def initialize(source)
--- a/templates/ext/prism/api_node.c.erb
+++ b/templates/ext/prism/api_node.c.erb
@@ -36,18 +36,26 @@ pm_string_new(pm_string_t *string, rb_encoding *encoding) {
     return rb_enc_str_new((const char *) pm_string_source(string), pm_string_length(string), encoding);
 }
 
-// Create a Prism::Source object from the given parser.
+// Create a Prism::Source object from the given parser, after pm_parse() was called.
 VALUE
 pm_source_new(pm_parser_t *parser, rb_encoding *encoding) {
-    VALUE source = rb_enc_str_new((const char *) parser->start, parser->end - parser->start, encoding);
+    VALUE source_string = rb_enc_str_new((const char *) parser->start, parser->end - parser->start, encoding);
+    VALUE source_argv[] = { source_string };
+    VALUE source = rb_class_new_instance(1, source_argv, rb_cPrismSource);
+
+    pm_source_init(source, parser);
+    return source;
+}
+
+void
+pm_source_init(VALUE source, pm_parser_t *parser) {
+    rb_funcall(source, rb_intern("start_line="), 1, LONG2NUM(parser->start_line));
+
     VALUE offsets = rb_ary_new_capa(parser->newline_list.size);
-
     for (size_t index = 0; index < parser->newline_list.size; index++) {
-        rb_ary_push(offsets, INT2FIX(parser->newline_list.offsets[index]));
+        rb_ary_push(offsets, ULONG2NUM(parser->newline_list.offsets[index]));
     }
-
-    VALUE source_argv[] = { source, LONG2NUM(parser->start_line), offsets };
-    return rb_class_new_instance(3, source_argv, rb_cPrismSource);
+    rb_funcall(source, rb_intern("offsets="), 1, offsets);
 }
 
 typedef struct pm_node_stack_node {
@@ -77,8 +85,7 @@ pm_node_stack_pop(pm_node_stack_node_t **stack) {
 }
 
 VALUE
-pm_ast_new(pm_parser_t *parser, pm_node_t *node, rb_encoding *encoding) {
-    VALUE source = pm_source_new(parser, encoding);
+pm_ast_new(pm_parser_t *parser, pm_node_t *node, rb_encoding *encoding, VALUE source) {
     ID *constants = calloc(parser->constant_pool.size, sizeof(ID));
 
     for (uint32_t index = 0; index < parser->constant_pool.size; index++) {
--- a/templates/ext/prism/api_node.c.erb
+++ b/templates/ext/prism/api_node.c.erb
@@ -40,22 +40,14 @@ pm_string_new(pm_string_t *string, rb_encoding *encoding) {
 VALUE
 pm_source_new(pm_parser_t *parser, rb_encoding *encoding) {
     VALUE source_string = rb_enc_str_new((const char *) parser->start, parser->end - parser->start, encoding);
-    VALUE source_argv[] = { source_string };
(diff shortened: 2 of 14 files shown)


eregon:
Should be green and ready for review now.

enebo (changes requested):
Something bothers me that we are putting Source into ParseResult.  I feel like we should consider only putting LineOffsets into it.  Our end goal is to not have any source by the time we get our serialized source back.  We just have not made it to that point yet.

If this is true then Source is nothing more than startLine and source bytes[].  Do we actually need startLine or could this be boiled into the linecalcs as well?  I did not read that but I wonder if we ever need that value after parsing is over.  If so then Source can go away as well.

enebo:
@eregon Also simple error in build-java failing in CI but you obviously will fix that.

eregon:
@enebo I think it's a lot cleaner and less error-prone this way, and it's also exactly how it works in the Ruby API (how `Prism.parse` takes a String of code and returns a ParseResult which includes the Prism::Source).
The main concern there is that if someone creates a Source object they would end up with invalid lineOffsets, unless the same object is also passed to Loader.
If the Source object is created by Loader it's not possible to have this issue.
Also creating the Source object in the caller felt awkward, I'd rather Loader takes just a `(byte[] serialize, byte[] sourceBytes)`.

> Do we actually need startLine or could this be boiled into the linecalcs as well?

It cannot be stored in the lineOffsets (which are offsets in the byte[]/char* source, so 0-based), it needs to be a separate field.

Regarding once we don't need the source bytes anymore for Loader, I think that's fine, then one would be able to just pass `null` for the `sourceBytes` (or have an overload doing that). And then the Source object would only contain startLine + lineOffsets, which seems all good (well except the name maybe, but its main purpose is to provide *source* line & column info).
In fact the only usage for `bytes` in the Source Java class currently is `bytes.length`, which we could do another way or avoid needing entirely in `findLine()`.

It is also something this PR helps significantly, since the lineOffsets now come from the serialized form and there is no longer a need to use the `sourceBytes` to find them.

If it's problematic later for some unforeseen we can of course always rework this, it's pretty easy to change, b836f031cad936826b5a5e6fcb1cb071cd665ded is a pretty small commit.

Are you OK with that?
If not I can remove that commit, but IMO it is significantly cleaner this way and also more convenient to call Loader. I was confused for a while how startLine was set before, if the Loader creates the Source there is no hidden magic and it's obvious.

enebo:
@eregon yeah I forgot we are just getting an index as well so I did not think through the startLine comment.

Will you be keeping the source?  I don't see anyone doing this with location fields removed but for FFI consumers who have all the location data they probably do want the source (although they do already have it). 

I am ok though with landing this.  We can argue out how we deal with source bytes once we actually don't need it for anything.  I am isolating this code outside the JRuby codebase so these APIs can change as much as they want and it won't require a new JRuby release.  Just a new gem release containing a newly made provider.

enebo (approved):

enebo:
@eregon I am only approving Java changes as I did not look very closely at the rest.

eregon:
> Will you be keeping the source?

Until TruffleRuby can cache the serialized form on disk (TBD where/how, not for the 24.0 release) it needs to read the source as a byte[] anyway to pass it to parseAndSerialize().
(and of course currently the source bytes are needed for Loader so it's not yet possible to skip reading the source file)

>  I am isolating this code outside the JRuby codebase so these APIs can change as much as they want and it won't require a new JRuby release. Just a new gem release containing a newly made provider.

Don't you vendor a specific version/commit of Prism in JRuby when using it as the parser for Ruby code?
That's what TruffleRuby and CRuby do. Then it doesn't matter which version(s) of Prism are installed as gems, it's independent.

Or do you also use the Java Loader when Prism is used as a gem? I wouldn't expect that since what's tested in CI on JRuby is the Ruby API using FFI (so not using the Java Loader).

enebo:
> Don't you vendor a specific version/commit of Prism in JRuby when using it as the parser for Ruby code? That's what TruffleRuby and CRuby do. Then it doesn't matter which version(s) of Prism are installed as gems, it's independent.
> 
> Or do you also use the Java Loader when Prism is used as a gem? I wouldn't expect that since what's tested in CI on JRuby is the Ruby API using FFI (so not using the Java Loader).

I am isolating the Java part as an artifact and it is included into a Ruby gem which is effectively a "barely" fork of prism gem that has been renamed.  That gem includes the artifact and will compile the source with the ENVs we want when the gem is installed (later we will probably pre-compile to remove compile requirement on main platforms).

The rationale for doing this is if there is a CVE or just a good fix there is a high likelihood I only have to do a gem release and not a full JRuby release.

eregon:
I see, makes sense. Could be worth documenting at https://github.com/ruby/prism/blob/main/docs/build_system.md#building-prism-as-part-of-jruby
We cannot do this for TruffleRuby because it needs to work for `--native` mode and that needs to know all Java code ahead-of-time.

eregon:

templates/src/serialize.c.erb
```diff
@@ -128,6 +128,17 @@ pm_serialize_node(pm_parser_t *parser, pm_node_t *node, pm_buffer_t *buffer) {
     }
 }
 
+void
```
kddnewton, line 131:
```suggestion
static void
```

templates/ext/prism/api_node.c.erb
```diff
@@ -36,18 +36,26 @@ pm_string_new(pm_string_t *string, rb_encoding *encoding) {
...
+    VALUE source_argv[] = { source_string };
+    VALUE source = rb_class_new_instance(1, source_argv, rb_cPrismSource);
+
+    pm_source_init(source, parser);
+    return source;
+}
 
+void
+pm_source_init(VALUE source, pm_parser_t *parser) {
+    rb_funcall(source, rb_intern("start_line="), 1, LONG2NUM(parser->start_line));
```
kddnewton, line 52: I'd prefer not to go through the ceremony of making two method calls in this function. We already have the start line when the source is created, and when we create the array for the offsets we can keep that reference around and push directly into it. That'll avoid the frame and everything else that comes from `rb_funcall`.

templates/lib/prism/serialize.rb.erb
```diff
@@ -82,6 +82,11 @@ module Prism
         source.start_line = load_varsint
       end
 
+      def load_line_offsets
+        offsets = load_varuint.times.map { load_varuint }
+        source.offsets = offsets
```
kddnewton, line 87:
```suggestion
        source.offsets.concat(load_varuint.times.map { load_varuint })
```

lib/prism/parse_result.rb
```diff
@@ -12,15 +12,13 @@ class Source
     attr_accessor :start_line
 
     # The list of newline byte offsets in the source code.
-    attr_reader :offsets
+    attr_accessor :offsets
```
kddnewton, line 15: I'd rather not have this be `attr_accessor`, it's not meant to be set later.

kddnewton (changes requested):
Generally fine with the changes, just a couple of things I'd like to see changed.

templates/lib/prism/serialize.rb.erb
eregon: `concat` makes an extra copy (memcpy/arraycopy of all elements) so that's worse performance-wise.
I'll remove the extra local variable though.

kddnewton: In that case can you 

```ruby
load_varuint.times do
  source.offsets << load_varuint
end
```

templates/ext/prism/api_node.c.erb
eregon: > We already have the start line when the source is created,

This is handled here to be consistent with how Loader.java and FFI are doing things on master (and there it's not known before what is the start line, we know it after deserializing it).
https://github.com/ruby/prism/blob/18ada3120231e5b7faa821b0710f5c19f79fa11a/templates/lib/prism/serialize.rb.erb#L81-L83

I think it's valuable to be consistent here to avoid errors and help understanding how it all works.

> when we create the array for the offsets we can keep that reference around and push directly into it

If we do that then we cannot use `rb_ary_new_capa(parser->newline_list.size)` and so it will occur extra resizes (so extra allocations and memory copies).

> I'd prefer not to go through the ceremony of making two method calls in this function.

I can change them to `rb_iv_set()` if you want. But that would be inconsistent with how FFI does things.

templates/lib/prism/serialize.rb.erb
eregon: But that will cause extra resizes, same as https://github.com/ruby/prism/pull/2381#discussion_r1488204019

lib/prism/parse_result.rb
eregon: I changed both start_line and offsets to be `attr_reader`.
And then used `#instance_variable_set/rb_iv_set()` to set them internally as needed.

eregon: Actually we could even go further and remove the `attr_reader` for `offsets`, like the Java Source object does.
Since offsets is an implementation detail of how Source#line works, and it might change.
But that seems out of scope for this PR.

eregon:
@kddnewton I think I addressed all your comments, could you merge?
Notably I have some changes on top of this PR so I need this to be merged to open the PR.

templates/ext/prism/api_node.c.erb
kddnewton: In that case you can call `rb_ary_resize`. I don't want us to be setting instance variables later if we don't have to.

eregon: That's not available for the corresponding piece of code in `serialize.rb.erb`.

I want `@start_line` and `@offsets` to be handled consistently for all languages and all places setting it (for readability, maintainability and avoiding bugs), and for `@start_line` it needs to be assigned (e.g. for `Parser.load`), there is no alternative.
I think that is reasonable and more important than avoiding to assign ivars (for this case).

BTW the Java code deserializing the offsets naturally returns a new array, so it needs to assign, no alternative possible there.

I do agree setting instance variables later is not ideal in general, but there is no way to avoid mutating the Source object (thinking of it as an object graph, which includes the offsets) after it's built here, because the offsets are only available after parsing. I think this mutation is harmless since the user cannot access the Source object in an incomplete state.
I refactored the Java API a bit specifically to have the same property of no incomplete Source object accessible by the user.

kddnewton: As an alternative, if you cannot have it be inconsistent, please make the Java side work like the CRuby side.

eregon: Done.
To my surprise, it is indeed measurably faster for `Prism.parse("1 + 2")` (see commit message for the results).
I would have expected these 2 calls to rb_iv_set()/rb_funcall() to be insignificant (and they might be for real files with more core/many lines, but not for "1 + 2").
Also interesting is that `rb_iv_set()` is quite slow on CRuby, to the point that `rb_funcall()` (for an attr_writer) is twice as fast: https://gist.github.com/eregon/a6404a71151b30021ca081904f7149a0

OTOH for serialize.rb, the fastest would be public `attr_writer`.

kddnewton (approved):

PR #2382, opened 2024-02-08
RipperCompat: add array-refs, assigns, symbols, strings

noahgibbs:
More Ripper support:

* assignment for variables
* array-refs
* assignment for array-refs
* simple and slightly-dynamic symbols
* plain and embedded strings

### DIFF
--- a/lib/prism/ripper_compat.rb
+++ b/lib/prism/ripper_compat.rb
@@ -351,7 +351,7 @@ module Prism
 
     # Visit a SymbolNode node.
     def visit_symbol_node(node)
-      if node.opening && ['"', "'", "("].include?(node.opening[-1])
+      if (opening = node.opening) && (['"', "'"].include?(opening[-1]) || opening.start_with?("%s"))
         bounds(node.value_loc)
         tstring_val = on_tstring_content(node.value.to_s)
         return on_dyna_symbol(on_string_add(on_string_content, tstring_val))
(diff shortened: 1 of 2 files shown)


lib/prism/ripper_compat.rb
```diff
@@ -299,6 +309,59 @@ def visit_rational_node(node)
...
+    end
+
+    # Visit an EmbeddedStatementsNode node.
+    def visit_embedded_statements_node(node)
+      visit(node.statements)
+    end
+
+    # Visit a SymbolNode node.
+    def visit_symbol_node(node)
+      if node.opening && ['"', "'", "("].include?(node.opening[-1])
```
kddnewton, line 354: I think we'll want to change this because there are a lot of possible bounds for `%s`.

```suggestion
      if (opening = node.opening) && (['"', "'"].include?(opening[-1]) || opening.start_with?("%s"))
```

lib/prism/ripper_compat.rb
```diff
@@ -406,6 +469,23 @@ def visit_binary_op_assign(node, operator: node.operator)
       on_opassign(on_var_field(ident_val), op_val, visit(node.value))
     end
 
+    # In Prism this is a CallNode with :[] as the operator.
+    # In Ripper it's an :aref.
+    def visit_aref_node(node)
+      first_arg_val = visit(node.arguments.arguments[0])
```
kddnewton, line 475: I don't think we can make the assumption that there's only one argument, but I'm fine delaying this if you want to handle it later. Same for aref_field.

noahgibbs: Oh, huh, Ruby will take up to two. But yeah, there's a *lot* here that I haven't done with full generality. Still adding things like... Hash tables at all :-)

kddnewton (approved):

Issue #2383, opened 2024-02-08
Support `node_id`


kddnewton:
We need to support an incrementing `node_id` integer on every node allocation. This is used to power `error_highlight`, but also other tools like [action view](https://github.com/rails/rails/blob/1007cc8c033f7713783faa02dc13f845640b464d/actionview/lib/action_view/template.rb#L232-L238).

eregon:
Is there maybe another way these 2 usages could do what they do once they migrate to Prism?
I guess the fundamental need is to identify a call node from a bytecode.

One idea to avoid the need to take/waste that space for every node would be to implicitly number all nodes by their order in the tree with some defined traversal (e.g. pre-order). Then to find node id N, we'd iterate nodes, incrementing our id counter for every node visited, and assuming they are the same nodes in the same order (which is very much what these usages need to assume to work), then return that nth node. E.g. by having some method on ParseResult making this easy.

https://github.com/rails/rails/blob/1007cc8c033f7713783faa02dc13f845640b464d/actionview/lib/action_view/template.rb#L395 seems to be almost doing this already, except it doesn't have a counter for every node visited.

PR #2384, opened 2024-02-08
Support node_id

kddnewton:
I did not want to support this, because I desperately did not want to have to add 4 bytes to every node. Unfortunately, in order to support all of the functionality of CRuby, this is a requirement. Hopefully we can mitigate this memory by reducing the size of our location fields at some point.

Fixes https://github.com/ruby/prism/issues/2383.

### DIFF
--- a/include/prism/parser.h
+++ b/include/prism/parser.h
@@ -503,6 +503,9 @@ static const uint8_t PM_FORWARDING_ALL = 0x8;
  * it's considering.
  */
 struct pm_parser {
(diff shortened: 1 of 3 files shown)


eregon:
Please take a look at https://github.com/ruby/prism/issues/2383#issuecomment-1934920712, I think that would be one way to have this functionality without degrading memory footprint.

kddnewton:
Closing for now.

PR #2386, opened 2024-02-09
Use ENV CC if present when compiling

kddnewton:
Trying to determine if this fixes ppc64le problems

### DIFF
--- a/ext/prism/extconf.rb
+++ b/ext/prism/extconf.rb
@@ -83,6 +83,10 @@ end
 # By default, all symbols are hidden in the shared library.
 append_cflags("-fvisibility=hidden")
 
+# We'll use the `CC` environment variable if it is set, otherwise we'll use the
+# default compiler that `mkmf` would have chosen.
+RbConfig::MAKEFILE_CONFIG["CC"] = ENV["CC"] if ENV["CC"]
+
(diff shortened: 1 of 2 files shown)


kddnewton:
@junaruga unfortunately it looks like maybe this doesn't fix it? Maybe I need to specify the compiler in the travis config somehow?

kddnewton:
Closing for now

junaruga:
> @junaruga unfortunately it looks like maybe this doesn't fix it? Maybe I need to specify the compiler in the travis config somehow?

Yes, the code in this PR is not enough to fix the issue. If you want to set the same compiler gcc-11 used in the ruby/ruby in the ppc64le, first you need to download the gcc-11 deb package, then you need to set a specific compiler `CC=gcc-11` somewhere in the `.travis.yml`.

The following code in the `language: c` mode means `export CC=gcc-11`.

https://github.com/ruby/ruby/blob/master/.travis.yml#L42
```
compiler: gcc-11
```

junaruga:
Below is an example. 
https://github.com/ruby/prism/pull/2387

junaruga:
I rebased the #2387 to use gcc-11. Sorry a way to modify the `extconf.rb` to set a specific C compiler didn't work. But I found another way.

PR #2387, opened 2024-02-09
Use gcc-11 to align with the ruby/ruby Travis CI.

junaruga:
This PR is to use the same version of GCC with the ruby/ruby's one in Travis CI.

The used gcc-11 version is 11.4.0, the same as the one used in ruby/ruby.

https://app.travis-ci.com/github/junaruga/ruby-prism/jobs/617607554#L172
> gcc-11 (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

---

However in the `rake compile`, the `cc` is used in the actual command line. That is not expected. I don't know why.
https://app.travis-ci.com/github/junaruga/ruby-prism/jobs/617607554#L219

Edit: solved the issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,11 @@ matrix:
     - arch: s390x
   fast_finish: true
 
+before_install:
+  - sudo apt update -yq
+  - sudo apt -yq install gcc-11
+  - gcc-11 --version
+
 before_script:
   # Enable the verbose option in mkmf.rb to print the compiling commands.
-  - export MAKEFLAGS="V=1"
+  - export MAKEFLAGS="V=1 CC=gcc-11"


junaruga:
Seeing another example, 
https://github.com/sparklemotion/nokogiri/blob/e67310bbc2fec06878b7dc34f58982e800fb42ab/ext/nokogiri/extconf.rb#L629-L631

```
if ENV["CC"]
  RbConfig::CONFIG["CC"] = RbConfig::MAKEFILE_CONFIG["CC"] = ENV["CC"]
end
```

Perhaps, the code below is not enough.

```
RbConfig::MAKEFILE_CONFIG["CC"] = ENV["CC"] if ENV["CC"]
```

junaruga:
OK. I found the solution. Let me rebase this PR.

junaruga:
Note that in this PR, the `gcc-11` is used to compile the prism native extension.
https://app.travis-ci.com/github/ruby/prism/jobs/617609010#L252

But I still see the failure in the ppc64le.

junaruga:
I will remove the 2nd commit below in this PR.

```
diff --git a/test/prism/library_symbols_test.rb b/test/prism/library_symbols_test.rb
index b10a367c..4a4d35b2 100644
--- a/test/prism/library_symbols_test.rb
+++ b/test/prism/library_symbols_test.rb
@@ -4,9 +4,6 @@ require_relative "test_helper"

 return if RUBY_PLATFORM !~ /linux/

-# TODO: determine why these symbols are incorrect on ppc64le
-return if RUBY_PLATFORM =~ /powerpc64le/
-
 module Prism
   #
   #  examine a prism dll or static archive for expected external symbols.
```

kddnewton:
@junaruga is this ready to go?

junaruga:
Yes! I think this PR is ready to go! It's up to you about merging this PR.

PR #2398, opened 2024-02-12
Fix `full_name` for constant path targets

vinistock:
We noticed that `full_name` was not working for constant targets. We probably forgot to add handling for them in the original PR.

### DIFF
--- a/lib/prism/node_ext.rb
+++ b/lib/prism/node_ext.rb
@@ -94,7 +94,7 @@ module Prism
 
     # Returns the full name of this constant. For example: "Foo"
     def full_name
-      name.name
+      name.to_s
     end
   end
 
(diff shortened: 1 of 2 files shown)


kddnewton (approved):

lib/prism/node_ext.rb
```diff
@@ -144,6 +154,19 @@ def full_name
...
+  class ConstantTargetNode < Node
+    # Returns the list of parts for the full name of this constant.
+    # For example: [:Foo]
+    def full_name_parts
+      [name]
+    end
+
+    # Returns the full name of this constant. For example: "Foo"
+    def full_name
+      name.name
```
kddnewton, line 166: This doesn't exist on Ruby 2.7 unfortunately. Can you switch to `#to_s`.

PR #2406, opened 2024-02-13
Add code unit APIs to location

kddnewton:
LSPs need this because the protocol dictates that you return code units for offsets. None of our existing APIs provided that information, and since we hid the source it's not nearly as useful for them. Now they can pass an encoding directly to:

* Location#start_code_units_offset
* Location#end_code_units_offset
* Location#start_code_units_column
* Location#end_code_units_column


eregon:
Nice, this should address the code unit stuff discussed in https://github.com/ruby/prism/issues/1788

PR #2409, opened 2024-02-13
Validate multibyte characters in strings

kddnewton:
Check that multibyte characters are valid using pm_strpbrk. We need to add a couple of codepaths to ensure all encodings are covered.

Importantly this doesn't check regular expressions, because apparently you're allowed to have invalid multibyte characters inside regular expression comment groups/extended mode.


nirvdrum (commented):
I had left off scanning previously because I was unsure of the impact in no longer using `strchr`. But, I think that's unavoidable. Thanks for taking the steps for that.

This is so close to tracking the code range. Can we add a new set of flags for that? It'd help avoid repeated byte scans later on. E.g., I think we could eliminate calls to  `pm_ascii_only_p` when calculating the encoding flags for a `Symbol`. It'd make validation of regular expressions a lot easier as well.

kddnewton:
I totally agree with you, but I just want to get this part out the door to unblock the eval PR. That being said, I opened https://github.com/ruby/prism/issues/2410 to track the code range stuff so it's not forgotten.

PR #2413, opened 2024-02-14
Bump to v0.23.0

kddnewton:


noahgibbs:
Man that's a lot of places to change. Glad we have release docs for it.

kddnewton:
Yeah, last time I did this I added the scripts, and this was a much better experience.

eregon:
It could be worth a shot to redo something like https://github.com/ruby/prism/pull/1295 to reduce the number of places in which the version is duplicated.
At least in templates that should be pretty safe.

PR #2420, opened 2024-02-15
Introduce `PRISM_HAS_MMAP` macro

hasumikin:
This is to know if the memory map interface is available in the target platform by using `_POSIX_MAPPED_FILES` and `_WIN32`

ref #2217

### DIFF
--- a/docs/build_system.md
+++ b/docs/build_system.md
@@ -73,6 +73,15 @@ and links to `libprism.a` (to avoid exporting symbols, so no conflict when insta
 
 TODO, similar to TruffleRuby.
 
+### Building prism for embedded system
+
+For instance, you can build a static library `libprism.a` targeting the Arm Cortex-M0+ embedded system by the commands below:
+
+* `templates/template.rb`
+* `CFLAGS="-mcpu=cortex-m0plus" make static CC=arm-none-eabi-gcc`
+
+The build process internally looks up `_POSIX_MAPPED_FILES` and `_WIN32` macros to determine whether the functions of the memory map are available on the target platform.
+
 ### Building prism from source as a C library
 
 All of the source files match `src/**/*.c` and all of the headers match `include/**/*.h`.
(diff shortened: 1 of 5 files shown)


kddnewton:
@hasumikin I think `_POSIX_MAPPED_FILES` is defined in `unistd.h`, so you probably need to include that before doing the check.

hasumikin:
@kddnewton 
Thank you.
My problem was fixed but another static check error was revealed that I couldn't figure out how to resolve :(

kddnewton:
Thanks @hasumikin you should be able to rebase now and it should be fixed.

My feedback on the PR, is that I think we should keep `PM_STRING_MAPPED` around, even if `PRISM_HAS_MMAP` is false. Maybe we should rename it to `PM_STRING_FILE` or something, but that way we could `malloc` and `read` the file into a `const uint8_t *` the same way you would `mmap` it.

kddnewton:
I'm also happy to make those changes after we merge this if you'd prefer. No problem! 😀

hasumikin:
@kddnewton 

> My feedback on the PR, is that I think we should keep PM_STRING_MAPPED around, even if PRISM_HAS_MMAP is false. Maybe we should rename it to PM_STRING_FILE or something, but that way we could malloc and read the file into a const uint8_t * the same way you would mmap it.

If I didn't get you wrong, we'd better create one in addition to `PM_STRING_MAPPED` and not include `_FILE` because (from the embedded system point of view,) we can not assume we have a filesystem.

How about `PM_STRING_GENERAL` for instance?

```c
/* pm_string.h */
typedef struct {
　(......)
    /** The type of the string. This field determines how the string should be freed. */
    enum {
    　(......)

        /** This string owns its memory, and should be freed using `pm_string_free`. */
        PM_STRING_OWNED,

#ifdef PRISM_HAS_MMAP
        /** This string is a memory-mapped file, and should be freed using `pm_string_free`. */
        PM_STRING_MAPPED,
#endif

        /** This is a general string, and should be taken care of by the user. */
        PM_STRING_GENERAL    // 👈
    } type;
} pm_string_t;
```

kddnewton:
Actually I realize now it should simply return a `PM_STRING_OWNED` if it doesn't have `mmap`, because that's effectively what it is. I'll merge this as-is and see what it takes to do that.

kddnewton:
Thank you for the contribution!

kddnewton (approved):

PR #2421, opened 2024-02-15
Update the error message format for Ruby 3.4

mame:
Ruby 3.4 will use a single quote instead of a backtick for a open quote.

https://bugs.ruby-lang.org/issues/16495

This is a draft PR because I don't know how to make the error message change between Ruby versions.

### DIFF
--- a/include/prism/diagnostic.h
+++ b/include/prism/diagnostic.h
@@ -8,6 +8,7 @@
 
 #include "prism/ast.h"
 #include "prism/defines.h"
+#include "prism/options.h"
 #include "prism/util/pm_list.h"
 
 #include <stdbool.h>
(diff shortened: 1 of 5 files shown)


mame:
FYI: I have already committed [a minimal patch](https://github.com/ruby/ruby/commit/d5c16ddfcb6c4cbb0209d66118ac23c7cb6c5149) directly to ruby/ruby to make the CI pass. Because this spec change has a large impact, more new failures may increase while waiting for your review, so I merged it in advance. I hope you understand.

kddnewton:
I would crease a new diagnostic for each of these and check the option in the parser before adding the diagnostic. That way diagnostic.c doesn't need to know about options.

eregon:
@kddnewton Wouldn't it be simpler to always use `'` in Prism?
Those messages only started to use backticks in https://github.com/ruby/prism/pull/2397 3 days ago.

mame:
@kddnewton Can I leave it to you? It makes sense that a good design of version-dependent message format is preferable. It would be difficult for me to design it well as I am not familiar with Prism. I am not particular about my ad-hoc patch.

kddnewton:
@mame absolutely! Happy to take over

kddnewton:
Thank you @mame for the PR! I've taken over in this one: https://github.com/ruby/prism/pull/2479

PR #2423, opened 2024-02-15
Aliases for Ripper Translation

noahgibbs:
Handle more aliases, including fully handling aliases.txt.

As part of handling more aliases, handle symbols that are keywords, constants or operations. Handle interpolated symbols. Handle global variable references.

<s>Better testing of prism ripper CLI and a test for it. I had broken it in renaming Translation::Ripper. Now it's tested.</s> It also prints better output when Prism or Ripper Translation gets an exception during parsing.

### DIFF
--- a/bin/prism
+++ b/bin/prism
@@ -220,7 +220,7 @@ module Prism
       source, filepath = read_source(argv)
 
       ripper = Ripper.sexp_raw(source)
-      prism = Prism::RipperCompat.sexp_raw(source)
+      prism = Prism::Translation::Ripper.sexp_raw(source) rescue :parse_error
 
       puts "Ripper:"
       pp ripper
--- a/test/prism/ripper_test.rb
+++ b/test/prism/ripper_test.rb
@@ -4,6 +4,50 @@ require_relative "test_helper"
 
 module Prism
   class RipperTest < TestCase
+    def truffleruby?
+      RUBY_ENGINE == "truffleruby"
+    end
+
+    def windows?
+      Gem.win_platform?
+    end
+
+    # Ripper produces certain ambiguous structures. For instance, it often
+    # adds an :args_add_block with "false" as the block meaning there is
+    # no block call. It can be hard to tell which of multiple equivalent
+    # structures it will produce. This method attempts to return a normalized
+    # comparable structure.
+    def normalized_sexp(parsed)
+      if parsed.is_a?(Array)
+        # For args_add_block, if the third entry is nil or false, remove it.
+        # Note that CRuby Ripper uses false for no block, while older JRuby
+        # uses nil. We need to do this for both.
+        return normalized_sexp(parsed[1]) if parsed[0] == :args_add_block && !parsed[2]
+
+        parsed.each.with_index do |item, idx|
+          if item.is_a?(Array)
+            parsed[idx] = normalized_sexp(parsed[idx])
+          end
+        end
+      end
+
+      parsed
+    end
+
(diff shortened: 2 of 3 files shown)


bin/prism
```diff
@@ -220,7 +220,7 @@ module Prism
       source, filepath = read_source(argv)
 
       ripper = Ripper.sexp_raw(source)
-      prism = Prism::RipperCompat.sexp_raw(source)
+      prism = Prism::Translation::Ripper.sexp_raw(source) rescue :parse_error
```
kddnewton, line 223: Can we specifically rescue `NotImplementedError` (I'm assuming that's what your handling)? I wouldn't want this to hide a different problem.

```suggestion
      prism =
        begin
          Prism::Translation::Ripper.sexp_raw(source)
        rescue NotImplementedError
          :parse_error
        end
```

test/prism/ripper_test.rb
```diff
@@ -4,6 +4,42 @@
 
 module Prism
   class RipperTest < TestCase
+    # Ripper produces certain ambiguous structures. For instance, it often
```
kddnewton, line 7:
```suggestion
    private

    # Ripper produces certain ambiguous structures. For instance, it often
```

test/prism/ripper_test.rb
```diff
@@ -4,6 +4,42 @@
...
+    # adds an :args_add_block with "false" as the block meaning there is
+    # no block call. It can be hard to tell which of multiple equivalent
+    # structures it will produce. This method attempts to return a normalized
+    # comparable structure.
+    def normalized_sexp(parsed)
```
kddnewton, line 12: I'm fine having this for now, but we'll definitely want to get rid of this before we close this ticket out.

kddnewton (changes requested):
A couple of small changes, but overall looks good.

test/prism/ripper_test.rb
noahgibbs: I'll definitely try. I'm having trouble figuring out when Ripper does and doesn't emit this, even looking through parse.y.

bin/prism
noahgibbs: I mostly use it to make sure that if something's not working yet in Translation::Ripper that I still see CRuby's Ripper output. When it says "parse_error" that should only hide another problem if I'm expecting a parse error. But if I'm midway through making a modification and it might get an error, this means that "prism ripper" is still useful. I guess I could add a mode that doesn't actually use Prism's ripper and just prints out CRuby's output without trying Prism?

noahgibbs:
Found a minor TruffleRuby Ripper incompatibility. Turning off that test on Truffle, filed a bug: https://github.com/oracle/truffleruby/issues/3457

noahgibbs:
Turning off the command line test on Windows. Aside from issues like not being able to just run bin/prism (need "ruby bin/prism" or rough equivalent) there's also not a great "which" equivalent.

noahgibbs:
Yeah, I'm just removing the command line test. That's annoying, I wanted to add it because the utility broke. But I'm not figuring out how to locate bin/prism to run it reliably, even if I exclude Windows. I can find it in dev checkouts (__dir__ + ../../bin/prism) and in installed gems (`which prism`). But in test-all it still doesn't seem to be there.

bin/prism
kddnewton: Ahh that's fine. We can keep it for now, I just want this to go away eventually.

kddnewton (approved):

Issue #2402, opened 2024-02-13
Optimize the FFI backend


eregon:
Currently it's slower.
I parse Prism's `lib/**/*.rb` files as a quick benchmark.
```
$ ruby -Ilib -rprism -rbenchmark -e '10.times { p Benchmark.realtime { Dir.glob("lib/**/*.rb") { |f| Prism.parse_file(f) } } }'
0.1726086970011238
0.1321143770001072
0.13798127100017155
0.147119974000816
0.14450125900111743
0.13631621999957133
0.128785977998632
0.15200302700031898
0.14583392900021863
0.14046238399896538

$ ruby --yjit -Ilib -rprism -rbenchmark -e '10.times { p Benchmark.realtime { Dir.glob("lib/**/*.rb") { |f| Prism.parse_file(f) } } }'
0.16340846000093734
0.11280311600057757
0.11687472400080878
0.11827147000076366
0.1305678140015516
0.11364481000055093
0.10130149700125912
0.10992031799833057
0.11991112200121279
0.13184594300037134

$ PRISM_FFI_BACKEND=true ruby -Ilib -rprism -rbenchmark -e '10.times { p Benchmark.realtime { Dir.glob("lib/**/*.rb") { |f| Prism.parse_file(f) } } }'
0.3652132739989611
0.3220108080022328
0.3004020099979243
0.3233778670000902
0.30800675899808994
0.31912518099852605
0.3041645149969554
0.29707527800201206
0.33191245600028196
0.29412771300121676

$ PRISM_FFI_BACKEND=true ruby --yjit -Ilib -rprism -rbenchmark -e '10.times { p Benchmark.realtime { Dir.glob("lib/**/*.rb") { |f| Prism.parse_file(f) } } }'
0.2793479979991389
0.21698390300298342
0.18590097600099398
0.20671838999987813
0.1923178380020545
0.1890159450013016
0.2051944070008176
0.18629215799956
0.1725655469999765
0.21039302300050622
```

```
$ PRISM_FFI_BACKEND=true ruby --yjit -Ilib -rprism -rbenchmark -rstackprof/autorun -e '10.times { p Benchmark.realtime { Dir.glob("lib/**/*.rb") { |f| Prism.parse_file(f) } } }'
0.27546833399901516
0.2049026849999791
0.1797586089996912
0.18482352300270577
0.19821838599818875
0.1960418029993889
0.18537640199792804
0.2127407010011666
0.18586592199790175
0.17573103299946524
[...]

PR #2426, opened 2024-02-15
Avoid extra String copies in the FFI backend

eregon:
* For Prism.parse_file the file contents would be read as native, then converted to a Ruby String, then converted to a native String for pm_serialize_parse().
* Refactor the logic to always use a pm_string for the source code and pass that to other native functions.
* Part of #2402


eregon:
Mmh, using `pm_string_constant_init` doesn't quite work on JRuby because the native String is only guaranteed alive during that call and not after. I'll use a FFI Pointer and copy the String there instead then.

kddnewton (approved):
Looks good!

PR #2427, opened 2024-02-15
Do not abort `bundle exec rake` if there are extra files

eregon:
This check turns out to be annoying in practice, I think it's fine it just prints/warns without exiting.
Not being able to run the test suite just because there are extra files seems rather extreme.

FWIW my extra files are flamegraphs and profile data from optimizing Prism (#2402).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -7,7 +7,7 @@ require "rake/clean"
 task compile: :make
 task compile_no_debug: :make_no_debug
 
-task default: [:check_manifest, :compile, :test]
+task default: [:compile, :test]
 
 require_relative "templates/template"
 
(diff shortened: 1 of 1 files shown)


kddnewton:
I'm generally okay with this change but I definitely want it to exit when running `rake build`. Can you make it do that? It has caught me a couple of times before releasing.

eregon:
OK, check_manifest is already done before `rake build`:
https://github.com/ruby/prism/blob/20b06022135ae1d22fafaceddaf0b94e230a6f55/Rakefile#L34-L35

So I think we can simply remove it from the rake default task then.

Rakefile
```diff
@@ -32,7 +32,7 @@ task make_no_debug: [:templates] do
 end
 
 # decorate the gem build task with prerequisites
-task build: [:templates, :check_annotations, :check_manifest]
+task build: [:check_manifest, :templates, :check_annotations]
```
eregon, line 35: I changed the order so `:check_manifest` is done first and fails quickly & early, before compiling stuff, which would make the error less obvious.

kddnewton (approved):
Nice

PR #2432, opened 2024-02-15
Make location methods thread-safe

eregon:
* Before it could result in NoMethodError if multiple threads were calling location methods: https://gist.github.com/eregon/b78b7f266d7ee0a278a389cfd1782232


kddnewton (approved):
Nice!

PR #2433, opened 2024-02-15
Visitor#visit should be called for every node

eregon:
* This way #visit is the "default visit" method and sees every node.
* This was no longer the case since 2e6baa3f19ab7d82c9980b5398cd3467fc26bf22.
* As a consequence EnsureEveryNodeOnceInAST was only visiting ProgramNode for `visitor.visit(ast)` and no nodes at all for `ast.accept(visitor)`.

### DIFF
--- a/templates/lib/prism/visitor.rb.erb
+++ b/templates/lib/prism/visitor.rb.erb
@@ -12,12 +12,12 @@ module Prism
 
     # Visits each node in `nodes` by calling `accept` on each one.
     def visit_all(nodes)
-      nodes.each { |node| node&.accept(self) }
+      nodes.each { |node| visit(self) }
     end
 
     # Visits the child nodes of `node` by calling `accept` on each one.
     def visit_child_nodes(node)
-      node.compact_child_nodes.each { |node| node.accept(self) }
+      node.compact_child_nodes.each { |node| visit(node) }
     end
   end
 
@@ -40,6 +40,8 @@ module Prism
   #       end
   #     end
   #
+  #     ast = Prism.parse(code).value
+  #     FooCalls.new.visit(ast)
   class Visitor < BasicVisitor
     <%- nodes.each_with_index do |node, index| -%>
 <%= "\n" if index != 0 -%>
--- a/test/prism/desugar_compiler_test.rb
+++ b/test/prism/desugar_compiler_test.rb
@@ -75,7 +75,7 @@ module Prism
       ast = Prism.parse(source).value.accept(DesugarCompiler.new)
       assert_equal expected, ast_inspect(ast.statements.body.last)
 
-      ast.accept(EnsureEveryNodeOnceInAST.new)
+      EnsureEveryNodeOnceInAST.new.visit(ast)
     end
 
     def assert_not_desugared(source, reason)
--- a/test/prism/visitor_test.rb
+++ b/test/prism/visitor_test.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+require_relative "test_helper"
+
+module Prism
+  class VisitorTest < TestCase
+    def test_visitor_visit_method_called_for_every_node
+      visited = []
+      visitor = Class.new(Visitor) {
+        define_method :visit do |node|
+          visited << node
+          super(node)
+        end
+      }.new
+      tree = Prism.parse("1 + 2").value
+
+      all_nodes = []
+      collect_all_children = -> node {
+        all_nodes << node
+        node.compact_child_nodes.each { |child| collect_all_children.call(child) }
+      }
+      collect_all_children.call(tree)
+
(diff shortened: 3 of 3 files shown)


eregon:
I'd like to get some feedback on this PR.
I'm not sure it's the right approach.

Maybe we should use a similar design to https://github.com/ruby/prism/blob/main/templates/java/org/prism/AbstractNodeVisitor.java.erb ?
That is have a `default_visit` abstract method in BasicVisitor,
and Visitor would define all `def visit_*_node(node) = default_visit(node)`.
On Visitor, `default_visit` could be `def default_visit(node) = visit_child_nodes(node)` but could be overwritten.
Then if one wants a visitor called for every node, they can just overwrite `default_visit`.
And also this pattern works nicely so that `default_visit` is only called if a specific `visit_foo_node` is not defined.
That's better than the `visit` in this PR, which is called for every node, before the `accept`.

eregon:
I'm also unsure if the BasicVisitor/Visitor separation is useful.
I think with `default_visit` it's unnecessary because one could just `def default_visit(node) = raise "not yet handled"` to achieve the same check that all visit methods are implemented.
Or maybe such a check is better done at code loading time than runtime and if so then using metaprogramming seems best (e.g. compare `.instance_methods` of the visitor and Visitor, I'm doing this in TruffleRuby).

kddnewton:
We moved away from this approach purposefully because it was hurting performance in ruby-lsp. This extra method call can really add up on big trees. I'm okay keeping this PR open if you want to fix the test, but we can't go back to calling `visit` on every node.

kddnewton (changes requested):
(Just marking this as "request changes")

eregon:
Alright, makes sense.

WDYT about the `default_visit` approach described in https://github.com/ruby/prism/pull/2433#issuecomment-1947475171 and the next comment?

kddnewton:
I just benchmarked it and it appears to lose about 20% speed with the `default_visit` approach. I don't want to pay that penalty on the default visitor. I would be open to a new kind of visitor that has an overridable `default_visit` method if you wanted to implement one. Something like `CentralizedVisitor` or something like that.

eregon:
Could you share your benchmark and your changes (even if just some snippet)?
Did you measure with the visitor doing nothing, or some specific task?

kddnewton:
Here's the benchmark:

```ruby
require "bundler/setup"
require "benchmark/ips"
require "prism"

class MyVisitor < Prism::Visitor
  attr_reader :count

  def initialize
    @count = 0
  end

  def visit_regular_expression_node(node)
    @count += 1
    super
  end
end

class MyOtherVisitor < Prism::OtherVisitor
  attr_reader :count

  def initialize
    @count = 0
  end

  def visit_regular_expression_node(node)
    @count += 1
    super
  end
end

ast = Prism.parse_file(File.expand_path("lib/prism/node.rb", __dir__)).value

Benchmark.ips do |x|
  x.report("Visitor") { ast.accept(MyVisitor.new) }
  x.report("OtherVisitor") { ast.accept(MyOtherVisitor.new) }
  x.compare!
end
```

Here's the diff for the template:

```ruby
diff --git a/templates/lib/prism/visitor.rb.erb b/templates/lib/prism/visitor.rb.erb
index 04156cc7a9..ad5d7505a3 100644
--- a/templates/lib/prism/visitor.rb.erb
+++ b/templates/lib/prism/visitor.rb.erb
@@ -47,4 +47,14 @@ module Prism
     alias visit_<%= node.human %> visit_child_nodes
     <%- end -%>
   end
+
+  class OtherVisitor < BasicVisitor
+    def default_visit(node) = visit_child_nodes(node)
+
+    <%- nodes.each_with_index do |node, index| -%>
+<%= "\n" if index != 0 -%>
+    # Visit a <%= node.name %> node
+    def visit_<%= node.human %>(node) = default_visit(node)
+    <%- end -%>
+  end
 end
```

Here are the results:

```
ruby 3.4.0dev (2024-02-07T11:34:48Z master 9ebaf7a8ef) [arm64-darwin23]
             Visitor     7.000 i/100ms
        OtherVisitor     6.000 i/100ms
             Visitor     72.283 (± 1.4%) i/s -    364.000 in   5.036294s
        OtherVisitor     62.040 (± 1.6%) i/s -    312.000 in   5.029632s

Comparison:
             Visitor:       72.3 i/s
        OtherVisitor:       62.0 i/s - 1.17x  slower
```

eregon:
OK, thanks for sharing and benchmarking that.
Given I don't have an immediate need for `default_visit` let's close this.

A simple workaround for a visitor visiting every node without implementing each `visit_foo_node` method is to use a recursive method like
```ruby
def visitor(node)
  do_sth(node)
  node.compact_child_nodes.each { visitor(_1) }
end
```

PR #2435, opened 2024-02-16
Fix lexing of `foo!` when it's a first thing to parse

maxprokopiev:
Running `Prism::Translation::Parser` on a file which starts with a method call with bang ends up with an error:

```
Error: test_single_method_call_with_bang.txt(Prism::ParserTest): NoMethodError: undefined method `[]' for nil:NilClass
lib/prism/translation/parser/lexer.rb:292:in `to_a'
lib/prism/translation/parser.rb:159:in `build_tokens'
lib/prism/translation/parser.rb:89:in `tokenize'
prism/parser_test.rb:111:in `assert_equal_parses'
test/prism/parser_test.rb:89:in `block (2 levels) in <class:ParserTest>'
```

because `tokens` is an empty array in that case

I've added a separate fixture for it to indicate that the error is happening only if it's a first thing to parse, but I also can move `a!` in the `fixtures/method_calls.txt` to the top instead.


kddnewton (approved):
Thanks @maxprokopiev!

PR #2439, opened 2024-02-16
Ripper Translation: much better begin/rescue/ensure/end

noahgibbs:
It may be easiest to review this as two separate commits. The first commit separates out a RipperCompiler class, which means moving a lot of methods. The second commit adds a lot of begin/rescue/etc functionality.

The file separation seems like a good idea regardless. But I did it right now because this PR needs to check multiple kinds of nodes (BeginNode, EnsureNode, BlockNode) to see if a semicolon follows immediately after the keyword. Ripper prepends a :void_stmt in those cases so we need to know. Separating the RipperCompiler roughly follows the structure of the Translation::Parser::Compiler, including giving it responsibility for the source code buffer and offsets. Currently Translation::Ripper's functionality there is simpler because this is all it needs.

This PR aims toward using the begin_rescue and begin_ensure fixtures but doesn't quite get there. Both fixtures use additional non-begin/rescue/ensure/end functionality that Translation::Ripper doesn't support yet.


kddnewton:
I like this a lot, but can we make it so that `Prism::Translation::Ripper` uses `Prism::Translation::Ripper::Compiler` instead of inheriting from it? As in, `Prism::Translation::Ripper` would define parse as:

```ruby
def parse
  result.magic_comments.each do |magic_comment|
    on_magic_comment(magic_comment.key, magic_comment.value)
  end
  if error?
    result.errors.each do |error|
      on_parse_error(error.message)
    end
    nil
  else
    result.value.accept(Compiler.new(self))
  end
end
```

That way the compiler itself doesn't have knowledge of the various bits and bobs of ripper state.

noahgibbs:
I'll see what I can do there. There's some oddity with the way SexpBuilder and SexpBuilderPP define on_ methods that may take a bit of dancing around. But it should be pretty straightforward to get Translation::Ripper not to inherit from it even if those two perhaps still do.

noahgibbs:
I've put together a quick additional commit that does that. Does it do what you want?

noahgibbs:
Hm, no, that's not what you were explaining in the comment. If Translation::Ripper defines on_* methods, that makes it a lot harder to call them in RipperCompiler -- and it does, pretty extensively, in order to emit Sexps. We could avoid a lot of the inheritance by passing in a sort of "Ripper emitter" object into RipperCompiler. Then SexpBuilder wouldn't need to inherit from it -- SexpBuilder would be (or define) a custom emitter, but wouldn't need to define on_* methods that RipperCompiler needed to call.

Maybe I could define a full set of on_* methods on RipperCompiler that instead calls through to the emitter? That would keep it from touching a huge number of lines. In the mean time I'll revert that third commit.

kddnewton (approved):

PR #2451, opened 2024-02-18
Split up comments between leading and trailing

kddnewton:
Also make them lazy to allocate the array, and also expose ParseResult#encoding.

### DIFF
--- a/lib/prism/parse_result.rb
+++ b/lib/prism/parse_result.rb
@@ -21,6 +21,12 @@ module Prism
       @offsets = offsets # set after parsing is done
     end
 
+    # Returns the encoding of the source code, which is set by parameters to the
+    # parser or by the encoding magic comment.
+    def encoding
+      source.encoding
+    end
+
     # Perform a byteslice on the source code using the given byte offset and
     # byte length.
     def slice(byte_offset, length)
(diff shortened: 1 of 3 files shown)


lib/prism/parse_result.rb
```diff
@@ -108,16 +114,46 @@ class Location
...
+
+    # Attach a comment to the trailing comments of this location.
+    def trailing_comment(comment)
+      trailing_comments << comment
+    end
+
+    # Returns all comments that are associated with this location (both leading
+    # and trailing comments).
+    def comments
+      (@leading_comments || []).concat(@trailing_comments || [])
```
noahgibbs, line 156: If @leading_comments and @trailing_comments are both non-empty, I think this is going to permanently add the trailing comments into the leading comments array. Concat isn't just +, it actually changes the first array.

lib/prism/parse_result.rb
```diff
@@ -108,16 +114,46 @@ class Location
...
       @source = source
       @start_offset = start_offset
       @length = length
-      @comments = []
+
+      # These are used to store comments that are associated with this location.
+      # They are initialized to `nil` to save on memory when there are no
+      # comments to be attached and/or the comment-related APIs are not used.
+      @leading_comments = nil
+      @trailing_comments = nil
```
eregon, line 128: I think in terms of memory usage, it's best to not assign these at all.
Assigning, even to `nil`, can cause extra memory to be used, because it's considered like any other value and e.g. the Shape will have `@leading_comments at OFFSET` and then in values[OFFSET] we'll have that `nil`.

kddnewton: This is going to have a consistent shape, so the callsites to `leading_comments`, `trailing_comments`, and `comments` don't become polymorphic, depending on call order. The `nil` itself is just going to be a tagged pointer, so it'll be a minimal impact.

eregon: Right, a monomorphic shape is also important.

For curiosity I measured it on `ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [x86_64-linux]` and there is a difference:
```
with this change:
$ bundle exec ruby -Ilib -rprism -robjspace -e 'loc = Prism.parse("1").value.location; p loc; p ObjectSpace.memsize_of(loc)'
#<Prism::Location @start_offset=0 @length=1 start_line=1>
80

without these 2 lines:
$ bundle exec ruby -Ilib -rprism -robjspace -e 'loc = Prism.parse("1").value.location; p loc; p ObjectSpace.memsize_of(loc)'
#<Prism::Location @start_offset=0 @length=1 start_line=1>
40
```

So it's a trade-off between memory footprint and monomorphic shape.
Given Location objects are lazy now the footprint likely matters less.

PR #1777, opened 2023-11-07
Provide Parameters#signature for mirroring Method#parameters

kddnewton:
Nice to have for client libraries.

@seven1m I thought you might be able to use this to implement `Method#parameters`

PR #2455, opened 2024-02-19
Enable ParametersSignatureTest on TruffleRuby

andrykonchin:
The tests now pass on TruffleRuby (except `test_key_ordering`). Follow-up of #1777.


eregon:
@kddnewton The tests fail because they use `truffleruby`, so the latest release, 23.1.
2 possibilities:
* Switch to truffleruby-head, then this PR should be green. It would also have the advantage to test compatibility with the latest truffleruby changes.
* Wait for 24.0 which comes out on April 16.

Which do you prefer?

kddnewton:
I'd prefer to wait for a stable release. I'll close this for now, and set a reminder to reopen it on April 16th.

eregon:
Looks like I got the date wrong somehow when copying it here, 24.0.0 was released on March 19, so now CI passes.

kddnewton (approved):

Issue #2374, opened 2024-02-07
Incompatibility with squiggly heredoc in `Prism::Translation::Parser`


koic:
## Steps to reproduce the problem

This incompatibility occurs when using string interpolation within squiggly heredoc. It can be reproduced with Prism 0.21.0 and the following code:

```ruby
require 'prism'
require 'prism/translation/parser'
require 'parser/ruby33'

string_interpolation_in_squiggly_heredoc = <<~'RUBY'
  <<~HEREDOC
    #{x}
  HEREDOC
RUBY

p Prism::Translation::Parser.parse(string_interpolation_in_squiggly_heredoc).loc.heredoc_body.source
p Parser::Ruby33.parse(string_interpolation_in_squiggly_heredoc).loc.heredoc_body.source
```

Both expect leading whitespace to be present, but the result from `Prism::Translation::Parser` does not have any whitespace:

```console
$ ruby example.rb
"\#{x}\n"
"  \#{x}\n"
```

## Expected

It expects the presence of leading whitespace, similar to `Parser`, suggesting an expectation for compatibility.

## Actual

In `Prism::Translation::Parser`, there is no leading whitespace, indicating an incompatibility with `Parser::Ruby33`.

## Context

During the process of attempting Prism integration with RuboCop, this regression was discovered with `Layout/HeredocIndentation` cop.

kddnewton:
Ahh I think I misunderstood. Parser doesn't strip common leading white space?

kddnewton:
Oh I misunderstood, the locations are incorrect. I see.

PR #2466, opened 2024-02-20
Draft: Parser heredocs

noahgibbs:
I started work on #2374. It's a small thing - the "source" location is a bit off on tilde-heredocs (a.k.a. squiggly heredocs). But it looks like we're also not lexing them quite like Parser does, and we do limited (or no) checks on some heredoc-related fixtures, especially for lexing (e.g. look at skip_all and skip_tokens in parser_test.rb). Our heredoc support for Translation::Parser is kind of limited and doesn't 100% match up.

I've been building better debugging support, such as unit tests on inline code and teaching "bin/prism parser" to emit source locations with its sexps. And I've improved the lexer a bit to match some source locations better. With tilde heredocs that means keeping a stack of string begin/ends, for instance, so that we can trim leading whitespace since Parser does that for lexing, not just parsing.

But I think maybe we don't care as much about the lexer. I had originally thought I'd need to fix that to fix the parser source locations. Now I think I was wrong about that.

This is a little rough. I'm just putting this up to show where I'm at and what I'm thinking for right now. Parser::Compiler#visit_heredoc is definitely going to need more work. I'm getting some basic tests working (various kinds of simple heredocs, nested heredocs, etc.) directly in parser_test.rb. And I can turn more fixtures on once more stuff works.

### DIFF
--- a/bin/prism
+++ b/bin/prism
@@ -225,13 +225,31 @@ module Prism
     # bin/prism parser [source]
     def parser(argv)
       require "parser/current"
+
+      with_locations = argv.delete("--source-locations")
+
       source, filepath = read_source(argv)
 
+      buffer = Parser::Source::Buffer.new(filepath, 1)
+      buffer.source = source
+
       puts "Parser:"
-      pp Parser::CurrentRuby.parse(source, filepath)
+      ast, _comments, _tokens = Parser::CurrentRuby.default_parser.tokenize(buffer)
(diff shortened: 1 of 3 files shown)


noahgibbs:
What's in this PR isn't the right solution to the issue in question, though some of it's still useful debug infrastructure. I'll reopen a new issue later.

PR #2469, opened 2024-02-20
Fallback for pm_string_mapped_init

kddnewton:
@hasumikin will this work, or will referencing `FILE` at all break compiling on embedded?

### DIFF
--- a/src/util/pm_string.c
+++ b/src/util/pm_string.c
@@ -136,10 +136,44 @@ pm_string_mapped_init(pm_string_t *string, const char *filepath) {
     *string = (pm_string_t) { .type = PM_STRING_MAPPED, .source = source, .length = size };
     return true;
 #else
-    (void) string;
-    (void) filepath;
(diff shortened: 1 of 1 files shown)


hasumikin:
@kddnewton
It will break. You can not assume an embedded system has `FILE`.
Some embedded systems have `FILE` but others don't.

Letting users read the entire Ruby script (whether it's a "file" or not) is enough, I think

kddnewton:
@hasumikin unfortunate. I was hoping to provide a fallback for posix platforms that don't have mmap. Do you think checking for either `_POSIX_VERSION` or `_FILE_OFFSET_BITS` would work?

hasumikin:
@kddnewton 
Sorry but I have no idea.
I don't think there is a robust way to know a POSIX that has no MMAP...

kddnewton:
No worries, I'll close this until it becomes a real use-case.

Issue #2145, opened 2024-01-09
eval encoding with magic comment still not correct


enebo:
This is something not quite right even after #1639:

```ruby
code = "# encoding: UTF-8\na = 'é'; puts a; a.encoding".b
p code.encoding
p eval(code)
```
This prints out:

```text
#<Encoding:ASCII-8BIT>
é
#<Encoding:ASCII-8BIT>
```

enebo:
Looking in debugger it appears like perhaps encoding from Loader is getting set to ASCII-8BIT and not pragma encoding?

eregon:
And it also works fine with Prism.parse on TruffleRuby (using FFI):
```
$ ruby -Ilib -rprism -e 'r = Prism.parse "# encoding: iso-8859-1\na = %{é}; puts a; a.encoding", encoding: Encoding::BINARY; p r.source.source.encoding'
#<Encoding:ISO-8859-1>
$ ruby -Ilib -rprism -e 'r = Prism.parse "a = %{é}; puts a; a.encoding", encoding: Encoding::BINARY; p r.source.source.encoding'           
#<Encoding:ASCII-8BIT>
```

Although `encoding: "BINARY"` doesn't work for the FFI backend, it needs `encoding: Encoding::BINARY`, probably we should fix that.

So not quite sure where is the bug, but it doesn't seem to be in Prism C code because serialization gets it right too.
Another way to show it:
```
$ ruby -Ilib -rprism -e 'p Prism.dump "# encoding: iso-8859-1\na = %{é}; puts a; a.encoding", encoding: Encoding::BINARY'
"PRISM\x00\x13\x00\x00\nISO-8859-1\x02\x01\x00\x00\x16\x01\x02\b\f\n\x00\x00\x00r\x00\x00\x00\x03w\x17\x1D\x01\x01\x89\x17\x1D\x03`\x17\t\x01\x00\x17\x01\x8A\e\x05\x00\x01\e\x02\x1D\x02\x01\x1F\x01\x01\x1D\x02\x19\x01\x13\"\x06\b\x00\x00\x02\x01\"\x04\x00\x05'\x01\x00\x01^'\x01\x01\x00\x00\x00\x13*\n\x00^*\x01\x01\x00\x01+\x01\x03\x01,\b\x00\x00\x00\x00\x17\x00\x00\x00\x01\x00\x00\x00\"\x00\x00\x00\x04\x00\x00\x00,\x00\x00\x00\b\x00\x00\x00\x00"
```

[...]

PR #2470, opened 2024-02-20
Fix eval encoding switching to UTF-8

kddnewton:
Fixes https://github.com/ruby/prism/issues/2145

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -6605,12 +6605,9 @@ parser_lex_magic_comment_encoding_value(pm_parser_t *parser, const uint8_t *star
     const pm_encoding_t *encoding = pm_encoding_find(start, end);
 
     if (encoding != NULL) {
-        if (encoding != PM_ENCODING_UTF_8_ENTRY) {
-            parser->encoding = encoding;
-            parser->encoding_changed = true;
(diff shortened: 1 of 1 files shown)


src/prism.c
```diff
@@ -6605,12 +6605,9 @@ parser_lex_magic_comment_encoding_value(pm_parser_t *parser, const uint8_t *star
...
     if (encoding != NULL) {
-        if (encoding != PM_ENCODING_UTF_8_ENTRY) {
-            parser->encoding = encoding;
-            parser->encoding_changed = true;
-            if (parser->encoding_changed_callback != NULL) parser->encoding_changed_callback(parser);
-        }
-
+        parser->encoding = encoding;
+        parser->encoding_changed = (encoding != PM_ENCODING_UTF_8_ENTRY);
+        if (parser->encoding_changed_callback != NULL) parser->encoding_changed_callback(parser);
```
noahgibbs, line 6610: This will call the callback even if the encoding is UTF-8. I'm assuming that's fine.

kddnewton: Good point! https://github.com/ruby/prism/pull/2474

PR #2475, opened 2024-02-21
Translation::Parser should process warnings, not just errors

### LINKED ISSUE #2454: `Prism::Translation::Parser` incompatibility with `Parser::Diagnostic`
An incompatibility issue in handling Ruby warnings has been identified between the Parser gem and Prism.

Here is an example of the difference in emulation for the following warnings:

```console
$ ruby -cwe "do_something *array"
-e:1: warning: `*' interpreted as argument prefix
Syntax OK
```

Below is the Gemfile common to expected and actual:

```ruby
$ cat Gemfile
# frozen_string_literal: true

source "https://rubygems.org"

gem 'prism'
gem 'rubocop'
```

## Expected Behavior

There is an expectation for a `diagnostics` array to return, similar to that provided by Parser gem:

```console
$ bundle exec ruby -rprism/translation/parser/rubocop -ve "p RuboCop::AST::ProcessedSource.new('do_something *array', 3.3).diagnostics"
[#<Parser::Diagnostic:0x000000011f05c7b8 @level=:warning, @reason=:ambiguous_prefix, @arguments={:prefix=>"*"}, @location=#<Parser::Source::Range (string) 13...14>, @highlights=[]>]
```

## Actual Behavior

When using Prism as the backend, the `diagnostics` return value becomes an empty array:

```console
$ bundle exec ruby -rprism/translation/parser/rubocop -ve "p RuboCop::AST::ProcessedSource.new('do_something *array', 80_82_73_83_77.33).diagnostics"
[]
```

It is expected to return the same result as `3.3`, even when the argument is `80_82_73_83_77.33`.

## Background

This difference in behavior was detected while I was working on the Prism integration, due to a failing test within RuboCop AST:
https://github.com/rubocop/rubocop-ast/blob/v1.30.0/spec/rubocop/ast/processed_source_spec.rb#L189-L197

noahgibbs:
We weren't registering warnings, just errors.

As a side effect, this will cause "bin/prism parser" to die when it sees a warning because we're running the Parser gem in that mode. I think that's probably fine, but it seems worth mentioning.

Should fix #2454

### DIFF
--- a/lib/prism/translation/parser.rb
+++ b/lib/prism/translation/parser.rb
@@ -17,9 +17,9 @@ module Prism
         attr_reader :message
 
         # Initialize a new diagnostic with the given message and location.
-        def initialize(message, level, location)
+        def initialize(message, level, reason, location)
           @message = message
-          super(level, :prism_error, {}, location, [])
+          super(level, reason, {}, location, [])
         end
       end
 
@@ -106,6 +106,12 @@ module Prism
         true
       end
 
+      # This is a hook to allow consumers to disable some warnings if they don't
+      # want them to block creating the syntax tree.
+      def valid_warning?(warning)
+        true
+      end
+
       # If there was a error generated during the parse, then raise an
       # appropriate syntax error. Otherwise return the result.
       def unwrap(result, offset_cache)
@@ -113,13 +119,13 @@ module Prism
           next unless valid_error?(error)
 
           location = build_range(error.location, offset_cache)
-          diagnostics.process(Diagnostic.new(error.message, :error, location))
+          diagnostics.process(Diagnostic.new(error.message, :error, :prism_error, location))
         end
         result.warnings.each do |warning|
           next unless valid_warning?(warning)
 
           location = build_range(warning.location, offset_cache)
-          diagnostics.process(Diagnostic.new(warning.message, :warning, location))
+          diagnostics.process(Diagnostic.new(warning.message, :warning, :prism_warning, location))
         end
 
         result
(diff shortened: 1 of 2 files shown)


lib/prism/translation/parser.rb
```diff
@@ -17,9 +17,9 @@ class Diagnostic < ::Parser::Diagnostic
         attr_reader :message
 
         # Initialize a new diagnostic with the given message and location.
-        def initialize(message, location)
+        def initialize(message, level, location)
           @message = message
-          super(:error, :prism_error, {}, location, [])
+          super(level, :prism_error, {}, location, [])
```
kddnewton, line 22: Can we accept a type here as well so that the warnings aren't `:prism_error`?

lib/prism/translation/parser.rb
```diff
@@ -113,7 +113,13 @@ def unwrap(result, offset_cache)
           next unless valid_error?(error)
 
           location = build_range(error.location, offset_cache)
-          diagnostics.process(Diagnostic.new(error.message, location))
+          diagnostics.process(Diagnostic.new(error.message, :error, location))
+        end
+        result.warnings.each do |warning|
+          next unless valid_error?(warning)
```
kddnewton, line 119: Can we introduce a `valid_warning?` here for this so that people can override that as well?

```suggestion
          next unless valid_warning?(warning)
```

kddnewton (changes requested):

kddnewton (approved):

Issue #2465, opened 2024-02-20
[Feature][Ruby API] Implement `==` (value comparison) for `Prism::Node`


exterm:
For testing, it would be great if we could compare ASTs.

It seems that value equality for nodes is just not defined. `Prism::Node#==` is inherited from BasicObject so it's only true if the two compared objects are the same object (object identity comparison). https://ruby-doc.org/3.2.2/BasicObject.html#method-i-3D-3D

It would be great to have value equality for nodes so we could do things like this:

```ruby
test "extract_argument_nodes extracts all argument nodes" do
  source = "class Entry; has_many :clowns, class_name: 'Order'; end"

  expected = Prism.parse(source).value.statements.body.first.body.body.first.arguments
  actual = extract_argument_nodes("class Entry; #{call}; end")
  
  assert_equal [expected], actual
end
```

Or even this: (ignoring node locations!)

```ruby
test "extract_argument_nodes extracts all argument nodes" do
  call = "has_many :clowns, class_name: 'Order'"

  expected = Prism.parse(call).body.first
  actual = extract_argument_nodes("class Entry; #{call}; end")
  
  assert_equal [expected], actual
end
```

From https://github.com/ruby/prism/discussions/2448

eregon:
I'm unsure if it's right to have Node#== ignores location fields, because then it could consider "different nodes" to be ==, as in they are in different position in the source and also they might be completely unrelated logically (e.g. a RequiredParameterNode with a given name for one method vs for another method).
There is also the question of `eql?`, structural comparison ignoring locations there feels wrong.

You might have seen https://github.com/ruby/prism/blob/71991592d4cf97f64b311485f5985c766022d0f4/test/prism/test_helper.rb#L26
Maybe such a method should be exposed by the Prism gem?

kddnewton:
I'm okay having a method for this, but not `#==`. Having that ignore location entirely would worry me for what consumers would accidentally find to be equal. I _would_ support this being `#===`, as that's closer semantics to "match"/"similar", which is what this request seems like it's asking for. That has the nice benefit of allowing you to use it in case expressions, as in your example:

```ruby
source = "class Entry; has_many :clowns, class_name: 'Order'; end"

case extract_argument_nodes("class Entry; #{call}; end")
when Prism.parse(source).value.statements.body.first.body.body.first.arguments
  # matched
end
```

exterm:
[...]

PR #2476, opened 2024-02-21
implement value equality and case equality on nodes

exterm:
Fix https://github.com/ruby/prism/issues/2465

I was hoping that deconstruct would be sufficiently specific, but it doesn't look like it (at least it doesn't include `value` for `IntegerNode`).

Looking at lib/prism/node_ext.rb, #value seems to be the only relevant property that is not a field.

It is derived from slice, but I don't want to do an exact comparison on slice since that would make formatting significant for the comparison.

So now I can either add value to the comparison in Node through a respond_to, or I can make it part of node_ext... which seems better. You can see both options in this PR (see the commits)

### DIFF
--- a/templates/lib/prism/node.rb.erb
+++ b/templates/lib/prism/node.rb.erb
@@ -219,10 +219,10 @@ module Prism
     def deconstruct_keys(keys)
       { <%= (node.fields.map { |field| "#{field.name}: #{field.name}" } + ["location: location"]).join(", ") %> }
     end
-
     <%- node.fields.each do |field| -%>
+
     <%- if field.comment.nil? -%>
-    # <%= "private " if field.is_a?(Prism::Template::FlagsField) %>attr_reader <%= field.name %>: <%= field.rbs_class %>
+    # <%= "protected " if field.is_a?(Prism::Template::FlagsField) %>attr_reader <%= field.name %>: <%= field.rbs_class %>
     <%- else -%>
     <%- field.each_comment_line do |line| -%>
     #<%= line %>
@@ -248,9 +248,8 @@ module Prism
       end
     end
     <%- else -%>
-    attr_reader :<%= field.name -%><%= "\n    private :#{field.name}" if field.is_a?(Prism::Template::FlagsField) %>
+    attr_reader :<%= field.name -%><%= "\n    protected :#{field.name}" if field.is_a?(Prism::Template::FlagsField) %>
     <%- end -%>
-
     <%- end -%>
     <%- node.fields.each do |field| -%>
     <%- case field -%>
@@ -348,6 +347,22 @@ module Prism
     # def self.type: () -> Symbol
     def self.type
       :<%= node.human %>
+    end
+
+    # Implements case-equality for the node. This is effectively == but without
+    # comparing the value of locations. Locations are checked only for presence.
+    def ===(other)
+      other.is_a?(<%= node.name %>)<%= " &&" if node.fields.any? %>
+        <%- node.fields.each_with_index do |field, index| -%>
+        <%- if field.is_a?(Prism::Template::LocationField) || field.is_a?(Prism::Template::OptionalLocationField) -%>
+        (<%= field.name %>.nil? == other.<%= field.name %>.nil?)<%= " &&" if index != node.fields.length - 1 %>
+        <%- elsif field.is_a?(Prism::Template::NodeListField) || field.is_a?(Prism::Template::ConstantListField) -%>
+        (<%= field.name %>.length == other.<%= field.name %>.length) &&
+        <%= field.name %>.zip(other.<%= field.name %>).all? { |left, right| left === right }<%= " &&" if index != node.fields.length - 1 %>
+        <%- else -%>
+        (<%= field.name %> === other.<%= field.name %>)<%= " &&" if index != node.fields.length - 1 %>
+        <%- end -%>
+        <%- end -%>
     end
   end
   <%- end -%>
--- a/test/prism/ruby_api_test.rb
+++ b/test/prism/ruby_api_test.rb
@@ -244,6 +244,21 @@ module Prism
       assert_equal 16, base[parse_expression("0x1")]
     end
 
+    def test_node_equality
+      assert_operator parse_expression("1"), :===, parse_expression("1")
+      assert_operator Prism.parse("1").value, :===, Prism.parse("1").value
+
+      complex_source = "class Something; @var = something.else { _1 }; end"
+      assert_operator parse_expression(complex_source), :===, parse_expression(complex_source)
+
+      refute_operator parse_expression("1"), :===, parse_expression("2")
+      refute_operator parse_expression("1"), :===, parse_expression("0x1")
+
+      complex_source_1 = "class Something; @var = something.else { _1 }; end"
+      complex_source_2 = "class Something; @var = something.else { _2 }; end"
+      refute_operator parse_expression(complex_source_1), :===, parse_expression(complex_source_2)
+    end
+
     private
 
     def parse_expression(source)


exterm:
cc @kddnewton

exterm:
Kevin said he'd

>  template out the method and make sure it compared against each of the fields.

I'm not quite sure what he meant by "template out", and I'm wondering where to find the full list of fields that should be compared (and are not covered by `deconstruct`).

I'll take another look.

exterm:
Ah, I guess "template out" would mean to not add it to the parent `Node` class, but to the specific child classes.

mmh. I was hoping that `deconstruct` would be sufficiently specific, but it doesn't look like it (at least it doesn't include `value`)

exterm:
Ah. I think for the comparison, I can just use `#slice` in addition to the fields already reflected in `deconstruct`

exterm:
I think this is ready for a review

templates/lib/prism/node.rb.erb
```diff
@@ -96,6 +96,39 @@ module Prism
     def type
       raise NoMethodError, "undefined method `type' for #{inspect}"
     end
+
+    # Value equality.
+    def ==(other)
+      return false unless other.is_a?(Node) && other.type == type
+
+      deconstructed_self = deconstruct_keys(nil)
```
kddnewton, line 104: Ahh, I really don't want to tie our pattern matching implementation to equality. We could easily add things to pattern matching that would could break this (like the ability to take out the `value` of an integer node, or the options of a regexp). This adds a dependency that we don't want to have.

This also doesn't account for source, which we definitely want to track. Nodes are tied to the source file they came from, but this doesn't account for that.

templates/lib/prism/node.rb.erb
```diff
@@ -96,6 +96,39 @@ module Prism
...
+    alias eql? ==
+
+    # Case equality. Implemented like #==, but ignores locations.
+    def ===(other)
+      return false unless other.is_a?(Node) && other.type == type
+
+      deconstructed_self = deconstruct_keys(nil)
+      deconstructed_other = other.deconstruct_keys(nil)
+
+      deconstructed_self.delete(:location)
```
kddnewton, line 123: `:location` is not the only key that includes locations, there can be others based on `config.yml`.

Overall I feel strongly that we should template this in `node.rb.erb`. As I mentioned above, I don't want to tie the pattern matching implementation to equality.

kddnewton (changes requested):

templates/lib/prism/node.rb.erb
exterm: I agree it makes sense to track source if we track location. Will look at making this independent from pattern matching. For context, I took inspiration from the node equality test helper - so that dependency is already there internally, might make sense fixing that. Maybe a good path forward becomes obvious when I find a better solution for this PR.

templates/lib/prism/node.rb.erb
exterm: If I understand you correctly, by "template it" you mean implement it in the templated subclasses of `Node`? What would be the advantage of doing that?

eregon: Notably it would be a lot faster by avoiding array allocations, and also more explicit what it compares (when looking at the generated `node.rb`)

exterm: > more explicit what it compares

I'm not sure I understand that part. Node equality seems to me like something that should be implemented once for all nodes.

The fact that some node subclasses have additional things that act like fields (`value`) makes that a little difficult, but I think the mixin in `node_ext` is an OK workaround.

On that note. If I were to implement equality in the templated subclasses, I'd have access to `fields` - but I still wouldn't know that for some types I need to compare `value` too. I don't think we want branching on `node.name` in the templating code. The value method is defined in `node_ext`, so that's probably where we should handle it for equality too. Meaning, I think I'd keep `EqualityWithValueMethod` around.

So I could do this two step thing:

1. move the equality methods to the templated subclasses
2. compare all fields from `node.fields` instead of iterating over `deconstruct_keys`

Or alternatively, we add a `fields` class method to each subclass, in which case we could still have only one implementation of equality in the parent class.

I'm struggling a bit to judge what's a good fit here since the templating seems to mean that standard OOP reasoning doesn't apply (otherwise I would strongly argue for just one implementation of node equality)

I'll put up another version and we can see what that looks like.

templates/lib/prism/node.rb.erb
exterm: > the ability to take out the value of an integer node

That would actually just make things more consistent. Why _isn't_ the value included in the deconstructed version of a node?

templates/lib/prism/node.rb.erb
eregon: I don't think it's necessary to compare `value` for IntegerNode, FloatNode, etc, *if* locations are compared too.
These `value` methods return derived values from other fields, and given things like `Kernel#Integer` are deterministic, if the other fields (including location) are equal, then the result `#value` must be equal too (because same location and `Integer(str) == Integer(str)`).

But for `===` indeed it's not so simple as that wouldn't compare locations.
I think prepending a module to do `def ===(o); super && value == o.value; end` would be a good way.
An alternative would be some predicate in template.rb.
WDYT @kddnewton ?

exterm: Interesting points.

exterm:
I'm not quite sure what the flags do. I've included them in the comparison for now and hope they're correct.

templates/lib/prism/node.rb.erb
exterm: I'd assume we want `Prism.parse("1") != Prism.parse("2")` though, and since locations are the same here and there is no source file, I think we need to compare the value?

exterm:
I also just realized this doesn't work for `ProgramNode` since it has the source, and that's still an identity comparison.

templates/lib/prism/node.rb.erb
exterm: Oh, also - this is a recursive comparison, but the source would be the same for the whole tree, right? We wouldn't want to compare the source for each node of the tree.

eregon:
> I also just realized this doesn't work for `ProgramNode` since it has the source, and that's still an identity comparison.

It's all nodes actually, not just `ProgramNode`.

That makes me think, is it expected that Location#== doesn't compare the `@source`s?
It feels weird because for 2 locations with different sources it can be different code and then it's not really the "same source code location" and in fact the source code bytes at those locations might differ.
So that actually breaks the property I mentioned above for IntegerNode, Location#== doesn't imply the same source code bytes for that location.
I think Location#== should compare the `@source` too, WDYT @kddnewton ?

BTW https://github.com/ruby/prism/pull/2483/files#diff-2ea89c2875e53b7d8ae10b02fb655b0a3a2b6304b8c565936a9c155805ec9c4b merged and that means no need to handle `value` specially for IntegerNode and FloatNode.

templates/lib/prism/node.rb.erb
eregon: > We wouldn't want to compare the source for each node of the tree.

If it's by identity it wouldn't matter much performance-wise. I wouldn't imagine to compare Source bytes, that sounds too expensive.
Which makes me think, maybe Node#== should be identity, just like it already is. Is there a use case for a `Node#==` which would compare locations & source & fields? Wouldn't the 2 nodes need to be the same to be == then?
One case I could think of is if using Prism::Compiler and transforming the tree, in that cases unmodified nodes could be ==. But I'm not sure there is any point to call == on these subtrees.

Based on these thoughts, my suggestion would be to focus on `===` (and that would ignore locations and so the source too, per the discussion on the issue) and leave `==` as identity until there a use case to do something different.

exterm: Yeah, I think that makes sense to me. I really want to know whether two different sources parse into equivalent ASTs.

kddnewton (changes requested):
I don't think we should skip comparing `source` on `==`. If they come from different sources, they're not the same node. I'm fine with saying they're _equivalent_ nodes, which is why I would like this PR to do one of the following:

* Include source comparisons in the `==` implementation
* Change the method being implemented to be `===`
* Change the method being implemented to be `=~`

I'm fine with any of these solutions. As a side-note, `EqualityWithValueMethod` should no longer be necessary now that we store the value as one of the fields.

exterm:
> EqualityWithValueMethod should no longer be necessary now that we store the value as one of the fields.

❤️ 

I agree with the other conclusions here. I'll change this to _only_ implement `===` or `=~` once I find some free time.

kddnewton:
Hey @exterm — do you have time to return to this?

kddnewton:
I've implemented `===` as per our discussion here. Let me know how it works out for you!

exterm:
Sorry, I have been partly super busy and partly travelling. Thank you for taking this on Kevin. I hope I can return to the packwerk work soon to try this out.

PR #2478, opened 2024-02-21
Support ItParametersNode

kddnewton:
So that compilers know they need to add to add an anonymous variable to the local table. Also, add a bunch of errors for the various combinations of parameters now, including:

* defining numbered, but ordinary already defined
* defining it, but ordinary already defined
* defining numbered, but it already defined
* defining it, but numbered already defined


eregon:
> So that compilers know they need to add to add an anonymous variable to the local table.

Isn't it the same as `_1`? Maybe CRuby names the hidden variable differently for `it` than for _1`?

kddnewton:
Yeah it's named for `_1` and it's not named for `it`.

kddnewton:
Regardless, this node is smaller anyway

PR #2483, opened 2024-02-22
Parse integer and float values, add them to the AST

kddnewton:


templates/java/org/prism/Loader.java.erb
```diff
@@ -310,6 +310,18 @@ public class Loader {
         return (short) flags;
     }
 
+    private VariableInteger loadVariableInteger() {
+        boolean negative = buffer.get() != 0;
+        int length = loadVarUInt();
+        int[] nodes = new int[length];
```
eregon, line 316: nodes -> parts or so

templates/java/org/prism/Loader.java.erb
```diff
@@ -310,6 +310,18 @@ public class Loader {
...
+
+        for (int i = 0; i < length; i++) {
+            nodes[i] = loadVarUInt();
+        }
+
+        return new VariableInteger(negative, nodes);
```
eregon, line 322: I think it would be far more useful if this returned `int`, `long` or `BigInteger` (and so an `Object` return type for this method).

That would also be a lot faster, allocating extra VariableInteger+int[] for the common case where it fits in `int` would be I think a significant overhead.

templates/java/org/prism/Loader.java.erb
```diff
@@ -310,6 +310,18 @@ public class Loader {
         return (short) flags;
     }
 
+    private VariableInteger loadVariableInteger() {
```
eregon, line 313: Minor: for the method name, "variable integer" seems a bit confusing with varint, which has a specific separate meaning (at least in how it's used here, like protobuf varint). GMP calls it "arbitrary precision integer". Not sure if there is a better term. "loadBigInteger" would be not ideal because it can be "small" too.

templates/java/org/prism/Loader.java.erb
eregon: In [rb_integer_pack](https://github.com/ruby/ruby/blob/e04146129ec6898dd6a9739dad2983c6e9b68056/include/ruby/internal/intern/bignum.h#L579-L679) terms that's called `words`.

eregon:
Could you update https://github.com/ruby/prism/blob/main/docs/serialization.md#types to include this new type, and whether it's most/least significant word first?

kddnewton:
> Could you update https://github.com/ruby/prism/blob/main/docs/serialization.md#types to include this new type, and whether it's most/least significant word first?

Done! This PR now includes `FloatNode` as well, which I've also updated.

eregon:
Very nice, looking forward to adopt this.

PR #2484, opened 2024-02-22
Fix parsing keyword params

haldun:
Fixes parsing code like:

```ruby
def foo(bar: bar = 1); end
```

@kddnewton I recall that there should be an issue for this, but I cannot find it now.


kddnewton (changes requested):
I think this works, but we'll also need to check for `,` and `)`, as in:

```ruby
def foo(bar: ((bar, *) = 1)); end
def foo(bar: ((*, bar) = 1)); end
```

PR #2485, opened 2024-02-22
all constant nodes should have full_name and full_name_parts methods

exterm:
In reference to https://github.com/ruby/prism/discussions/2434#discussioncomment-8549837

There's no specific test for node extensions, it seems? Not sure where I'd add a test for this.


kddnewton (approved):
Looks good. There are tests in ruby_api test but I'm okay merging this as is

PR #2487, opened 2024-02-23
Fix loadInteger() to always return the most optimal representation

eregon:
* It was non-optimal for -2**31 and -2**63, where it returned respectiely long and BigInteger instead of int and long.
* There was a missing negation for the wordsLength > 2 case.

Found while adopting it in TruffleRuby which asserts the optimal representation is always used.


kddnewton (approved):
Ahh thanks, my lack of Java knowledge is showing

PR #2496, opened 2024-02-25
Follow RuboCop's parser engine support

koic:
With the introduction of `Prism::Translation::Parser` to RuboCop (RuboCop AST), the number of arguments for `RuboCop::AST::ProcessedSource#parser_class` internal API will be changed:
https://github.com/rubocop/rubocop-ast/pull/277

## Before

As a result, the following error will occur starting from the next release of RuboCop AST (< 1.30.0) :

```console
$ bundle exec ruby -rrubocop/ast -rprism -rprism/translation/parser/rubocop -ve "p RuboCop::AST::ProcessedSource.new('42', 80_82_73_83_77.33).ast"
/Users/koic/src/github.com/ruby/prism/lib/prism/translation/parser/rubocop.rb:25:in `parser_class':
wrong number of arguments (given 2, expected 1) (ArgumentError)
        from /Users/koic/src/github.com/rubocop/rubocop-ast/lib/rubocop/ast/processed_source.rb:309:in `create_parser'
        from /Users/koic/src/github.com/rubocop/rubocop-ast/lib/rubocop/ast/processed_source.rb:219:in `parse'
        from /Users/koic/src/github.com/rubocop/rubocop-ast/lib/rubocop/ast/processed_source.rb:47:in `initialize'
        from -e:1:in `new'
        from -e:1:in `<main>'
```

## After

This PR prevents the above error by updating the monkey patch to support the new argument:

```console
$ bundle exec ruby -rrubocop/ast -rprism -rprism/translation/parser/rubocop -ve "p RuboCop::AST::ProcessedSource.new('42', 80_82_73_83_77.33).ast"
s(:int, 42)
```

Moreover, to ensure compatibility with the existing RuboCop AST, conditional logic has been implemented to maintain backward compatibility.

### DIFF
--- a/lib/prism/translation/parser/rubocop.rb
+++ b/lib/prism/translation/parser/rubocop.rb
@@ -20,17 +20,34 @@ module Prism
 
       # This module gets prepended into RuboCop::AST::ProcessedSource.
       module ProcessedSource
-        # Redefine parser_class so that we can inject the prism parser into the
-        # list of known parsers.
-        def parser_class(ruby_version)
-          if ruby_version == Prism::Translation::Parser::VERSION_3_3
-            require "prism/translation/parser33"
-            Prism::Translation::Parser33
-          elsif ruby_version == Prism::Translation::Parser::VERSION_3_4
-            require "prism/translation/parser34"
-            Prism::Translation::Parser34
-          else
-            super
+        # This condition is compatible with rubocop-ast versions up to 1.30.0.
+        if RuboCop::AST::ProcessedSource.instance_method(:parser_class).arity == 1
+          # Redefine parser_class so that we can inject the prism parser into the
+          # list of known parsers.
+          def parser_class(ruby_version)
+            if ruby_version == Prism::Translation::Parser::VERSION_3_3
+              require "prism/translation/parser33"
+              Prism::Translation::Parser33
+            elsif ruby_version == Prism::Translation::Parser::VERSION_3_4
+              require "prism/translation/parser34"
+              Prism::Translation::Parser34
+            else
+              super
+            end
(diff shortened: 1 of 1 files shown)


lib/prism/translation/parser/rubocop.rb
```diff
@@ -20,17 +20,34 @@ class Parser
...
+              require "prism/translation/parser34"
+              Prism::Translation::Parser34
+            else
+              super
+            end
+          end
+        else
+          # Redefine parser_class so that we can inject the prism parser into the
+          # list of known parsers.
+          def parser_class(ruby_version, _parser_engine)
```
kddnewton, line 41: If we get here, should we check `parser_engine` to verify that it's prism?

koic: Yes, it is possible to narrow down condition with `parser_engine`. However, by default, the Parser gem's value (`:parser_whitequark`) is passed to `parser_engine` argument. Therefore, it might be preferable not to narrow down conditions for the following reasons:

1. While it is possible to specify Prism in RuboCop AST, but RuboCop core is still WIP to support this. So, passing the value of Prism `:parser_prism` to `parser_engine` is not yet possible.
2. The special value `TargetRubyVersion: 80_82_73_83_77.33` might be intended to signify the use of Prism.

It might be better to hold off on adding `parser_engine` condition at least until RuboCop Core has finished updating. Especially due to the first reason, there might come a time when this patch would not be operational.

kddnewton: That sounds good!

kddnewton (approved):

PR #2497, opened 2024-02-25
Give a kind to all node[] fields and better types in Java-generated code

eregon:
* Fixes https://github.com/ruby/prism/issues/2058

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1556,6 +1556,9 @@ nodes:
               ^
       - name: elements
         type: node[]
+        kind:
+          - AssocNode
+          - AssocSplatNode
         comment: |
           The elements of the hash. These can be either `AssocNode`s or `AssocSplatNode`s.
 
@@ -1903,6 +1906,10 @@ nodes:
         type: location
       - name: parts
         type: node[]
+        kind:
+          - StringNode
+          - EmbeddedStatementsNode
+          - EmbeddedVariableNode
       - name: closing_loc
         type: location
     newline: parts
@@ -1920,6 +1927,10 @@ nodes:
         type: location
       - name: parts
         type: node[]
+        kind:
+          - StringNode
+          - EmbeddedStatementsNode
+          - EmbeddedVariableNode
       - name: closing_loc
         type: location
     newline: parts
@@ -1934,6 +1945,11 @@ nodes:
         type: location?
       - name: parts
         type: node[]
+        kind:
+          - StringNode
+          - EmbeddedStatementsNode
+          - EmbeddedVariableNode
+          - InterpolatedStringNode # `"a" "#{b}"`
       - name: closing_loc
         type: location?
     newline: parts
@@ -1948,6 +1964,10 @@ nodes:
         type: location?
       - name: parts
         type: node[]
+        kind:
+          - StringNode
+          - EmbeddedStatementsNode
+          - EmbeddedVariableNode
       - name: closing_loc
         type: location?
     newline: parts
@@ -1962,6 +1982,10 @@ nodes:
         type: location
       - name: parts
         type: node[]
+        kind:
+          - StringNode
+          - EmbeddedStatementsNode
+          - EmbeddedVariableNode
       - name: closing_loc
         type: location
     newline: parts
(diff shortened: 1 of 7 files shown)


eregon:
I was wondering if CI would catch that InterpolatedStringNode#parts could be not only:
```
        kind:
          - StringNode
          - EmbeddedStatementsNode
```
But also EmbeddedVariableNode (which I figured out by checking prism.c).
But unfortunately the CI does not catch this.
I guess we would need some kind of runtime type checks for this, and `steep` seems to only do static type checks.

eregon:
I added my own runtime checks in nodes `initialize` constructors, only added by the template if `$CHECK_FIELD_KIND` is set.

templates/lib/prism/node.rb.erb
```diff
@@ -98,6 +98,9 @@ module Prism
       raise NoMethodError, "undefined method `type' for #{inspect}"
     end
   end
+  <%- if ENV["CHECK_FIELD_KIND"] -%>
+  CHECK_FIELD_KIND = ENV["CHECK_FIELD_KIND"]
```
eregon, line 102: It's a build-time + run-time check to make it still possible to run e.g. `bin/parse -e 'code'` when investigating a type failure. If it was build-time only, that would require to `rm lib/prism/node.rb && bundle exec rake compile && bin/parse`.

kddnewton: I think it could make it easier to test as well.

config.yml
```diff
@@ -1934,6 +1945,11 @@ nodes:
         type: location?
       - name: parts
         type: node[]
+        kind:
+          - StringNode
+          - EmbeddedStatementsNode
+          - EmbeddedVariableNode
+          - InterpolatedStringNode # example in parse_line_evstr_after_break.txt
```
kddnewton, line 1952: This one is by design. It represents string concatenation (we replaced `StringConcatNode` with this).

config.yml
```diff
@@ -1948,6 +1964,11 @@ nodes:
         type: location?
       - name: parts
         type: node[]
+        kind:
+          - StringNode
+          - EmbeddedStatementsNode
+          - EmbeddedVariableNode
+          - SymbolNode # %I example in spanning_heredoc.txt line 48
```
kddnewton, line 1971: This is a bug, will be addressed by https://github.com/ruby/prism/pull/2517, so this can be removed.

config.yml
```diff
@@ -2238,10 +2292,32 @@ nodes:
...
+          - LocalVariableTargetNode
+          - InstanceVariableTargetNode
+          - ClassVariableTargetNode
+          - GlobalVariableTargetNode
+          - ConstantTargetNode
+          - ConstantPathTargetNode
+          - CallTargetNode
+          - IndexTargetNode
+          - MultiTargetNode
+          - LocalVariableReadNode # On parsing error of `-> { _1, = 0 }`
```
kddnewton, line 2305: This is a bug, will be addressed by https://github.com/ruby/prism/pull/2517.

config.yml
```diff
@@ -2238,10 +2292,32 @@ nodes:
...
+          - LocalVariableTargetNode
+          - InstanceVariableTargetNode
+          - ClassVariableTargetNode
+          - GlobalVariableTargetNode
+          - ConstantTargetNode
+          - ConstantPathTargetNode
+          - CallTargetNode
+          - IndexTargetNode
+          - MultiTargetNode
+          - LocalVariableReadNode # On parsing error of `-> { *,_1 = 0 }`
```
kddnewton, line 2320: This is a bug, will be addressed by https://github.com/ruby/prism/pull/2517.

templates/lib/prism/node.rb.erb
eregon: I agree it's verbose in this template.
Generating another file seems quite heavy though, and we probably wouldn't want anything to use these checks besides that specific test added in CI (because it would be a lot slower).
I'm thinking to add some helper methods in template.rb to make this shorter and more readable, then it would only be a couple lines added in this file.

eregon: Done, I simplified this a lot in the last commit.

kddnewton (approved):

PR #2507, opened 2024-02-26
Parse huge integer faster

tompng:
Parsing huge integer literal is a bit slow.
`Prism.parse('1'*1000000)` takes about 67.8s, `eval('1'*1000000)` takes only 0.32s. This pull request improves it to 1.46s.

Constructing a bigint with `digits.each{v=v*10+_1}` is `O(n^2)`.
To make it faster, I think we need a fast bigint multiplication algorithm.
(I'm not sure this is worth introducing complexity)

### Bigint and multiplication
Implement karatsuba algorithm that ruby's Bignum uses internally.
Implement radix conversion using fast multiplication.

### Parse digits
If base is power of two, just pack bits to bigint.
If base is 10, peform radix conversion.

### pm_integer_t to ::Integer
`"0x...".to_i`  is  faster than `|` `<<` for large integer.

### pm_intger_t to string
Use radix conversion

### Comparison

| Prism.parse(code) | before | after | ratio |
| --- | --- | --- | --- |
| '1'*1000000 | 67.798872s | 1.462340s | x46.3 |
| '1'*100000 | 0.723387s | 0.061709s | x11.7 |
| '0x'+'1'*100000 | 0.908699s | 0.002813s | x323 |
| (['1'*100]*100000).join(';') | 0.322435s | 0.189371s | x1.70 |
| (['1'*20]*100000).join(';') | 0.084510s | 0.088816s | x0.95 slower |

### DIFF
--- a/ext/prism/extension.c
+++ b/ext/prism/extension.c
@@ -992,30 +992,16 @@ integer_parse(VALUE self, VALUE source) {
     pm_integer_t integer = { 0 };
     pm_integer_parse(&integer, PM_INTEGER_BASE_UNKNOWN, start, start + length);
 
-    VALUE number;
-
-    if (integer.values == NULL) {
-        number = UINT2NUM(integer.value);
-    } else {
-        number = UINT2NUM(0);
-        for (size_t i = 0; i < integer.length; i++) {
-            VALUE receiver = rb_funcall(UINT2NUM(integer.values[i]), rb_intern("<<"), 1, ULONG2NUM(i * 32));
-            number = rb_funcall(receiver, rb_intern("|"), 1, number);
-        }
-    }
-
-    if (integer.negative) number = rb_funcall(number, rb_intern("-@"), 0);
-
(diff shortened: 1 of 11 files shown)


kddnewton:
Amazing work @tompng — I'll get this reviewed this week

kddnewton (changes requested):
Overall this looks really good!

The only change I would like to see: right now this PR introduces `bigint_t` in addition to `pm_integer_t`. When necessary, you parse into a `bigint_t` and then convert everything into a `pm_integer_t` afterward.

Could you change this instead so that `pm_integer_t` looks like `bigint_t` and we just use one struct? As in:

```c
typedef struct {
    uint32_t value;
    size_t length;
    uint32_t *values;
} pm_integer_t;
```

That way we don't have to convert all of the bigints back into integers at the end, but we still can avoid allocations if the number is < 32 bits.

tompng:
Updated 😄

kddnewton (approved):

Issue #2002, opened 2023-12-06
Support `-n`, `-p`, `-l`, and `-a` command-line options


kddnewton:
These options all mutate the AST. We either need to move these into the compiler in CRuby and then it won't be relevant here, or we need to support them in prism. Personally I'm a fan of moving them into the compiler in CRuby (I've opened a ticket here: https://bugs.ruby-lang.org/issues/20035). But we'll see if they go for it.

PR #2511, opened 2024-02-27
Support -p, -n, -a, and -l command line options

kddnewton:
The current CRuby parser supports these options by mutating the AST. I don't want to do that, and instead would prefer to do it in the compiler. But that work is for another day, and this will be far easier than wading into those waters. So instead, we now support the following options:

* `command_line_p`
* `command_line_n`
* `command_line_a`
* `command_line_l`

These correspond with the respective command line flags. They mutate the AST such that you get back what would be executed (this pretty much exactly matches what the current parser does).

cc @enebo, @eregon, @seven1m — you may want to use these or you may not. I'm assuming all of you currently have your own solutions for this. It's not that much code at the end of the day so entirely up to you.

Fixes https://github.com/ruby/prism/issues/2002

### DIFF
--- a/docs/serialization.md
+++ b/docs/serialization.md
@@ -195,7 +195,10 @@ The final argument to `pm_serialize_parse` is an optional string that controls t
 | `4`     | the length the encoding    |
 | ...     | the encoding bytes         |
 | `1`     | frozen string literal      |
-| `1`     | suppress warnings          |
+| `1`     | `-p` command line flag     |
+| `1`     | `-n` command line flag     |
+| `1`     | `-l` command line flag     |
+| `1`     | `-a` command line flag     |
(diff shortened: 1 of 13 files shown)


eregon:
This looks convenient, we should try it, cc @andrykonchin.
I remember when we implemented (more of) these flags in TruffleRuby (years ago) we were wondering if we could do it by just prepending/appending code around the input code but that didn't fly for some reason (I don't recall why, maybe simply because it would cause all node locations to be wrong, but with this that shouldn't be an issue).

enebo:
@kddnewton Nice!  I manually write the syntax tree around the contents of -e but being able to delete that code would be a nice simplification. 

 JRuby-proper is written around wrapping these nodes around something already parsed so I have to look at how easy it will be to achieve this as part of the original parse call (obviously old parser does not work this way so I need to consider that as well).

PR #2525, opened 2024-02-29
Fix an incorrect parsing for `Prism::Translation::Parser`

koic:
This PR fixes an incorrect parsing for `Prism::Translation::Parser` when one-line pattern mathing with Ruby 2.7 runtime.

## Expected

Parsing should be done based on the specified Ruby parsing version, independent of the Ruby runtime version. When parsing for Ruby 3.3, it should return `:match_pattern_p` node:

```console
$ ruby -rprism -rprism/translation/parser33 -ve 'p Prism::Translation::Parser33.parse("foo in bar")'
ruby 3.0.6p216 (2023-03-30 revision 23a532679b) [x86_64-darwin19]
s(:match_pattern_p,
s(:send, nil, :foo),
  s(:match_var, :bar))
```

## Actual

When parsing with Ruby 2.7 runtime, `match_pattern` node is returned, even though it is expected to parse for Ruby 3.3:

```console
$ ruby -rprism -rprism/translation/parser33 -ve 'p Prism::Translation::Parser33.parse("foo in bar")'
ruby 2.7.8p225 (2023-03-30 revision 1f4d455848) [x86_64-darwin19]
s(:match_pattern,
s(:send, nil, :foo),
  s(:match_var, :bar))
```

The cause was the use of `RUBY_VERSION` for condition logic, which made it dependent on runtime Ruby version.
This PR resolves the issue by depending on `Prism::Translation::Parser#version` instead, using parsing Ruby version instead of runtime Ruby version.

## Background

Found due to incompatibility with RuboCop's `Layout/SpaceAroundKeyword` and `Style/TernaryParentheses` cops.

### DIFF
--- a/lib/prism/translation/parser/compiler.rb
+++ b/lib/prism/translation/parser/compiler.rb
@@ -1070,22 +1070,12 @@ module Prism
 
         # foo in bar
         # ^^^^^^^^^^
-        if RUBY_VERSION >= "3.0"
-          def visit_match_predicate_node(node)
-            builder.match_pattern_p(
-              visit(node.value),
-              token(node.operator_loc),
-              within_pattern { |compiler| node.pattern.accept(compiler) }
-            )
-          end
-        else
-          def visit_match_predicate_node(node)
-            builder.match_pattern(
-              visit(node.value),
-              token(node.operator_loc),
-              within_pattern { |compiler| node.pattern.accept(compiler) }
-            )
-          end
+        def visit_match_predicate_node(node)
+          builder.match_pattern_p(
+            visit(node.value),
+            token(node.operator_loc),
+            within_pattern { |compiler| node.pattern.accept(compiler) }
+          )
         end
 
         # foo => bar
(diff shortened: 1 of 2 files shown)


kddnewton:
Ahh interesting. In this case won't `parser.version >= 3.0` always be true? We only have `3.3` and `3.4`. So maybe the issue is that we shouldn't be comparing to `Parser::CurrentRuby` in the test suite?

koic:
> Ahh interesting. In this case won't `parser.version >= 3.0` always be true? 

Sure. This PR has been updated to remove what is essentially an unnecessary conditional branch.

> So maybe the issue is that we shouldn't be comparing to `Parser::CurrentRuby` in the test suite?

How about comparing the results of `Parser::Ruby33` and `Parser::Translation::Parser33` instead of `Parser::CurrentRuby` (which depends on runtime Ruby version) and `Parser::Translation::Parser` (fixed to Ruby 3.4)? This would allow for testing of parsing that is independent of runtime Ruby version.

Eventually, it will be necessary to consider comparison logic for Ruby 3.4 as well, but the approach would likely involve comparing the results of `Parser::Ruby34` and `Parser::Translation::Parser34`.

I will update the PR with this approach for now. Please let me know if there are any concerns.

kddnewton:
I think that makes the most sense. Thank you!

PR #2528, opened 2024-02-29
Use regular method to alias behavior of `visit_child_nodes`

Morriar:
The way the `visit_` methods are aliased to the base visitor makes it hard to override the behavior of `visit`:

```rb
module Prism
  class Visitor
    def visit(node)
      puts "BaseVisitor"
    end

    alias :visit_foo :visit
  end
end

class MyVisitor < Prism::Visitor
  def visit(node)
    puts "MyVisitor"
  end
end

v = Prism::Visitor.new
v.visit_foo(42) # => BaseVisitor

v = MyVisitor.new
v.visit_foo(42) # => BaseVisitor :(
```

One would need to also override each alias to `visit` so the call resolves to `MyVisitor#visitor`:

```rb
class MyVisitor < Prism::Visitor
  def visit(node)
    puts "MyVisitor"
  end

  def visit_foo(node)
    visit(node)
  end

  # ... all the other nodes
end

v = MyVisitor.new
v.visit_foo(42) # => MyVisitor :)
```

We can fix this with using normal methods rather than aliases:

```rb
module Prism
  class Visitor
    def visit(node)
      puts "BaseVisitor"
    end

    def visit_foo(node)
      visit(node)
    end
  end
end

class MyVisitor < Prism::Visitor
  def visit(node)
    puts "MyVisitor"
  end
end

v = MyVisitor.new
v.visit_foo(42) # => MyVisitor :)
```

### DIFF
--- a/templates/lib/prism/visitor.rb.erb
+++ b/templates/lib/prism/visitor.rb.erb
@@ -47,7 +47,9 @@ module Prism
     <%- nodes.each_with_index do |node, index| -%>
 <%= "\n" if index != 0 -%>
     # Visit a <%= node.name %> node
-    alias visit_<%= node.human %> visit_child_nodes
+    def visit_<%= node.human %>(node)
+      visit_child_nodes(node)
+    end
     <%- end -%>
   end
 end


kddnewton (changes requested):
Just marking as "request changes"

eregon:
Seems quite related to #2433.

I wonder if for clarity maybe we should remove BasicVisitor#visit, as it seems not called anyway (except maybe manually on the root node).

eregon:
I see. Performance-wise it's probably better to:
```ruby
class MyVisitor < Prism::BasicVisitor
  def visit_program_node(node)
    node.statements&.accept(self)
  end
end
```
to improve the inline cache for the `accept` call (which then would be only StatementsNode and not many different nodes, assuming an example with more `visit_*` implemented). TruffleRuby would be able to split that to do it automatically but that's the only Ruby implementation supporting splitting AFAIK.
But it's more verbose/less elegant code obviously.

I think one thing that could help is more extensive documentation for the existing visitors with examples on how to use them and what they can be used for (e.g. visiting a subset or specific nodes, but no easy way to have a method called for each node, for that one way is the recursive method I posted above).

Issue #1247, opened 2023-08-14
Serialization of piped Ruby source code


enebo:
Almost all Ruby source is processed as an in-memory block of bytes.  I just remembered ```some_piped_output | ruby -``` is not.  In this case the parsing will be to make ```$stdin``` into an ```IO``` and using ```IO#gets``` to process each line.  How do we support this?  MRI obviously also needs this and we perhaps just need a second entry point where instead of passing in a source blob we pass in an fd (and then leverage the same code MRI does in the lexer for this).

eregon:
Are you sure it's incrementally reading from $stdin?
AFAIK Ruby reads all of `$stdin` and then only parses:

When using a TTY Ruby or the shell waits for the whole input:
```
$ ruby
p 1
sleep 1
p 2
Ctrl+D
1 # only printed after Ctrl+D, i.e. end of input
2
```

And on
```
$ sh -c "echo 'p 1'; sleep 1; echo 'p 2'" | ruby
1
2
```
nothing appears before one second, so at least the `p 1` is not executed and probably not parsed before end-of-input.

enebo:
@eregon I am befuddled.

enebo:
Ah so this is a problem after all:

```text
sh -c "echo 'DATA.each_line do |line| puts line; end'; echo '__END__'; echo '1'; sleep 5; echo '2'"; sleep 5; echo '3' | ruby -
```

In this case when we hit __END__ the parsing is finished so the code executes but stdin is still open and can be piping the data after __END__.

kddnewton:
Oh man, I am scared about this. This violates one of my fundamental assumptions that it was okay to replace the line-fetching code with mmap. I really did not realize this was a path. There is a ton of bookkeeping that goes into making this happen in CRuby right now, particularly around heredocs where you have to jump back and forth between multiple lines. I'll take a look to see if there's an easy way, but I'm nervous.

kddnewton:
Actually @enebo is the _only_ case where this is an issue when we hit `__END__`? In that case we could specially handle it with just that condition.

enebo:
@kddnewton HAHA.  I hope so.  It took me like an hour to remember this case.  I think the fundamentals though would be that no AST is produced until ^D or ```__END__``` so any streaming behavior would be after that point (who cares about seeing inline syntax warns/errors while it actively streams).  All I can brainstorm is there could be some super esoteric command-line option?

kddnewton:
[...]

PR #2534, opened 2024-03-01
Support parsing streams

kddnewton:
Closes https://github.com/ruby/prism/issues/1247 (our second-oldest open issue!)

Prism can now parse streams. In the C API you can use it like this:

```c
#include "prism.h"

int main(void) {
    pm_parser_t parser;
    pm_options_t options = { .line = 1, 0 };
    pm_node_t *node = pm_parse_stream(&parser, stdin, (pm_parse_stream_fgets_t *) fgets, &options);

    pm_buffer_t buffer;
    pm_prettyprint(&buffer, &parser, node);
    printf("%.*s\n", (int) pm_buffer_length(&buffer), pm_buffer_value(&buffer));

    pm_node_destroy(&parser, node);
    pm_parser_free(&parser);
    pm_buffer_free(&buffer);
}
```

Relevant signatures are:

```c
typedef char * (pm_parse_stream_fgets_t)(char *restrict string, int size, void *restrict stream);

PRISM_EXPORTED_FUNCTION pm_node_t *
pm_parse_stream(pm_parser_t *parser, void *stream, pm_parse_stream_fgets_t *fgets, const pm_options_t *options);
```

We never touch the stream itself, so it's treated as an opaque pointer. It is the responsibility of the given function pointer to load one line at a time into its first parameter. From the Ruby API you can call it like:

```ruby
Prism.parse_stream(stream)
```

where `stream` is any object that responds to `gets(length)`.

This is suitable for implementing streaming from stdin into the prism parser, while appropriately stopping the stream at the point that it hits an `__END__` marker, as in:

```ruby
stream = StringIO.new("foo\n__END__\nbar")
Prism.parse_stream(stream)
stream.read # => "bar"
```

cc @enebo @eregon @seven1m

### DIFF
--- a/docs/ruby_api.md
+++ b/docs/ruby_api.md
@@ -20,6 +20,7 @@ The full API is documented below.
 * `Prism.lex_file(filepath)` - parse the tokens corresponding to the given source file and return them as an array within a parse result
 * `Prism.parse(source)` - parse the syntax tree corresponding to the given source string and return it within a parse result
 * `Prism.parse_file(filepath)` - parse the syntax tree corresponding to the given source file and return it within a parse result
+* `Prism.parse_stream(io)` - parse the syntax tree corresponding to the source that is read out of the given IO object using the `#gets` method and return it within a parse result
 * `Prism.parse_lex(source)` - parse the syntax tree corresponding to the given source string and return it within a parse result, along with the tokens
 * `Prism.parse_lex_file(filepath)` - parse the syntax tree corresponding to the given source file and return it within a parse result, along with the tokens
 * `Prism.load(source, serialized)` - load the serialized syntax tree using the source as a reference into a syntax tree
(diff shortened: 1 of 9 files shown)


enebo:
@kddnewton I need to see if jffi and project Panama are capable of passing a pointer to a function which returns what you want.  Others will want this I am sure but I am not sure if I can use it at this point or not.

kddnewton:
@enebo any chance you can pass `stdin` and `fgets` directly? (That was kind of my hope in making the function pointer have the same signature as `fgets`.).

kddnewton:
I could also provide a version of this that assumes stdin and fgets?

enebo:
@kddnewton It looks like I can make a function in JNR.  Sometimes this is hit or miss but I will try and figure this out next week.  So long as I can make the right signature this should work out ok.  stdin+fgets might be possible too but I need to still bind it using JNR (JNR is in Java and not C)

enebo:
Err I misread your second suggestion but let's assume I can make a proper pointer to function and see where that goes.

PR #2540, opened 2024-03-02
Fix serialising `command_line` parsing option in Java

andrykonchin:
Java enum ordinals start from `0` value. So the first enum constant (`A`) has ordinal `0` and should be serialised as `1<<0` to equal `PM_OPTIONS_COMMAND_LINE_A = 0x1`. The same for enum constant `E` - its ordinal is `1` and it should be serialised as `1<<1` to match `PM_OPTIONS_COMMAND_LINE_E = 0x2`, etc.


kddnewton (approved):
Thanks!

PR #2542, opened 2024-03-03
[For embedded support] PRISM_XALLOCATOR

hasumikin:
ref https://github.com/ruby/prism/discussions/2217

This PR is to make the alloc interface replaceable:

- After this patch is merged, we are all going to add `x` prefix to malloc, calloc, realloc, and free (eg: `xmalloc` instead of `malloc`)
- By default, they are replaced with stdlib's functions at build by `include/prism/defines.h`
- If you compile Prism against a platform where you aren't able (or you don't want) to use stdlib's malloc, you can use custom allocator functions by defining `-D PRISM_XALLOCATOR` macro at build time
- Please see the `docs/build_system.md` in this patch for further information

BTW, I could casually make Prism work in PicoRuby with this branch https://twitter.com/hasumikin/status/1763133869321257287


hasumikin:
@kddnewton Sorry but could you help me with the CI "Documentation / check (pull_request) " failure?

kddnewton (approved):
Thanks @hasumikin! This is great. I think I'll use this in CRuby as well!

PR #2561, opened 2024-03-06
Run Sorbet type-checking with `typed: true`

paracycle:
This PR makes the Sorbet type-checking task work with `typed: true` for all files by default, except for a few files that we revert back to `typed: false`. We also suppress a handful of classes of errors that are hard to solve without inline type assertions and are not essential errors for what we are trying to do here.

I've also had to make some code changes to get type-checking to work at this level. The major refactor was to move all template related methods and classes under the `Prism::Template` namespace, so that there is no name collision with types under the `Prism` namespace.

### DIFF
--- a/lib/prism/translation/parser/compiler.rb
+++ b/lib/prism/translation/parser/compiler.rb
@@ -1799,7 +1799,7 @@ module Prism
 
         # Visit a heredoc that can be either a string or an xstring.
         def visit_heredoc(node)
-          children = []
+          children = Array.new
           node.parts.each do |part|
             pushing =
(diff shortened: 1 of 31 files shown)


kddnewton (approved):

lib/prism/translation/parser/compiler.rb
```diff
@@ -1799,7 +1799,7 @@ def visit_block(call, block)
 
         # Visit a heredoc that can be either a string or an xstring.
         def visit_heredoc(node)
-          children = []
+          children = Array.new
```
eregon, line 1802: Is this change necessary?

eregon (commented):
Nice!

Issue #2074, opened 2023-12-14
Heredocs not properly escaping with backslash followed by newline


enebo:
```ruby
puts <<~HERE
      a
      b\
      c
    HERE
```

This should be:

```text
a
bc
```

but is:
 
```text
a
b      c
```

So it appears to remove the `\n` but then doesn't eat the extra whitespace on the next line.

kddnewton:
... what. I had no idea it was supposed to eat the extra whitespace. That's... so weird. I will get this fixed, but probably not by 3.3.

enebo:
@kddnewton heredocs is a scary place.  the dedent feature is extra scary.

rwstauner:
I was curious so I looked into this only to realize that it needs to eat the leading whitespace that matches the ultimately common whitespace which might be decided after this line (which seems substantially more complicated):

```
$ cat heredoc.rb
def main
  puts <<~HERE.inspect
    a
    b\
    c
   d
  HERE
end
main
$ ruby -v heredoc.rb
ruby 3.4.0dev (2024-01-18T15:35:46Z master 00814fd672) [arm64-darwin23]
" a\n b c\nd\n"
$ ruby --parser=prism heredoc.rb
" a\n b    c\nd\n"
```

kddnewton:
That's okay that's already what we do anyway.

PR #2565, opened 2024-03-07
Fix up tilde heredoc line continuations

kddnewton:
Fixes https://github.com/ruby/prism/issues/2074

cc @enebo

### DIFF
--- a/test/prism/unescape_test.rb
+++ b/test/prism/unescape_test.rb
@@ -230,6 +230,8 @@ module Prism
       else
         assert_equal expected.bytes, actual.bytes, message
       end
+    rescue Exception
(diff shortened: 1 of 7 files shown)


test/prism/unescape_test.rb
```diff
@@ -230,6 +230,8 @@ def assert_unescape(context, escape)
       else
         assert_equal expected.bytes, actual.bytes, message
       end
+    rescue Exception
+      binding.irb
```
eregon, line 234: Leftover debug code?

kddnewton: This was cleaned up later

PR #2577, opened 2024-03-12
Change `frozen_string_literal` to be a tri-state

casperisfine:
An explicit `false` is not equivalent to the comment being missing, because the default can be switched with a runtime flag:

```bash
$ ruby --enable-frozen-string-literal -e 'p "foo".frozen?'
true
```

NB: I chose to make it a second boolean flag as I'm not sure what the API compatibility guarantees are, but maybe a three state enum might be better?

@kddnewton

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -665,7 +665,9 @@ flags:
       - name: FORCED_BINARY_ENCODING
         comment: "internal bytes forced the encoding to binary"
       - name: FROZEN
-        comment: "frozen by virtue of a `frozen_string_literal` comment"
+        comment: "frozen by virtue of a `frozen_string_literal: true` comment or `--enable-frozen-string-literal`"
+      - name: MUTABLE
+        comment: "mutable by virtue of a `frozen_string_literal: false` comment or `--disable-frozen-string-literal`"
     comment: Flags for string nodes.
   - name: SymbolFlags
     values:
--- a/include/prism/options.h
+++ b/include/prism/options.h
@@ -14,6 +14,22 @@
 #include <stdint.h>
 
 /**
+ * String literals should be made frozen.
+ */
+#define PM_OPTIONS_FROZEN_STRING_LITERAL_DISABLED   ((int8_t) -1)
+
+/**
+ * String literals may be frozen or mutable depending on the implementation
+ * default.
+ */
+#define PM_OPTIONS_FROZEN_STRING_LITERAL_UNSET      ((int8_t)  0)
+
+/**
+ * String literals should be made mutable.
+ */
+#define PM_OPTIONS_FROZEN_STRING_LITERAL_ENABLED    ((int8_t)  1)
+
+/**
  * A scope of locals surrounding the code that is being parsed.
  */
 typedef struct pm_options_scope {
@@ -85,8 +101,14 @@ typedef struct {
     /** A bitset of the various options that were set on the command line. */
     uint8_t command_line;
 
-    /** Whether or not the frozen string literal option has been set. */
-    bool frozen_string_literal;
+    /**
+    * Whether or not the frozen string literal option has been set.
+    * May be:
+    *  - PM_OPTIONS_FROZEN_STRING_LITERAL_DISABLED
+    *  - PM_OPTIONS_FROZEN_STRING_LITERAL_ENABLED
+    *  - PM_OPTIONS_FROZEN_STRING_LITERAL_UNSET
+    */
+    int8_t frozen_string_literal;
 } pm_options_t;
 
 /**
--- a/src/prism.c
+++ b/src/prism.c
@@ -5836,8 +5836,13 @@ pm_string_node_create_unescaped(pm_parser_t *parser, const pm_token_t *opening, 
     pm_string_node_t *node = PM_ALLOC_NODE(parser, pm_string_node_t);
     pm_node_flags_t flags = 0;
 
-    if (parser->frozen_string_literal) {
-        flags = PM_NODE_FLAG_STATIC_LITERAL | PM_STRING_FLAGS_FROZEN;
+    switch (parser->frozen_string_literal) {
+        case PM_OPTIONS_FROZEN_STRING_LITERAL_DISABLED:
+            flags = PM_STRING_FLAGS_MUTABLE;
+            break;
+        case PM_OPTIONS_FROZEN_STRING_LITERAL_ENABLED:
+            flags = PM_NODE_FLAG_STATIC_LITERAL | PM_STRING_FLAGS_FROZEN;
+            break;
     }
 
     *node = (pm_string_node_t) {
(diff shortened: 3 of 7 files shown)


kddnewton:
Yeah I'd prefer a uint8_t since it has multiple states as opposed to two bools if you don't mind

casperisfine:
Alright, doing that then.

src/prism.c
```diff
@@ -6236,8 +6241,13 @@ pm_symbol_node_to_string_node(pm_parser_t *parser, pm_symbol_node_t *node) {
     pm_string_node_t *new_node = PM_ALLOC_NODE(parser, pm_string_node_t);
     pm_node_flags_t flags = 0;
 
-    if (parser->frozen_string_literal) {
-        flags = PM_NODE_FLAG_STATIC_LITERAL | PM_STRING_FLAGS_FROZEN;
+    switch (parser->frozen_string_literal) {
+        case 1:
+            flags = PM_NODE_FLAG_STATIC_LITERAL | PM_STRING_FLAGS_FROZEN;
+            break;
+        case 2:
```
eregon, line 6248:
```suggestion
        case -1:
```
Right?

casperisfine: Yes 🤦

casperisfine: Well, `0` but similarly broken.

Also perhaps I don't understand the meaning of `PM_NODE_FLAG_STATIC_LITERAL` correctly, but I think it applies regardless of frozen string literals.

As long as there is no interpolation, I think this flag should be set cause the string can be created eagerly. Frozen string literals will just define whether we have to `dupstring` or `putobject`.

casperisfine:
Alright I think I fixed all issues. I changed `unspecified` to be `0` so it's easier to initialize strucs.

Also I use `int8_t` not `uint8_t` as I think it makes a bit more sense to use `-1 / 0 / 1` than `0 / 1 / 2`.

kddnewton:
Okay I'm trying to get my head around how to compile this and I'm now not so sure about these changes. I'm going to lay out a bunch of information, and maybe you can find where I'm missing something? Here's a table of the various combinations of options in Ruby:

| `# frozen_string_literal` | `--{enable,disable}-frozen-string-literal` | `"".frozen?` |
| --- | --- | --- |
| unspecified | unspecified | `false` |
| `false` | unspecified | `false` |
| `true` | unspecified | `true` |
| unspecified | `disable` | `false` |
| `false` | `disable` | `false` |
| `true` | `disable` | `true` |
| unspecified | `enable` | `true` |
| `false` | `enable` | `false` |
| `true` | `enable` | `true` |

We need to map these options from the combination of `pm_parser_t#frozen_string_literal`, `pm_options_t#frozen_string_literal`, and `pm_string_node_t#flags`.

From the table above, it seems like not specifying a CLI option and passing `--disable-frozen-string-literal` both result in the same triplet. So I think `pm_options_t#frozen_string_literal` can be a boolean. Then looking at only the bottom 6 (because we consider the top 6 two triplets of the same), the only difference is an unspecified comment changes the value based on the CLI. In that case, if we set `pm_parser_t#frozen_string_literal` to `pm_options_t#frozen_string_literal` and then change `pm_parser_t#frozen_string_literal` based on the comment, it seems like we'll get the right values.

So I guess what I'm saying is I'm not understanding what we're changing here.

casperisfine:
Alright, I'll try to explain. So first for the context, I'm working on changing the default value of `frozen_string_literal`: https://bugs.ruby-lang.org/issues/20205

Which I'm implementing in https://github.com/Shopify/ruby/pull/549 for now.

So MRI now needs to know if a literal string is `frozen/mutable/unspecified`, and each will result in a different instruction ('putobject/putstring/putchilledstring' respectively).

Does that make sense now?

src/prism.c
```diff
@@ -5836,8 +5836,13 @@ pm_string_node_create_unescaped(pm_parser_t *parser, const pm_token_t *opening,
     pm_string_node_t *node = PM_ALLOC_NODE(parser, pm_string_node_t);
     pm_node_flags_t flags = 0;
 
-    if (parser->frozen_string_literal) {
-        flags = PM_NODE_FLAG_STATIC_LITERAL | PM_STRING_FLAGS_FROZEN;
+    switch (parser->frozen_string_literal) {
+        case -1:
+            flags = PM_NODE_FLAG_STATIC_LITERAL | PM_STRING_FLAGS_MUTABLE;
```
kddnewton, line 5841: It shouldn't be seen as a static literal if it is mutable. This changes stuff like `["foo"]` which will `newarray`/`duparray` depending on the contents.

src/prism.c
```diff
@@ -6236,8 +6241,13 @@ pm_symbol_node_to_string_node(pm_parser_t *parser, pm_symbol_node_t *node) {
     pm_string_node_t *new_node = PM_ALLOC_NODE(parser, pm_string_node_t);
     pm_node_flags_t flags = 0;
 
-    if (parser->frozen_string_literal) {
-        flags = PM_NODE_FLAG_STATIC_LITERAL | PM_STRING_FLAGS_FROZEN;
+    switch (parser->frozen_string_literal) {
+        case -1:
+            flags = PM_NODE_FLAG_STATIC_LITERAL | PM_STRING_FLAGS_MUTABLE;
```
kddnewton, line 6246: Same here, no static literal.

config.yml
```diff
@@ -665,7 +665,9 @@ flags:
       - name: FORCED_BINARY_ENCODING
         comment: "internal bytes forced the encoding to binary"
       - name: FROZEN
-        comment: "frozen by virtue of a `frozen_string_literal` comment"
+        comment: "frozen by virtue of a `frozen_string_literal: true` comment"
```
kddnewton, line 668:
```suggestion
        comment: "frozen by virtue of a `frozen_string_literal: true` comment or `--enable-frozen-string-literal`"
```

config.yml
```diff
@@ -665,7 +665,9 @@ flags:
...
+      - name: MUTABLE
+        comment: "mutable by virtue of a `frozen_string_literal: false` comment"
```
kddnewton, line 670:
```suggestion
        comment: "mutable by virtue of a `frozen_string_literal: false` comment or `--disable-frozen-string-literal`"
```

include/prism/options.h
```diff
@@ -85,8 +85,13 @@ typedef struct {
...
 
-    /** Whether or not the frozen string literal option has been set. */
-    bool frozen_string_literal;
+    /**
+    * Whether or not the frozen string literal option has been set.
+    *   -1 is explicitly disabled.
+    *    0 is not set (default).
+    *    1 is explicitly enabled.
+    */
+    int8_t frozen_string_literal;
```
kddnewton, line 94: Can you make some constants so we don't rely on the specific values everywhere? Something like:

```c
static const int8_t PM_OPTIONS_FROZEN_STRING_LITERAL_DISABLED = -1;
static const int8_t PM_OPTIONS_FROZEN_STRING_LITERAL_UNSET = 0;
static const int8_t PM_OPTIONS_FROZEN_STRING_LITERAL_ENABLED = 1;
```

kddnewton (changes requested):
Ahh I think I understand now, thanks for the explanation. In this case I think I understand that the table is now:

So if I understand your proposal correctly, it would change the table to be:

| `# frozen_string_literal` | `--{enable,disable}-frozen-string-literal` | `""` |
| --- | --- | --- |
| unspecified | unspecified | chilled |
| `false` | unspecified | mutable |
| `true` | unspecified | frozen |
| unspecified | `disable` | mutable |
| `false` | `disable` | mutable |
| `true` | `disable` | frozen |
| unspecified | `enable` | frozen |
| `false` | `enable` | mutable |
| `true` | `enable` | frozen |

And for compiling we should check:

* if `FROZEN` -> frozen
* if `MUTABLE` -> mutable
* else -> chilled

Is that right?

I've added a couple of minor comments on the code itself.

include/prism/options.h
casperisfine: I made them `#define` otherwise they can't be used in `switch/case`.

kddnewton (approved):

PR #2599, opened 2024-03-13
Warn for maximum number variables

kddnewton:
This one will impact implementations. Now `NumberedReferenceReadNode` has `0` as the number if the value is too large. This matches CRuby's expectations. We also use this for adding an appropriate warning.

cc @enebo @eregon @seven1m

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -247,6 +247,7 @@ warnings:
   - FLOAT_OUT_OF_RANGE
   - INTEGER_IN_FLIP_FLOP
   - INVALID_CHARACTER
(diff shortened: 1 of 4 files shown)


eregon:
It would be nicer if `$4294967296` is a parse error instead of a warning.
Then the value of the node field wouldn't matter much, because errors should be handled first.
But I guess that would need a CRuby ticket, because currently it is indeed just a warning:
```
$ ruby -ve 'p $4294967296'
-e:1: warning: `$4294967296' is too big for a number variable, always nil
nil
```

PR #2607, opened 2024-03-17
Use `require_relative` in the Prism codebase

koic:
If there are many searches in the `$LOAD_PATH` in the user environment, require will perform unnecessary searches that are not needed. In contrast, `require_relative` is efficient because it uses a relative path.


koic:
The CI failure started occurring on Ruby's master branch. Below is the bug report for that issue:
https://bugs.ruby-lang.org/issues/20343

kddnewton (approved):

PR #2613, opened 2024-03-19
[Compatibility] Improve printf format

hasumikin:
For better compatibility, this patch suggests using __MINGW_PRINTF_FORMAT that GCC provides for MinGW environment.


kddnewton (approved):
Great!

PR #2622, opened 2024-03-23
Use `Sexp#line_max` not `Sexp#max_line` for RubyParser translator

presidentbeef:
Because RubyParser's `Sexp` uses `method_missing` for [search/delete](https://github.com/seattlerb/sexp_processor/blob/4081209706e81317fde170b327b73f2a69df4507/lib/sexp.rb#L253-L255), it swallows up calls to non-existent methods. That makes it harder to catch issues like this.

Took me a little while to convince myself that all the tests were actually running and passing, but `max_line` really was `nil` the whole time.

Happily, the tests still pass with `line_max` being set appropriately.

### DIFF
--- a/lib/prism/translation/ruby_parser.rb
+++ b/lib/prism/translation/ruby_parser.rb
@@ -192,19 +192,19 @@ module Prism
 
           if node.opening == "("
             result.line = node.opening_loc.start_line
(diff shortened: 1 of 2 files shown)


kddnewton (approved):
Oh wow, that's amazing we got this far without noticing.

PR #2624, opened 2024-03-25
Avoid using `else` for `peek_at` function in src/prism.c

francktrouillez:
I was reading `src/prism.c` and found out this situation.

This allows to avoid using `else` since the `if` branch has a return statement. This case occurred in the `peek_at` function in `src/prism.c`.


kddnewton:
This is stylistic and doesn't have any impact on the resulting code. The current code is fine as it is.

PR #2668, opened 2024-04-02
Fix parser translation's heredoc whitespace calculation

### LINKED ISSUE #2374: Incompatibility with squiggly heredoc in `Prism::Translation::Parser`

st0012:
Given this example:

```rb
<<~HEREDOC
  #{1} a
HEREDOC
```

Both the parser gem and Prism's translation layer would generate the following AST:

```
s(:dstr,
  s(:begin,
    s(:int, 1)),
  s(:str, " a\n"))
```

However, the parser gem inserts a empty string node into this node's location, like:

```
<Parser::Source::Map::Heredoc:0x0000000104ce73b8
 @expression=#<Parser::Source::Range (string) 0...10>,
 @heredoc_body=#<Parser::Source::Range (string) 11...20>,
 @heredoc_end=#<Parser::Source::Range (string) 20...27>,
 @node=s(:dstr,
  s(:str, ""),
  s(:begin,
    s(:int, 1)),
  s(:str, " a\n"))>
```

This is required to calculate the correct whitespace for the heredoc body.

We need to adjust the translation layer to account for this.

With this fix, we also won't need to ignore the tilde heredoc fixture anymore.

Closes #2374

### DIFF
--- a/lib/prism/translation/parser/compiler.rb
+++ b/lib/prism/translation/parser/compiler.rb
@@ -953,8 +953,35 @@ module Prism
         def visit_interpolated_string_node(node)
           if node.heredoc?
             children, closing = visit_heredoc(node)
+            opening = token(node.opening_loc)
 
-            return builder.string_compose(token(node.opening_loc), children, closing)
+            start_offset = node.opening_loc.end_offset + 1
+            end_offset = node.parts.first.location.start_offset
+
+            # In the below case, the offsets should be the same:
+            #
+            # <<~HEREDOC
+            #   a #{b}
+            # HEREDOC
+            #
+            # But in this case, the end_offset would be greater than the start_offset:
+            #
+            # <<~HEREDOC
+            #   #{b}
+            # HEREDOC
+            #
+            # So we need to make sure the result node's heredoc range is correct, without updating the children
+            result = if start_offset < end_offset
+              # We need to add a padding string to ensure that the heredoc has correct range for its body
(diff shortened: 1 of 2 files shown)


andyw8:
Possibly related? https://github.com/rubocop/rubocop/issues/12834

kddnewton (approved):

Issue #2688, opened 2024-04-09
Failure when calling `transpose` in `Prism::Translation::Parser`


andyw8:
https://github.com/ruby/prism/pull/2602 seems to have introduced a regression. Given a file containing:

```
"a
b\
c"
```

running RuboCop with `ParserEngine: parser_prism` fails with:

```
element size differs (2 should be 3)
/home/spin/.bundle/shopify/ruby-3.3.0/bundler/gems/prism-3ae54cb4dbb3/lib/prism/translation/parser/compiler.rb:1521:in `transpose'
/home/spin/.bundle/shopify/ruby-3.3.0/bundler/gems/prism-3ae54cb4dbb3/lib/prism/translation/parser/compiler.rb:1521:in `visit_string_node'
```

andyw8:
FYI @koic

andyw8:
(I bsected to identify this commit. It works with the prior commit, 29b05579a8a966964fcd351ec630b38351ec0ef6).

andyw8:
This seems to be resolved in latest Prism (v0.27.0).

PR #2689, opened 2024-04-09
WIP: Fix `transpose` failure for `Prism::Translation::Parser`

andyw8:
Closes https://github.com/ruby/prism/issues/2688

Added a failing test case fixture to illustrate the problem:

```
Error: test_dstring.txt(Prism::ParserTest): IndexError: element size differs (2 should be 3)
/Users/andyw8/src/github.com/ruby/prism/lib/prism/translation/parser/compiler.rb:1516:in `transpose'
/Users/andyw8/src/github.com/ruby/prism/lib/prism/translation/parser/compiler.rb:1516:in `visit_string_node'
```

For the fixture added, the mismatch is due to the differing number of elements:

`node.content.lines:   ["\n", "b\\\n", "c\n"]`

and

`node.unescaped.lines: ["\n", "bc\n"]`


andyw8:
(I've pushed an in-progress attempt from pairing with @vinistock)

PR #2693, opened 2024-04-11
Add debug gem

andyw8:
To make debugging a little more convenient.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,7 @@ platforms = %i[mri mswin mingw x64_mingw]
 gem "ffi", platform: platforms
 gem "parser", platform: platforms
 gem "ruby_parser", platform: platforms
+gem "debug", platforms: platforms
 gem "benchmark-ips"
 
 group :memcheck do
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     - name: Set up Ruby
       uses: ruby/setup-ruby@v1
       with:
-        ruby-version: "2.7.0"
+        ruby-version: "3.3"
```
andyw8, line 33: (unsure if this was deliberately set to an old Ruby?)

It was failing with:

```
 ERROR:  Error installing bundler:
  	There are no versions of bundler (= 2.5.5) compatible with your Ruby & RubyGems
  	bundler requires Ruby version >= 3.0.0. The current ruby version is 2.7.0.0.
```

kddnewton: This is very intentional, it needs to be able to run with BASERUBY, which is the ruby that is compiling the current Ruby. This cannot change.

kddnewton:
Do you need some features that aren't bundled with ruby 3.3? Or can you just rely on the bundled version. I'm reticent to add another dependency that will need to be updated.

andyw8:
@kddnewton it's not so much about the feature. If we run the a via `bundle exec`, e.g. as the Ruby LSP Code Lens test runner does, then bundled gems such as `debug` aren't available.

PR #2705, opened 2024-04-14
`extconf.rb`: Send `RbConfig` preferences to `#make` via env vars

ParadoxV5:
These record the preferences Ruby was built with.
As a library that aims to be canonical, Prism should therefore match those, especially as a part of a C extension gem.

Using `RbConfig` also enables cross-compiling using https://bugs.ruby-lang.org/issues/20345 or the mentioned `rake-compiler`.

Of course, `MAKEDIRS` and `RMALL` are redundant since it’s the same for all of *nix and MSYS.
But if [`mkmf` supports challengers who venture without popular gear](https://github.com/ruby/ruby/blob/v3_3_0/lib/mkmf.rb#L2092-L2099), so should we.


Makefile
```diff
@@ -8,13 +8,16 @@ ECHO1 = $(V:1=@ :)
 ECHO = $(ECHO1:0=@ echo)
 FUZZ_OUTPUT_DIR = $(shell pwd)/fuzz/output
 
-SOEXT := $(shell ruby -e 'puts RbConfig::CONFIG["SOEXT"]')
+SOEXT ?= $(shell ruby -e 'puts RbConfig::CONFIG["SOEXT"]')
```
ParadoxV5, line 11: C extensions actually use `DLEXT`. But, considering this is the `Makefile` for the Prism core lib, _not_ the Prism extension gem, I kept it as `SOEXT`.

Both `DLEXT` and `SOEXT` are `so` on Linux Ruby, but Windows RubyInstaller has `so` for `DLEXT` – matching Linux’s – and `dll` for `SOEXT` – matching Windows conventions.

Makefile
```diff
@@ -8,13 +8,16 @@ ECHO1 = $(V:1=@ :)
...
 
 CPPFLAGS := -Iinclude $(CPPFLAGS)
 CFLAGS := -g -O2 -std=c99 -Wall -Werror -Wextra -Wpedantic -Wundef -Wconversion -Wno-missing-braces -fPIC -fvisibility=hidden $(CFLAGS)
```
ParadoxV5, line 14: Both my Ubuntu Ruby and Windows RubyInstaller includes `-O3` in the `CFLAGS`, while this `Makefile` specifies `-O2`.
Could this matter?

ext/prism/extconf.rb
```diff
@@ -40,17 +40,22 @@ def generate_templates
...
+require "rbconfig"
+
 # Runs `make` in the root directory of the project. Note that this is the
 # `Makefile` for the overall project, not the `Makefile` that is being generated
 # by this script.`
 def make(target)
   Dir.chdir(File.expand_path("../..", __dir__)) do
-    system(RUBY_PLATFORM.include?("openbsd") ? "gmake" : "make", target, exception: true)
+    system(
+      RbConfig::CONFIG.slice(*%w[SOEXT CPPFLAGS CFLAGS CC AR ARFLAGS MAKEDIRS RMALL]), # env
```
ParadoxV5, line 51: Should we extract this operation out so it’s not re-`#slice`d on every call?

Makefile
```diff
@@ -8,13 +8,16 @@ ECHO1 = $(V:1=@ :)
...
-CC := cc
+CC ?= cc
 WASI_SDK_PATH := /opt/wasi-sdk
 
+MAKEDIRS ?= mkdir -p
+RMALL ?= rm -f -r
+
```
ParadoxV5, line 20: Of course, `MAKEDIRS` and `RMALL` are redundant since it’s the same for all of *nix and MSYS.
But if [`mkmf` supports challengers who venture without popular gear](https://github.com/ruby/ruby/blob/v3_3_0/lib/mkmf.rb#L2092-L2099), so should we.

ParadoxV5:
`RbConfig` also these entries
```ruby
{
  "LIBEXT"=>"a",
  "LDSHARED"=>"gcc -shared",
  "DLDSHARED"=>"gcc -shared",
  "OUTFLAG"=>"-o ",
  "COUTFLAG"=>"-o "
}
```
but I’m uncertain how to integrate them.

ParadoxV5:
> (`clang-analyzer` CI failure)

I suppose they arose because the Ruby `RbConfig::CONFIG['CFLAGS']` enabled more diagnostics?
Not sure how it works.

Makefile
kddnewton: Would you suggest changing this? I'm not familiar enough with windows dev to know if it would break something.

Makefile
kddnewton: I don't think it's going to end up mattering truthfully — O2 keeps it from getting very long compile times.

ext/prism/extconf.rb
kddnewton: That's okay, it's not a big deal

kddnewton (approved):
This is great, can you just rebase? I fixed the clang-analyzer issues

Makefile
ParadoxV5: I briefly reviewed `extconf.rb` and would recomment _against_ changing this, at least in this PR for scope reasons.

I have experimented before.
Extension Gems use `so` even on Windows because, appearently, the Ruby ecosystem _expects `.so` and ignores `.dll`_.
But OTOH, MinGW `-l` expects `.dll` to be the implicit file extension _but not `.so`_.

__TL;DR: Changing file extensions can break build pipelines.__

kddnewton:
Sorry, if you don't mind rebasing one more time, looks like I had a different version of clang analyzer running locally.

kddnewton (approved):

ext/prism/extconf.rb
ParadoxV5: Actually, on hindsight, should we send `RbConfig::MAKEFILE_CONFIG` (unexpanded `Makefile` includes) over rather than `RbConfig::CONFIG` (expanded `Makefile` values)?

kddnewton: I assume it wouldn't make a difference right?

ParadoxV5: The difference would be if the `RbConfig::MAKEFILE_CONFIG` values were expanded from variables that
* our hand-written `Makefile` does not cover / cover differently, or
* interpret users’ env vars on a finer level

Makefile
eregon: On macOS one cannot link a .bundle against another .bundle AFAIK, so SOEXT is correct for libprism.

PR #2706, opened 2024-04-14
Avoid shelling *nix commands

ParadoxV5:
* Sibling of #2705

Some of the commands used in the `Makefile` are not covered by `RbConfig`.
Coïncidentally, they’re all `$(shell …)` calls.

This PR replaces them with alternatives that do not assume the availability of *nix commands.


Makefile
```diff
@@ -15,8 +15,8 @@ CFLAGS := -g -O2 -std=c99 -Wall -Werror -Wextra -Wpedantic -Wundef -Wconversion
 CC := cc
 WASI_SDK_PATH := /opt/wasi-sdk
 
-HEADERS := $(shell find include -name '*.h')
-SOURCES := $(shell find src -name '*.c')
```
ParadoxV5, line 19: `Makefile` doesn’t support recursive `wildcard`, so shelling out was accustomed.
The alternative is to write a custom function.

I’m fine with abandoning this change if it’s not worth to not shell-out.

For reference, `mkmf`-generated `Makefile`s does __not__ expand recursively; that is, `mkmf` does not support source code subfolders.

Makefile
```diff
@@ -6,7 +6,7 @@ Q1 = $(V:1=)
 Q = $(Q1:0=@)
 ECHO1 = $(V:1=@ :)
 ECHO = $(ECHO1:0=@ echo)
-FUZZ_OUTPUT_DIR = $(shell pwd)/fuzz/output
+FUZZ_OUTPUT_DIR = $(CURDIR)/fuzz/output
```
ParadoxV5, line 9: `CURDIR` is technically different from `shell pwd` in that the latter isn’t stuck constant if the `Makefile` somehow `cd`s halfway in the script.
That said, do we really want the build to base on the working directory rather than fixed at the repo root?

Makefile
```diff
@@ -15,8 +15,8 @@ CFLAGS := -g -O2 -std=c99 -Wall -Werror -Wextra -Wpedantic -Wundef -Wconversion
 CC := cc
 WASI_SDK_PATH := /opt/wasi-sdk
 
-HEADERS := $(shell find include -name '*.h')
-SOURCES := $(shell find src -name '*.c')
+HEADERS := $(wildcard include/*.h include/*/*.h include/*/*/*.h')
+SOURCES := $(wildcard src/*.c src/*/*.c)
```
ParadoxV5, line 19: `Makefile` doesn’t support recursive `wildcard`, so shelling out was accustomed.
The alternative is to write a custom function.

I’m fine with abandoning this change if it’s not worth to not shell-out.

For reference, `mkmf`-generated `Makefile`s does __not__ expand recursively; that is, `mkmf` does not support source code subfolders.

kddnewton (approved):

PR #2709, opened 2024-04-17
Add a reflection API for determining the fields of a node

kddnewton:


ext/prism/extension.c
```diff
@@ -987,6 +987,18 @@ parse_success_p(int argc, VALUE *argv, VALUE self) {
     return result;
 }
 
+/**
+ * call-seq:
+ *   Prism::parse_failure?(source, **options) -> Array
```
froydnj, line 992: Presumably this is not correct, given that the below code only returns true or false.

ext/prism/extension.c
```diff
@@ -1008,6 +1020,18 @@ parse_file_success_p(int argc, VALUE *argv, VALUE self) {
     return result;
 }
 
+/**
+ * call-seq:
+ *   Prism::parse_file_failure?(filepath, **options) -> Array
```
froydnj, line 1025: Same comment here.

PR #2711, opened 2024-04-18
Disable GNU `make` implicit variables and Quiet `ar` with `ARFLAGS` rather than `> null`

ParadoxV5:
`CC`, `AR` and `ARFLAGS` (and also `RM` and `CXX`) are all [implicit variables in GNU `make`](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html).
* `CC` and `AR` already defaults to `cc` and `ar`, which means `CC ?= cc` does nothing *since before I started tinkering*.
* `ARFLAGS` defaults to `-rv`, where `v` stands for `v`erbose.
  * https://stackoverflow.com/a/40516532 hints that changing it to `r` avoids printing to `null`,
    and *printing to `/dev/null` was the final obstacle for me to build nightly Prism on my [MSYS-less Windows setup](https://github.com/ParadoxV5/ruby-mingw-make)!*

Rather than adapting the `Makefile` and `extconf.rb` around them, I disabled them with `--no-builtin-variables`. While this flag may be GNU-exclusive,  the `README` implies that we only support GNU `make`.
* `CURDIR` from #2706 is also possibly GNU-exclusive.
* The popularity of GNU means it’s a-okay to be toolset-specific rather than platform-specific.

---

* Previously conflicted with #2716

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ CFLAGS := -g -O2 -std=c99 -Wall -Werror -Wextra -Wpedantic -Wundef -Wconversion 
 JAVA_WASM_CFLAGS := -g -Oz -std=c99 -Wall -Werror -Wextra -Wpedantic -Wundef -Wconversion -Wno-missing-braces -fPIC -fvisibility=hidden -Wimplicit-fallthrough $(JAVA_WASM_CFLAGS)
 CC ?= cc
 AR ?= ar
+ARFLAGS ?= -r$(V0:1=v)
 WASI_SDK_PATH := /opt/wasi-sdk
 
 MAKEDIRS ?= mkdir -p
@@ -38,7 +39,7 @@ build/libprism.$(SOEXT): $(SHARED_OBJECTS)
 
 build/libprism.a: $(STATIC_OBJECTS)
 	$(ECHO) "building $@ with $(AR)"
-	$(Q) $(AR) $(ARFLAGS) $@ $(STATIC_OBJECTS) $(Q1:0=>/dev/null)
+	$(Q) $(AR) $(ARFLAGS) $@ $(STATIC_OBJECTS)
 
 javascript/src/prism.wasm: Makefile $(SOURCES) $(HEADERS)
 	$(ECHO) "building $@"


Makefile
```diff
@@ -12,7 +12,7 @@ SOEXT ?= $(shell ruby -e 'puts RbConfig::CONFIG["SOEXT"]')
 
 CPPFLAGS := -Iinclude $(CPPFLAGS)
 CFLAGS := -g -O2 -std=c99 -Wall -Werror -Wextra -Wpedantic -Wundef -Wconversion -Wno-missing-braces -fPIC -fvisibility=hidden $(CFLAGS)
-CC ?= cc
+ARFLAGS := -r
```
ParadoxV5, line 15: I introduced `?=` in #2705 to be explicit that the variable may be overwritten by env vars sent from `extconf.rb`.
But a quick test shows that `<implicit variable> ?= <anything>` doesn’t do anything because the variable was already set – implicitly by GNU `make`! :sweat_smile:
This is why this assignment must be either `=` or `:=` but not `?=`. It shouldn’t affect receiving the `rbconfig` `ARFLAGS` through the aforementioned env vars.

Makefile
```diff
@@ -36,7 +36,7 @@ build/libprism.$(SOEXT): $(SHARED_OBJECTS)
 
 build/libprism.a: $(STATIC_OBJECTS)
 	$(ECHO) "building $@"
-	$(Q) $(AR) $(ARFLAGS) $@ $(STATIC_OBJECTS) $(Q1:0=>/dev/null)
+	$(Q) $(AR) $(ARFLAGS) $@ $(STATIC_OBJECTS)
```
ParadoxV5, line 39: A quick text find shows that this was the only recipe where `Q1` is used, and the only other use of `Q1` is https://github.com/ruby/prism/blob/17762fa4c0b2c069c1a4283cb47bc5276d9dd0f5/Makefile#L6, which may now be flattenend. (I’m not aware if the `V`s and `Q`s are a convention.)

ParadoxV5:
> (Mac Ruby 2.7 CI failure)

Intriguing.
`CC` and co. should be set from the `rbconfig` in practice. Considering this was never an issue in my previous `Makefile` PRs, does this mean that the Mac Ruby 2.7 was set with `libtool` for `AR`, *nothing* for `ARFLAGS`, *and* `make` had a different default for Mac?
I do not know how to investigate a Mac problem.

*Easy solution: drop Ruby 2.7 support (we’re still in 0.x aren’t we?)*

kddnewton:
Unfortunately we cannot drop 2.7 support, as that is BASERUBY and we need it to build Ruby itself.

kddnewton:
> CC and AR already defaults to cc and ar, so setting CC to cc does nothing since before I started tinkering.

I'm not sure this is true. Before, we always set `CC` to `cc`, so it couldn't be overridden by env vars. Now it's going to be the same CC that Ruby is compiled with.

ParadoxV5:
> Before, we always set `CC` to `cc`, so it couldn't be overridden by env vars.

Like with `ARFLAGS` in the PR, I expect the original `CC := cc` to be overridable by env vars.

> Now it's going to be the same CC that Ruby is compiled with.

This though is intended.

Makefile
ParadoxV5: * [ ] TODO: circument __only__ for `ar -rv`; do not change for `libtool ` (such as used in that *one* failing CI)

ParadoxV5:
> Like with `ARFLAGS` in the PR, I expect the original `CC := cc` to be overridable by env vars.

Oof, I had it mixed up. It’s only overridable with `make KEY=VALUE`, not from env vars.

kddnewton:
I think I'd prefer to leave it because I'd like to support all of the Ruby versions in the range that we support, which includes patch revisions.

eregon:
Yeah agreed let's keep that workaround for existing releases and link to the relevant issues to make it easy to figure out from the code why that is needed

Makefile
ParadoxV5: For this matter, ba82ce9 introduces an acutal use for `V0`:
https://github.com/ruby/prism/blob/ba82ce961516d2495e94c496c8c0c8820344abcb/Makefile#L17

Makefile
ParadoxV5: Switched to `--no-builtin-variables`. Explicit is better than implicit anyways.

ext/prism/extconf.rb
```diff
@@ -51,6 +51,7 @@ def make(env, target)
     system(
       env,
       RUBY_PLATFORM.include?("openbsd") ? "gmake" : "make",
+      "--no-builtin-variables", # don't let GNU make implicit variables override variable fallbacks in the Makefile
```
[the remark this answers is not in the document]
ParadoxV5: This flag disables [implicit variables for GNU `make`](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html).
As noted on the previous posts (such as the updated [top post](#issue-2249483820)), those implicit variables are taking precedence over https://github.com/ruby/prism/blob/9bb871092d658bfc9137e21bf87ed466327ae719/Makefile#L15-L17, thus they – especially the new `ARFLAGS ?= -r$(V0:1=v)` – aren’t applied without this flag.

kddnewton:
@ParadoxV5 now that the build system has changed because it's not shelling out to make anymore, can you check if this is still necessary? If it is can you rebase?

Makefile
```diff
@@ -14,6 +14,7 @@ CPPFLAGS := -Iinclude $(CPPFLAGS)
 CFLAGS := -g -O2 -std=c99 -Wall -Werror -Wextra -Wpedantic -Wundef -Wconversion -Wno-missing-braces -fPIC -fvisibility=hidden $(CFLAGS)
 CC ?= cc
 AR ?= ar
+ARFLAGS ?= -r$(V0:1=v)
```
ParadoxV5, line 17: According to #2716, the original `?=`s were once a workaround for #2716.
That issue’s root cause’s separately fixed and 7c2461f purged its related `extconf.rb` code. Like this PR, I wonder if those lines are still necessary?

ParadoxV5 (commented):
> now that the build system has changed because it's not shelling out to make anymore, can you check if this is still necessary?

Interesting, @kddnewton.

At #3273, @mame was practically requesting Prism-gem support for MSVC, and it was approved by the decision to stop intercepting the platform-agnostic `mkmf`.
FWIW, MSVC is *also* not MSYS.

My goal was to support [MSYS (DevKit in RubyInstaller’s jargon)-less MinGW](https://github.com/ParadoxV5/ruby-mingw-make).
I built [the latest `HEAD`](https://github.com/ruby/prism/tree/cdf702b) and can confirm that, because the `RbConfig`-based `mkmk` absorbs all setup differences, that decision covers my case as well and it no longer requires this PR.

The only purpose left in this patch, then, is to improve the general code quality.

kddnewton (approved):

Issue #2716, opened 2024-04-19
Update makefile to use / fall back on built-in "ar"


forthrin:
```
$ gem install prism
<snip>
/bin/sh: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ar: No such file or directory
$ ack Toolchains
ext/prism/Makefile
99:AR = /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ar
$ which -a ar
/usr/bin/ar

# WORKAROUND
$ mkdir -p /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
$ ln -s $(which ar) /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/
Successfully installed prism-0.26.0
```

kddnewton:
To be clear, did this work with 0.25.0?

forthrin:
Running `gem update` daily for quite a while and first time this happened, so assume yes.

kddnewton:
Also @forthrin that value is coming from `RbConfig::CONFIG`. Is it possible that you compiled Ruby with an `ar` and then somehow removed it?

ParadoxV5:
> Also @forthrin that value is coming from `RbConfig::CONFIG`. Is it possible that you compiled Ruby with an `ar` and then somehow removed it?

⬆. The new release switched from `make`-default `AR` & etc. to those specified by `RbConfig::CONFIG`, so this sort of problem arise if the current setup differs that built Ruby with.

This might be the same issue as the *one* failing CI in #2711.
I’ll try some research on the Mac building environment.
In the mean time, `p RUBY_DESCRIPTION, RbConfig::CONFIG.slice('AR', 'ARFLAGS', 'CC')` data could help us with confirmations.

forthrin:
I'm on macOS 14. `gem` says it's a "default gem". I have never installed it manually or used it directly. So everything that has happened with it has done so by default behind the scenes without my intervention or knowledge.

```
["ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [arm64-darwin23]",
 {"AR"=>
   "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ar",
  "ARFLAGS"=>"rcu ",
  "CC"=>"clang"}]
```

[...]

PR #2725, opened 2024-04-22
Fall back to default AR and CC

kddnewton:


Makefile
```diff
@@ -13,6 +13,7 @@ SOEXT ?= $(shell ruby -e 'puts RbConfig::CONFIG["SOEXT"]')
 CPPFLAGS := -Iinclude $(CPPFLAGS)
 CFLAGS := -g -O2 -std=c99 -Wall -Werror -Wextra -Wpedantic -Wundef -Wconversion -Wno-missing-braces -fPIC -fvisibility=hidden $(CFLAGS)
 CC ?= cc
+AR ?= ar
```
ParadoxV5, line 16: As discovered in https://github.com/ruby/prism/pull/2711#discussion_r1569712785, `CC ?= cc` and `AR ?= ar` has no effect, whether run from `make` directly (because GNU `make` already preconfigures them to `cc` and `ar` respectively) or from `extconf.rb` (as it’ll override them with `rbconfig` values)
```suggestion
```

ext/prism/extconf.rb
```diff
@@ -40,15 +40,34 @@ def generate_templates
   end
 end
 
+# We're going to need to run `make` using prism's `Makefile`. We want to match
+# up as much of the configuration to the configuration that built the current
+# version of Ruby as possible.
 require "rbconfig"
+env = RbConfig::CONFIG.slice("SOEXT", "CPPFLAGS", "CFLAGS", "CC", "AR", "ARFLAGS", "MAKEDIRS", "RMALL")
```
ParadoxV5, line 47: * Resolves https://github.com/ruby/prism/pull/2705#discussion_r1564910429

ParadoxV5 (commented):
* Fixes #2716

---

`prism-0.26.0.gem` was accidentally checked in; the CI is also complaining about it.

---

I put https://github.com/ruby/prism/pull/2711#discussion_r1573048983 on hold for observing the progress of #2716 first.
In order to override GNU `make`’s default `ARFLAGS` without affecting `extconf.rb`, we’ll have to either
* send these preferences via CLI args rather than env vars (CLI: `make KEY=VALUE` rather than `KEY=VALUE make`), or
* rename the `ARFLAGS` variable in the `Makefile`

Makefile
kddnewton: I agree with you in theory, but something else must be going on here, because after lots of back and forth with @forthrin this is what finally resolved the issue. I'm assuming there's something else here going on.

PR #2727, opened 2024-04-23
Use GC-able symbols for AST

kddnewton:


templates/ext/prism/api_node.c.erb
```diff
@@ -108,17 +108,17 @@ pm_node_stack_pop(pm_node_stack_node_t **stack) {
 
 VALUE
 pm_ast_new(const pm_parser_t *parser, const pm_node_t *node, rb_encoding *encoding, VALUE source) {
-    ID *constants = xcalloc(parser->constant_pool.size, sizeof(ID));
+    ID *constants = xcalloc(parser->constant_pool.size, sizeof(VALUE));
```
froydnj, line 111: This should now be a value pointer?

kddnewton: Yup, updated

PR #2736, opened 2024-04-25
Remove need for Natalie patches

kddnewton:
@seven1m this should remove the need for those patches you're keeping in https://github.com/natalie-lang/natalie/tree/master/ext/prism-patches


seven1m (commented):
Wow, thank you!!

PR #2754, opened 2024-04-29
Specialize code unit computing for ASCII only sources

vinistock:
### Motivation

The new code unit APIs work really well and they achieved the behaviour we wanted in the Ruby LSP. Unfortunately, they are significantly slower than the much simpler computation for columns. Invoking `byteslice` repeatedly for large documents slows down the LSP by a lot.

Thankfully, we don't have to pay this price unnecessarily. If we specialize the implementation of the code unit API for ASCII-only sources, then we can take the simpler route which is about 5x faster on my machine.

### Implementation

I created a specialized version of `Source` called `ASCIISource`. Creating an equivalent `MultiByteSource` felt unnecessary, since we can have that as the default behaviour of `Source` (or we can flip if we prefer the faster option as the default).

To decide which kind of source to create, I introduced a `Source.create` method, which I honestly don't love, but also can't think of anything simpler.

### DIFF
--- a/ext/prism/extension.c
+++ b/ext/prism/extension.c
@@ -32,6 +32,7 @@ ID rb_option_id_frozen_string_literal;
 ID rb_option_id_line;
 ID rb_option_id_scopes;
 ID rb_option_id_version;
+ID rb_prism_source_id_for;
 
 /******************************************************************************/
 /* IO of Ruby code                                                            */
@@ -599,8 +600,7 @@ parse_lex_input(pm_string_t *input, const pm_options_t *options, bool return_nod
 
     VALUE source_string = rb_str_new((const char *) pm_string_source(input), pm_string_length(input));
     VALUE offsets = rb_ary_new();
-    VALUE source_argv[] = { source_string, LONG2NUM(parser.start_line), offsets };
-    VALUE source = rb_class_new_instance(3, source_argv, rb_cPrismSource);
+    VALUE source = rb_funcall(rb_cPrismSource, rb_prism_source_id_for, 3, source_string, LONG2NUM(parser.start_line), offsets);
 
     parse_lex_data_t parse_lex_data = {
         .source = source,
@@ -1378,6 +1378,8 @@ Init_prism(void) {
     rb_option_id_line = rb_intern_const("line");
     rb_option_id_scopes = rb_intern_const("scopes");
     rb_option_id_version = rb_intern_const("version");
+
+    rb_prism_source_id_for = rb_intern("for");
 
     /**
      * The version of the prism library.
--- a/lib/prism/parse_result.rb
+++ b/lib/prism/parse_result.rb
@@ -5,6 +5,14 @@ module Prism
   # conjunction with locations to allow them to resolve line numbers and source
   # ranges.
   class Source
+    # Create a new source object with the given source code. This method should
+    # be used instead of `new` and it will return either a `Source` or a
+    # specialized and more performant `ASCIISource` if no multibyte characters
+    # are present in the source code.
+    def self.for(source, start_line = 1, offsets = [])
+      source.ascii_only? ? ASCIISource.new(source, start_line, offsets): new(source, start_line, offsets)
+    end
+
     # The source code that this source object represents.
     attr_reader :source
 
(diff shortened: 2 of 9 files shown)


kddnewton:
Just to confirm you benchmarked this locally and it solves the issue for ascii only?

ext/prism/extension.c
```diff
@@ -599,8 +599,7 @@ parse_lex_input(pm_string_t *input, const pm_options_t *options, bool return_nod
 
     VALUE source_string = rb_str_new((const char *) pm_string_source(input), pm_string_length(input));
     VALUE offsets = rb_ary_new();
-    VALUE source_argv[] = { source_string, LONG2NUM(parser.start_line), offsets };
-    VALUE source = rb_class_new_instance(3, source_argv, rb_cPrismSource);
+    VALUE source = rb_funcall(rb_cPrismSource, rb_intern("create"), 3, source_string, LONG2NUM(parser.start_line), offsets);
```
kddnewton, line 602: Since we're coming from C, I think I would prefer here that we do the ascii_only check ourselves, because it will also force a coderange check, which will be nice.

That would be something like:

```c
VALUE source = rb_class_new_instance(3, source_argv, rb_str_is_ascii_only_p(source_string) ? rb_cPrismASCIISource : rb_cPrismSource);
```

lib/prism/parse_result.rb
```diff
@@ -5,6 +5,11 @@ module Prism
   # conjunction with locations to allow them to resolve line numbers and source
   # ranges.
   class Source
+    # Create a new source object with the given source code.
+    def self.create(source, start_line = 1, offsets = [])
```
kddnewton, line 9: How about we override `new` instead, and leave the callsites as they are? That should make it transparent to the consumer.

lib/prism/parse_result.rb
```diff
@@ -106,6 +111,19 @@ def find_line(byte_offset)
...
+  # cannot be applied to sources that include multibyte characters. Sources that
+  # include multibyte characters are represented by the Prism::Source class.
+  class ASCIISource < Source
+    # Specialized version of `code_units_column` that does not depend on
+    # `code_units_offset`, which is a more expensive operation. This is
+    # essentialy the same as `Prism::Source#column`.
+    def code_units_column(byte_offset, _encoding)
+      byte_offset - line_start(byte_offset)
+    end
+  end
```
kddnewton, line 125: Can we specialize a couple of other methods here as well?

```ruby
# Return the character offset for the given byte offset.
def character_offset(byte_offset)
  byte_offset
end

# Return the column number in characters for the given byte offset.
def character_column(byte_offset)
  byte_offset - line_start(byte_offset)
end

# Returns the offset from the start of the file for the given byte offset
# counting in code units for the given encoding.
#
# This method is tested with UTF-8, UTF-16, and UTF-32. If there is the
# concept of code units that differs from the number of characters in other
# encodings, it is not captured here.
def code_units_offset(byte_offset, encoding)
  byte_offset
end

# Returns the column number in code units for the given encoding for the
# given byte offset.
def code_units_column(byte_offset, encoding)
  byte_offset - line_start(byte_offset)
end
```

kddnewton (changes requested):
Some minor stylistic stuff, otherwise really good!

lib/prism/parse_result.rb
vinistock: If we override `new` here, won't `rb_class_new_instance` end up performing the ascii only check twice? Once in C and the other one in Ruby?

vinistock:
> Just to confirm you benchmarked this locally and it solves the issue for ascii only?

Yes, it does. About 5x faster than the version that handles multibyte.

lib/prism/parse_result.rb
kddnewton: `rb_class_new_instance` doesn't call `::new`, it calls `allocate` and `initialize` directly.

lib/prism/parse_result.rb
```diff
@@ -106,6 +114,45 @@ def find_line(byte_offset)
...
+    # Returns the column number in code units for the given encoding for the
+    # given byte offset.
+    def code_units_column(byte_offset, encoding)
+      byte_offset - line_start(byte_offset)
+    end
+
+    # Specialized version of `code_units_column` that does not depend on
+    # `code_units_offset`, which is a more expensive operation. This is
+    # essentialy the same as `Prism::Source#column`.
+    def code_units_column(byte_offset, encoding)
```
kddnewton, line 151: Looks like you've got this method twice

vinistock: Oops. Bad copy and paste.

ext/prism/extension.c
```diff
@@ -599,8 +599,7 @@ parse_lex_input(pm_string_t *input, const pm_options_t *options, bool return_nod
 
     VALUE source_string = rb_str_new((const char *) pm_string_source(input), pm_string_length(input));
     VALUE offsets = rb_ary_new();
-    VALUE source_argv[] = { source_string, LONG2NUM(parser.start_line), offsets };
-    VALUE source = rb_class_new_instance(3, source_argv, rb_cPrismSource);
+    VALUE source = rb_funcall(rb_cPrismSource, rb_intern("for"), 3, source_string, LONG2NUM(parser.start_line), offsets);
```
kddnewton, line 602: Last change, and then this looks good.

I don't want to `rb_intern("for")` every time this method is called, because it's called whenever we parse. Instead can you eagerly get the ID when prism is initialized? There's prior art here with stuff like `rb_option_id_command_line`.

vinistock: Done! We're still calling `rb_intern` in `api_node.c.erb`, not sure if you want to get rid of that one too.

kddnewton (approved):

lib/prism/parse_result.rb
```diff
@@ -106,6 +114,39 @@ def find_line(byte_offset)
...
 
+  # Specialized version of Prism::Source for source code that includes ASCII
+  # characters only. This class is used to apply performance optimizations that
+  # cannot be applied to sources that include multibyte characters. Sources that
+  # include multibyte characters are represented by the Prism::Source class.
+  class ASCIISource < Source
+    # Return the character offset for the given byte offset.
+    def character_offset(byte_offset)
+      byte_offset
+    end
```
eregon, line 125: I think we could use this logic directly in `Source` like `return byte_offset if @source.ascii_only?` (the result of `ascii_only?` is cached via String code range) or `return byte_offset if @ascii_only`.
Then there wouldn't be the concern that `Source.new` might be used instead of `Source.for` and that would then be needlessly slower.

OTOH, maybe users aren't supposed to use `Source.new`/`Source.for` directly anyway? (since e.g. `Prism.parse` takes a String and not a Source)

kddnewton: Yeah `Source` would be package-private if that were allowed in Ruby. I wanted us to specialize on creation of the object because the `source` field is also not meant to change.

PR #2762, opened 2024-05-02
Test support for C++

egiurleo:
This PR adds a test to CI ensuring that the Prism C API is functional in a C++ environment.

### DIFF
--- a/.github/workflows/cpp-bindings.yml
+++ b/.github/workflows/cpp-bindings.yml
@@ -1,0 +1,33 @@
+name: C++ bindings
+
+on:
+  push:
+    paths:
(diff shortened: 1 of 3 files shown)


kddnewton:
Thanks @egiurleo! I ended up going a slightly different way because after living with `operator_` being different in Ruby vs C it was weirding me out. So I renamed those fields to `binary_operator` in all layers here: https://github.com/ruby/ruby/pull/10758. 

I still would like to merge the github action and the test script though.

egiurleo:
@kddnewton Removed the commit that introduced `operator_` and just kept the testing.

PR #2763, opened 2024-05-03
Remove generics from sorbet types

kddnewton:


froydnj:
Can you elaborate on the rationale here, ideally with examples of things that are motivating this?  I assume it's because of ergonomic concerns when developing with them, but I'm curious if there's something that could be addressed on Sorbet's end.

kddnewton:
Yeah there are a couple of things going on here:

* There's an ecosystem problem. There are gems that use prism to parse that don't use sorbet. They inherit from `Compiler`. Then a third gem consumes the second gem, but uses sorbet. So sorbet complains that the subclass doesn't re-declare the type member, breaking their build.
* Because prism doesn't use the runtime, the `[]` method doesn't exist, so you can't actually use the generics. (I don't want to/can't depend on sorbet-runtime for this.)

Issue #1568, opened 2023-09-20
Type ConstantPathNode#child as ConstantReadNode for Java


eregon:
It seems to be the case, but it's not typed as such.
I've found I have a few `final String name = ((Nodes.ConstantReadNode) constantPathNode.child).name;` in TruffleRuby.
Due to this it's a bit awkward to deal with ConstantPathNode. And the same for `ConstantPathTargetNode`
Maybe the child should even be a `constant` and not a node if it's always just a literal constant anyway?

kddnewton:
It could be a missing node in the case of error recovery. But you can definitely cast it if you ensure you have no errors in the source.

eregon:
Ah yes, I vaguely remembered something like that but I couldn't find it when searching in this repo.
For Java I believe it's fine to just return a dummy node or a single MissingNode in case there is any errors, or maybe even throw a Java exception so we don't bother deserializing anything but errors.
Then we could type that field properly at least for Java nodes, which seems valuable enough.
I'll give that a try, and reopen this issue to track that, I hope you don't mind.

eregon:
@kddnewton One more thought, could that field be typed as a `constant?` instead of being a ConstantReadNode|MissingNode?
Then we'd save the extra node so we would save quite a bit of memory, and it could be nil/null for the error case (and that needs nothing special in Java, we'd just get a NullPointerException if it happens).

eregon:
Maybe the node is useful to get the location of the child name, but that's I believe always the same as ConstantPathNode#delimiter_loc.end_offset to ConstantPathNode#end_offset.

kddnewton:
Yeah we want the location of the child node for linters/formatters.

I think we really shouldn't do this for consistency's sake. It would be super confusing for this one piece of the AST to be different for just a single consumer. I'd like to publish the Java docs in addition to the C/Ruby/JS/Rust ones, and seeing a single deviation would be really strange. I'm going to close this one for now for that reason.

eregon:
> I'd like to publish the Java docs in addition to the C/Ruby/JS/Rust ones, and seeing a single deviation would be really strange. I'm going to close this one for now for that reason.

There are going to be significant differences for Java nodes/docs, e.g. Java nodes only have semantic fields, at least currently.
I think it'd be a good cleanup to type this field properly as ConstantReadNode for Java (and just e.g. null ParseResult#value if any error), but it's not high priority.

PR #2765, opened 2024-05-03
Change ConstantPathNode#child to ConstantPathNode#{name,name_loc}

kddnewton:
This has been requested for a long time, and I'm finally doing it now. Unfortunately this is a breaking change for all of the APIs.

I've added in a Ruby method for `#child` that is deprecated so that existing usage doesn't break, but for everyone else this is going to be a bit of a pain.


eregon:
Link: https://github.com/ruby/ruby/pull/10716

eregon (approved):
Looks good.
I found a discussion about this in https://github.com/ruby/prism/issues/1568 and this PR seems to pretty much address that, great!

PR #2773, opened 2024-05-05
Fix typecheck CI job

eregon:
* Update typecheck/Gemfile.lock to use the Bundler shipped with Ruby 3.3.1.

Previous failure: https://github.com/ruby/prism/actions/runs/8943853950/job/24569472569
which I guess it the reason for https://github.com/ruby/prism/commit/926c2bce25862a6b013ded252e7fbc0376540f1b


eregon:
Mmh https://github.com/ruby/prism/actions/runs/8957377538/job/24600406961?pr=2773 still gives:
```
/opt/hostedtoolcache/Ruby/3.3.1/x64/bin/bundle install --jobs 4
Your lockfile does not satisfy dependencies of "rbi", but the lockfile can't be
updated because frozen mode is set
```

And looking at the Gemfile lock there is:
https://github.com/ruby/prism/blob/e418265166975d879efefdf91e868ad35e68e593/gemfiles/typecheck/Gemfile.lock#L3-L4
https://github.com/ruby/prism/blob/e418265166975d879efefdf91e868ad35e68e593/gemfiles/typecheck/Gemfile.lock#L54-L55
I wonder where that constraint comes from.

eregon:
Removing the Gemfile.lock allows it to get to the next step:
https://github.com/ruby/prism/actions/runs/8957414293/job/24600490815?pr=2773

eregon:
https://github.com/Shopify/rbi/pull/298 might solve that constraint issue.

eregon:
It passed now: https://github.com/ruby/prism/actions/runs/8957490030/job/24600645943?pr=2773
So I think this is worth merging.

eregon:
> I don't want to delete the Gemfile.lock because I'm trying to use the latest versions of these things because we rely on it internally in Shopify.

Doesn't deleting the Gemfile.lock actually ensures the latest versions are used?

eregon:
Given the change in `rakelib/typecheck.rake` I wonder if the Prism from the repo was actually used with the checked-in Gemfile.lock, or maybe some other version was used (maybe the default gem since this runs on 3.3?).
In my local Gemfile.lock I have:
```
PATH
  remote: ../..
  specs:
    prism (0.28.0)
...
    rbi (0.1.12)
      prism (>= 0.18.0, < 0.28)
```
Which is kind of fun that Bundler/RubyGems doesn't complain about it.

kddnewton:
They've relaxed the version requirements on these things now, which is nice. There's still an RBI failure, but I'm re-enabling the other checks in CI.

PR #2774, opened 2024-05-05
Do not use values from RbConfig if we are not building a C extension

eregon:
* Notably, RbConfig::CONFIG["CC"] would be the GraalVM LLVM toolchain on TruffleRuby < 24.0 and we specifically do not want to use that as already explained in comments.

Related: #2705 #2725
cc FYI @ParadoxV5

### DIFF
--- a/ext/prism/extconf.rb
+++ b/ext/prism/extconf.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "rbconfig"
+
 if ARGV.delete("--help")
   print(<<~TEXT)
     USAGE: ruby #{$PROGRAM_NAME} [options]
@@ -40,26 +42,6 @@ def generate_templates
   end
 end
 
-# We're going to need to run `make` using prism's `Makefile`. We want to match
-# up as much of the configuration to the configuration that built the current
-# version of Ruby as possible.
-require "rbconfig"
-env = RbConfig::CONFIG.slice("SOEXT", "CPPFLAGS", "CFLAGS", "CC", "AR", "ARFLAGS", "MAKEDIRS", "RMALL")
-
-# It's possible that the Ruby that is being run wasn't actually compiled on this
-# machine, in which case the configuration might be incorrect. In this case
-# we'll need to do some additional checks and potentially fall back to defaults.
-if env.key?("CC") && !File.exist?(env["CC"])
-  env.delete("CC")
-  env.delete("CFLAGS")
-  env.delete("CPPFLAGS")
-end
-
-if env.key?("AR") && !File.exist?(env["AR"])
-  env.delete("AR")
-  env.delete("ARFLAGS")
-end
-
 # Runs `make` in the root directory of the project. Note that this is the
 # `Makefile` for the overall project, not the `Makefile` that is being generated
 # by this script.`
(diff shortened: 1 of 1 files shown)


ext/prism/extconf.rb
```diff
@@ -77,15 +57,38 @@ def make(env, target)
...
   File.write("Makefile", "all install clean:\n\t@#{RbConfig::CONFIG["NULLCMD"]}\n")
   return
 end
 
+# We're going to need to run `make` using prism's `Makefile`.
+# We want to use the same toolchain (compiler, flags, etc) to compile libprism.a and
+# the C extension since they will be linked together.
+# The C extension uses RbConfig, which contains values from the toolchain that built the running Ruby.
+require "rbconfig"
+env = RbConfig::CONFIG.slice("SOEXT", "CPPFLAGS", "CFLAGS", "CC", "AR", "ARFLAGS", "MAKEDIRS", "RMALL")
```
ParadoxV5, line 76: * Keep: `SOEXT`
* Don’t use: `CC` & `CFLAGS` & `CPPFLAGS`, probably `AR` & `ARFLAGS` too
* What about?: `MAKEDIRS`, `RMALL`

ParadoxV5 (commented):
:up:. We should’ve foreseen those mistakes when I originally sent my patch, so we didn’t need all of these follow-ups post release.

ext/prism/extconf.rb
eregon: I just moved that code around, I didn't change any logic for CRuby (just trying to clarify the comments a bit), so I would like to not change that in this PR.

FWIW reusing the same CC for building the extension and libprism.a seems best, so I think it makes sense to keep it there.
`RbConfig::CONFIG["CC"]` not existing seems very weird, how can any gem install of a gem with C extension work? Does mkmf have a fallback if `RbConfig::CONFIG["CC"]` does not exist?

kddnewton: It's possible that people are using all pre-compiled gems + a Ruby that was built on a different system.

ext/prism/extconf.rb
```diff
@@ -77,15 +57,38 @@ def make(env, target)
...
 # On non-CRuby we only need the shared library since we'll interface with it
 # through FFI, so we'll build only that and not the C extension. We also avoid
-# `require "mkmf"` as that prepends the LLVM toolchain to PATH on TruffleRuby,
-# but we want to use the native toolchain here since libprism is run natively.
+# `require "mkmf"` as that prepends the GraalVM LLVM toolchain to PATH on TruffleRuby < 24.0,
+# but we want to use the system toolchain here since libprism is run natively.
 if RUBY_ENGINE != "ruby"
   generate_templates
-  make(env, "build/libprism.#{RbConfig::CONFIG["SOEXT"]}")
+  soext = RbConfig::CONFIG["SOEXT"]
```
kddnewton, line 64: We're no longer requiring `rbconfig` in this file before this block, is that okay?

kddnewton (approved):
Just a small question first!

ext/prism/extconf.rb
eregon: Good catch, I'll move the `require`.
RubyGems requires `rbconfig` IIRC but best to not rely on that, and it could be an issue if someone has `RUBYOPT=--disable-gems` in their environment.

eregon: Fixed

ext/prism/extconf.rb
ParadoxV5: > > It's possible that people are using all pre-compiled gems + a Ruby that was built on a different system.

Is it time for us to start shipping prebuilt gems as well?

> Ah right, so in that case Prism happens to be the one to fail because it's the first C extension built from source on that target system? And so any other C extension built from source would fail too (because mkmf would use `RbConfig::CONFIG["CC"]`). It seems it's a user error in that case, they should not expect to be able to install native extensions from source without a native toolchain or with RbConfig pointing at a non-existing toolchain.

I previously thought about that too.

> I just wondered why the workaround is there

To resolve #2716. Perhaps @kddnewton just wanted to make the custom happy ASAP.

kddnewton: Yeah this was just to get a release out.

eregon: One last thought about this is while this code might help with `make build/libprism.a` I don't see how it could help for mkmf and building the C extension, and so `gem install prism` should still fail. That's a mystery.
Unless mkmf check for the existance of `RbConfig::CONFIG["AR"]` but I don't see anything about that in https://github.com/ruby/ruby/blob/v3_3_0/lib/mkmf.rb (CC_WRAPPER is checked but not CC/AR).
I think maybe in #2716 `RbConfig::CONFIG["CC"]` exists but `RbConfig::CONFIG["AR"]` does not. And then this workaround would help, since the root Makefile will use `ar` instead and C extensions very rarely use `AR`.

PR #2820, opened 2024-05-17
Reuse pm_buffer_append function in pm_buffer_append_zeroes

S-H-GAMELINKS:
pm_buffer_append  and pm_buffer_append_zeroes functions has similar code.
May be more simpler resue `pm_buffer_append` function in `pm_buffer_append_zeroes` function.

### DIFF
--- a/src/util/pm_buffer.c
+++ b/src/util/pm_buffer.c
@@ -84,10 +84,8 @@ pm_buffer_append(pm_buffer_t *buffer, const void *source, size_t length) {
  */
 void
 pm_buffer_append_zeroes(pm_buffer_t *buffer, size_t length) {
-    size_t cursor = buffer->length;
-    if (pm_buffer_append_length(buffer, length)) {
-        memset(buffer->value + cursor, 0, length);
-    }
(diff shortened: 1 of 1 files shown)


src/util/pm_buffer.c
```diff
@@ -84,10 +84,8 @@ pm_buffer_append(pm_buffer_t *buffer, const void *source, size_t length) {
  */
 void
 pm_buffer_append_zeroes(pm_buffer_t *buffer, size_t length) {
-    size_t cursor = buffer->length;
-    if (pm_buffer_append_length(buffer, length)) {
-        memset(buffer->value + cursor, 0, length);
-    }
+    size_t source = 0;
+    pm_buffer_append(buffer, (const void *) source, length);
```
kddnewton, line 88: Unfortunately, this is going to read off the end of `source` if `length` is greater than the number of bytes in a `size_t`, so I don't think we can do this here.

S-H-GAMELINKS: Thanks for the review!
I understood.

PR #2837, opened 2024-05-24
Add Prism::Node#deprecated method type

S-H-GAMELINKS:
Add `Prism::Node#deprecated` method type and supress `deprecated` method type error.

### DIFF
--- a/templates/sig/prism/node.rbs.erb
+++ b/templates/sig/prism/node.rbs.erb
@@ -19,6 +19,7 @@ module Prism
     def pretty_print: (untyped q) -> untyped
     def to_dot: () -> String
     def tunnel: (Integer line, Integer column) -> Array[Prism::node]
+    def deprecated: (*String) -> void
   end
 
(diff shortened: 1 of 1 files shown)


templates/sig/prism/node.rbs.erb
```diff
@@ -33,6 +33,7 @@ module Prism
     def fields: () -> Array[Prism::Reflection::Field]
     def inspect: () -> String
     def type: () -> Symbol
+    def deprecated: (*String) -> void
```
kddnewton, line 36: Can you put this on `class Node`? `_Node` is for the stuff that is implemented on each node.

kddnewton (changes requested):

templates/sig/prism/node.rbs.erb
S-H-GAMELINKS: OK, I'll try it!

kddnewton (approved):

PR #2854, opened 2024-06-01
Add `data_loc` method to Rust bindings

ryym:
Closes https://github.com/ruby/prism/discussions/2621

Currently Prism recognizes the `__END__` content and provides a way to access it in Ruby, C, and Java APIs, but Rust bindings does not have it. This PR adds the `data_loc` method to Rust API.

### DIFF
--- a/rust/ruby-prism/src/lib.rs
+++ b/rust/ruby-prism/src/lib.rs
@@ -198,6 +198,18 @@ impl<'pr> ParseResult<'pr> {
         }
     }
 
+    /// Returns an optional location of the __END__ marker and the rest of the content of the file.
+    #[must_use]
+    pub fn data_loc(&self) -> Option<Location<'_>> {
+        let root = self.source.as_ptr();
+        let location = unsafe { &(*self.parser.as_ptr()).data_loc };
+        if location.start >= root {
+            Some(Location::new(self.parser, location))
+        } else {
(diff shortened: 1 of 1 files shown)


rust/ruby-prism/src/lib.rs
```diff
@@ -198,6 +198,18 @@ impl<'pr> ParseResult<'pr> {
         }
     }
 
+    /// Returns an optional location of the __END__ marker and the rest of the content of the file.
+    #[must_use]
+    pub fn data_loc(&self) -> Option<Location<'_>> {
+        let root = self.source.as_ptr();
+        let location = unsafe { &(*self.parser.as_ptr()).data_loc };
+        if location.start >= root {
```
kddnewton, line 206: In other optional locations we're checking `start.is_null()` instead, can you do that here?

kddnewton (changes requested):

rust/ruby-prism/src/lib.rs
ryym: Thank you, I changed the implementation to use `is_null()` .
https://github.com/ruby/prism/pull/2854/commits/be34f68fa502dfdbb2ba43454d343a2b8cd7c0f7

kddnewton (approved):

PR #2861, opened 2024-06-03
Enable typechecking in CI

KaanOzkan:


rakelib/typecheck.rake
```diff
@@ -2,6 +2,8 @@
 
 namespace :typecheck do
   task tapioca: :templates do
+    Rake::Task["compile:prism"].invoke
```
KaanOzkan, line 5: This is supposed to run task dependencies and the ask only if they haven't been ran. Looked like what we want.

kddnewton (approved):
Thanks @KaanOzkan

PR #2887, opened 2024-06-13
Make Rake compile task a dependency of test task

peterzhu2118:
We should make the compile task a dependency of the test task so that:

- On new clones of prism, running `rake test` will not fail with `'Kernel#require_relative': cannot load such file -- prism/lib/prism/node (LoadError)`.
- When a C file changes, it will be recompiled before the tests are ran.

### DIFF
--- a/rakelib/test.rake
+++ b/rakelib/test.rake
@@ -8,7 +8,7 @@ config = lambda do |t|
   t.test_files = FileList["test/**/*_test.rb"]
 end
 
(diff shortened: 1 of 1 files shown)


peterzhu2118:
Ah looks like on CI you want to run tests without compiling prism?

kddnewton:
Yeah we want to be able to install the gem and then test that we can run the tests with the install files to make sure we have all the right stuff bundled.

PR #2891, opened 2024-06-15
Document MultiTargetNode fields

herwinw:
This should check off another box in #2123.

According to https://en.wikipedia.org/wiki/Bracket#Parentheses the singular form of parentheses in EN-US is parenthesis, I'm not sure if this is the form we should standardi{s,z}e. I tried looking for other examples with the name of `(` or `)`, but couldn't find anything except for `paren`.

I think the example code covers all use cases.


kddnewton (approved):

kddnewton:
This looks really great, thanks!

PR #2899, opened 2024-06-20
Set `#line" directives to respect the path in ruby/ruby repo

mame:
ruby/ruby measures test coverage of C code, but the `#line` directive generated by prism points to a file that does not exist, so coverage is not taken properly.

This changeset specifies the location of the source files as a relative path in terms of ruby/ruby repo.


kddnewton (approved):
Sounds good, thanks @mame!

PR #2909, opened 2024-06-25
Fix documentation comment in generated with templates files

andrykonchin:
Fix warning
```
/******************************************************************************/
^
warning: [dangling-doc-comments] documentation comment is not attached to any declaration

```
in Nodes.java and AbstractNodeVisitor.java files. `/**` is treated as a starting of a javadoc comment.


kddnewton (changes requested):
This is a small thing, but can you make that specific for Java files? I'd like to keep it the same for .c/.h files

eregon:
[Doxygen](https://www.doxygen.nl/manual/docblocks.html) seems to also consider `/**` as a documentation comment.
How about using `/*--------------------------------*/` or so instead to avoid any issue?

andrykonchin:
Done

kddnewton (approved):

PR #2911, opened 2024-06-25
Parse break under the correct context when used with until/while modifiers

st0012:
In the example of `break while false`, Prism currently parse the `break` under the main context, which causes it to raise an error. This commit fixes this issue by parsing the `break` under the correct context when seeing a `while` or `until` modifier.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -18342,6 +18342,14 @@ parse_expression_prefix(pm_parser_t *parser, pm_binding_power_t binding_power, b
                 }
             }
 
+            if (match1(parser, PM_TOKEN_KEYWORD_WHILE_MODIFIER)) {
+                parser->current_context->context = PM_CONTEXT_WHILE;
+            }
+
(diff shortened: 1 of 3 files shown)


src/prism.c
```diff
@@ -18342,6 +18342,14 @@ parse_expression_prefix(pm_parser_t *parser, pm_binding_power_t binding_power, b
                 }
             }
 
+            if (match1(parser, PM_TOKEN_KEYWORD_WHILE_MODIFIER)) {
+                parser->current_context->context = PM_CONTEXT_WHILE;
+            }
+
+            if (match1(parser, PM_TOKEN_KEYWORD_UNTIL_MODIFIER)) {
+                parser->current_context->context = PM_CONTEXT_UNTIL;
```
st0012, line 18350: To me honest, I'm not sure if changing context like this is allowed or a good idea.

kddnewton:
Superseded by https://github.com/ruby/prism/pull/2912

PR #2913, opened 2024-06-26
Arena allocator

kddnewton:


kddnewton (commented):
Waiting for a chance to rebase this

Cluster444:
I'd love to help get this merged in! Working on a zig interface and this seems like it would make life easier.

I did a quick rebase, the conflicts were light, mostly around node_id which was new and the node_alloc function/macro pair.

The bigger part I'm guessing is around the use of xmalloc and friends since they don't take an allocator. Is the goal to replace xalloc/xfree with a call that takes an allocator?

Cluster444:
Quick update, I did get this to rebase and all of the tests pass.

Though I had to kill xfree almost entirely as there's a few xfree calls that need to go away, just need to hunt those down.

Cluster444:
So I think I have this sorted out, though I'm not sure how to get the results back to you here.

https://github.com/Cluster444/prism/tree/57b85f897d9e4c7ce60d725f0ff071fbbb59eae8

Here's the fork of this PR, rebased on main.

The only extra bit was lifting up all the field types that are being ignored in node destroy to make it a little more clear what those big lists were for.

flavorjones:
@Cluster444 I think it would be OK to create a new PR with your rebased branch.

kddnewton:
Mostly the blocker here is that we need to migrate CRuby over to this new API and ensure it still works, which is what I haven't had time to do yet.

Cluster444:
It's looking like I have some free time tomorrow evening so I could have a look at that. I've got some basic experience with the CRuby internals, so I could start poking around and see what blows up. If you have any specific off-hand that come to mind that would help!

Also is the intention to get rid of most/all uses of heap allocs in favor of the arena? It seems mostly straightforward enough to do, but I also get the sense this is the first of a few optimization steps?

kddnewton:
The goal is to get rid of the heap allocs that are done to allocate parts of the AST. So that freeing the AST doesn't require walking the tree. Basically the goal is to get rid of `pm_node_destroy`.

kddnewton:
Closing in favor of #3231

PR #2914, opened 2024-06-28
Document some nodes related to pattern matching

herwinw:
Part of  #2123


herwinw:
`MatchRequiredNode` has some additional information on what types of nodes it will generate:
```yaml
- name: pattern
  type: node
  comment: |
    Represents the right-hand side of the operator. The type of the node depends on the expression.

    Anything that looks like a local variable name (including `_`) will result in a `LocalVariableTargetNode`.

        foo => a # This is equivalent to writing `a = foo`
               ^

    Using an explicit `Array` or combining expressions with `,` will result in a `ArrayPatternNode`. This can be preceded by a constant.

        foo => [a]
               ^^^

        foo => a, b
               ^^^^

        foo => Bar[a, b]
               ^^^^^^^^^

    If the array pattern contains at least two wildcard matches, a `FindPatternNode` is created instead.

        foo => *, *a
               ^^^^^

    Using an explicit `Hash` or a constant with square brackets and hash keys in the square brackets will result in a `HashPatternNode`.

        foo => { a: 1, b: }

        foo => Bar[a: 1, b:]

        foo => Bar[**]

    Anything else will result in the regular node for that expression, for example a `ConstantReadNode`.

        foo => CONST
```
But this information applies to other locations as well. It is recursively applied to all possible fields (e.g. something that looks like a variable name inside an `ArrayPatternNode` will result in a `LocalVariableTargetNode`. And this applies to all other pattern matching related parse nodes as well (like `foo in [a]`). There probably is a better way to document this than to copy-paste it everywhere or referring back to other nodes for details.
Marking this one as draft until this issue has been resolved/answered.

config.yml
```diff
@@ -2856,10 +3010,55 @@ nodes:
...
+              foo => a, b
+                     ^^^^
+
+              foo => Bar[a, b]
+                     ^^^^^^^^^
+
+          If the array pattern contains at least two wildcard matches, a `FindPatternNode` is created instead.
+
+              foo => *, *a
+                     ^^^^^
```
herwinw, line 3042: This example probably needs to be updated, wait for results in #2915

config.yml
```diff
@@ -903,16 +903,64 @@ nodes:
...
+
+              foo in 1, 2, *
+                           ^
+
+              foo in 1, 2, *a, 3
+                           ^^
       - name: posts
         type: node[]
+        comment: |
+          This represents all the entries after the wildcard. This list will be empty if there is no wildcard, or the
```
herwinw, line 943: This line is incomplete.

config.yml
```diff
@@ -1833,27 +1881,81 @@ nodes:
...
+
+              foo in Foo(*bar, baz, *qux)
+                                        ^
     comment: |
       Represents a find pattern in pattern matching.
 
           foo in *bar, baz, *qux
                  ^^^^^^^^^^^^^^^
 
+          foo in *bar, *qux
```
herwinw, line 1948: Another use of syntax that is not accepted by MRI

kddnewton: This has been fixed.

config.yml
kddnewton: This has been fixed.

kddnewton (approved):
This looks great, @herwinw is this good to go?

kddnewton:
Oh sorry I didn't see your comment. Typically what I have done in those cases is put the documentation into the `parsing_rules.md` file and linked to it. Could you do the same?

herwinw:
I will have a look at this to finish this one up, but that's going to be this weekend (or maybe later), I'm a bit short on time lately.

kddnewton:
No problem take your time!

PR #2941, opened 2024-07-17
Implement string field accessors in Rust crate

FnControlOption:
For example, here's `SymbolNode::unescaped`:

```rs
pub fn unescaped(&self) -> &str {
    unsafe {
        let source = (*self.pointer).unescaped.source;
        let length = (*self.pointer).unescaped.length;
        let slice = std::slice::from_raw_parts(source, length);
        std::str::from_utf8_unchecked(slice)
    }
}
```

This assumes (possibly incorrectly) that string fields will always contain valid UTF-8. As an alternative, `from_utf8_unchecked` could be omitted by returning the plain `&[u8]` slice instead.


rust/ruby-prism/build.rs
```diff
@@ -51,6 +51,7 @@ enum NodeFieldType {
 #[serde(untagged)]
 enum NodeFieldKind {
     Concrete(String),
+    #[allow(dead_code)]
```
kddnewton, line 54: Why is the necessary?

rust/ruby-prism/build.rs
```diff
@@ -304,8 +305,13 @@ fn write_node(file: &mut File, flags: &[Flags], node: &Node) -> Result<(), Box<d
...
             },
             NodeFieldType::String => {
-                writeln!(file, "    pub const fn {}(&self) -> &str {{", field.name)?;
-                writeln!(file, "        \"\"")?;
+                writeln!(file, "    pub fn {}(&self) -> &str {{", field.name)?;
+                writeln!(file, "        unsafe {{")?;
+                writeln!(file, "            let source = (*self.pointer).{}.source;", field.name)?;
+                writeln!(file, "            let length = (*self.pointer).{}.length;", field.name)?;
+                writeln!(file, "            let slice = std::slice::from_raw_parts(source, length);")?;
+                writeln!(file, "            std::str::from_utf8_unchecked(slice)")?;
```
kddnewton, line 313: Yeah we can't do this. Strings can be contents of regexps, xstrings, strings, and symbols, all of which can have any of the 90 source encodings that Ruby supports. Better to return a u8 slice and have the consumer handle it.

kddnewton (changes requested):

kddnewton (approved):

PR #2943, opened 2024-07-18
Do not read directory

kddnewton:


src/util/pm_string.c
```diff
@@ -116,6 +116,13 @@ pm_string_mapped_init(pm_string_t *string, const char *filepath) {
         return false;
     }
 
+    // Ensure it is a file and not a directory
+    if (S_ISDIR(sb.st_mode)) {
```
eregon, line 120: How about: `if (!S_ISREG(sb.st_mode)) {` here?
If it's a block or character device or socket we'd likely want to error too, no?

PR #2944, opened 2024-07-19
Fallback to pm_string_file_init on platforms without memory-mapped files

kateinoigakukun:
> ..., and on other POSIX systems we'll use `read`.

As `pm_string_mapped_init`'s doc comment says, it should fall back to `read(2)`-based implementation on platforms without memory-mapped files like WASI, but it didn't. This commit fixes it by calling `pm_string_file_init` in the fallback case.
Also `defined(_POSIX_MAPPED_FILES)` check for `read(2)`-based path is unnecessary, and it prevents the fallback from being executed, so this change removes it.

### DIFF
--- a/src/util/pm_string.c
+++ b/src/util/pm_string.c
@@ -143,10 +143,7 @@ pm_string_mapped_init(pm_string_t *string, const char *filepath) {
     *string = (pm_string_t) { .type = PM_STRING_MAPPED, .source = source, .length = size };
     return true;
 #else
-    (void) string;
-    (void) filepath;
-    perror("pm_string_mapped_init is not implemented for this platform");
-    return false;
+    return pm_string_file_init(string, filepath);
 #endif
 }
 
@@ -205,7 +202,7 @@ pm_string_file_init(pm_string_t *string, const char *filepath) {
     CloseHandle(file);
     *string = (pm_string_t) { .type = PM_STRING_OWNED, .source = source, .length = (size_t) file_size };
     return true;
-#elif defined(_POSIX_MAPPED_FILES)
+#else
     FILE *file = fopen(filepath, "rb");
     if (file == NULL) {
         return false;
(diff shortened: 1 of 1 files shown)


kddnewton (commented):
@hasumikin I assume this won't work for you. Is there something else we can check?

kddnewton:
@kateinoigakukun for context, the reason it's checking `_POSIX_MAPPED_FILES` is that on picoruby, it's possible to not have a file system at all, so `fread` won't be defined and it won't be able to compile. Do you know if there's anything else we can check for this?

kateinoigakukun:
@kddnewton Thank you for the detailed context. Hmm, I don't think we have something to rely on to decide if the given libc supports fread or not. Do you think it makes sense to add checks for `PRISM_HAS_NO_FILESYSTEM`, then ask picoruby to set the macro?

hasumikin:
I guess there are three options:
(1) Filesystem with mmap
(2) Filesystem without mmap
(3) No filesystem

(3) is no longer POSIX-like system and this is what PicoRuby mainly targets on.
But to be exact, (2) is also assumable for embedded systems including PicoRuby.

I'm fine removing an automatic filesystem detection by `_POSIX_MAPPED_FILES`.
Instead, explicitly defining macro like the following, to distinguish three options, is doable:

```
PRISM_NO_FILESYSTEM
PRISM_NO_MAPPED_FILESYSTEM
```

kateinoigakukun:
@hasumikin Thanks for your feedback! Do we have any target platform that defines `_POSIX_MAPPED_FILES` but doesn't want to use mmap? If there isn't, we might not need `PRISM_NO_MAPPED_FILESYSTEM` for now.

hasumikin:
@kateinoigakukun 
>Do we have any target platform that defines _POSIX_MAPPED_FILES but doesn't want to use mmap?

AFAIK, we don't.

>If there isn't, we might not need PRISM_NO_MAPPED_FILESYSTEM for now.

You're right. Your patch should work

kddnewton (approved):

PR #2956, opened 2024-07-24
Optimize serialized size for PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS=1

eregon:
From https://github.com/ruby/prism/pull/2924#issuecomment-2248749015, `average` went from 0.799 before #2924 to 0.950 after, with this PR it goes back to 0.800.

As an aside this makes the `flags` field public again, which is necessary as TruffleRuby uses it like that (the error can be seen at https://github.com/oracle/truffleruby/actions/workflows/prism.yml).

### DIFF
--- a/templates/java/org/prism/Loader.java.erb
+++ b/templates/java/org/prism/Loader.java.erb
@@ -363,7 +363,7 @@ public class Loader {
             <%-
             params = ["startOffset", "length"]
             params << "buffer.getInt()" if node.needs_serialized_length?
-            params << "loadFlags()"
+            params << "loadFlags()" if node.flags
             params.concat node.semantic_fields.map { |field|
               case field
               when Prism::Template::NodeField then "#{field.java_cast}loadNode()"
--- a/templates/java/org/prism/Loader.java.erb
+++ b/templates/java/org/prism/Loader.java.erb
@@ -353,7 +353,6 @@ public class Loader {
 
     private Nodes.Node loadNode() {
         int type = buffer.get() & 0xFF;
-        loadVarUInt(); // skip over node_id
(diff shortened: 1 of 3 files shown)


kddnewton (approved):
I'm okay with this, but I was under the impression you wanted the flags for the static literals maybe? Do you want to just exclude node id?

eregon:
> I was under the impression you wanted the flags for the static literals maybe?

I thought initially that would be nice mostly for the Ruby API (https://github.com/ruby/prism/issues/1531) as a module which guarantees a `value` methods (which some of these nodes already have).

I suppose it could be useful for Java compilation to know if a node is a static literal for which the value can be potentially cached in the IR, but in practice that still depends on implementation and representation details so I think the value is fairly limited.

eregon:
I did a quick search and for instance the only place we do `instanceof Nodes.IntegerNode` in TruffleRuby is for https://github.com/oracle/truffleruby/blob/3cd422433deebe3fa664f8c4540811c42ca02e93/src/main/java/org/truffleruby/parser/YARPTranslator.java#L3769-L3785 which doesn't map so nicely to the static literal flag (it covers more than that like ivar reads).

eregon:
And for completeness, regarding the `newline` flag, I looked a bit into it today (previous comment about it in https://github.com/ruby/prism/pull/2924#issuecomment-2206751291) but there seems to be no value to it currently in the Java API since it wouldn't really save any work from `MarkNewlinesVisitor`. (It could be helpful when using something else than MarkNewlinesVisitor to assign the real newline flags though I suppose, but Loader currently always uses MarkNewlinesVisitor).

Issue #2960, opened 2024-07-29
Allow generating ctags via prism


teoljungberg:
Projects such as [ripper-tags](https://github.com/tmm1/ripper-tags) have been used by the community to get better integration to generate tags files for ctags(1).

Could, and should, ruby/prism support a similar functionality? I assume it would go into a separate project, or to help any library author write it themselves.

PR #2980, opened 2024-08-12
Add sample for generating tags

kddnewton:
Fixes https://github.com/ruby/prism/issues/2960

Hey @teoljungberg — happy to provide a sample of code that would do what you're asking (included in this PR). I don't have the bandwidth to maintain it as a separate project, but you're more than welcome to grab it up and publish it as a gem if you like. Alternatively, you can grab the script out of this repo that will now contain it.


teoljungberg:
Thank you! Will investigate and give it a whirl.

PR #2982, opened 2024-08-14
Tweak inspect representation of `Prism::Location`

koic:
This PR tweaks inspect representation of `Prism::Location`.

## Before

During debugging, the meaning of `@location=21474836481` was unclear:

```console
$ ruby -Ilib -rprism -e 'p Prism.lex("puts :hi").value.map(&:first)[1]'
#<Prism::Token:0x000000010cd74e40 @source=#<Prism::ASCIISource:0x000000010cb5f808 @source="puts :hi", @start_line=1, @offsets=[0]>,
@type=:SYMBOL_BEGIN, @value=":", @location=21474836481>
```

## After

This PR clarifies the contents of the location object, aligning with what I think user expects:

```console
$ ruby -Ilib -rprism -e 'p Prism.lex("puts :hi").value.map(&:first)[1]'
#<Prism::Token:0x000000010e174d50 @source=#<Prism::ASCIISource:0x000000010df5efe8 @source="puts :hi", @start_line=1, @offsets=[0]>,
@type=:SYMBOL_BEGIN, @value=":", @location=#<Prism::Location @start_offset=5 @length=1 start_line=1>>
```

Although it is uncertain whether Prism will accept this change in the inspect representation, it is submitted here as a suggestion.

### DIFF
--- a/lib/prism/parse_result.rb
+++ b/lib/prism/parse_result.rb
@@ -711,5 +711,11 @@ module Prism
         other.type == type &&
         other.value == value
     end
+
+    # Returns a string representation of this token.
+    def inspect
+      location
+      super
+    end
   end
 end


eregon:
LGTM, seems an obvious fix.

lib/prism/parse_result.rb
```diff
@@ -711,5 +711,10 @@ def ==(other)
         other.type == type &&
         other.value == value
     end
+
+    # Returns a string representation of this token.
+    def inspect
+      "#<Prism::Token @source=#{@source.inspect}, @type=#{@type.inspect}, @value=#{@value.inspect}, location=#{location.inspect}>"
```
kddnewton, line 717: What do you think about using `Kernel#inspect` instead? It seems like the issue is that `location` hasn't been reified yet, which we could do here?

```suggestion
      location
      super
```

koic: Thank you for the feedback. I've updated this PR.

The implementation looks simpler and better. One of the remaining differences is whether to include the object identifier in the inspect representation. I don't have a strong opinion on whether it should be included or not :-)

## Before

```console
$ ruby -Ilib -rprism -e 'p Prism.lex("puts :hi").value.map(&:first)[1]'
#<Prism::Token:0x000000010e174d50 @source=#<Prism::ASCIISource:0x000000010df5efe8 @source="puts :hi", @start_line=1, @offsets=[0]>,
@type=:SYMBOL_BEGIN, @value=":", @location=#<Prism::Location @start_offset=5 @length=1 start_line=1>>
```

## After

```console
$ ruby -Ilib -rprism -e 'p Prism.lex("puts :hi").value.map(&:first)[1]'
#<Prism::Token @source=#<Prism::ASCIISource:0x000000010df5efe8 @source="puts :hi", @start_line=1, @offsets=[0]>,
@type=:SYMBOL_BEGIN, @value=":", @location=#<Prism::Location @start_offset=5 @length=1 start_line=1>>
```

kddnewton (approved):
I think it's probably fine to include the object id.

PR #2983, opened 2024-08-14
Callback on shebang switches

kddnewton:
Add the ability to receive a callback when the parser encounters a shebang that contains additional switches after the Ruby engine. This is necessary because some command-line flags may be present there that will alter the parse.


nobu:
I don't think shebang handling is a role of parsers.

kddnewton:
I agree! Unfortunately I couldn't find another way to alert the runtime of the extra flags. Any suggestions?

To be clear, this isn't actually parsing the shebang it's just alerting the caller that one was found.

PR #3007, opened 2024-08-22
Fix bug when `#!` ends with carriage return

### LINKED ISSUE #2996: CRuby test failure: not setting external encoding on shebang flag
The `TestRubyOptions#test_shebang` test is failing, because it is not properly setting the default external encoding based on the flag found in the shebang. Specifically this assertion is failing:

```ruby
    warning = /mswin|mingw/ =~ RUBY_PLATFORM ? [] : /shebang line ending with \\r/
    assert_in_out_err([{'RUBYOPT' => nil}], "#!ruby -KU -Eutf-8\r\np \"\u3042\"\r\n",
                      ["\"\u3042\""], warning,
                      encoding: Encoding::UTF_8)
```

In order to test changes, you will need to apply to following diff to `ruby/ruby` (copy into a file like `prism.patch` and run `git apply prism.patch`):

```diff
diff --git a/ruby.c b/ruby.c
index 3c78ada936..d9b7824a98 100644
--- a/ruby.c
+++ b/ruby.c
@@ -1428,10 +1428,10 @@ proc_long_options(ruby_cmdline_options_t *opt, const char *s, long argc, char **
     }
     else if (is_option_with_arg("parser", Qfalse, Qtrue)) {
         if (strcmp("prism", s) == 0) {
-            (*rb_ruby_prism_ptr()) = true;
+            // default behavior
         }
         else if (strcmp("parse.y", s) == 0) {
-            // default behavior
+            (*rb_ruby_prism_ptr()) = false;
         }
         else {
             rb_raise(rb_eRuntimeError, "unknown parser %s", s);
@@ -3077,8 +3077,6 @@ ruby_process_options(int argc, char **argv)
     VALUE iseq;
     const char *script_name = (argc > 0 && argv[0]) ? argv[0] : ruby_engine;
 
-    (*rb_ruby_prism_ptr()) = false;
-
     if (!origarg.argv || origarg.argc <= 0) {
         origarg.argc = argc;
         origarg.argv = argv;
diff --git a/vm.c b/vm.c
index c7ccbc9550..8a66428eeb 100644
--- a/vm.c
+++ b/vm.c
@@ -4461,7 +4461,7 @@ rb_ruby_verbose_ptr(void)
     return &cr->verbose;
 }
 
-static bool prism;
+static bool prism = true;
 
 bool *
 rb_ruby_prism_ptr(void)
```

Then, to run the individual tests, run:

```
make -j
make test-all TESTS=test/ruby/test_rubyoptions.rb TESTOPTS='-n /test_shebang/'
```

amomchilov:
Related to #2996

Fixes the second of the two assertion failures here:

```
TestRubyOptions#test_shebang [/Users/alex/src/github.com/Ruby/ruby/test/ruby/test_rubyoptions.rb:514]:
pid 49518 exit 0.

1. [1/2] Assertion for "stdout"
   | <["\"あ\""]> expected but was
   | <["\"\\u3042\""]>.

2. [2/2] Assertion for "stderr"
   | Expected /shebang line ending with \\r/ to match "".

Finished tests in 0.030572s, 32.7097 tests/s, 359.8064 assertions/s.
1 tests, 11 assertions, 1 failures, 0 errors, 0 skips

ruby -v: ruby 3.4.0dev (2024-08-22T14:01:55Z master 56a34b5af5) +PRISM [arm64-darwin23]
make: *** [yes-test-all] Error 1
```

## Root cause

The "string" (pointer+length combo) given to `pm_parser_warn_shebang_carriage_return` ended on the `\r`, right before the `\n`. So the check for `\r\n` actually found `8\r` instead.

## Solution approach

When there _is_ a newline, bump the `length` by 1 to include it as part of the string passed to `pm_parser_warn_shebang_carriage_return`, which can then accurately detect the `\r\n`.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -21711,7 +21711,7 @@ pm_strnstr(const char *big, const char *little, size_t big_length) {
  */
 static void
 pm_parser_warn_shebang_carriage_return(pm_parser_t *parser, const uint8_t *start, size_t length) {
-    if (length > 2 && start[length - 1] == '\n' && start[length - 2] == '\r') {
+    if (length > 2 && start[length - 2] == '\r' && start[length - 1] == '\n') {
         pm_parser_warn(parser, start, start + length, PM_WARN_SHEBANG_CARRIAGE_RETURN);
     }
 }
@@ -21904,11 +21904,17 @@ pm_parser_init(pm_parser_t *parser, const uint8_t *source, size_t size, const pm
 
         const char *engine;
         if ((engine = pm_strnstr((const char *) parser->start, "ruby", length)) != NULL) {
-            pm_parser_warn_shebang_carriage_return(parser, parser->start, length);
-            if (newline != NULL) parser->encoding_comment_start = newline + 1;
+            if (newline != NULL) {
+                size_t length_including_newline = length + 1;
+                pm_parser_warn_shebang_carriage_return(parser, parser->start, length_including_newline);
+
+                parser->encoding_comment_start = newline + 1;
+            }
+
             if (options != NULL && options->shebang_callback != NULL) {
                 pm_parser_init_shebang(parser, options, engine, length - ((size_t) (engine - (const char *) parser->start)));
             }
+
             search_shebang = false;
         } else if (!parser->parsing_eval) {
             search_shebang = true;
@@ -21938,17 +21944,20 @@ pm_parser_init(pm_parser_t *parser, const uint8_t *source, size_t size, const pm
 
             size_t length = (size_t) ((newline != NULL ? newline : parser->end) - cursor);
             if (length > 2 && cursor[0] == '#' && cursor[1] == '!') {
-                if (parser->newline_list.size == 1) {
-                    pm_parser_warn_shebang_carriage_return(parser, cursor, length);
-                }
-
                 const char *engine;
                 if ((engine = pm_strnstr((const char *) cursor, "ruby", length)) != NULL) {
                     found_shebang = true;
-                    if (newline != NULL) parser->encoding_comment_start = newline + 1;
+                    if (newline != NULL) {
+                        size_t length_including_newline = length + 1;
+                        pm_parser_warn_shebang_carriage_return(parser, cursor, length_including_newline);
+
+                        parser->encoding_comment_start = newline + 1;
+                    }
+
                     if (options != NULL && options->shebang_callback != NULL) {
                         pm_parser_init_shebang(parser, options, engine, length - ((size_t) (engine - (const char *) cursor)));
                     }
+
                     break;
                 }
             }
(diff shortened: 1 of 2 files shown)


src/prism.c
```diff
@@ -21711,7 +21711,7 @@ pm_strnstr(const char *big, const char *little, size_t big_length) {
  */
 static void
 pm_parser_warn_shebang_carriage_return(pm_parser_t *parser, const uint8_t *start, size_t length) {
-    if (length > 2 && start[length - 1] == '\n' && start[length - 2] == '\r') {
```
amomchilov, line 21714: I could have only renumbered instead:

```diff
-    if (length > 2 && start[length - 1] == '\n' && start[length - 2] == '\r') {
+    if (length > 2 && start[length] == '\n' && start[length - 1] == '\r') {
```

but IMHO it's more clear to have the `\r` and `\n` in that order, as they would be in the line ending.

kddnewton (changes requested):
This isn't quite right, because this is going to read off the end of the buffer. In general when we're passing around slices (`const uint8_t *` + `size_t`) it means that's the only part that has been checked as being safe.

So probably this needs to be fixed on the caller side, where we're missing a `+ 1` somewhere when we have found newlines.

amomchilov:
@kddnewton Makes sense.

Just to clarify, the last "valid" index of this slice is at `[length - 1]`, and `[length]` is one past the end (at least the part of it that we're guaranteed to exist), right?

src/prism.c
```diff
@@ -21711,7 +21711,7 @@ pm_strnstr(const char *big, const char *little, size_t big_length) {
...
+    if (length > 2 && start[length - 2] == '\r' && start[length - 1] == '\n') {
```
amomchilov, line 21714: Small stylistic change, but I think it's a bit more clear to put the `\r` and `\n` "in order"

src/prism.c
```diff
@@ -21939,7 +21945,8 @@ pm_parser_init(pm_parser_t *parser, const uint8_t *source, size_t size, const pm
             size_t length = (size_t) ((newline != NULL ? newline : parser->end) - cursor);
             if (length > 2 && cursor[0] == '#' && cursor[1] == '!') {
                 if (parser->newline_list.size == 1) {
```
amomchilov, line 21947: @kddnewton I don't think this case can every be hit.

`parser->newline_list` starts out with `size = 1` (with `offset[0] = 0`). The `pm_newline_list_append` on line 21934 above will cause `parser->newline_list.size` here to always be >2

kddnewton: In this case, can you add some tests and verify that we have the correct behavior in these cases?

kddnewton (changes requested):

src/prism.c
amomchilov: Do you know what Ruby code even triggers this case? I couldn't figure it out, so I had just forced it to reach this path by editing the `if` predicates.

amomchilov: Done.

kddnewton (approved):

PR #3018, opened 2024-08-27
Optimize Array allocations

amomchilov:
Just some minor perf opportunities I spotted while working on other stuff.

### DIFF
--- a/ext/prism/extension.c
+++ b/ext/prism/extension.c
@@ -372,7 +372,7 @@ dump_file(int argc, VALUE *argv, VALUE self) {
  */
 static VALUE
 parser_comments(pm_parser_t *parser, VALUE source) {
-    VALUE comments = rb_ary_new();
+    VALUE comments = rb_ary_new_capa(parser->comment_list.size);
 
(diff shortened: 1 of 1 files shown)


ext/prism/extension.c
```diff
@@ -607,7 +608,7 @@ parse_lex_input(pm_string_t *input, const pm_options_t *options, bool return_nod
     pm_parser_register_encoding_changed_callback(&parser, parse_lex_encoding_changed_callback);
 
     VALUE source_string = rb_str_new((const char *) pm_string_source(input), pm_string_length(input));
-    VALUE offsets = rb_ary_new();
+    VALUE offsets = rb_ary_new_capa(parser.newline_list.size);
     VALUE source = rb_funcall(rb_cPrismSource, rb_id_source_for, 3, source_string, LONG2NUM(parser.start_line), offsets);
```
amomchilov, line 612: This is also a bit odd, that we pass `offsets` here on line 611, but don't actually fill it in until line 636 below.

It's a bit more clear this way, but it might be an issue if `Prism::Source.for` ever decides to copy the array (and copies its emptiness).

kddnewton (approved):

PR #3022, opened 2024-08-28
Type every node field and mark on-error-only types explicitly

eregon:
* For Loader.java, do not deserialize the AST if there are errors, so then Java nodes only have non-error types for fields.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -793,8 +793,8 @@ nodes:
           - GlobalVariableReadNode
           - BackReferenceReadNode
           - NumberedReferenceReadNode
-          - SymbolNode # On parsing error of `alias $a b`
-          - MissingNode # On parsing error of `alias $a 42`
+          - on error: SymbolNode # alias $a b
+          - on error: MissingNode # alias $a 42
         comment: |
           Represents the old name of the global variable that can be used before aliasing.
 
@@ -824,8 +824,8 @@ nodes:
         kind:
           - SymbolNode
           - InterpolatedSymbolNode
-          - GlobalVariableReadNode # On parsing error of `alias a $b`
-          - MissingNode # On parsing error of `alias a 42`
+          - on error: GlobalVariableReadNode # alias a $b
+          - on error: MissingNode # alias a 42
       - name: keyword_loc
         type: location
     comment: |
@@ -837,8 +837,10 @@ nodes:
     fields:
       - name: left
         type: node
+        kind: pattern expression
       - name: right
         type: node
+        kind: pattern expression
       - name: operator_loc
         type: location
     comment: |
@@ -850,6 +852,7 @@ nodes:
     fields:
       - name: left
         type: node
+        kind: non-void expression
         comment: |
           Represents the left side of the expression. It can be any [non-void expression](https://github.com/ruby/prism/blob/main/docs/parsing_rules.md#non-void-expression).
 
@@ -860,8 +863,9 @@ nodes:
               ^
       - name: right
         type: node
+        kind: Node
         comment: |
-          Represents the right side of the expression. It can be any [non-void expression](https://github.com/ruby/prism/blob/main/docs/parsing_rules.md#non-void-expression).
+          Represents the right side of the expression.
 
               left && right
                       ^^^^^
@@ -885,6 +889,7 @@ nodes:
     fields:
       - name: arguments
         type: node[]
+        kind: non-void expression
     comment: |
       Represents a set of arguments to a method or a keyword.
 
@@ -895,6 +900,7 @@ nodes:
     fields:
       - name: elements
         type: node[]
+        kind: non-void expression
         comment: Represent the list of zero or more [non-void expressions](https://github.com/ruby/prism/blob/main/docs/parsing_rules.md#non-void-expression) within the array.
       - name: opening_loc
         type: location?
@@ -923,12 +929,18 @@ nodes:
     fields:
       - name: constant
         type: node?
+        kind:
+          - ConstantReadNode
+          - ConstantPathNode
       - name: requireds
         type: node[]
+        kind: pattern expression
       - name: rest
         type: node?
+        kind: pattern expression
       - name: posts
         type: node[]
+        kind: pattern expression
       - name: opening_loc
         type: location?
       - name: closing_loc
@@ -954,6 +966,7 @@ nodes:
     fields:
       - name: key
         type: node
+        kind: non-void expression
         comment: |
           The key of the association. This can be any [non-void expression](https://github.com/ruby/prism/blob/main/docs/parsing_rules.md#non-void-expression).
 
@@ -967,6 +980,7 @@ nodes:
                 ^^^^^^^^^^
       - name: value
         type: node
+        kind: non-void expression
         comment: |
           The value of the association, if present. This can be any [non-void expression](https://github.com/ruby/prism/blob/main/docs/parsing_rules.md#non-void-expression).
 
@@ -991,6 +1005,7 @@ nodes:
     fields:
       - name: value
         type: node?
+        kind: non-void expression
         comment: |
           The value to be splatted, if present. Will be missing when keyword rest argument forwarding is used.
 
@@ -1053,6 +1068,7 @@ nodes:
     fields:
       - name: expression
         type: node?
+        kind: non-void expression
       - name: operator_loc
         type: location
     comment: |
@@ -1157,6 +1173,7 @@ nodes:
     fields:
       - name: receiver
         type: node?
+        kind: non-void expression
       - name: call_operator_loc
         type: location?
       - name: message_loc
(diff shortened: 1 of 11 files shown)


config.yml
```diff
@@ -1766,6 +1825,7 @@ nodes:
         type: location?
       - name: value
         type: node
+        kind: non-void expression # TODO: not sure if this is correct, e.g. defined?(break/return/redo/next/yield) is allowed, but are those "void expressions"?
```
eregon, line 1828: Are `break/return/redo/next/yield` void expressions?
If yes, we should type this as `Node` to be exact (although the difference is only for documentation purpose in this file).

As a note, `defined?(BEGIN{})` is an error, but the PreExecutionNode ends up in the AST. So technically `on error: Node` but that seems needlessly complex.

config.yml
```diff
@@ -2300,12 +2405,15 @@ nodes:
         kind: ArgumentsNode
       - name: closing_loc
         type: location
-      - name: block
+      - name: block # TODO: should this field even exists if it's only filled on error?
```
eregon, line 2408: What about these? I think it's not possible to have them non-nil and not have a SyntaxError, if so I think these fields shouldn't exist and always be "dropped" in the AST.
Probably best to do separately though.

templates/template.rb
```diff
@@ -417,6 +418,31 @@ def initialize(config, flags)
...
+                  REMOVE_ON_ERROR_TYPES ? nil : kind
+                else
+                  kind
+                end
+              end.compact
+              kinds = kinds.first if kinds.size == 1
+              options[:kind] = kinds
+            else
+              if field.fetch("type").include?("node")
+                raise "Missing kind in config.yml for field #{@name}##{options.fetch(:name)}"
```
eregon, line 442: This will ensure every node field going forward will have a type

config.yml
eregon: Mmh, CI is failing because of this case + `PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS=1`:
https://github.com/ruby/prism/actions/runs/10605781865/job/29395273994?pr=3022
Which makes sense, they have an empty type then, and even though we could default to just `Node` in that case that still wouldn't work for `prism.c` to compile.
So maybe best to fix this in this PR.
Do we want some deprecated methods returning `nil` for compatibility for those?

config.yml
eregon: According to pm_check_value_expression() they seem to be, except `yield`, which makes sense `yield` doesn't unwind and it returns a value (my mistake in the comment).

eregon: I changed this to `kind: Node # More than non-void expression as defined?(return) is allowed, yet defined?(BEGIN{}) is SyntaxError`

eregon:
Rust is failing, and the reason is it doesn't reuse the logic in template.rb but duplicates some of it.
That's rather unfortunate, is there a reason it can't be generated by template.rb like for other languages?

I guess I will have to give up on `- on error: SymbolNode` (which is a nested Hash) as Rust is much stricter with types.

config.yml
eregon: It's actually valid on Ruby < 3.4:
```ruby
foo = Object.new

def foo.[](*args, &b)
  p [:[], args, b]
end

def foo.[]=(*args, &b)
  p [:[]=, args, b]
end

value = 42
b = -> {}

foo[&b] &&= value
```
gives
```
[:[], [], #<Proc:0x00007ff5434ece48 test.rb:12 (lambda)>]
[:[]=, [42], #<Proc:0x00007ff5434ece48 test.rb:12 (lambda)>]
```

templates/template.rb
```diff
@@ -417,6 +419,31 @@ def initialize(config, flags)
...
+            if kinds = options[:kind]
+              kinds = [kinds] unless kinds.is_a?(Array)
+              kinds = kinds.map do |kind|
+                case kind
+                when "non-void expression"
+                  # the actual list of types would be way too long
+                  "Node"
+                when "pattern expression"
+                  # TODO: can we be more precise, is the list small enough?
+                  "Node"
```
eregon, line 431: FWIW this is the full list I got from running the test suite:
```ruby
AlternationPatternNode
ArrayPatternNode
CapturePatternNode
FindPatternNode
HashPatternNode
PinnedExpressionNode
PinnedVariableNode
LocalVariableTargetNode

ImplicitRestNode
SplatNode

IfNode
UnlessNode

SymbolNode
InterpolatedSymbolNode
TrueNode
FalseNode
NilNode
IntegerNode
FloatNode
RangeNode
StringNode
InterpolatedStringNode
XStringNode
RegularExpressionNode
InterpolatedRegularExpressionNode
SourceFileNode
SourceLineNode
SourceEncodingNode
ParenthesesNode
ArrayNode
RationalNode
ImaginaryNode

SelfNode
ConstantReadNode
ConstantPathNode
LambdaNode

MissingNode
```
A 37-wide union type is probably not very practical so I'll leave it as `Node` here and remove the TODO

templates/template.rb
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env ruby
-# typed: false
+# typed: ignore
```
eregon, line 2: Sorbet would complain about `NodeKindField#initialize`.
I tried to declare & pass name and comment kwargs explicitly, but it still wasn't happy and only suggested using `T.unsafe`.
It seems weird that Sorbet would fail for something so simple, but anyway it seems like a bug it even complains about this in a `typed: false` file.

templates/template.rb
```diff
@@ -417,6 +419,34 @@ def initialize(config, flags)
...
+                when Hash
+                  kind = kind.fetch("on error")
+                  REMOVE_ON_ERROR_TYPES ? nil : kind
+                else
+                  kind
+                end
+              end.compact
+              if kinds.size == 1
+                kinds = kinds.first
+                kinds = nil if kinds == "Node"
```
eregon, line 441: I initially tried for the kind to just be `"Node"` if there is nothing more precise but that causes a bunch of subtle bugs in templates, so I make it `nil` in that case like before. It's necessary to treat `Node` specially in several cases and comparing against a String doesn't seem any better.

eregon:
This should be green now and ready for review.
I had to learn a bit of Rust but it wasn't too bad (well, except the crazy String types complications).

config.yml
```diff
@@ -3578,6 +3762,7 @@ nodes:
     fields:
       - name: body
         type: node[]
+        kind: Node # fun fact: the only place where absolutely every node is valid
```
kddnewton, line 3765: This comment isn't entirely correct. There are lots of inner nodes that wouldn't be valid here, like hash contents or parameters. I think we should probably remove this comment.

kddnewton (changes requested):
Yeah I think I'm good with the general direction of this. I need to take a minute to do a full review (I'm not 100% sure all of the places where you marked non-void are actually non-void, they may allow void expressions like `break`). Only one comment to change at the moment.

config.yml
eregon: OK, removed.

eregon:
> I'm not 100% sure all of the places where you marked non-void are actually non-void, they may allow void expressions like `break`.

FWIW `non-void expression` is typed as `Node` so whether it's `non-void expression` or more it doesn't matter in terms of the given types, it's just for documentation purposes in this file.

I checked most `non-void expression` I added, except when the field was called `value` where I just assumed that should be `non-void expression`.

eregon:
I'm worried it's going to get conflicts soon otherwise.
As said above even if a `non-void expression` should be `Node` in the config there is no functional issue as they both become `Node` in types.

I could also re-check every use of `non-void expression` if you want and it doesn't take too long (based on the parser rules/grammar or based on trying code example with e.g. `break`). Let me know.

eregon:
I made a pass over all `kind: non-void expression`.
Almost all of them are correct, except for the following things:
* First, even though e.g. `foo(return)` is a SyntaxError, Prism parses it to `CallNode(ArgumentsNode(ReturnNode))` so it's possible to have a void expression as a node in the AST returned by Prism. So `kind: non-void expression` means non-void expression on success, any node on error.
* `AndNode` which was already documented to have left and right as non-void expression is not quite as simple. Left is always non-void expression, that's fine. But right has no such requirement, for example `a and return` is valid. OTOH `a and BEGIN{}` is invalid. So I think we should change AndNode & OrNode right to be `Node` and not `non-void expression`, since some void expressions are allowed there.
* RescueModifierNode it seems both the expression and rescue_expression can be void expressions: `loop { break rescue a }` is valid and `a rescue return` is valid but `a rescue BEGIN{}` is not.
I'll change these 3 field kinds to Node.

The way I checked is looking at transitive callers of `pm_check_value_expression` in prism.c to find the checks.
So that's
```
pm_check_value_expression
pm_assert_value_expression
parse_value_expression
parse_starred_expression
parse_predicate
parse_assignment_value -> all or/and/operator write nodes
```
And for the rest I tried with examples with `bin/parse`.

eregon:
`CRuby bindings / test-all` has some errors, which I guess means some types need to be changed in `prism_compile.c`.

eregon:
CI is green now after rebase.

PR #2558, opened 2024-03-06
Deprecate `TargetRubyVersion: 80_82_73_83_77_xx`

koic:
Prism has been directly supported as a parser engine since RuboCop 1.62: https://github.com/rubocop/rubocop/releases/tag/v1.62.0

This makes specifying `TargetRubyVersion` with special values like `80_82_73_83_77.33` using the `prism/translation/parser/rubocop` file unnecessary. As a result, it would be possible to deprecate this approach.

OTOH, early users might be surprised if `prism/translation/parser/rubocop` were to be suddenly removed. Therefore, this PR deprecates the parameters related to `prism/translation/parser/rubocop`.

```console
$ bundle exec ruby -rrubocop -rprism/translation/parser/rubocop -e "RuboCop::AST::ProcessedSource.new('42', 80_82_73_83_77.33).ast"
WARN: Prism is directly supported since RuboCop 1.62. The `prism/translation/parser/rubocop` file is deprecated.
WARN: Setting `TargetRubyVersion: 80_82_73_83_77.33` is deprecated. Set to `ParserEngine: parser_prism` and `TargetRubyVersion: 3.3` instead.

$ bundle exec ruby -rrubocop -rprism/translation/parser/rubocop -e "RuboCop::AST::ProcessedSource.new('42', 80_82_73_83_77.34).ast"
WARN: Prism is directly supported since RuboCop 1.62. The `prism/translation/parser/rubocop` file is deprecated.
WARN: Setting `TargetRubyVersion: 80_82_73_83_77.34` is deprecated. Set to `ParserEngine: parser_prism` and `TargetRubyVersion: 3.4` instead.
```

Eventually, it will be possible to remove it at some point.

Regarding documentation, it has been updated to not show the old, discouraged usage but rather the new way of specifying it in RuboCop.

PR #3023, opened 2024-08-29
Remove deprecated lib/prism/translation/parser/rubocop.rb file

koic:
Follow up https://github.com/ruby/prism/pull/2558.

This PR removes deprecated lib/prism/translation/parser/rubocop.rb file.

The file was a workaround to allow setting `TargetRubyVersion: 80_82_73_83_77.xx` until Prism (`Prism::Translation::Parser`) is integrated into RuboCop. RuboCop already supports Prism (`Prism::Translation::Parser`) as of https://github.com/rubocop/rubocop/pull/12724.

It has been several months since the file was deprecated in ruby/prism#2558. And, yesterday, Prism 1.0.0 was released, but perhaps the file should have been removed before then.

Although this might be seen as incompatible with semver 1.0.0, I think there is no longer a reason to keep the file.


kddnewton (approved):
Works for me!

PR #3026, opened 2024-08-30
Fix rendering of type overview docs

amomchilov:
| Before | After |
|--------|-------|
| image | image |

### DIFF
--- a/templates/include/prism/ast.h.erb
+++ b/templates/include/prism/ast.h.erb
@@ -159,7 +159,7 @@ typedef struct pm_node {
  *
  * Type: ::<%= node.type %>
 <% if (node_flags = node.flags) %>
- * Applicable flags (#pm_<%= node_flags.human %>):
+ * Flags (#pm_<%= node_flags.human %>):
 <%- node_flags.values.each do |value| -%>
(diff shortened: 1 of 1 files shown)


templates/include/prism/ast.h.erb
```diff
@@ -157,11 +157,11 @@ typedef struct pm_node {
  *<%= line %>
 <%- end -%>
  *
- * Type: <%= node.type %>
-<%- if (node_flags = node.flags) -%>
- * Flags:
+ * Type: ::<%= node.type %>
+<% if (node_flags = node.flags) %>
+ * Applicable flags (#pm_<%= node_flags.human %>):
```
kddnewton, line 162:
```suggestion
 * Flags (#pm_<%= node_flags.human %>):
```

kddnewton (changes requested):
Just one small thing then good

templates/include/prism/ast.h.erb
amomchilov: Hmmm, are you sure? It seems out-of-context on its own, IMO.

kddnewton (approved):

PR #3027, opened 2024-08-30
Add Doxygen `\memberof` annotations

amomchilov:
Doxygen's [`\memberof`](https://www.doxygen.nl/manual/commands.html#cmdmemberof) command lets you associate C functions to types, to make them render as if they were method. This is in a similar spirit to `\extends`, which we use to simulate inheritance from `pm_node`.

Demo:

image

It's possible to hide the `PRISM_EXPORTED_FUNCTION` macro spam from these return types, but I'll do that in a separate PR.


include/prism.h
```diff
@@ -43,6 +43,7 @@
  * The prism version and the serialization format.
  *
  * @returns The prism version as a constant string.
+ * \public \memberof pm_parser
```
kddnewton, line 46: I would prefer that this not be a member of parser, since it does not accept it as a parameter/doesn't correspond to a parser.

include/prism.h
```diff
@@ -77,13 +81,15 @@ PRISM_EXPORTED_FUNCTION void pm_parser_free(pm_parser_t *parser);
  *
  * @param parser The parser to use.
  * @return The AST representing the source.
+ * \public \memberof pm_parser
  */
 PRISM_EXPORTED_FUNCTION pm_node_t * pm_parse(pm_parser_t *parser);
 
 /**
- * This function is used in pm_parse_stream to retrieve a line of input from a
+ * This function is used in pm_parse_stream() to retrieve a line of input from a
```
kddnewton, line 89:
```suggestion
 * This function is used in pm_parse_stream to retrieve a line of input from a
```

include/prism.h
```diff
@@ -77,13 +81,15 @@ PRISM_EXPORTED_FUNCTION void pm_parser_free(pm_parser_t *parser);
...
  * stream. It closely mirrors that of fgets so that fgets can be used as the
  * default implementation.
+ * \public \memberof pm_parser
```
kddnewton, line 92: This shouldn't be marked as a member of parser, as it's a user-defined callback that does not correspond to a parser.

include/prism.h
```diff
@@ -200,6 +216,7 @@ PRISM_EXPORTED_FUNCTION void pm_serialize_parse_lex(pm_buffer_t *buffer, const u
  * @param size The size of the source.
  * @param data The optional data to pass to the parser.
  * @return True if the source parses without errors or warnings.
+ * \public \static \memberof pm_parser
```
kddnewton, line 219: I don't think this one is related to parser either.

include/prism/node.h
```diff
@@ -53,6 +57,7 @@ void pm_node_list_free(pm_node_list_t *list);
  *
  * @param parser The parser that owns the node.
  * @param node The node to deallocate.
+ * \public \memberof pm_node \relatedalso pm_parser
```
kddnewton, line 60: I think this is the opposite.

```suggestion
 * \public \memberof pm_parser \relatedalso pm_node
```

include/prism/node.h
```diff
@@ -112,6 +118,7 @@ PRISM_EXPORTED_FUNCTION const char * pm_node_type_to_str(pm_node_type_t node_typ
  * @param node The root node to start visiting from.
  * @param visitor The callback to call for each node in the subtree.
  * @param data An opaque pointer that is passed to the visitor callback.
+ * \public \memberof pm_node
```
kddnewton, line 121: This one I disagree with, as it's more related to the visitor (the function callback). So I would exclude this.

include/prism/node.h
```diff
@@ -123,6 +130,7 @@ PRISM_EXPORTED_FUNCTION void pm_visit_node(const pm_node_t *node, bool (*visitor
  * @param node The node to visit the children of.
  * @param visitor The callback to call for each child node.
  * @param data An opaque pointer that is passed to the visitor callback.
+ * \public \memberof pm_node
```
kddnewton, line 133: Same here, this is more related to the visitor, so I don't think this should be associated with a particular node.

include/prism/regexp.h
```diff
@@ -17,12 +17,14 @@
 #include <string.h>
 
 /**
- * This callback is called when a named capture group is found.
+ * This callback is called by pm_regexp_parse() when a named capture group is found.
+ * \public \memberof pm_parser
```
kddnewton, line 21: Not related to parser, this can be called by anything.

include/prism/regexp.h
```diff
@@ -17,12 +17,14 @@
...
  */
 typedef void (*pm_regexp_name_callback_t)(const pm_string_t *name, void *data);
 
 /**
- * This callback is called when a parse error is found.
+ * This callback is called by pm_regexp_parse() when a parse error is found.
+ * \public \memberof pm_parser
```
kddnewton, line 27: Also not related to parser, this can be called by anything.

include/prism/regexp.h
```diff
@@ -37,6 +39,7 @@ typedef void (*pm_regexp_error_callback_t)(const uint8_t *start, const uint8_t *
  * @param name_data The optional data to pass to the name callback.
  * @param error_callback The callback to call when a parse error is found.
  * @param error_data The data to pass to the error callback.
+ * \public \memberof pm_parser
```
kddnewton, line 42: I don't know about this one. It's only public so we can fuzz it, and while I understand that it's got a parser as the first parameter, it's not really meant as a member function, it's more a joining function between two modules.

kddnewton (changes requested):
A couple comments. Also it seems like sometimes we have `memberof pm_*_t` and sometimes we have `memberof pm_*`. I think I would prefer to use the typedef.

include/prism/node.h
amomchilov: What do you think about switching it to "related"?

It'll still help discoverability, since someone looking at this API would probably look in `pm_node` to start

amomchilov:
@kddnewton Ready for re-review

include/prism.h
Morriar: Could you use `\relates` instead?

include/prism.h
```diff
@@ -147,6 +157,7 @@ void pm_serialize_content(pm_parser_t *parser, pm_node_t *node, pm_buffer_t *buf
  * @param parser The parser to serialize.
  * @param node The node to serialize.
  * @param buffer The buffer to serialize to.
+ * \public \memberof pm_node
```
Morriar, line 160: Why are we pointing to `pm_buffer_t` in the previous one and `pm_node` instead of `pm_node_t` in this one?

include/prism.h
```diff
@@ -200,6 +215,7 @@ PRISM_EXPORTED_FUNCTION void pm_serialize_parse_lex(pm_buffer_t *buffer, const u
  * @param size The size of the source.
  * @param data The optional data to pass to the parser.
  * @return True if the source parses without errors or warnings.
+ * \public \memberof pm_parser
```
Morriar, line 218: Same question about `pm_parser_t` vs. `pm_parser`?

include/prism/node.h
```diff
@@ -22,6 +23,7 @@
  *
  * @param list The list to append to.
  * @param node The node to append.
+ * \private \memberof pm_node_list
```
Morriar, line 26: Why are this one and the following ones private?

include/prism/node.h
```diff
@@ -112,6 +119,7 @@ PRISM_EXPORTED_FUNCTION const char * pm_node_type_to_str(pm_node_type_t node_typ
  * @param node The root node to start visiting from.
  * @param visitor The callback to call for each node in the subtree.
  * @param data An opaque pointer that is passed to the visitor callback.
+ * \public \related pm_node
```
Morriar, line 122: `\relatedalso`?

include/prism/node.h
```diff
@@ -123,6 +131,7 @@ PRISM_EXPORTED_FUNCTION void pm_visit_node(const pm_node_t *node, bool (*visitor
  * @param node The node to visit the children of.
  * @param visitor The callback to call for each child node.
  * @param data An opaque pointer that is passed to the visitor callback.
+ * \public \related pm_node
```
Morriar, line 134: `\relatedalso`

kddnewton:
Superseded by https://github.com/ruby/prism/pull/3641

PR #3028, opened 2024-08-30
Escape chars to fix HTML rendering errors

amomchilov:
All the special characters in the tokens' docs result in garbled HTML:

image

Fixed:

image


kddnewton (changes requested):
This is currently failing doxygen. (https://github.com/ruby/prism/actions/runs/10638854883/job/29495664260?pr=3028#step:8:155).

I think in general I would prefer that the templates themselves perform this kind of escaping as opposed to the YAML file itself. This is meant to be read by anything, so it's not really a concerns of the YAML file whether or not the C docs render correctly.

PR #3054, opened 2024-09-12
Fix node flags values for Java

andrykonchin:
Recently common flags were added for every node in #2924. So each flag value was increased by 2. There were changes for Ruby and JS but not for Java.

Also I have found helpful to have similar formatting/code style in templates (at least for the code to generate flags) to easily find difference. So I changed it a bit in several templates what is quite subjective.

### DIFF
--- a/templates/java/org/prism/Nodes.java.erb
+++ b/templates/java/org/prism/Nodes.java.erb
@@ -135,19 +135,19 @@ public abstract class Nodes {
         protected abstract String toString(String indent);
     }
 <%# FLAGS -%>
-    <%- flags.each do |group| -%>
+    <%- flags.each do |flag| -%>
 
     /**
-     * <%= Prism::Template::JavaDoc.escape(group.comment) %>
+     * <%= Prism::Template::JavaDoc.escape(flag.comment) %>
      */
-    public static final class <%= group.name %> implements Comparable<<%= group.name %>> {
-        <%- group.values.each_with_index do |value, index| -%>
+    public static final class <%= flag.name %> implements Comparable<<%= flag.name %>> {
+        <%- flag.values.each_with_index do |value, index| -%>
 
         // <%= value.comment %>
         public static final short <%= value.name %> = 1 << <%= index + 2 %>;
         <%- end -%>
 
-        <%- group.values.each do |value| -%>
+        <%- flag.values.each do |value| -%>
         public static boolean is<%= value.camelcase %>(short flags) {
             return (flags & <%= value.name %>) != 0;
         }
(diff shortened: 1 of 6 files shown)


andrykonchin:
The only concern I have is testing. I've tested only the Java template.

templates/rbi/prism/node.rbi.erb
```diff
@@ -148,7 +148,7 @@ end
 module Prism::<%= flag.name %>
   <%- flag.values.each_with_index do |value, index| -%>
   # <%= value.comment %>
-  <%= value.name %> = T.let(1 << <%= index %>, Integer)
+  <%= value.name %> = T.let(1 << <%= index + 2 %>, Integer)
```
eregon, line 151: I think it'd be good to add `COMMON_FLAGS = 2` in https://github.com/ruby/prism/blob/main/templates/template.rb to avoid hardcoding that value in a bunch of places.

templates/java/org/prism/Nodes.java.erb
```diff
@@ -135,27 +135,27 @@ public abstract class Nodes {
...
+     * <%= Prism::Template::JavaDoc.escape(flag.comment) %>
      */
-    public static final class <%= group.name %> implements Comparable<<%= group.name %>> {
-        <%- group.values.each_with_index do |value, index| -%>
+    public static final class <%= flag.name %> implements Comparable<<%= flag.name %>> {
+        <%- flag.values.each_with_index do |value, index| -%>
 
         // <%= value.comment %>
-        public static final short <%= value.name %> = 1 << <%= index %>;
+        public static final short <%= value.name %> = 1 << <%= index + 2 %>;
```
eregon, line 147: Ah yes this makes sense.
I did https://github.com/ruby/prism/pull/2956/commits/cf6334e7992ede75fc15ca663c8d0bca379cd4a2 so common flags are not serialized for all nodes, but kept the shift for consistency&simplicity, so we still need to `+ 2` when checking them.

templates/rbi/prism/node.rbi.erb
andrykonchin: Done

templates/lib/prism/node.rb.erb
```diff
@@ -389,13 +389,13 @@ module Prism
     end
   end
   <%- end -%>
-  <%- flags.each_with_index do |flag, flag_index| -%>
+  <%- flags.each_with_index do |flag| -%>
```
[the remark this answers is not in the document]
andrykonchin: Done

eregon (approved):
LGTM

kddnewton (approved):

kddnewton:
Yeah this looks good, would you mind @andrykonchin just rebasing so we can make sure the errors go away?

andrykonchin:
Yes, sure. I will rebase.

andrykonchin:
Done

PR #3055, opened 2024-09-12
Check errno for parsing directory

kddnewton:


include/prism/util/pm_string.h
```diff
@@ -93,6 +93,26 @@ void pm_string_owned_init(pm_string_t *string, uint8_t *source, size_t length);
  */
 void pm_string_constant_init(pm_string_t *string, const char *source, size_t length);
 
+/**
+ * Represents the result of calling pm_string_mapped_init or
+ * pm_string_file_init. We need this additional information because there is
+ * not a platform-agnostic way to indicate that the file that was attempted to
+ * be opened was a directory.
```
eregon, line 100: Couldn't we use fstat() or stat() before calling `pm_string_file_init()` to check if directory and raise there?
That way `pm_string_file_init()` wouldn't have to deal with these complications.

kddnewton: We could on linux but on windows we need the wide filepath in order to check, which is being allocated within this function. We also already are stat-ing the filepath inside this function, so it's just a matter of reusing it.

PR #3066, opened 2024-09-16
Document lifetime of `pm_options_t`

amomchilov:
Depends on #3027


kddnewton:
Superceded by https://github.com/ruby/prism/pull/3638

PR #3088, opened 2024-09-26
Add fields documentation for AliasMethodNode and AlternationPatternNode

dinsley:
This adds some documentation to the fields for `AliasMethodNode` and `AlternationPatternNode`. I was going to go through and do more, but wanted to get a pull-request up for review before moving forward with any others just incase. I'm trying to get more familiar with all of the node types and fields for a project that utilizes Prism, so this looks like a better (and more productive!) way to do it.

I also wanted to check if there was any other smaller intro issues/tickets where there might be an opportunity to jump in on. I took a brief look through and wasn't too sure on the scope of the existing ones on the issues board. If there was any other high-level ideas thrown around for more examples/guides for Prism I'd be happy to explore those too!

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -821,6 +821,17 @@ nodes:
         kind:
           - SymbolNode
           - InterpolatedSymbolNode
+        comment: |
+          Represents the new name of the method that will be aliased.
+
+              alias foo bar
+                    ^^^
+
+              alias :foo :bar
+                    ^^^^
+
+              alias :"#{foo}" :"#{bar}"
+                    ^^^^^^^^^
       - name: old_name
         type: node
         kind:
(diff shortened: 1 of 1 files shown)


config.yml
```diff
@@ -840,11 +862,26 @@ nodes:
       - name: left
         type: node
         kind: pattern expression
+        comment: |
+          Represents the left side of the expression. It can be any [non-void expression](https://github.com/ruby/prism/blob/main/docs/parsing_rules.md#non-void-expression).
```
eregon, line 866: > It can be any [non-void expression]

No, it's always a pattern expression, so that's incorrect.

kddnewton:
Hey @dinsley! Thanks so much for the contribution! If you don't mind doing that one change that @eregon mentioned we can get this merged.

config.yml
dinsley: Thanks! Sorry about that -- I totally missed seeing those keys.

dinsley:
Removed the incorrect references to those being able to be any non-void expresssions :)

kddnewton (approved):

Issue #3112, opened 2024-09-30
Comma allowed just after block arguments


tompng:
These are rejected in parse.y
```ruby
p(&a,)
a.(&a,)
a[&a,]
def f(&)
  p(&,)
  a.(&,)
  a[&,]
end
```

PR #3119, opened 2024-10-01
Fix 3112 - disallow commas after block arg

eileencodes:
Prism was already disallowing arguments after block args, but in parse.y, any comma after a block arg is a syntax error. This moves the error handling into `PM_TOKEN_UAMPERSAND` where we can check if the current type is `PM_TOKEN_COMMA`then raise an error. I've also updated the tests to include the examplesfrom ruby/prism#3112.

Fixes: ruby/prism#3112

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -14127,9 +14127,6 @@ parse_arguments(pm_parser_t *parser, pm_arguments_t *arguments, bool accepts_for
     bool parsed_forwarding_arguments = false;
 
     while (!match1(parser, PM_TOKEN_EOF)) {
-        if (parsed_block_argument) {
-            pm_parser_err_current(parser, PM_ERR_ARGUMENT_AFTER_BLOCK);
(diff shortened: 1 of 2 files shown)


src/prism.c
```diff
@@ -14178,6 +14175,10 @@ parse_arguments(pm_parser_t *parser, pm_arguments_t *arguments, bool accepts_for
                     arguments->block = argument;
                 }
 
+                if (parser->current.type == PM_TOKEN_COMMA) {
```
kddnewton, line 14178:
```suggestion
                if (match1(parser, PM_TOKEN_COMMA)) {
```

We use this match1 function in other places, so just updating to match that.

eileencodes: Ok, works for me 👍🏼 

Not sure what's up with the rust build.

kddnewton (approved):

PR #3130, opened 2024-10-03
Use `partial_script` for the parser translators

Earlopain:
Followup to https://github.com/ruby/prism/pull/3079


kddnewton (approved):

Earlopain:
I just noticed typecheck failures. Seems like that option is missing from `prism.rbi`. Should I add it in this PR?

Issue #3134, opened 2024-10-03
Prism accepts comma where it is not allowed after newline


tompng:
These are rejected in --parser=parse.y but accepted in --parser=prism
```ruby
p(a,b
,)

[a
,]

def f(a
,b);end
```

PR #3139, opened 2024-10-04
Emit error if comma is after newline in array literals

haldun:
Related https://github.com/ruby/prism/issues/3134


kddnewton:
Thanks @haldun — I ended up solving this in a similar way here https://github.com/ruby/prism/pull/3148.

PR #3168, opened 2024-10-08
Avoid breaking code units offset on binary encoding

vinistock:
When a file is using binary encoding and contains multibyte characters, trying to call `encode` with one of the accepted LSP encodings (UTF-8, UTF-16 or UTF-32) will fail because it's not a valid conversion.

In situations like these, we want to avoid breaking, even if we can't provide the correct locations for nodes. In the included test, you can see that the locations are all using the number of bytes, rather than the code units.

### DIFF
--- a/test/prism/ruby/location_test.rb
+++ b/test/prism/ruby/location_test.rb
@@ -140,6 +140,25 @@ module Prism
       assert_equal 7, location.end_code_units_column(Encoding::UTF_32LE)
     end
 
+    def test_code_units_handles_binary_encoding_with_multibyte_characters
+      # If the encoding is set to binary and the source contains multibyte
+      # characters, we avoid breaking the code unit offsets, but they will
+      # still be incorrect.
+
+      program = Prism.parse(<<~RUBY).value
+        # -*- encoding: binary -*-
+
+        😀 + 😀
+      RUBY
+
+      # first 😀
+      location = program.statements.body.first.receiver.location
+
+      assert_equal 4, location.end_code_units_column(Encoding::UTF_8)
+      assert_equal 4, location.end_code_units_column(Encoding::UTF_16LE)
+      assert_equal 4, location.end_code_units_column(Encoding::UTF_32LE)
+    end
+
     def test_chop
       location = Prism.parse("foo").value.location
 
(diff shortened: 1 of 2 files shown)


eregon:
`# encoding: binary` files with non-US-ASCII characters are kind of a broken concept/nasty edge case.
In both the example here and in https://github.com/Shopify/ruby-lsp/issues/2674 the actual encoding of the file is UTF-8, otherwise we wouldn't be able to show the ∂φ/∂x or smileys.

So maybe the best in that case is to consider the source as UTF-8 bytes, because BINARY is simply inaccurate and UTF-8 is by far the most likely actual encoding of the file?
I think it would then return the correct offsets for those two examples, yes?

test/prism/ruby/location_test.rb
```diff
@@ -140,6 +140,25 @@ def test_code_units
...
 
+    def test_code_units_handles_binary_encoding_with_multibyte_characters
+      # If the encoding is set to binary and the source contains multibyte
+      # characters, we avoid breaking the code unit offsets, but they will
+      # still be incorrect.
+
+      program = Prism.parse(<<~RUBY).value
+        # -*- encoding: binary -*-
+
+        😀 + 😀\n😍 ||= 😍
```
eregon, line 151: A `\n` in a heredoc seems strange, I think you don't need it here, you could just end the line after the second 😀

kddnewton: @vinistock I agree here, can we make this a newline?

eregon:
> The issue is, if the file is Shift_JIS encoded (or some other encoding) then forcing it into UTF-8 is going to result in problems, even more so than replacing invalid/undef.

True, but does this ever happen?
At least so far all `# encoding: binary` files with non-US-ASCII characters I have seen are actually UTF-8.
And the reason for that `# encoding: binary` BTW in at least some of these ruby/spec files is to workaround a bug in the parser gem (used by RuboCop, used in ruby/spec CI), which couldn't handle things like "\xAA\xBB" with AA and BB some broken UTF-8:
https://github.com/whitequark/parser?tab=readme-ov-file#invalid-characters-inside-comments-and-literals

I wonder what editors do in such a case (since most would still display non-ASCII characters). Maybe they just assume the filesystem encoding which is UTF-8? (at least on Unix, not sure on Windows)
I think Ideally the LSP in this case would ask the editor which encoding it guessed and that would be passed to Prism, but I suspect there is no easy way to find that out.

I like your `def for(source)` approach, although I think we should go straight to `BrokenSource` if it's not valid UTF-8.

kddnewton:
I'll follow up in another PR

PR #3172, opened 2024-10-09
Attempt to assume binary is UTF-8

kddnewton:


eregon (approved):
LGTM.
I'm a bit worried there might be some confusion of which encoding should be used in the `encoding: binary` case with valid UTF-8 bytes.
For example StringNode#value should still give a BINARY string, not a UTF-8 String, and the C parser should not consider anything to be UTF-8.
Basically in that case, I think UTF-8 should only be used for code_unit & codepoint offsets, and otherwise BINARY remains used as before.

eregon:
(IMO CRuby should raise for `encoding: binary` files with non-US-ASCII characters in it, TruffleRuby already does that, but it would be a breaking change)

kddnewton:
I think that's basically what we're going to have. This is only impacting the encoding of the actual Source string itself, not the encoding of the individual strings in the result. They will still match their behavior.

Agreed though that this is very weird and niche.

PR #3176, opened 2024-10-10
Do not warn for unused variables on negative lines

kddnewton:
Fixes [Bug #20788]

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -962,17 +962,19 @@ pm_locals_order(PRISM_ATTRIBUTE_UNUSED pm_parser_t *parser, pm_locals_t *locals,
             pm_constant_id_list_insert(list, (size_t) local->index, local->name);
 
             if (warn_unused && local->reads == 0) {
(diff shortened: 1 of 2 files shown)


src/prism.c
```diff
@@ -962,17 +962,19 @@ pm_locals_order(PRISM_ATTRIBUTE_UNUSED pm_parser_t *parser, pm_locals_t *locals,
             pm_constant_id_list_insert(list, (size_t) local->index, local->name);
 
             if (warn_unused && local->reads == 0) {
-                pm_constant_t *constant = pm_constant_pool_id_to_constant(&parser->constant_pool, local->name);
+                if (pm_newline_list_line(&parser->newline_list, local->location.start, parser->start_line) >= 0) {
```
eregon, line 965: One quick thought is we could skip this maybe-no-so-cheap (?) offset to line conversion if start_line >= 0, because then the line of that offset cannot be negative if the first line is >= 0.

kddnewton: Yeah great idea

PR #3178, opened 2024-10-11
Prism::StringQuery

kddnewton:
Introduce StringQuery to provide methods to access some metadata about the Ruby lexer.


ext/prism/extension.c
```diff
@@ -1133,6 +1134,67 @@ parse_file_failure_p(int argc, VALUE *argv, VALUE self) {
...
+string_query(pm_string_query_t result) {
+    switch (result) {
+        case PM_STRING_QUERY_ERROR:
+            rb_raise(rb_eArgError, "Invalid or non ascii-compatible encoding");
+            return Qfalse;
+        case PM_STRING_QUERY_FALSE:
+            return Qfalse;
+        case PM_STRING_QUERY_TRUE:
+            return Qtrue;
+    }
```
k0kubun, line 1155: `-Wreturn-type` warns this:

```
../prism/extension.c: In function ‘string_query’:
../prism/extension.c:1156:1: warning: control reaches end of non-void function [-Wreturn-type]
 1156 | }
      | ^
```

But I'm not sure how you'd like to fix this. Should it be `default: rb_bug("unreachable");`?

kddnewton: Ahh sorry about that, @S-H-GAMELINKS fixed this in a follow-up PR. But yeah, it should be unreachable, I'm not sure why the compiler can't figure out that the enum is exhaustive.

PR #3185, opened 2024-10-17
Speedup memcheck job

eregon:
* By using ubuntu-24.04 which has a recent enough valgrind: https://github.com/ruby/prism/pull/3182#issuecomment-2420294315

cc @peterzhu2118

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -204,30 +204,13 @@ jobs:
       run: bundle exec rake serialized_size:topgems
 
   memcheck:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v4
-    - name: Install libc6-dbg
-      run: |
-        sudo apt-get update
-        sudo apt-get install -y libc6-dbg
-    - name: Cache valgrind
-      uses: actions/cache@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -204,30 +204,9 @@ jobs:
...
 
   memcheck:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v4
-    - name: Install libc6-dbg
-      run: |
-        sudo apt-get update
-        sudo apt-get install -y libc6-dbg
```
eregon, line 213: I'm not sure if we still need this and/or if we need to install valgrind from apt, letting the CI find out

eregon: valgrind needs to be installed, it's not in the image and it depends on `libc6-dbg`: https://github.com/ruby/prism/actions/runs/11391934879/job/31696848683?pr=3185#step:3:35

eregon:
Success: https://github.com/ruby/prism/actions/runs/11391934879/job/31696848683?pr=3185 took 6min 23s and this older build https://github.com/ruby/prism/actions/runs/11369080496/job/31625692265 took 9min 24s.

kddnewton (approved):

Issue #3186, opened 2024-10-21
`Prism::Translation::RubyParser` - Nested string interpolations again (compatibility regression)


presidentbeef:
RubyParser 3.20.3:

```ruby
RubyParser.new.parse('"#{"#{B} C"} D"')
 => s(:dstr, "", s(:evstr, s(:const, :B)), s(:str, " C"), s(:str, " D"))
```

Prism 0.29.0 (matches RubyParser):
```ruby
Prism::Translation::RubyParser.parse('"#{"#{B} C"} D"')
 => s(:dstr, "", s(:evstr, s(:const, :B)), s(:str, " C"), s(:str, " D"))
```

Prism 0.30.0+:
```ruby
Prism::Translation::RubyParser.parse('"#{"#{B} C"} D"')
 => s(:dstr, "", s(:evstr, s(:const, :B)), s(:str, " C D"))
```

PR #3200, opened 2024-11-04
Fix up regression in ruby parser translation

kddnewton:
Fixes https://github.com/ruby/prism/issues/3186


presidentbeef:
Appreciate your hard work on this!

PR #3210, opened 2024-11-04
Bump Chicory to 1.0.0-M1

andreaTP:
:wave: we just released a first Milestone for Chicory, since it's breaking in many ways, I wanted to check that everything remains aligned.

Notes:
- we now have a `wasi` implementation, so, some more code-paths should be usable (for filesystem operation we suggest to use Jimfs though)
- we have a new AOT compiler, I would be happy to set up here, basically we can ship a "ready to run" Jar package for Prism, wdyt? (example: [wabt](https://github.com/dylibso/chicory/tree/main/wabt))

### DIFF
--- a/java-wasm/pom.xml
+++ b/java-wasm/pom.xml
@@ -15,14 +15,30 @@
     <maven.compiler.source>11</maven.compiler.source>
     <maven.compiler.target>11</maven.compiler.target>
 
+    <chicory.version>1.0.0-M1</chicory.version>
     <junit.version>5.11.3</junit.version>
   </properties>
+
+  <dependencyManagement>
+    <dependencies>
+          <dependency>
+              <groupId>com.dylibso.chicory</groupId>
+              <artifactId>bom</artifactId>
+              <version>${chicory.version}</version>
+              <type>pom</type>
+              <scope>import</scope>
+          </dependency>
+      </dependencies>
+  </dependencyManagement>
 
   <dependencies>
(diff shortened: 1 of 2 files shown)


kddnewton (approved):

kddnewton:
Yeah that sounds great! Do you mean add it as an artifact after the build kind of thing? @enebo would you be able to use that?

andreaTP:
I think @enebo will be able to use it easily, I would require some help in setting up the automated release CI,.
If that's good for you @kddnewton I'll move forward in the next few days. 😊 

BTW, a wonderful experience contributing to this repo! Great job!

kddnewton:
Yeah that's fine. I think just uploading it as an artifact to the build to start would be good. That's what we have for JS and it works fine for now https://github.com/ruby/prism/blob/efd3362a91f9fef296e42e931ef637e60cea18d1/.github/workflows/javascript-bindings.yml#L37-L40

andreaTP:
Thanks for the feedback, @kddnewton , two comments here:

- I was aiming to publish a full-fledged Jar to Maven Central as we are already uploading the wasm file to artifacts
- I need more time to finalize this proposal as we are hitting some Java bytecode limitations (we are already working on fixing the issue)

All in all, the current setup is fine for now, I'll come up with something when we are ready :slightly_smiling_face:

kddnewton:
Sounds great!

PR #3223, opened 2024-11-10
Use RbConfig to locate libprism and headers when it is a default gem

eregon:
* This is notably necessary on TruffleRuby, which is updating to Ruby 3.3 which introduces Prism as a default gem.
* Using the existing path is not an option as it would end up in truffleruby/lib/build/libprism.so and "truffleruby/lib/include/#{header}" which are not good places for such files.

### DIFF
--- a/lib/prism/ffi.rb
+++ b/lib/prism/ffi.rb
@@ -13,7 +13,15 @@ module Prism
 
     # Define the library that we will be pulling functions from. Note that this
     # must align with the build shared library from make/rake.
-    ffi_lib File.expand_path("../../build/libprism.#{RbConfig::CONFIG["SOEXT"]}", __dir__)
+    libprism_in_build = File.expand_path("../../build/libprism.#{RbConfig::CONFIG["SOEXT"]}", __dir__)
+    libprism_in_libdir = "#{RbConfig::CONFIG["libdir"]}/prism/libprism.#{RbConfig::CONFIG["SOEXT"]}"
+    if File.exist? libprism_in_build
+      INCLUDE_DIR = File.expand_path("../../include", __dir__)
+      ffi_lib libprism_in_build
+    else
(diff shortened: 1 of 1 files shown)


lib/prism/ffi.rb
```diff
@@ -13,7 +13,15 @@ module LibRubyParser # :nodoc:
...
     # Define the library that we will be pulling functions from. Note that this
     # must align with the build shared library from make/rake.
-    ffi_lib File.expand_path("../../build/libprism.#{RbConfig::CONFIG["SOEXT"]}", __dir__)
+    libprism_in_build = File.expand_path("../../build/libprism.#{RbConfig::CONFIG["SOEXT"]}", __dir__)
+    libprism_in_libdir = "#{RbConfig::CONFIG["libdir"]}/prism/libprism.#{RbConfig::CONFIG["SOEXT"]}"
+    if File.exist? libprism_in_build
+      INCLUDE_DIR = File.expand_path("../../include", __dir__)
+      ffi_lib libprism_in_build
+    else
+      INCLUDE_DIR = "#{RbConfig::CONFIG["libdir"]}/prism/include"
```
eregon, line 22: As a note I used `libdir` here and not `includedir` on purpose, those headers should not be reachable with C include statements and should not be mixed with Ruby C API headers.

eileencodes (approved):

kddnewton:
@eregon I'm not sure if I'm reading this right, but it seems like when on the FFI backend this would favor the builtin Prism instead of the one bundled with the gem. Does that mean that on the FFI backend you can no longer upgrade the Prism gem version?

eregon:
It prefers the libprism in `../../build/libprism` so that will get used when installing the gem, same as before.
It's only when it doesn't find it there, i.e. when prism is a default gem in TruffleRuby that it will use RbConfig.

kddnewton:
Ahhh I see now, thanks!

PR #3229, opened 2024-11-20
Add RDoc to development dependencies to use its new theme

st0012:
RDoc has started receiving more updates, especially around the default theme. By having it as a development dependency, we can use the latest theme for Prism's Ruby documentation too.

### Before

Screenshot 2024-11-20 at 22 55 25

### After

Screenshot 2024-11-20 at 22 54 56


kddnewton (approved):
Sounds good!

PR #3231, opened 2024-11-21
Arena Allocator

Cluster444:
Rebased from #2913 into main

### DIFF
--- a/templates/src/node.c.erb
+++ b/templates/src/node.c.erb
@@ -93,18 +93,22 @@ pm_node_list_destroy(pm_parser_t *parser, pm_node_list_t *list) {
  * using the parser argument, but it's there to allow for the future possibility
  * of pre-allocating larger memory pools.
  */
+<%- ignored_node_fields_for_destroy = [
+  Prism::Template::LocationField, Prism::Template::OptionalLocationField, Prism::Template::UInt8Field, Prism::Template::UInt32Field,
+  Prism::Template::ConstantField, Prism::Template::OptionalConstantField, Prism::Template::DoubleField, Prism::Template::ConstantListField,
+  Prism::Template::IntegerField] -%>
 PRISM_EXPORTED_FUNCTION void
 pm_node_destroy(pm_parser_t *parser, pm_node_t *node) {
     switch (PM_NODE_TYPE(node)) {
         <%- nodes.each do |node| -%>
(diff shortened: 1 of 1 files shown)


include/prism/defines.h
```diff
@@ -164,6 +164,7 @@
  * #endif
  * ```
  */
+
```
tenderlove, line 167:
```suggestion
```

src/prism.c
```diff
@@ -21505,6 +21441,7 @@ parse_expression_infix(pm_parser_t *parser, pm_node_t *node, pm_binding_power_t
         }
         case PM_TOKEN_QUESTION_MARK: {
             context_push(parser, PM_CONTEXT_TERNARY);
+
```
tenderlove, line 21444:
```suggestion
```

src/prism.c
```diff
@@ -22161,6 +22098,10 @@ pm_parser_init(pm_parser_t *parser, const uint8_t *source, size_t size, const pm
         .warn_mismatched_indentation = true
     };
 
+    // TODO: find a better starting size
```
tenderlove, line 22101: 👍 

4k is probably a fine starting size, maybe we can add some logging later and figure out a better start.

templates/ext/prism/api_node.c.erb
```diff
@@ -100,7 +100,7 @@ pm_node_stack_pop(pm_node_stack_node_t **stack) {
     const pm_node_t *visit = current->visit;
 
     *stack = current->prev;
-    xfree(current);
+    // xfree(current);
```
tenderlove, line 103: Should we just delete this?

tenderlove (commented):
I think this looks good. The arena implementation seems fine, I left a few minor commit suggestions but I don't see any actual problems.

@jhawthorn do you mind taking a look at this (if you have time)? Otherwise I'm going to merge it.

kddnewton:
@tenderlove make sure you don't merge this before we get this working in ruby/ruby using the new API, otherwise this is going to break a bunch of stuff. I've been holding off on merging this because I want to make sure we can do some benchmarking on both speed and memory from within CRuby before we go this route, since it's a such an invasive change.

kddnewton:
I'll start testing this in ruby/ruby and see what changes are necessary. In the meantime I'm going to label this PR as "blocked" just to indicate that it should not be merged until it is fully tested over there.

Issue #3171, opened 2024-10-09
`a rescue b in c .method` and `a and b in c .method` is wrongly accepted


tompng:
```ruby
'a' rescue 2 => 3.upcase
#=> "A"
1 and 2 in 3.inspect
#=> "false"
'a' or 1 => 1 + 'b'
#=> "ab"
```

PR #3242, opened 2024-11-27
Reject invalid dot method call after match predicate or after match required

ydah:
Partially fixes:  https://github.com/ruby/prism/issues/3171


kddnewton:
Thank you so much for the fix here! (And the other fixes also!) I won't have time to review this week (Thanksgiving tomorrow) but I will take a look first thing next week.

tenderlove (approved):

Issue #3106, opened 2024-09-30
Assigning command call is allowed where rejected in parse.y


tompng:
These command calls are rejected in parse.y but accepted in prism
```ruby
1 if a = b c
1 and a = b c
a = b c and 1
```

Earlopain:
I fixed the first two cases in https://github.com/ruby/prism/pull/3654 but the last one is still up for grabs

Earlopain:
Sadyly the last fix had some issues. I reverted that, so reopening.

PR #3245, opened 2024-11-27
Reject assigning command call without parentheses

ydah:
Fixes: https://github.com/ruby/prism/issues/3106


eileencodes:
It looks like this change has a legit test failure, can you take a look?

kddnewton (changes requested):
I think this is something we'll want to fix with the `accepts_command_call` variable as opposed to validating later.

PR #3272, opened 2024-12-02
GitHub Action that uploads builds as release artifacts

egiurleo:
This GitHub Action builds Prism for Linux and MacOS and then uploads the builds to the latest release.

## Motivation
I am working on adopting Prism in a third-party library (Sorbet) where Ruby is not available during the build process. In order to avoid adding Ruby to the Sorbet build pipeline, I would like to have access to pre-built Prism binaries that Sorbet can pull and use in its build process.

### DIFF
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -1,0 +1,39 @@
+name: Build Library Artifacts
+
+concurrency:
+  group: "${{github.workflow}}-${{github.ref}}"
+  cancel-in-progress: true
+
+on:
+  release:
+    types: [created]
+  workflow_dispatch:
+
+jobs:
+  build:
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-artifacts.yml
```diff
@@ -0,0 +1,40 @@
...
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: '3.2'
```
kddnewton, line 25: Can we make this 3.3?

kddnewton (changes requested):
One quick change, looks great!

tenderlove (approved):

kddnewton (approved):

PR #3274, opened 2024-12-03
`isinf` is available on MSVC

nobu:
At least Visual Studio 2015 (MSVC 14.0) provides the macro conforming to C99.
And, float.h is necessary to use `_finite()` like the current code.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -4142,14 +4142,7 @@ pm_double_parse(pm_parser_t *parser, const pm_token_t *token) {
 
     // If errno is set, then it should only be ERANGE. At this point we need to
     // check if it's infinity (it should be).
-    if (
-        errno == ERANGE &&
-#ifdef _WIN32
-        !_finite(value)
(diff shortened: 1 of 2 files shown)


nobu:
I see, the problem is mingw, not mswin.

https://github.com/ruby/prism/actions/runs/12141353366/job/33853261523?pr=3274#step:4:37
```
src/prism.c:4145:34: error: conversion from 'double' to 'float' may change value [-Werror=float-conversion]
 4145 |     if (errno == ERANGE && isinf(value)) {
      |                                  ^~~~~
```

kddnewton:
Yeah, I couldn't get the build to pass using `isinf` directly. I think mingw is non-conforming.

kddnewton:
Unless there's a way around it, I think we need to keep it as-is.

kddnewton:
Superseded by https://github.com/ruby/prism/pull/3307

PR #3307, opened 2024-12-12
Use isinf on non-mingw windows

kddnewton:
Supersedes https://github.com/ruby/prism/pull/3274


kddnewton:
@nobu I think this solves the problem for non-mingw

PR #3308, opened 2024-12-13
added nearest_target function for comments

melon64:
Added binary search to find node closest to comment, added test for it 

TODO
- Add logic for more node types for deeper traversal (to improve location accuracy)
- Add attachment logic


kddnewton:
I'm generally okay with the direction, but can you talk a little about the motivation here?

PR #3311, opened 2024-12-16
Blocks & kwargs are not valid in index assignments

eightbitraptor:
This PR is replicating the changes made in ruby/ruby#12343 and ruby/ruby#12342 in order to sync them to the upstream Prism repo.

I was under the impression that Prism was set up with a bi-directional sync, so that changes made to Prism inside the ruby repo would be mirrored here - but that doesn't seem to be the case?


kddnewton (approved):

kddnewton:
It's only 1 way unfortunately.

PR #3324, opened 2024-12-18
Enable implicit fall-through errors

amomchilov:
This PR enables [`-Wimplicit-fallthrough`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough) to catch any potentially unintended fall-through between `case` bodies.

The exiting `/* fallthrough */` comments are now formalized with a new `PRISM_FALLTHROUGH` macro which:

* Uses C23's standardized `[[fallthrough]]` attribute, if available,
* Uses `__attribute__((fallthrough))` on GCC and Clang
* Uses to `__fallthrough` on MSVC

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -9110,7 +9110,7 @@ lex_global_variable(pm_parser_t *parser) {
         case '-':
             parser->current.end++;
             allow_multiple = false;
-            /* fallthrough */
+            PRISM_FALLTHROUGH
         default: {
             size_t width;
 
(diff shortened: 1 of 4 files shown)


src/prism.c
```diff
@@ -21893,6 +21893,7 @@ parse_expression(pm_parser_t *parser, pm_binding_power_t binding_power, bool acc
             if (pm_symbol_node_label_p(node)) {
                 return node;
             }
+            break;
```
amomchilov, line 21896: This one had an implicit fall-through that wasn't annotated.

Since the `default` case it falls into only does a `break`, I just added its own `break` instead of `PRISM_FALLTHROUGH`.

tenderlove (approved):

PR #3325, opened 2024-12-19
Fix `it` lexing

tenderlove:
Given the following code:

```ruby
i=2; 42.tap { it /1/i }
```

Before this commit, the tokenizer would treat `it /1/i` as a method call with a regular expression parameter.  `it` is to be treated as a local variable, so this needs to be tokenized as division rather than a regular expression.  In other words it should be tokenized as `it / 1 / i`

This commit fixes tokenization.

[Bug #20970]

Bug is here: https://bugs.ruby-lang.org/issues/20970#change-111104

This makes the AST correct, but I haven't tested it with `prism_compile.c`

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -11874,7 +11874,8 @@ parser_lex(pm_parser_t *parser) {
                         !(last_state & (PM_LEX_STATE_DOT | PM_LEX_STATE_FNAME)) &&
                         (type == PM_TOKEN_IDENTIFIER) &&
                         ((pm_parser_local_depth(parser, &parser->current) != -1) ||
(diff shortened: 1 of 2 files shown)


tenderlove:
I _think_ we're treating `it` as a local variable.  Even after this change, `it` won't appear in the local variable lists in Prism, though I believe https://github.com/ruby/ruby/pull/12398 takes care of the case. I think we should probably be fixing Prism to include `it` in it's tables, and then we can probably remove https://github.com/ruby/ruby/pull/12398 ?

I'm not sure.

tenderlove:
Seems like we don't need this anymore.

PR #3336, opened 2024-12-21
Support 3.5 for version option

kddnewton:


eregon:
@kddnewton Currently there are bunch of versions comparisons like https://github.com/ruby/prism/blob/3c4851cbdc001f6c02720819c357400f4251f30c/src/prism.c#L2969 in prism.c
I think they are correct because all conditions use PM_OPTIONS_VERSION_CRUBY_3_3 (and it is the lowest version) and none use 3_4/LATEST (which would most likely result in a bug) but it feels brittle.
WDYT about giving enum values that increase with newer versions?
That way one could do e.g. `parser->version >= PM_OPTIONS_VERSION_CRUBY_3_4` instead of `parser->version != PM_OPTIONS_VERSION_CRUBY_3_3`, which seems a lot clearer about intent (behavior for 3.4+).

PR #3337, opened 2024-12-21
Support leading logical operators

kddnewton:
Per https://bugs.ruby-lang.org/issues/20925


src/prism.c
```diff
@@ -10842,14 +10842,37 @@ parser_lex(pm_parser_t *parser) {
...
+                                if (!lexed_comment) parser_lex_ignored_newline(parser);
+                                lexed_comment = false;
+                                goto lex_next_token;
+                            }
+
+                            // If we hit a '.' or a '&.' we will lex the ignored
+                            // newline.
+                            if (following && (
+                                (peek_at(parser, following) == '.') ||
+                                (peek_at(parser, following) == '&' && peek_at(parser, following + 1) == '.')
```
Morriar, line 10849: Would it be better to use store `peek_at(parser, following)` into an intermediate variable?

src/prism.c
```diff
@@ -10842,14 +10842,37 @@ parser_lex(pm_parser_t *parser) {
...
-                                (following && (
-                                    (peek_at(parser, following) == '.') ||
-                                    (peek_at(parser, following) == '&' && peek_at(parser, following + 1) == '.')
-                                ))
+                                (parser->version == PM_OPTIONS_VERSION_LATEST) &&
+                                following && (
+                                    (peek_at(parser, following) == '&' && peek_at(parser, following + 1) == '&') ||
+                                    (peek_at(parser, following) == '|' && peek_at(parser, following + 1) == '|') ||
+                                    (peek_at(parser, following) == 'a' && peek_at(parser, following + 1) == 'n' && peek_at(parser, following + 2) == 'd' && !char_is_identifier(parser, following + 3, parser->end - (following + 3))) ||
+                                    (peek_at(parser, following) == 'o' && peek_at(parser, following + 1) == 'r' && !char_is_identifier(parser, following + 2, parser->end - (following + 2)))
```
Morriar, line 10866: Same question here about all the `peek_at(parser, following)`

src/prism.c
```diff
@@ -10842,14 +10842,37 @@ parser_lex(pm_parser_t *parser) {
...
+                            // If we are parsing as CRuby 3.5 or later and we
+                            // hit a '&&' or a '||' then we will lex the ignored
+                            // newline.
                             if (
-                                lex_state_ignored_p(parser) ||
-                                (following && (
-                                    (peek_at(parser, following) == '.') ||
-                                    (peek_at(parser, following) == '&' && peek_at(parser, following + 1) == '.')
-                                ))
+                                (parser->version == PM_OPTIONS_VERSION_LATEST) &&
```
Earlopain: I found this version check a bit strange since `LATEST` is a moving target and openend https://github.com/ruby/prism/pull/3605 to talk about it.

Earlopain: PR above was merged, this should now be this

```suggestion
                                (parser->version >= PM_OPTIONS_VERSION_CRUBY_3_5) &&
```

src/prism.c
```diff
@@ -10889,6 +10912,63 @@ parser_lex(pm_parser_t *parser) {
                             parser->next_start = NULL;
                             LEX(PM_TOKEN_AMPERSAND_DOT);
                         }
+
+                        if (parser->version == PM_OPTIONS_VERSION_LATEST) {
```
Earlopain:
```suggestion
                        if (parser->version >= PM_OPTIONS_VERSION_CRUBY_3_5) {
```

XrXr:
FYI Matz approved this feature a week ago in https://bugs.ruby-lang.org/issues/20925#note-11

ric2b:
Hey @kddnewton, this change seems to turn methods called `and?` and `or?` into syntax errors when called, is that intended?

image

Earlopain:
Hm, no. I don't think that is intentional. Do you have an account on https://bugs.ruby-lang.org to report this? If not, I can do it for you. It will need to be backported and be fixed in parse.y as well.

If you need to work around this now, you can add empty parens to your method like `def logical?()`

ric2b:
I don't have an account there, no, feel free to report yourself. 

Wrapping with parenthesis solves it, but it did look unintentional so I wanted to mention it 👍

Earlopain:
I reported at https://bugs.ruby-lang.org/issues/21946, https://github.com/ruby/prism/pull/3966 will fix this. It will probably be backported.

Issue #2539, opened 2024-03-02
Extra rubocop offenses when running rubocop on ruby/spec


eregon:
From https://github.com/rubocop/rubocop/pull/12724#issuecomment-1974766497
I noticed there are 3 more offenses when running with `ParserEngine: parser_prism`:
```
core/objectspace/define_finalizer_spec.rb:166:1: W: Lint/ImplicitStringConcatenation: Combine  and  into a single string literal, rather than using implicit string concatenation.
    RUBY

language/fixtures/bytes_magic_comment.rb:2:26: F: Lint/Syntax: invalid multibyte character 0xA0
(Using Ruby 3.3 parser; configure using TargetRubyVersion parameter, under AllCops)
$magic_comment_result = '你好'.bytes.inspect
                         ^^
language/fixtures/bytes_magic_comment.rb:2:27: F: Lint/Syntax: invalid multibyte character 0xBD
(Using Ruby 3.3 parser; configure using TargetRubyVersion parameter, under AllCops)
$magic_comment_result = '你好'.bytes.inspect
                           ^^
```
I guess there are due to some differences between parser gem and `Prism::Translation::Parser`.

kddnewton:
Something very odd is happening with `Parser::Source::Buffer`. We're passing in a big5-encoded string, and it looks like it's transcoding it to utf-8. So then when it hits the multi-byte characters in the source, it has reencoded them as utf-8 which is invalid in big5. I think we need to tell it to please stop doing that.

CKolkey:
I'm not sure if this is the right place to be tracking these, but I also found that the following cops were being triggered incorrectly when using implicit keyword values:

```ruby
# Lint/UnusedMethodArgument

def foo(bar)
  Baz.new(bar:)
end
```

```ruby
# Lint/UnusedBlockArgument

[1, 2, 3].each do |n|
  Foo.new(n:)
end
```

That said, it found these (incorrect) issues _very_ quickly, so, nice work :D

kddnewton:
Thanks @CKolkey I've opened https://github.com/ruby/prism/pull/2563 to fix this. If you find more, feel free to open new tickets.

andyw8:
Here's a minimal example that results in a false positive for `Layout/ExtraSpacing` due to the emoji:

```ruby
# 👎
<<-RUBY.squish
  a
  b
RUBY
```

kddnewton:
[...]

PR #3338, opened 2024-12-21
Fix an incompatibility with the parser translator

Earlopain:
For https://github.com/rubocop/rubocop/issues/12982

The offset cache contains an entry for each byte so it can't be accessed via the string length. For `’` the bytesize is 3, so it's two bytes off. Adds tests for all variants except for this:
```
"fo
o" "ba
’"
```

For some reason, this still has the wrong offset. Also, one test fails in `RubyParser` but that is certainly out of my area of expertise.


kddnewton (approved):

Earlopain:
Actually, I think the reason why that other test I wanted to add doesn't work is for the same reason as https://github.com/ruby/prism/issues/2539#issuecomment-2052598076

In isolation the test works but when there is more than one multi-byte character it breaks. The offset is not correct when there were previous multibyte characters in the file.

PR #3341, opened 2024-12-23
Implement squiggly heredocs for the parser translator

Earlopain:
In parser, the string content is dedented. This implements these rules as far as I was able to understand them. It's all a bit confusing with spaces/tabs, always learning more funny things about them.

I refered to the prism implementation, parser seems to adhere to it very closely.


lib/prism/translation/parser/lexer.rb
```diff
@@ -321,17 +321,39 @@ def to_a
...
+                heredoc_stack.push(heredoc)
               end
             when :tSTRING_CONTENT
-              unless (lines = token.value.lines).one?
+              if (lines = token.value.lines).one?
+                # Heredoc interpolation can have multiple STRING_CONTENT nodes on the same line.
+                is_first_token_on_line = lexed[index - 1] && token.location.start_line != lexed[index - 2][0].location&.start_line
+                # The parser gem only removes indentation when the heredoc is not nested
+                not_nested = heredoc_stack.size == 1
+                if is_first_token_on_line && not_nested && (heredoc = heredoc_stack[0]).common_whitespace > 0
```
Earlopain, line 353: Any tip for getting this through sorbet? It complains that `heredoc` can be nil but that will never be the case because the previous condition checked that the first element is present.

When rewriting this into something that proves to sorbet that it can't be nil, it's telling me that the condition is always falsy.

Earlopain: Got a workaround from https://github.com/sorbet/sorbet/issues/8401#issuecomment-2559830955 that seems to work

kddnewton (approved):
Wow, amazing work. I don't know that I can properly review this because obviously the code is quite complex. But the tests are passing and you've got even more tests passing, so I'm assuming this is working. Nice job!

PR #3343, opened 2024-12-24
Better handle heredoc escaping in the parser translator

Earlopain:
Builds on top of https://github.com/ruby/prism/pull/3341, will fix https://github.com/rubocop/rubocop/issues/12878

This does a few things:
* After a line continuation in a heredoc, emit only a single token
* Dedent ` heredocs
* Properly handle escapes in single and doublequoted heredocs

Some of the heredoc escaping logic should be applied to strings as well,
but I'll leave that for a different PR.


kddnewton (approved):
Makes sense. I think we will need to share this escaping logic with the parser translation as well, but that can be done later.

PR #3347, opened 2024-12-26
Fix parser translator ast for heredoc with written newlines

Earlopain:
Heredocs don't don't start a new string node when encountering "\\\n".

Builds on top of #3343, but only because the tokens would otherwise not line up. It's otherwise totally standalone. Will actually fix https://github.com/rubocop/rubocop/issues/12878

### DIFF
--- a/lib/prism/translation/parser/compiler.rb
+++ b/lib/prism/translation/parser/compiler.rb
@@ -2079,27 +2079,49 @@ module Prism
 
                 escaped_lengths = []
                 normalized_lengths = []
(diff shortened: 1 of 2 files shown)


kddnewton:
@Earlopain just to see this working as expected, would you mind rebasing this branch now that the other stuff is merged?

Earlopain:
Happy new year to you! Thanks for taking a look at the other PRs.

Apologies for the scope and mishmash of things in those, I was largely working through tests and just seeing what they need to pass. Looks like this one is still good to go

kddnewton (approved):

PR #3361, opened 2025-01-03
Implement more string token escaping in the parser translator

Earlopain:
This leaves `\c` and `\M` escaping but I don't understand how these should even work yet. Maybe later, they seem uncommon enough.

It seems a bit redundant to implement this again when prism already knows how to do all this, and will probably do a better job as well. Maybe it's ok? The implementation is not too complicated I think. Open to hear different approaches.


kddnewton:
I could be convinced to go this way, but definitely would prefer to share the logic instead of reimplementing. It would involve exposing the unescaping logic from `prism.c` as `Prism.*` functions that we could reuse. Maybe we can save that for a later refactoring.

Earlopain:
Yes, I totally agree that that should be shared, `escape_read` looks like it is doing the bulk of the heavy lifting and I certainly don't want to replicate all of that. But I don't think I'm able to do so myself, I never did much C and it looks very interconnected with doing actual parsing (which makes sense).

For now I would say this is good enough but I'd also understand if you don't want to bother with this. I'll open an issue for such an API if that is OK with you.

kddnewton:
On second thought, I'm okay with this for now. I don't think it's going to be duplicated anywhere else, and I don't mind having a Ruby implementation around. If you don't mind just rebasing, I'm happy to implement the \C and \M escapes.

Earlopain:
Thanks! I'll rebase the other PR after this is merged since I'm pretty sure all the skips I added there will pass with this change.

Issue #3345, opened 2024-12-25
Lexing a global variable will read beyond the end of the source given to the parser, causing an assertion failure


lopopolo:
I've hooked up prism to a rudimentary libFuzzer harness using `cargo-fuzz` and got a crash:

```console
$ cargo +nightly fuzz run prism -- -rss_limit_mb=512
    Finished `release` profile [optimized + debuginfo] target(s) in 0.00s
    Finished `release` profile [optimized + debuginfo] target(s) in 0.00s
     Running `fuzz/target/aarch64-apple-darwin/release/prism -artifact_prefix=/Users/lopopolo/dev/artichoke/prism-fuzz/fuzz/artifacts/prism/ -rss_limit_mb=512 /Users/lopopolo/dev/artichoke/prism-fuzz/fuzz/corpus/prism`
prism(38610,0x1ebdbf840) malloc: nano zone abandoned due to inability to reserve vm space.
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 1494678647
INFO: Loaded 1 modules   (1256 inline 8-bit counters): 1256 [0x104916ab0, 0x104916f98),
INFO: Loaded 1 PC tables (1256 PCs): 1256 [0x104916f98,0x10491be18),
INFO:        2 files found in /Users/lopopolo/dev/artichoke/prism-fuzz/fuzz/corpus/prism
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
INFO: seed corpus: files: 2 min: 1b max: 1b total: 2b rss: 43Mb
#3      INITED cov: 23 ft: 23 corp: 2/2b exec/s: 0 rss: 44Mb
#1048576        pulse  cov: 23 ft: 23 corp: 2/2b lim: 4096 exec/s: 524288 rss: 464Mb
#2097152        pulse  cov: 23 ft: 23 corp: 2/2b lim: 4096 exec/s: 524288 rss: 468Mb
Assertion failed: (parser->current.end <= parser->end), function parser_lex, file prism.c, line 10614.
==38610== ERROR: libFuzzer: deadly signal
    #0 0x000104e340f0 in __sanitizer_print_stack_trace+0x28 (librustc-nightly_rt.asan.dylib:arm64+0x5c0f0)
    #1 0x0001047e6060 in fuzzer::PrintStackTrace()+0x30 (prism:arm64+0x100062060)
[...]

PR #3365, opened 2025-01-05
Fix global variable read off end

kddnewton:
Fixes https://github.com/ruby/prism/pull/3365


lopopolo:
Linking https://github.com/ruby/prism/issues/3345

@kddnewton your fixes comment uses the same issue number as this PR.

lopopolo:
Are there tests that could be added here? This fixed a (or several?) memory safety violation, correct?

Issue #3381, opened 2025-01-08
`a => [-2**b]` should be SyntaxError but parse succeeds


tompng:
Parsing `a in [-2**b]` succeeds and executing it crash with this error
```
$ ruby -e "a in [-2**b]"
-e: [BUG] Unexpected node type in pattern matching expression: PM_CALL_NODE
ruby 3.5.0dev (2025-01-06T01:50:53Z master c8e3d745fa) +PRISM [x86_64-linux]
```

Found this code failed to `RubyVM::InstructionSequence.compile` while compiling every ruby code fragments in ruby repository
```ruby
# spec/ruby/core/marshal/fixtures/marshal_data.rb:328
"Integer -2**8" => [-2**8,
                       "\004\bi\377\000"]
```

Earlopain:
Should be reopened, it was closed again by accident

PR #3382, opened 2025-01-08
Reject pattern match with unexpected double splat inside array

ydah:
`a => [-2**b]` should be SyntaxError

Fixes: https://github.com/ruby/prism/issues/3381


Earlopain:
This seems to have some unintended consequence: https://github.com/ruby/ruby/actions/runs/12675763001/job/35327482624

```rb
[-2**32,2**32]
=> [4294967296, 4294967296] # Both are positive
```

kddnewton:
I've opened a PR to revert this. Unfortunately doing this in the general expression parser does not take into account that this is a pattern match and should be rejected. We can't solve this in that place.

PR #3383, opened 2025-01-08
Only use ffi on Ruby >= 3.0.0

tenderlove:
This is for #3368


kddnewton:
I think this is probably fine, but this isn't the gemfile that has the issue (it's the 2.7 gemfile).

eregon:
Ref: https://github.com/ruby/prism/pull/3386

PR #3388, opened 2025-01-09
Add one more token to lparen conversion for the parser translator

Earlopain:
Before:
```
  0...3: tIDENTIFIER     => "foo"
  4...5: tLPAREN_ARG     => "("
  5...6: tLPAREN2        => "("
  6...7: tINTEGER        => 1
  7...9: tDOT2           => ".."
 9...10: tINTEGER        => 1
10...11: tRPAREN         => ")"
11...12: tRPAREN         => ")"
12...13: tNL             => nil
```

After:
```
  0...3: tIDENTIFIER     => "foo"
  4...5: tLPAREN_ARG     => "("
  5...6: tLPAREN         => "("
  6...7: tINTEGER        => 1
  7...9: tDOT2           => ".."
 9...10: tINTEGER        => 1
10...11: tRPAREN         => ")"
11...12: tRPAREN         => ")"
12...13: tNL             => nil
```


kddnewton (approved):

kddnewton:
You are doing amazing work here, btw. Thank you so much for digging into this, it has been on my list for so long and now it's actually getting really close!

Earlopain:
Thanks! I hope to have the translator be the only parser for RuboCop on new Ruby versions starting with 3.5 (https://github.com/rubocop/rubocop/issues/13617) since it will be pretty hard to keep up otherwise with the parser deprecation and all that.

I think it's doing a damn well job already, so kudos to you for starting it, especially the ast is in a very good state. Just a handful of discrepancies I notice, in addition to the tests available here.

It'll be an interesting question on how to handle `it` and other new node types when `parser` doesn't support that itself. Maybe the simplest thing would be to just inline the builder since it's doubtful that it'll get any changes anymore, but I'll leave that as a thing for the not-so-distant future for now.

PR #3389, opened 2025-01-10
WIP: Improve the `build-artifacts` GitHub action

egiurleo:
Ignore this for now -- working on the best approach!

Follow up to #3272

After doing some testing on the Sorbet project, we realized that we want to package pre-built headers along with a binary (because some of the headers are only generated during the compilation process), and also that Bazel works a lot better with static libraries, hence uploading `libprism.a` rather than the `dylib` or `so` file. 

I tested this job on the [Shopify fork](https://github.com/Shopify/prism/actions/runs/12712675711) to make sure it still works.

### DIFF
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -28,12 +28,17 @@ jobs:
     - name: Build library
       run: bundle exec rake compile
 
+    - name: Package files
+      run: |
+        mkdir dist
+        cp -r include dist/
+        cp build/libprism.a dist/
+        tar -czf prism-${{ matrix.os }}.tar.gz -C dist .
+
     - name: Upload to release
       if: github.event_name == 'release'
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       uses: softprops/action-gh-release@v2
       with:
-        files: |
-          build/libprism.so
-          build/libprism.dylib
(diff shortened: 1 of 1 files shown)


.github/workflows/build-artifacts.yml
```diff
@@ -28,12 +28,17 @@ jobs:
...
     - name: Upload to release
       if: github.event_name == 'release'
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       uses: softprops/action-gh-release@v2
       with:
-        files: |
-          build/libprism.so
-          build/libprism.dylib
+        files: prism-${{ matrix.os }}.tar.gz
```
paracycle, line 44: I am not sure if this will be enough, since for mac and linux we still need to differentiate between `arm64` and `amd64` platforms, etc.

Besides, IMO `prism-macos-latest.tar.gz` is a weird name for the package.

st0012: Based on [this chart](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories), it looks like to run amd64 macOS we need to add `macos-13` to the `os` matrix too.

egiurleo:
Closed in favor of #3391

PR #3390, opened 2025-01-10
Fix binary encoding for parser translator

Earlopain:
Skipping detecting the encoding is almost always right, just for binary it should actually happen.

A symbol containing escapes that are invalid in utf-8 would fail to parse since symbols must be valid in the script encoding.
Additionally, the parser gem would raise an exception somewhere during string handling.

The parser gem reject code where a string is invalid in the script encoding. RuboCop however monkey-patches that behaviour out, I have mirrored that logic for the tests here.

### DIFF
--- a/lib/prism/translation/parser.rb
+++ b/lib/prism/translation/parser.rb
@@ -51,7 +51,7 @@ module Prism
         source = source_buffer.source
 
         offset_cache = build_offset_cache(source)
-        result = unwrap(Prism.parse(source, filepath: source_buffer.name, version: convert_for_prism(version), partial_script: true, encoding: false), offset_cache)
+        result = unwrap(Prism.parse(source, **prism_options), offset_cache)
 
         build_ast(result.value, offset_cache)
       ensure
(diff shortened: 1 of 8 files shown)


kddnewton:
It seems like this is causing a bunch of new files to fail. I'm not sure if I understand fully what's going on there. Why are we adding a bunch of skips?

Earlopain:
Ah, yes. That is from the first commit, they were previously rejected by the parser gem as invalid syntax.

I made that change since I noticed that the `encoding_shift_jis` file I added supposedly passed tests even though I was certain that at least the tokens must not line up currently.

RuboCop accepts such code via monkeypatch here: https://github.com/rubocop/rubocop-ast/blob/02b8d0faeeebc3ba6b5284c9e66569bbae836941/lib/rubocop/ast/builder.rb#L112-L114
And this is the parser issue with a bit more context https://github.com/whitequark/parser/issues/283

kddnewton:
@Earlopain if you don't mind just rebasing here I can merge this

kddnewton (approved):

Earlopain:
Merged the two commits since all newly running tests from that monkeypatch pass now.

PR #3391, opened 2025-01-10
Modify `build-artifacts.yml` to package libprism source instead

st0012:
Instead of uploading the built `libprism.so` and `libprism.dylib`, which still requires headers files to be built, Prism can upload the full source, headers, and the Makefile as a tarball that can be used to build `libprism` from source.

The reason default release artifacts are not enough is because some source and header files are generated through Prism's rake tasks.

Example release: https://github.com/Shopify/prism/releases/tag/v1.3.2

Note: This is @paracycle's brilliant idea 🚀

### DIFF
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -11,10 +11,7 @@ on:
 
 jobs:
   build:
-    strategy:
-      matrix:
-        os: [ubuntu-latest, macos-latest]
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


paracycle (approved):
Looks good to me, thank you!

.github/workflows/build-artifacts.yml
```diff
@@ -11,10 +11,7 @@ on:
 
 jobs:
   build:
-    strategy:
-      matrix:
-        os: [ubuntu-latest, macos-latest]
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
```
st0012, line 14: Since we're not building binary files anymore, we don't need to run this on multiple systems.

PR #3401, opened 2025-01-12
Fix parser translator when unescaping invalid utf8

Earlopain:
1. The string starts out as binary
2. `ち` is appended, forcing it back into utf-8
3. Some invalid byte sequences are tried to append

> incompatible character encodings: UTF-8 and BINARY (ASCII-8BIT)

This makes use of my wish to use `append_as_bytes`. Unfortunatly that method is rather new so it needs a fallback.

It's for https://github.com/ruby/prism/issues/2539. Since ruby-spec has some rather odd-looking code it runs into it.


kddnewton (changes requested):
Could you add this to the polyfills directory? We already have one for byteindex and one for unpack1, so may as well add another one here

Earlopain:
Works for me, didn't know that was a thing.

I really appreciate how fast you're getting to these. Thanks!

kddnewton (approved):

PR #3407, opened 2025-01-13
Fix parser translator tokens for %-arrays with whitespace escapes

Earlopain:
Also fixes a token incompatibility for the word separator. parser only considers whitespace until the first newline


kddnewton (approved):
That's kind of wild.

PR #3408, opened 2025-01-13
Be a bit stricter when comparing parser translator tokens

Earlopain:
Now it's possible to receive an assertion like this:
```
expected: []
actual: [:tNL, [nil, #<Parser::Source::Range comment_single.txt 8...9>]]
<[]> expected but was
<[:tNL, [nil, #<Parser::Source::Range comment_single.txt 8...9>]]>
```

All tests pass so it's no big deal but in https://github.com/ruby/prism/pull/3399 I almost introduced a different bug where this check would then have failed like this instead of silently passing.


kddnewton (approved):
I can't wait to have `assert_equal expected_tokens, actual_tokens`.

PR #3417, opened 2025-01-16
Allow to test a custom fixtures path during testing

Earlopain:
Of course, these won't really be fixtures, but it allows to test against whole codebases without copying them, doing symlinks or something like that.

For example, I can tell that over the whole RuboCop codebase, there are only 8 files that produce mismatched ast. Telling what the problem is is a different problem. The ast for real files can and will be huge so I haven't checked (maybe parser bug) but it's nice for discoverability regardless


test/prism/snapshots_test.rb
```diff
@@ -60,7 +60,7 @@ def assert_snapshot(fixture)
         # If the snapshot file exists, then assert that the printed value
         # matches the snapshot.
         assert_equal(saved, printed)
-      else
+      elsif !Fixture.custom_base_path?
```
kddnewton, line 63: Can you change this to just skip this whole test if there's a custom base path? I found it confusing to have this case in here.

Something like `return if Fixture.custom_base_path?` at the top after the test helper require would probably work.

kddnewton (changes requested):

test/prism/snapshots_test.rb
Earlopain: Agreed, that's cleaner

kddnewton (approved):

PR #3418, opened 2025-01-16
Handle control and meta escapes in parser translation

kddnewton:


Earlopain:
Cool! I feared it would be a bit more involved. I haven't used the scanner before so thanks for cleaning that up a bit also (:

PR #3419, opened 2025-01-17
Fix parser translator ast for regex with fake newlines

Earlopain:
These newlines are preserved, so in prism they will not create an extra line in the unescaped content.
Additionally, apply the same logic to interpolated regex nodes.


lib/prism/translation/parser/compiler.rb
```diff
@@ -2144,6 +2145,7 @@ def string_nodes_from_line_continuations(unescaped, escaped, start_offset, openi
               .each do |lines|
                 escaped_lengths << lines.sum(&:bytesize)
                 unescaped_lines_count = lines.sum do |line|
+                  next 0 if regex # Will always be preserved as is
```
kddnewton, line 2148: Can we put this if statement around the map here instead of checking each time?

kddnewton (changes requested):

kddnewton (approved):

PR #3441, opened 2025-01-20
Fix parser translator rescue location with semicolon body

Earlopain:
There are a few other locations that should be included in that check. I think the end location must always be present but I left the fallback in to be safe (maybe implicit begin somehow?)


kddnewton:
Nice work! How are you finding these inconsistencies at this point? Working through the parser test suite?

kddnewton (approved):

Issue #3449, opened 2025-01-28
Reducing noise from dependabot PRs


eregon:
The [current dependabot config](https://github.com/ruby/prism/blob/a46edfd951e1d6e2d1411261aed84fc613511468/.github/dependabot.yml) creates a [ridiculous amount of PRs](https://github.com/ruby/prism/issues?q=is%3Apr+author%3Aapp%2Fdependabot) and that turns into tons of spam-like emails for everyone watching this repo (which I would presume are interested in Prism development but not in tons of dependency bumps).
FWIW I have tried to setup an email filter to exclude those but it's tricky and not perfect, there seems to be no easy way to filter those.

I can imagine some ways to fix this:
* Convince dependabot to create one PR weekly with all updates instead of a dozen PRs.
* Drop dependabot, if there not enough values in these updates. For example, are #3405 #3438 #3440 useful?
* Remove the Gemfile.lock files entirely, like many gems do, and just automatically use the latest gems without a ton of extra PRs/spam. The CI logs will anyway show which versions have been used.
* Consolidate the various Gemfile.lock files to have a smaller amount of them.

Thoughts?

eregon:
A quick search found https://slar.se/dependabots-dependency-grouping.html and https://github.blog/changelog/2023-08-24-grouped-version-updates-for-dependabot-are-generally-available/
That seems an easy solution.

eregon:
Here is a try: https://github.com/ruby/prism/pull/3450

eregon:
https://github.com/ruby/prism/pull/3450 seems good enough, I'll close this then

PR #3450, opened 2025-01-28
Group dependabot updates in fewer PRs to minimize noise

eregon:
* See #3449

I followed https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates#prioritizing-meaningful-updates and https://slar.se/dependabots-dependency-grouping.html

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,10 +4,18 @@ updates:
     directory: '/'
     schedule:
       interval: 'weekly'
+    groups:
+      single-pr:
+        patterns:
+          - "*"
   - package-ecosystem: 'maven'
     directory: '/java-wasm'
     schedule:
       interval: 'weekly'
+    groups:
+      single-pr:
+        patterns:
+          - "*"
   - package-ecosystem: 'bundler'
     directories:
       - '/gemfiles/2.7'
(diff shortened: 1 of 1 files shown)


kddnewton:
Makes sense to me let's try it

kddnewton (approved):

kddnewton:
They usually come in on monday so we'll see this week

eregon:
It seems better indeed: https://github.com/ruby/prism/pull/3453 bumped actions in 2 workflows
and https://github.com/ruby/prism/pull/3454 bumped 2 different gems in 2 different gemfiles :tada: 

The name `single-pr` seems not the best group name looking at these PR titles, I'll do a quick PR to change that to `actions-deps:`/`ruby-deps:`/etc

eregon:
There now seems to be a way to group all dependabot updates in a single PR:
https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#multi-ecosystem-groups-

Do we want that? (if so I can make a PR)
Or we prefer one PR per ecosystem as currently?

Earlopain:
I don't really see much value in multiple PRs at once and would welcome this change.

kddnewton:
I like one PR per ecosystem because the Java ones tend to break, and it's nice to be able to merge the Ruby ones without having to worry about the Java ones.

PR #3458, opened 2025-02-06
Move rust code out of build.rs and into lib.rs

segiddins:
It is easier to follow non-templated rust code when it is in the source file rather than a string


kddnewton:
Hey @segiddins! I'm not opposed to this change, but can you take a look at the failures on this PR?

kddnewton (approved):
Looks good, thanks!

PR #3463, opened 2025-02-13
Expose prettyprint/serialize/parse options in rust

segiddins:


kddnewton:
I'm okay with prettyprint and the parse options, but I'm not sure I understand the use case for the serialization API in rust. Could you explain the motivation?

segiddins:
The motivation is the ability to parse the AST into a more natural shape for use in rust, similar to what the prism ruby gem does (rather than maintaining pointers to the c structure nodes). I am using it in https://github.com/segiddins/gemfile-rs/blob/main/src/prism_ast/generated.rs, an in-progress project of mine

kddnewton:
@segiddins Thanks for your patience.

I'm fine accepting the options from the rust side (I definitely think we should do that). I'm much less keen to expose the serialization format. We have only done that for languages that didn't have direct C interop, which I'd like to keep that way. That's why we have the ruby-prism (as opposed to the ruby-prism-sys) crate. Does that crate work for your use case?

Either way, would you be able to split up this PR? The first would be the options, which I'd be happy to merge immediately. The other would be the other APIs you're trying to expose.

PR #3481, opened 2025-02-26
Support `itblock` for `Prism::Translation::Parser`

koic:
## Summary

`itblock` node is added to support the `it` block parameter syntax introduced in Ruby 3.4.

```console
$ ruby -Ilib -rprism -rprism/translation/parser34 -e 'buffer = Parser::Source::Buffer.new("path"); buffer.source = "proc { it }"; \
                                                      p Prism::Translation::Parser34.new.tokenize(buffer)[0]'
s(:itblock,
  s(:send, nil, :proc), :it,
  s(:lvar, :it))
```

This node design is similar to the `numblock` node, which was introduced for the numbered parameter syntax in Ruby 2.7.

```
$ ruby -Ilib -rprism -rprism/translation/parser34 -e 'buffer = Parser::Source::Buffer.new("path"); buffer.source = "proc { _1 }"; \
                                                      p Prism::Translation::Parser34.new.tokenize(buffer)[0]'
s(:numblock,
  s(:send, nil, :proc), 1,
  s(:lvar, :_1))
```

The difference is that while numbered parameters can have multiple parameters, the `it` block parameter syntax allows only a single parameter.

In Ruby 3.3, the conventional node prior to the `it` block parameter syntax is returned.

```console
$ ruby -Ilib -rprism -rprism/translation/parser33 -e 'buffer = Parser::Source::Buffer.new("path"); buffer.source = "proc { it }"; \
                                                      p Prism::Translation::Parser33.new.tokenize(buffer)[0]'
s(:block,
  s(:send, nil, :proc),
  s(:args),
  s(:send, nil, :it))
```

## Development Note

The Parser gem does not yet support the `it` block parameter syntax. This is the first case where Prism's node design precedes that of the Parser gem. When implementing https://github.com/whitequark/parser/issues/962, this node design will need to be taken into consideration.

cc @whitequark @iliabylich

### DIFF
--- a/test/prism/ruby/parser_test.rb
+++ b/test/prism/ruby/parser_test.rb
@@ -6,6 +6,7 @@ begin
   verbose, $VERBOSE = $VERBOSE, nil
   require "parser/ruby33"
   require "prism/translation/parser33"
+  require "prism/translation/parser34"
 rescue LoadError
   # In CRuby's CI, we're not going to test against the parser gem because we
   # don't want to have to install it. So in this case we'll just skip this test.
@@ -144,6 +145,22 @@ module Prism
       end
     end
 
+    def test_it_block_parameter_syntax
+      it_fixture_path = Pathname(__dir__).join("../../../test/prism/fixtures/it.txt")
+
+      buffer = Parser::Source::Buffer.new(it_fixture_path)
+      buffer.source = it_fixture_path.read
+      actual_ast = Prism::Translation::Parser34.new.tokenize(buffer)[0]
+
+      it_block_parameter_sexp = parse_sexp {
+        s(:itblock,
+          s(:send, nil, :x), :it,
+          s(:lvar, :it))
+      }
+
+      assert_equal(it_block_parameter_sexp, actual_ast.to_sexp)
+    end
+
     private
 
     def assert_equal_parses(fixture, compare_asts: true, compare_tokens: true, compare_comments: true)
@@ -241,6 +258,10 @@ module Prism
         "expected: #{expected_comments.inspect}\n" \
         "actual: #{actual_comments.inspect}"
       }
+    end
+
+    def parse_sexp(&block)
+      Class.new { extend AST::Sexp }.instance_eval(&block).to_sexp
     end
   end
 end
--- a/lib/prism/translation/parser/builder.rb
+++ b/lib/prism/translation/parser/builder.rb
@@ -6,7 +6,55 @@ module Prism
       # A builder that knows how to convert more modern Ruby syntax
       # into whitequark/parser gem's syntax tree.
       class Builder < ::Parser::Builders::Default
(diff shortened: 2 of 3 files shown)


test/prism/fixtures/whitequark/it_arg_after_34.txt
```diff
@@ -0,0 +1,7 @@
+-> do it + 42 end
```
Earlopain, line 1: This code is actually not really tested, only `Parser33` is run in the test suite. I thought about it a bit before, and probably a separate directory for new syntax would make sense (or something along those lines).

You probably need to assert specifically against the ast anyways, as it only compares to the parser gem, and there is nothing to compare against in this case.

koic: For now, I've added a test to check the `:itblock` node type, which is currently provided only by `Prism::Translation::Parser`. Since it is essentially an extension of the Parser gem, I've added it to `parser_test.rb`.  

Additionally, I've implemented a mechanism to compare node compatibility with the `Parser::Ruby34`, but since the Parser gem doesn't yet support the `it` block parameter, it is currently skipped.

test/prism/ruby/parser_test.rb
```diff
@@ -242,5 +261,11 @@ def assert_equal_comments(expected_comments, actual_comments)
         "actual: #{actual_comments.inspect}"
       }
     end
+
+    def assert_prism_only_node(actual_ast, fixture_path)
```
Earlopain, line 265: It's doubtful if parser will ever support 3.4 and this seems highly specific. It also only works because this is valid syntax to the `parser` gem, if some new syntax gets added you can't test this via this mechanism.

I think I would just add new specific tests for this syntax. If the parser gem does end up supporting it, then these test cases can just be imported. Something like this:

```rb
# Pseudo-code like
def test_it_syntax
  parser = Prism:Parser::Translator::Ruby34.new
  expected = s(:itblock, s(:begin, s(:itarg)))
  assert_equal(expected, parser.tokenize(it_fixture)[0].to_sexp)
end
```

Then, you can extend the existing `it.txt` (stuff in `fixtures/whitequark` is pulled from upstream) and explicitly assert against it. What do you think?

koic: It is currently unclear how the Parser gem will support Ruby 3.4 and later, but I agree with the idea of focusing on the necessary test for now. And using S-expressions in `Prism::Translation::Parser` to test preceding syntax is a good idea, as it clarifies the expected syntax tree. I have updated the test accordingly.

Earlopain: This looks good to me, thanks!

It might be a bit annoying to have to update the expectation when the fixture changes, inlining the code so the test is entirely self-contained might be worth a thought. But I think for now this is fine.

kddnewton (approved):

PR #3482, opened 2025-02-27
Use a locale-insensitive version of tolower

byroot:
[[Bug #21161]](https://bugs.ruby-lang.org/issues/21161)

The `tolower` function provided by the libc is locale dependent and can behave in ways you wouldn't expect for some value of `LC_CTYPE`.


kddnewton (approved):

kddnewton:
Thanks!

PR #3491, opened 2025-03-12
Mark extension as Ractor-safe

kddnewton:
@tenderlove does this look okay to you? I'm less familiar with Ractors, but wanted to make sure this was okay. From the documentation on Ractor safety, I believe this is okay?

### DIFF
--- a/ext/prism/extension.c
+++ b/ext/prism/extension.c
@@ -1331,6 +1331,11 @@ Init_prism(void) {
         );
     }
 
+#ifdef HAVE_RB_EXT_RACTOR_SAFE
+    // Mark this extension as Ractor-safe.
+    rb_ext_ractor_safe(true);
+#endif
+
     // Grab up references to all of the constants that we're going to need to
     // reference throughout this extension.
     rb_cPrism = rb_define_module("Prism");
--- a/lib/prism/ffi.rb
+++ b/lib/prism/ffi.rb
@@ -7,6 +7,10 @@
 require "rbconfig"
 require "ffi"
 
+# We want to eagerly load this file if there are Ractors so that it does not get
+# autoloaded from within a non-main Ractor.
+require "prism/serialize" if defined?(Ractor)
+
 module Prism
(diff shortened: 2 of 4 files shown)


tenderlove (approved):
Change makes sense to me. I guess there's some tests failing though?

kddnewton:
Yeah, just because I forgot to check `HAVE_RB_EXT_RACTOR_SAFE` for Ruby 2.7.

XrXr:
A bunch of weird test failures started happening in ruby/ruby when this landed:

```
  
    1) Failure:
  TestObjectSpace#test_each_object [/Users/runner/work/ruby/ruby/src/test/ruby/test_objectspace.rb:226]:
  <1> expected but was
  <0>.
  
    2) Failure:
  TestObjectSpace#test_each_object_enumerator [/Users/runner/work/ruby/ruby/src/test/ruby/test_objectspace.rb:237]:
  <1> expected but was
  <0>.
  
    3) Error:
  TestObjectSpace#test_id2ref_32:
  Test::Unit::ProxyError: "32616" is id of the unshareable object on multi-ractor
      /Users/runner/work/ruby/ruby/src/test/ruby/test_objectspace.rb:32:in 'ObjectSpace._id2ref'
      /Users/runner/work/ruby/ruby/src/test/ruby/test_objectspace.rb:32:in 'block in TestObjectSpace#deftest_id2ref'
```

```
    klass = Class.new
    new_obj = klass.new

    found = []
    count = ObjectSpace.each_object(klass) do |obj|
      found << obj
    end
#>> assert_equal(1, count)
```

Seems something is missing. https://github.com/ruby/ruby/actions/runs/13819663391/job/38661667031#step:14:1953

eregon:
I suspect that's because this adds tests for Ractor in the same process, and Ractor simply isn't compatible enough (e.g. it changes the semantics of `bjectSpace.each_object`), and so no Ractor must be created in the main test process in CRuby test-all.

PR #3494, opened 2025-03-13
Ensure backwards compatibility with the custom parser builder

Earlopain:
Temoprary backwards-compat code so that current users don't break. I noticed this by accident while doing some testing

Eventually the `Translation::Parser` initializer should asser that the correct class is passed in. I would start with a warning after https://github.com/rubocop/rubocop-ast/pull/354 is merged and released.

cc @koic

### DIFF
--- a/lib/prism/translation/parser/compiler.rb
+++ b/lib/prism/translation/parser/compiler.rb
@@ -1157,7 +1157,17 @@ module Prism
         # -> { it }
         # ^^^^^^^^^
         def visit_it_parameters_node(node)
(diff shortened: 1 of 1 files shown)


kddnewton (approved):

Earlopain:
Thanks @kddnewton!

If I may ask, when would the next release be? It would help RuboCop further integrating prism, as I don't think there are any outstanding changes than can be made on the RuboCop side without a new release.

It would also be nice for all the translation bugfixes that have happened since.

PR #3501, opened 2025-03-19
Warn when the parser translator receives an incompatible builder class

Earlopain:
In https://github.com/ruby/prism/pull/3494 I added a bit of code so that using the new builder doesn't break stuff. This code can be dropped when it is enforced that builder is _always_ the correct subclass (and makes future issues like that unlikely).

### DIFF
--- a/lib/prism.rb
+++ b/lib/prism.rb
@@ -68,6 +68,7 @@ module Prism
 end
 
 require_relative "prism/polyfill/byteindex"
+require_relative "prism/polyfill/warn"
 require_relative "prism/node"
 require_relative "prism/node_ext"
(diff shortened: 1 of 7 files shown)


Earlopain:
I was using `category: :deprecation` but tests told be _no_. It was only added in Ruby 3.0, so that makes sense.

I copied that from one other deprecation I found. Since this is broken on Ruby 2.7, I removed that keyword in a separate commit. I guess that code is never tested.

kddnewton:
I ended up adding it as another polyfill here, so we could continue to use the `category` keyword without worrying about it.

kddnewton (approved):

PR #3503, opened 2025-03-19
Mark Prism as ractor-safe

kddnewton:


test/prism/ractor_test.rb
```diff
@@ -0,0 +1,77 @@
...
+    def with_ractor(*arguments, &block)
+      reader, writer = IO.pipe
+
+      pid = fork do
+        reader.close
+        writer.puts(ignore_warnings { Ractor.new(*arguments, &block) }.take)
+      end
+
+      writer.close
+      result = reader.gets.chomp
```
nobu, line 71: `reader` should be closed.
See ruby/ruby@a51364f54b6

nobu: Or, just use `IO.popen`.

```ruby
    def with_ractor(*arguments, &block)
      IO.popen("-") do |reader|
        next reader.gets.chomp if reader

        puts(ignore_warnings { Ractor.new(*arguments, &block) }.take)
      end
    end
```

test/prism/ractor_test.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+return unless defined?(Ractor)
+
+require_relative "test_helper"
+
+return if Prism::TestCase.windows?
```
nobu, line 7: I think this condition should be whether `fork` is implemented or not, instead of the platform name.

```ruby
return  unless defined?(fork)
```

kddnewton: https://github.com/ruby/prism/pull/3505

PR #3504, opened 2025-03-20
Enforce a minimum `parser` version for the parser translator

Earlopain:
There hasn't been much that would actually affect parsers usage of it. But, when adding new node types, these usually appear in the `Parser::Meta::NODE_TYPES`.

`itblock` was added, gets emitted by prism, and then `rubocop-ast` blindly delegates to `on_itblock`. These methods are dynamically created through `NODE_TYPES`, which means that it will error if it doesn't contain `itblock`. Ref https://github.com/rubocop/rubocop-ast/issues/366

This is unfortunate because in `rubocop-ast` these methods are eagerly defined but the prism translator is lazily loaded on demand.
The simplest solution is to add them on the `parser` side (even if they are not emitted directly), and require that a version that contains those be used.

In summary when adding a new node type:
* Add it to `Parser::Meta::PRISM_TRANSLATION_PARSER_NODE_TYPES` (gets included in `NODE_TYPES`), see https://github.com/whitequark/parser/pull/1071
* Bump the minimum `parser` version used by `prism` to a version that contains the above change
* Actually emit that node type in `prism`

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -15,7 +15,7 @@ GEM
       mini_portile2 (~> 2.8.2)
       racc (~> 1.4)
     onigmo (0.1.0)
-    parser (3.3.7.0)
+    parser (3.3.7.2)
       ast (~> 2.4.1)
       racc
     power_assert (2.0.5)
--- a/gemfiles/2.7/Gemfile.lock
+++ b/gemfiles/2.7/Gemfile.lock
@@ -8,7 +8,7 @@ GEM
   specs:
     ast (2.4.2)
     onigmo (0.1.0)
-    parser (3.3.7.1)
+    parser (3.3.7.2)
       ast (~> 2.4.1)
       racc
     power_assert (2.0.5)
--- a/gemfiles/3.0/Gemfile.lock
+++ b/gemfiles/3.0/Gemfile.lock
@@ -13,7 +13,7 @@ GEM
       mini_portile2 (~> 2.8.2)
       racc (~> 1.4)
     onigmo (0.1.0)
-    parser (3.3.7.1)
+    parser (3.3.7.2)
       ast (~> 2.4.1)
       racc
     power_assert (2.0.4)
--- a/gemfiles/3.1/Gemfile.lock
+++ b/gemfiles/3.1/Gemfile.lock
@@ -13,7 +13,7 @@ GEM
       mini_portile2 (~> 2.8.2)
       racc (~> 1.4)
     onigmo (0.1.0)
-    parser (3.3.7.1)
+    parser (3.3.7.2)
       ast (~> 2.4.1)
       racc
     power_assert (2.0.5)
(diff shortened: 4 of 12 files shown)


Earlopain:
Test failure is from https://github.com/ruby/ruby/commit/820c541671d9485144d81138bb029f2da8379edd, needs reverse sync

kddnewton (commented):
Does this mean that we will always need to be on the latest parser every time we emit new node types? That feels not great. Is there any way we could do some method missing shenanigans and then warn?

Earlopain:
Yeah, the moment a new node type gets added the minimum parser version would have to be bumped. `parser` actually has a test that all node types it emits are part of this list, and with prism there's an obvious disconnect there.

So it's definitly part of the api that this list is exhaustive but I have no real idea on how prism can achieve that when these parts of the code are only lazily loaded.

Maybe with `const_added` hook somehow, and put that in the main entrypoint? But that is only Ruby >= 3.2 and adding one of those to `Module` doesn't make me very comfortable.

Earlopain:
It would be possible to work around this in rubocop-ast with something like you mentioned where it defines such methods with method_missing when it encounters them but that would only be my second-favorite solution. I did actually experiment with that already but it may not only impact rubocop which is why I went here instead.

kddnewton:
Would it be possible to check the version of parser that gets loaded and then include modules that contain the newest node supports?

Earlopain:
Ah, you mean something like this here and branch on the parser version? https://github.com/ruby/prism/blob/640b1e71d3adb85711ea434c49afef1a46b6cc1d/lib/prism/translation/parser/compiler.rb#L1159-L1170

If so, I don't really think so since it would then emit ast that doesn't represent what it actually encountered.

Earlopain:
There aren't actually that many users of this constant https://github.com/search?q=%2FParser%3A%3AMeta%3A%3ANODE_TYPES%2F+lang%3Aruby+-is%3Afork+-path%3A*.rbi&type=code&p=2

There's `mutant` from @\mbj which I am vaguely familiar with but I can't tell how that would be affected by missing node types if it starts to use prism. The rest seems to be almost exclusivly non-linked forks or vendored code. So maybe being more defensive in `rubocop-ast` to accomodate prism is the solution.

Or do you maybe have an idea on how to solve this differently @koic?

kddnewton:
I was thinking something like:

```ruby
if Parser::VERSION >= "..."
 def visit_it_parameters_node(node)
   if builder.is_a?(Translation::Parser::Builder)
     builder.itarg
   else
     builder.args(nil, [], nil, false)
   end
  end
else
 def visit_it_parameters_node(node)
   builder.args(nil, [], nil, false)
 end
end
```

Would something like that work?

kddnewton:
What if we polyfill the new stuff in, in case it's not yet there? Then the translation layer doesn't have to worry about it.

Earlopain:
That doesn't really work, it would require prism to be loaded immediatly after parser, so that no code depending on that constant is executed before.

kddnewton (approved):
Okay, after all of this discussion I think I'm okay with this approach. I don't think adding new node types is all that common anyway. It will probably be fine.

PR #3505, opened 2025-03-20
Update ractor_test.rb per review

kddnewton:


test/prism/ractor_test.rb
```diff
@@ -1,11 +1,9 @@
 # frozen_string_literal: true
 
-return unless defined?(Ractor)
+return if !defined?(Ractor) || !defined?(fork)
```
eregon, line 3: `defined?(fork)` doesn't work, since the method exists even on Windows, on TruffleRuby, etc.
The correct way is `Process.respond_to?(:fork)` (https://github.com/oracle/truffleruby/blob/master/doc/user/compatibility.md#fork)

kddnewton: Are you sure? It appears that this passed the windows and truffleruby CIs?

kddnewton: Ahh, it doesn't matter here because Ractor will never be defined here.

eregon: I'm sure for TruffleRuby and JRuby and just checked locally:
```
$ ruby -ve 'p defined?(Process.fork)'
jruby 9.4.2.0 (3.1.0) 2023-03-08 90d2913fda OpenJDK 64-Bit Server VM 17.0.12+7 on 17.0.12+7 +jit [x86_64-linux]
"method"
$ ruby -ve 'p defined?(Process.fork)'
truffleruby 25.0.0-dev-be122da7, like ruby 3.3.5, GraalVM CE Native [x86_64-linux]
"method"
```
I'm not sure for CRuby on Windows, maybe `defined?` handles unimplemented methods specially.

Regardless, it would be good to fix this, because TruffleRuby/JRuby might possibly define Ractor one day.

eregon: Indeed `defined?(fork)` returns nil on CRuby on Windows:
https://github.com/eregon/actions-shell/actions/runs/13975003694/job/39126439100#step:4:569
But "method" on JRuby/TruffleRuby, so not a good way to test that.

nobu: Shouldn’t JRuby/TruffleRuby fix it?

eregon: Probably, yes. But it won't help for existing releases so I think the value is fairly low, given `Process.respond_to?(:fork)` works well and is already used in most places.
It seems completely untested and maybe arguable behavior whether an unimplemented method is "defined" or not (but it probably makes sense to just follow CRuby behavior there).
FWIW I checked and `Process.singleton_class.method_defined?(:fork)` behaves consistently with `defined?(fork)`, i.e., also doesn't work.

PR #3508, opened 2025-03-20
Add `Prism::Translation::ParserCurrent`

Earlopain:
It's not my favorite api but for users that currently use the same thing from `parser`, moving over is more difficult
than it needs to be.

If you plan to support both old and new ruby versions, you definitly need to branch somewhere on the ruby version
to either choose prism or parser.
But with prism you then need to enumerate all the versions again and choose the correct one.

Also, don't recommend to use `Prism::Translation::Parser` in docs. It's version-less but actually always just uses Ruby 3.4 which is probably not what the user intended.

Note: parser also warns when the patch version doesn't match what it expects. But I don't think prism has such a concept, and anyways it would require releases anytime ruby releases, which I don't think is very desirable.

I wrote something on how to migrate to parser/prism in https://github.com/whitequark/parser/pull/1072 and it contains the following snippet:

```rb
code_version = RUBY_VERSION.to_f

if code_version <= 3.3
  require 'parser/current'
  parser = Parser::CurrentRuby
else
  require 'prism'
  case code_version
  when 3.3
    Prism::Translation::Parser33
  when 3.4
    Prism::Translation::Parser34
  else
    warn "Unknown Ruby version #{code_version}, using 3.4 as a fallback"
    Prism::Translation::Parser34
  end
end
```

With this change, it can be written like this (and doesn't require changes from the programmer when new rubies get released)
```rb
code_version = RUBY_VERSION.to_f


if code_version <= 3.3
  require 'parser/current'
  parser = Parser::CurrentRuby
else
  require 'prism'
  parser = Prism::Translation::ParserCurrent
end
```

### DIFF
--- a/docs/parser_translation.md
+++ b/docs/parser_translation.md
@@ -6,17 +6,23 @@ Prism ships with the ability to translate its syntax tree into the syntax tree u
 
 The `parser` gem provides multiple parsers to support different versions of the Ruby grammar. This includes all of the Ruby versions going back to 1.8, as well as third-party parsers like MacRuby and RubyMotion. The `prism` gem provides another parser that uses the `prism` parser to build the syntax tree.
 
-You can use the `prism` parser like you would any other. After requiring the parser, you should be able to call any of the regular `Parser::Base` APIs that you would normally use.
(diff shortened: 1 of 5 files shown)


kddnewton (approved):
Yeah this seems fine

Earlopain:
Thanks!

Issue #3502, opened 2025-03-19
node_id missing for JRuby/TruffleRuby serialization


enebo:
#2383 added support for node_id but if the other implementations wants to be able to support this same error reporting then we also need it.

This is a pretty unfortunate feature request as it is a lot of extra space/memory just for a reproducible id.  Serialized files could avoid it with a second predictable tree-walk to build up the list but it is probably faster to have those ids packed as fields even if the serialized file is bigger.

eregon:
What do you need node_id for?

enebo:
For this:

https://github.com/rails/rails/blob/main/actionview/lib/action_view/template.rb#L231

I can see us making a more standard way of getting node_id than RubyVM::AbstractSyntaxTree but without coming up with another solution we need node_id.

eregon:
I think we need to come with another solution and a proper API for this anyway.
JRuby and TruffleRuby can't define RubyVM, it would break many gems which e.g. assume `RubyVM::InstructionSequence` exists if `defined?(RubyVM)`.
For example a method which takes a Thread::Backtrace::Location and return the Prism node for that (that's all standard, no RubyVM involved).
An example implementation on TruffleRuby for that is to reparse and find the node which has the same start/end offset.

If there is the problem that e.g. offsets might change due to an added magic comment or something, I think that should be handled in the caller to make sure it passes the same source code (I guess reading from disk is not an option in that usage as it's generated template code).

enebo:
@eregon I was not implying we should support RubyVM but we need a mechanism to get the node and once we had that we can PR something more sensible there.  This code definitely needs something other than relying on MRI internal APIs (no argument there).

Your location-based idea I think is ok?    It does make me wonder why Rails did it this way though.

kddnewton:
I tried the location based approach for a long time, but it ended up not working when I tried to port it to error highlight. Once everything was in place, there ended up being problems getting it to locate the actual nodes, particularly within the Rails error handling code. I ended up switching to node ids and it because much, much simpler.

The added benefit is that if you provided a compatible API for `RubyVM::AbstractSyntaxTree.node_id_for_backtrace_location`, then everything in error highlight would just work out of the box.

I'll add the node_id to the serialization.

enebo:
@eregon if you can get something to work with Location then I think if you made a PR for that then I would be happy to remove node_id.  I take @kddnewton at his word that he tried to make using only Location work and there were problems.  Having said that perhaps you can figure out what he didn't.

[...]

PR #3513, opened 2025-03-21
Add node_id to Java serialization

kddnewton:
Fixes https://github.com/ruby/prism/issues/3502

### DIFF
--- a/templates/java/org/prism/Loader.java.erb
+++ b/templates/java/org/prism/Loader.java.erb
@@ -358,6 +358,7 @@ public class Loader {
 
     private Nodes.Node loadNode() {
         int type = buffer.get() & 0xFF;
+        int nodeId = loadVarUInt();
         int startOffset = loadVarUInt();
         int length = loadVarUInt();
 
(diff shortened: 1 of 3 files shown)


eregon:
Could you add it only for JRuby?
See https://github.com/ruby/prism/blob/804d2c62c242209265eb2c3c1bbf6f6dfb65a16e/templates/template.rb#L15 for how to check that.

For TruffleRuby we already keep exact start/end lin+column information at runtime, so we should have all the information needed without needing `node_id`.
Using `node_id` in TruffleRuby seems a bad idea since it would make all Prism nodes bigger and all Truffle nodes bigger, when I believe there is no need (at least currently).

eregon:
Before: https://github.com/ruby/prism/actions/runs/13990575114/job/39173423666
```
Total sizes for top 100 gems:
total serialized size:  66147373
total serialized/total source: 0.733

average: 0.800
median:  0.779
1st quartile: 0.568
3rd quartile: 1.007
min - max: 0.076 - 3.675
```

After:
```
Total sizes for top 100 gems:
total serialized size:  82954905
total serialized/total source: 0.920

average: 0.907
median:  0.890
1st quartile: 0.638
3rd quartile: 1.149
min - max: 0.078 - 3.948
```

I definitely don't want such a big increase in TruffleRuby for something that won't be used.

PR #3516, opened 2025-03-22
Compile Prism to Bytecode with Chicory

andreaTP:
Finally got it! :tada: :tada: :tada: 

@enebo I'm not 100% sure about the API that you'd like to expose, happy to do any required change or handover to you, but we should be able to publish this as a self-contained jar.

### DIFF
--- a/.github/workflows/java-wasm-bindings.yml
+++ b/.github/workflows/java-wasm-bindings.yml
@@ -24,7 +24,7 @@ jobs:
           bundler-cache: true
 
       - name: rake templates
-        run: PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS=1 bundle exec rake templates
+        run: PRISM_EXCLUDE_PRETTYPRINT=1 PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS=1 bundle exec rake templates
 
       - name: Set up WASI-SDK
         run: |
@@ -32,7 +32,7 @@ jobs:
           tar xvf wasi-sdk-25.0-x86_64-linux.tar.gz
 
       - name: Build the project
-        run: make java-wasm WASI_SDK_PATH=$(pwd)/wasi-sdk-25.0-x86_64-linux PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS=1
+        run: make java-wasm WASI_SDK_PATH=$(pwd)/wasi-sdk-25.0-x86_64-linux
 
       - name: Set up Java
         uses: actions/setup-java@v4
--- a/java-wasm/src/main/java/org/prism/Prism.java
+++ b/java-wasm/src/main/java/org/prism/Prism.java
@@ -1,5 +1,6 @@
 package org.prism;
 
+import com.dylibso.chicory.runtime.ByteArrayMemory;
 import com.dylibso.chicory.runtime.ExportFunction;
 import com.dylibso.chicory.runtime.ImportValues;
 import com.dylibso.chicory.runtime.Instance;
@@ -25,6 +26,7 @@ public class Prism implements AutoCloseable {
     public Prism(WasiOptions wasiOpts) {
         wasi = WasiPreview1.builder().withOptions(wasiOpts).build();
         instance = Instance.builder(PrismModule.load())
+            .withMemoryFactory(ByteArrayMemory::new)
             .withMachineFactory(PrismModule::create)
             .withImportValues(ImportValues.builder().addFunction(wasi.toHostFunctions()).build())
             .build();
--- a/java-wasm/src/main/java/org/prism/Prism.java
+++ b/java-wasm/src/main/java/org/prism/Prism.java
@@ -1,0 +1,68 @@
+package org.prism;
+
+import com.dylibso.chicory.runtime.ExportFunction;
+import com.dylibso.chicory.runtime.ImportValues;
+import com.dylibso.chicory.runtime.Instance;
+import com.dylibso.chicory.wasi.WasiOptions;
+import com.dylibso.chicory.wasi.WasiPreview1;
+
+import java.nio.charset.StandardCharsets;
+
+public class Prism implements AutoCloseable {
+    private final ExportFunction calloc;
+    private final ExportFunction pmSerializeParse;
+    private final ExportFunction pmBufferInit;
+    private final ExportFunction pmBufferSizeof;
+    private final ExportFunction pmBufferValue;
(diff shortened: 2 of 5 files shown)


java-wasm/src/main/java/org/prism/Prism.java
```diff
@@ -0,0 +1,68 @@
...
+        pmBufferInit.apply(bufferPointer);
+
+        pmSerializeParse.apply(
+            bufferPointer[0], sourcePointer[0], source.length(), optionsPointer[0]);
+
+        var result = instance.memory().readBytes(
+            (int) pmBufferValue.apply(bufferPointer[0])[0],
+            (int) pmBufferLength.apply(bufferPointer[0])[0]);
+
+        return Loader.load(result, sourceBytes);
```
andreaTP, line 59: We should probably call the corresponding `free` after the `calloc`.

kddnewton (approved):
Amazing! I'll leave this to @enebo to review, but it looks fine to me.

enebo:
@kddnewton I will likely look this over on Tuesday.  @andreaTP Fantastic that you figured this out.

kddnewton:
@andreaTP purely out of curiosity, would you mind explaining the work here? I _think_ I understand that it's providing the ability to compile it AOT, but I'd love to know more.

andreaTP:
@kddnewton thanks for the question, what is happening here [is this](https://github.com/roastedroot/sqlite4j?tab=readme-ov-file#how-it-works) `sed 's/sqlite/prism/g`, basically:

- we compile Prism to WASM with high size optimization and excluding the pretty print to avoid JVM's code size limits
- at build time we use the Chicory AOT compiler and obtain from it 3 things:
    - some `.class` files containing the Java bytecode translated from the Wasm functions
    - a shrinked wasm module(with the function bodies stripped down) to be used for the wasm module initialization
    - a little source wrapper to make things fit together
- the resulting library is pure Java bytecode and can be distributed to run on any JVM (and even compiled back to native with GraalVM `native-image`)

You can check the result by running `mvn clean package` in the `java-wasm` folder, the resulting `target/java-prism-<version>.jar` is the self contained Java Prism library, depending only on the runtime datastructures of Chicory.

I let @enebo review here as the exposed API is still kind of "dummy" but I have a couple of possible steps in mind:

- provide all the automated exports automatically using the [`WasmModuleInterface`](https://chicory.dev/docs/experimental/host-modules) - (I need to fix a bug and improve a little before being able to do it)
- publish the library to Maven Central for general consumption, we have at least 2 ways of doing it:
    - publish "officially" under the `org.prism` groupId, if you are not already publishing something we can go through the process together
    - publish under a third party organization( [e.g.](https://github.com/roastedroot) ) I'm autonomous here but we need a mechanism to keep things in sync
    
In this way Java developers will be able to use Prism directly(for whatever reason, e.g. tooling etc.) as a plain Java dependency, e.g.:

```xml
<dependency>
    <groupId>org.prism</groupId>
    <artifactId>prism</artifactId>
    <version>0.0.1</version>
</dependency>
```
Please note that the API will be pretty low level(`malloc` and `free` included), so, offering an higher level API for the most common operations (like I demonstrated [here](https://github.com/ruby/prism/pull/3516/files#diff-704247145288a00b0f0d6ab143f0d8a38f7c97f748f654000fb25ec510d9a26dR40) ) is something we should aim for in the longer run.

I hope this answer the question!

andreaTP:
For anyone following I prepared the mentioned [next iteration](https://github.com/ruby/prism/commit/5c7d25e13b459e33ca38c941140b15937bccf870) based on [this PR](https://github.com/dylibso/chicory/pull/811).

.github/workflows/java-wasm-bindings.yml
```diff
@@ -24,15 +24,15 @@ jobs:
...
-        run: PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS=1 bundle exec rake templates
+        run: PRISM_EXCLUDE_PRETTYPRINT=1 PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS=1 bundle exec rake templates
 
       - name: Set up WASI-SDK
         run: |
           wget https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-25/wasi-sdk-25.0-x86_64-linux.tar.gz
           tar xvf wasi-sdk-25.0-x86_64-linux.tar.gz
 
       - name: Build the project
-        run: make java-wasm WASI_SDK_PATH=$(pwd)/wasi-sdk-25.0-x86_64-linux PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS=1
```
enebo, line 35: `PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS=1` has to be specified somewhere or the resulting AST created will include syntax tool info in addition to what we need for the parser so serialization will not work.  Is this maybe happening elsewhere?

enebo:
@kddnewton to back up @andreaTP comments above this is the plan for how we deal with platforms we cannot compile on (or haven't compiled on yet until we figure out how to pre-bundle common platform prism binaries).  If it runs well enough then perhaps we never have to compile to a native platform but I am not optimistic.  In theory, JRuby can still run without any foreign calls (running in "pure Java mode") so it also helps with that.

.github/workflows/java-wasm-bindings.yml
andreaTP: Correct, this one was redundant as already applied here:
https://github.com/ruby/prism/pull/3516/files#diff-236daf15d21928259835dcaf94f20bd485d8b3f33ed4df0ccece2278f1d0262bR27

enebo:
@kddnewton This PR is good.  I will let you merge since you asked the last question :smile:

andreaTP:
> but I am not optimistic

@enebo would you be able to provide some numbers?
I expect to have improved performances, at least, of one order of magnitude compared to the interpreter.

enebo:
@andreaTP I was just saying I was not optimistic it would be faster than us making a foreign call to the C dynamic library.  I am very hopeful it will be a lot faster than the non-AOT version.  I am pretty hopeful that it will perform similar or better to the existing LALR legacy parser.

I will try and get some numbers this week though.  I remember the interpreter was multiples slower than the C callout but I do not think it was >10 so if it is an order that will be very exciting indeed!

andreaTP:
Thanks for merging @enebo , I'll follow up with the improvements after the next release of Chicory.

kddnewton:
Ahh sorry I missed your last comment @enebo but thanks for merging! This stuff looks great!

PR #3520, opened 2025-03-24
Mention the prism/parser guide from the `parser` gem

Earlopain:
I wrote something up on how to use both if you need to also parse older ruby versions. https://github.com/whitequark/parser/blob/master/doc/PRISM_TRANSLATION.md

### DIFF
--- a/docs/parser_translation.md
+++ b/docs/parser_translation.md
@@ -20,6 +20,9 @@ Prism::Translation::ParserCurrent.parse("puts 'Hello World!'")
 
 All the parsers are autoloaded, so you don't have to worry about requiring them yourself.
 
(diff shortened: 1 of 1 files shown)


docs/parser_translation.md
```diff
@@ -20,6 +20,9 @@ Prism::Translation::ParserCurrent.parse("puts 'Hello World!'")
 
 All the parsers are autoloaded, so you don't have to worry about requiring them yourself.
 
+If you also need to parse Ruby versions below 3.3 (which `prism` has no support for), check out
```
Morriar, line 23: I was surprised by this but indeed it looks like Prism has no support < 3.3: https://github.com/ruby/prism/pull/2364.

Morriar (approved):

PR #3523, opened 2025-03-25
feat(rust): add `Integer::to_u32_digits`

Stranger6667:
This PR adds `Integer::to_u32_digits`, which is similar to [BigInt::to_u32_digits](https://docs.rs/num-bigint/latest/num_bigint/struct.BigInt.html#method.to_u32_digits) from the `num_bigint` crate, except it does not allocate a vector for digits and returns a boolean for the sign part to be more in line with the current set of types in the `prism` crate.

The main benefit of this change is that it allows for almost direct construction of `BigInt`


kddnewton (approved):

kddnewton:
Yeah, looks good, thanks for the contribution!

Stranger6667:
Thank you, @kddnewton !

Let me fix the lints :)

PR #3529, opened 2025-03-30
Fix parser translator when splatting in pattern matching pin

Earlopain:
Fixes https://github.com/rubocop/rubocop/issues/14056 because it ends up treating it as a local variable, and `a.x` is not a valid local variable name.

I'm not big on pattern matching, but conceptually it makes sense to me to treat anything inside ^() to not be
pattern matching syntax?


kddnewton (approved):
Yes, this is definitely the correct fix. Would you mind rebasing?

PR #3537, opened 2025-04-01
[java-wasm] Allow serialize to be called separately from serializeParse.

enebo:
JRuby internally uses loader differently and also encodes options differently.

Also added MemoryPages options since in testing one file complained about memory usage.

Lastly, original method was making a String from bytes so allow serialize to accept bytes and
sourceLength since sometimes you have source smaller than the byte[] backing store.

### DIFF
--- a/java-wasm/src/main/java/org/prism/Prism.java
+++ b/java-wasm/src/main/java/org/prism/Prism.java
@@ -2,11 +2,11 @@ package org.prism;
 
 import com.dylibso.chicory.runtime.ByteArrayMemory;
 import com.dylibso.chicory.experimental.hostmodule.annotations.WasmModuleInterface;
-import com.dylibso.chicory.runtime.ByteArrayMemory;
 import com.dylibso.chicory.runtime.ImportValues;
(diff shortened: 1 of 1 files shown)


java-wasm/src/main/java/org/prism/Prism.java
```diff
@@ -22,7 +22,7 @@ public Prism() {
     public Prism(WasiOptions wasiOpts) {
         wasi = WasiPreview1.builder().withOptions(wasiOpts).build();
         instance = Instance.builder(PrismModule.load())
-            .withMemoryFactory(ByteArrayMemory::new)
+            .withMemoryFactory(limits -> new ByteArrayMemory(new MemoryLimits(10, MemoryLimits.MAX_PAGES)))
```
andreaTP, line 25: - nitpick - it will be good if the memory limits can be tuned when creating the object.

Maybe in the next iteration we can make use of a `Builder` to lock the constructor down.

andreaTP (commented):
LGTM, thanks @enebo !

kddnewton (approved):

PR #3538, opened 2025-04-02
Bump deprecated Ubuntu runners

andreaTP:
We experienced failures due to the GH brownout, this should be the fix. ( cc. @enebo )

### DIFF
--- a/.github/workflows/cruby-bindings.yml
+++ b/.github/workflows/cruby-bindings.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   test-all:
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


enebo (approved):

enebo:
@kddnewton This has all jobs green again.  If there is a specific requirement here for the older Ubuntu then this can be reverted or changed.  Going to optimistically merge.

enebo:
Thanks @andreaTP!

kddnewton:
I assume this is fine. There's a thing where we want to be running with the same packages as BASE_RUBY, but I think they may have updated it recently so we're probably fine.

PR #3551, opened 2025-04-17
Be more defensive in the parser translator lexer

Earlopain:
Fixes https://github.com/rubocop/rubocop/issues/14100

Generally I have been good about safely accessing the tokens but failed to properly guard against no tokens in places
where it could theoretically happen through invalid syntax.

I added a test case for one occurance, other changes are theoretical only.

Also, TIL that this works:
```rb
a = [[1,2], [3,4]]
i = 0

while (foo, _ = a[i])
  puts foo
  i += 1
end
# => 1
# => 3
```

I make use of that construct to safely access the data without much changes. Makes some code easier actually.

### DIFF
--- a/lib/prism/translation/parser/lexer.rb
+++ b/lib/prism/translation/parser/lexer.rb
@@ -275,7 +275,7 @@ module Prism
             when :tCOMMENT
               if token.type == :EMBDOC_BEGIN
 
-                while !((next_token = lexed[index][0]) && next_token.type == :EMBDOC_END) && (index < length - 1)
+                while !((next_token, _ = lexed[index]) && next_token.type == :EMBDOC_END) && (index < length - 1)
                   value += next_token.value
                   index += 1
                 end
@@ -287,8 +287,8 @@ module Prism
                 is_at_eol = value.chomp!.nil?
                 location = range(token.location.start_offset, token.location.end_offset + (is_at_eol ? 0 : -1))
 
-                prev_token = lexed[index - 2][0] if index - 2 >= 0
-                next_token = lexed[index][0]
+                prev_token, _ = lexed[index - 2]
+                next_token, _ = lexed[index]
 
                 is_inline_comment = prev_token&.location&.start_line == token.location.start_line
                 if is_inline_comment && !is_at_eol && !COMMENT_CONTINUATION_TYPES.include?(next_token&.type)
@@ -307,7 +307,7 @@ module Prism
                 end
               end
             when :tNL
-              next_token = next_token = lexed[index][0]
+              next_token, _ = lexed[index]
               # Newlines after comments are emitted out of order.
               if next_token&.type == :COMMENT
                 comment_newline_location = location
@@ -344,8 +344,8 @@ module Prism
               location = range(token.location.start_offset, token.location.start_offset + percent_array_leading_whitespace(value))
               value = nil
             when :tSTRING_BEG
-              next_token = lexed[index][0]
-              next_next_token = lexed[index + 1][0]
+              next_token, _ = lexed[index]
+              next_next_token, _ = lexed[index + 1]
               basic_quotes = value == '"' || value == "'"
 
               if basic_quotes && next_token&.type == :STRING_END
@@ -413,7 +413,7 @@ module Prism
                 while token.type == :STRING_CONTENT
                   current_length += token.value.bytesize
                   # Heredoc interpolation can have multiple STRING_CONTENT nodes on the same line.
-                  is_first_token_on_line = lexed[index - 1] && token.location.start_line != lexed[index - 2][0].location&.start_line
+                  is_first_token_on_line = lexed[index - 1] && token.location.start_line != lexed[index - 2]&.first.location&.start_line
                   # The parser gem only removes indentation when the heredoc is not nested
                   not_nested = heredoc_stack.size == 1
                   if is_percent_array
@@ -427,7 +427,7 @@ module Prism
                     tokens << [:tSTRING_CONTENT, [current_string, range(start_offset, start_offset + current_length)]]
                     break
                   end
-                  token = lexed[index][0]
+                  token, _ = lexed[index]
                   index += 1
                 end
               else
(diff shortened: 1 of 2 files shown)


Earlopain:
Hi @kddnewton, sorry for the bother but can you check this out?

lib/prism/translation/parser/lexer.rb
```diff
@@ -482,7 +482,7 @@ def to_a
               end
 
               if percent_array?(quote_stack.pop)
-                prev_token = lexed[index - 2][0] if index - 2 >= 0
+                prev_token, _ = lexed[index - 2]
```
kddnewton: This is going to wrap around now, but I don't think it should

lib/prism/translation/parser/lexer.rb
```diff
@@ -287,8 +287,8 @@ def to_a
                 is_at_eol = value.chomp!.nil?
                 location = range(token.location.start_offset, token.location.end_offset + (is_at_eol ? 0 : -1))
 
-                prev_token = lexed[index - 2][0] if index - 2 >= 0
-                next_token = lexed[index][0]
+                prev_token, _ = lexed[index - 2]
```
kddnewton: Same here, we don't want it to wrap

kddnewton (changes requested):

lib/prism/translation/parser/lexer.rb
Earlopain: Oops, indeed. I should not have removed that

lib/prism/translation/parser/lexer.rb
```diff
@@ -413,7 +413,7 @@ def to_a
                 while token.type == :STRING_CONTENT
                   current_length += token.value.bytesize
                   # Heredoc interpolation can have multiple STRING_CONTENT nodes on the same line.
-                  is_first_token_on_line = lexed[index - 1] && token.location.start_line != lexed[index - 2][0].location&.start_line
+                  is_first_token_on_line = lexed[index - 1] && token.location.start_line != lexed[index - 2]&.first.location&.start_line
```
kddnewton: Is it possible for lexed[index - 2] to be nil here? If it is, then we need &.location as well. If it's not, then we shouldn't use &.first

kddnewton (changes requested):

lib/prism/translation/parser/lexer.rb
Earlopain: I meant this to be "safe" safe navigation. I don't believe this matters for valid syntax, but with invalid one I can't say so wanted to be on the safe side.

rafaelfranca (approved):

Earlopain:
Any chance to get this merged? This is also a problem for rubocop extensions that try to extract ruby code from templating like erb or slim (https://github.com/r7kamura/rubocop-slim/issues/31)

lib/prism/translation/parser/lexer.rb
```diff
@@ -275,7 +275,7 @@ def to_a
             when :tCOMMENT
               if token.type == :EMBDOC_BEGIN
 
-                while !((next_token = lexed[index][0]) && next_token.type == :EMBDOC_END) && (index < length - 1)
```
Morriar, line 280: I think you can also do the same line 284?

lib/prism/translation/parser/lexer.rb
```diff
@@ -413,7 +413,7 @@ def to_a
                 while token.type == :STRING_CONTENT
                   current_length += token.value.bytesize
                   # Heredoc interpolation can have multiple STRING_CONTENT nodes on the same line.
-                  is_first_token_on_line = lexed[index - 1] && token.location.start_line != lexed[index - 2][0].location&.start_line
+                  is_first_token_on_line = lexed[index - 1] && token.location.start_line != lexed[index - 2]&.first&.location&.start_line
```
Morriar: Can we avoid the mix of index accesses and `first` with something like this?

```suggestion
                  prev_prev_token, _ = lexed[index - 2] if index - 2 >= 0
                  is_first_token_on_line = lexed[index - 1] && token.location.start_line != prev_prev_token&.location&.start_line
```

lib/prism/translation/parser/lexer.rb
Earlopain: Yeah, that works (if you meant what I changed that is)

lib/prism/translation/parser/lexer.rb
Earlopain: Yeah, sure. It's a bit better. It's actually just `prev_token`, even if it is`[index -2]`. Right at the top of the loop `index` already gets incremented so you need to unintuitively look back a bit further.

I should move most of the state from the huge `to_a` into the class itself to make writing helper methods more convenient sometime.

Morriar (approved):

lib/prism/translation/parser/lexer.rb
```diff
@@ -595,9 +596,9 @@ def calculate_heredoc_whitespace(heredoc_token_index)
           previous_line = -1
           result = Float::MAX
 
-          while (lexed[next_token_index] && next_token = lexed[next_token_index][0])
+          while (next_token, _ = lexed[next_token_index])
```
paracycle: I don't think the behaviour of these two constructs (old and new) will be the same in every instance. For example:
```ruby
irb> foo = [nil, 1]
#=> [nil, 1]

irb> while (i, _ = foo); puts "42"; break; end
42
#=> nil

irb> while (i = foo&.first); puts "42"; break; end
#=> nil
```

The same would be true for any assignments in `if` conditions as well.

I would very much like to see `&.first` instead of destructuring used to safely get these values, since its behaviour would be the same as `[0]` (but safe against `nil` values) and I also find that it is more readable.

Morriar: Good catch, I was expecting the destructure to be falsy

Earlopain: The token being `nil` would be a bug in prism I believe (and probably blow up in a different place later). I can see now how this isn't quite the same but practially the difference should be zero.

Anyways, I changed it.

Earlopain:
Thanks everyone

kddnewton:
Thanks for your patience @Earlopain — I'll be much faster when I'm back from paternity leave.

Earlopain:
@kddnewton don't worry about it at all. Your labour on the previous PRs more than makes up for it

lib/prism/translation/parser/lexer.rb
paracycle: Thanks for changing it, but I don't get why you think it would be a bug. The loop is explicitly conditional on the `next_token` being truthy, and if it isn't it should break.

Earlopain: I was just saying that this difference shouldn't matter for the data that prism provides. It's always [token, some_internal_state], if there's no token (`nil`, then that should not have been part of `lexed` in the first place.

Issue #2915, opened 2024-06-28
Prism accepts double wildcard with no required fields in pattern match


herwinw:
Slightly related to  #2914

Prism parser the following code:
```ruby
[1,2,3,4] => *a, *b
```
This results in a `FindPatternNode` with an empty list `requireds`.

MRI's parser rejects this syntax (I've tried this with Ruby 3.3.3 and a 3.4.0-dev of today):
```
ruby -e '[1,2,3,4] => *a, *b; p a; p b'
-e:1: syntax error, unexpected *
[1,2,3,4] => *a, *b; p a; p b
ruby: compile error (SyntaxError)
```
The behaviour is ambiguous: if we enable the prism parser in MRI, it turns out that the whole array is put into the right splat:
```
ruby --parser=prism -e '[1,2,3,4] => *a, *b; p a; p b'
ruby: warning: The compiler based on the Prism parser is currently experimental and compatibility with the compiler based on parse.y is not yet complete. Please report any issues you find on the `ruby/prism` issue tracker.
[]
[1, 2, 3, 4]
```
But the result `a = [1, 2, 3, 4]; b = []` would have been just as valid. Other combinations like `a = [1]; b = [2, 3, 4]` would be valid too, but kind of random.

I would guess a `FindPatternNode` with no requireds should be a syntax error.

herwinw:
The same thing can happen with an `ArrayPatternNode`:
```ruby
[1,2,3,4] => *a, *b, 4
```
This too gets rejected by MRI
```
ruby -e '[1,2,3,4] => *a, *b, 4'
-e:1: syntax error, unexpected *
[1,2,3,4] => *a, *b, 4
ruby: compile error (SyntaxError)
```
Switching the parser to Prism (ruby 3.4-dev, the default included prism 0.30) results in a segfault.

kddnewton:
Thanks @herwinw! I'll get these fixed

PR #3564, opened 2025-05-19
[Bug #21345] Fix accepting multiple rest patterns with leading match

Earlopain:
Related:
* https://bugs.ruby-lang.org/issues/20765
* https://github.com/ruby/prism/issues/2915


Earlopain:
Openend https://bugs.ruby-lang.org/issues/21352 for valgrind

rafaelfranca:
Code is correct, but as this also closes a bug in the ruby issue tracker, and I don't know if that requires us to commit to Ruby, I'm not going to merge. @tenderlove @eileencodes what is the process here?

eileencodes:
It will sync to Ruby so it's fine to merge here.

PR #3566, opened 2025-05-22
Monomorphise visitor methods

paracycle:
The current implementation of the visitor pattern in Prism uses a single method (`visit_child_nodes`) to handle all node types. This can lead to performance issues since the `node` argument will end up being polymorphic, and will prevent effective use of inline caches, which in CRuby are monomorphic.

This commit generates an inlined version of the previous code for each node type, thus making the calls inside visitor methods monomorphic. This should improve performance, especially in cases where the visitor is called frequently.

Thanks to @tenderlove for finding that this was slower than it needed to be.

Benchmark results:
```
❯ ruby bench.rb
ruby 3.4.3 (2025-04-14 revision d0b7e5b6a0) +PRISM [arm64-darwin20]
             vanilla     1.000 i/100ms
              inline     1.000 i/100ms
             vanilla      0.840 (± 0.0%) i/s     (1.19 s/i) -      5.000 in   5.957006s
              inline      0.951 (± 0.0%) i/s     (1.05 s/i) -      5.000 in   5.256137s

Comparison:
              inline:        1.0 i/s
             vanilla:        0.8 i/s - 1.13x  slower
```

<details>
<summary>Benchmark code</summary>

```ruby
require "benchmark/ips"
require "prism"

class InlinedVisitor < Prism::Visitor
  eval Prism::Visitor
    .instance_methods(false)
    .map { |method| "def #{method}(node); node.compact_child_nodes.each { |node| node.accept(self) }; end" }
    .join("\n")
end

class VanillaVisitor < Prism::Visitor
end

ast = Prism.parse(<<~RUBY).value
  class Foo
    def bar
      baz&.to_s { |foo|  42 }
    end

    def baz
      3 + 4
    end

    def qux
      5 + 6
    end

    def quux
      7 + 8
    end

    def corge
      9 + 10
    end

    def grault
      11 + 12
    end

    def garply
      13 + 14
    end

    def waldo
      15 + 16
    end

    def fred
      17 + 18
    end

    def plugh
      19 + 20
    end

    def xyzzy
      21 + 22
    end

    def thud
      23 + 24
    end

    def foo_bar_baz_qux_quux_corge_grault_garply_waldo_fred_plugh_xyzzy_thud; end
  end
RUBY

vanilla_visitor = VanillaVisitor.new
inlined_visitor = InlinedVisitor.new

Benchmark.ips do |b|
  b.report("vanilla") do
    10_000.times do
    vanilla_visitor.visit(ast)
    vanilla_visitor.visit(ast)
    vanilla_visitor.visit(ast)
    vanilla_visitor.visit(ast)
    vanilla_visitor.visit(ast)
    vanilla_visitor.visit(ast)
    vanilla_visitor.visit(ast)
    vanilla_visitor.visit(ast)
    vanilla_visitor.visit(ast)
    vanilla_visitor.visit(ast)
    end
  end

  b.report("inline") do
    10_000.times do
    inlined_visitor.visit(ast)
    inlined_visitor.visit(ast)
    inlined_visitor.visit(ast)
    inlined_visitor.visit(ast)
    inlined_visitor.visit(ast)
    inlined_visitor.visit(ast)
    inlined_visitor.visit(ast)
    inlined_visitor.visit(ast)
    inlined_visitor.visit(ast)
    inlined_visitor.visit(ast)
    end
  end

  b.compare!
end
```
</details>

### DIFF
--- a/templates/lib/prism/compiler.rb.erb
+++ b/templates/lib/prism/compiler.rb.erb
@@ -35,7 +35,9 @@ module Prism
     <%- nodes.each_with_index do |node, index| -%>
 <%= "\n" if index != 0 -%>
     # Compile a <%= node.name %> node
-    alias visit_<%= node.human %> visit_child_nodes
+    def visit_<%= node.human %>(node)
+      node.compact_child_nodes.map { |node| node.accept(self) }
+    end
     <%- end -%>
   end
 end
--- a/templates/lib/prism/visitor.rb.erb
+++ b/templates/lib/prism/visitor.rb.erb
@@ -47,7 +47,9 @@ module Prism
     <%- nodes.each_with_index do |node, index| -%>
 <%= "\n" if index != 0 -%>
     # Visit a <%= node.name %> node
-    alias visit_<%= node.human %> visit_child_nodes
+    def visit_<%= node.human %>(node)
+      node.compact_child_nodes.each { |node| node.accept(self) }
+    end
     <%- end -%>
   end
 end


kddnewton (approved):
I'm cool with approving this as it is, but if we're going this direction, I think we should go all the way and inline compact_child_nodes as well. That way we avoid the intermediate array and the block and all that business. For some nodes, that will make this whole function empty.

tenderlove:
> I'm cool with approving this as it is, but if we're going this direction, I think we should go all the way and inline compact_child_nodes as well. That way we avoid the intermediate array and the block and all that business. For some nodes, that will make this whole function empty.

I think that would make sense. Though I don't think we could remove the `compact_child_nodes` method because that wouldn't be backwards compatible.

eregon:
I think it would be quite valuable to add something like `Node#each_child` (e.g. I already wanted this before this PR).
That should be basically as efficient as inlining the `compact_child_nodes` but also be generally useful for more things than just the visitor.
`Node#each_compact_child` seems too low value and probably no faster than `node.each_child { if node; ...; end }`.

In the Compiler, `.map` is used, but `Node#each_child` is efficient there too because we can't predict the exact size of the returned Array since some children might be nil. So `compiled = []; node.each_child { |node| compiled << node.accept(self) if node }; compiled` should be basically as fast but not need to manually inline the logic another time.

eregon:
It would notably address
> We can see walking the AST:
> * causes extra Array allocations for Prism because each call to child_nodes allocates an Array.

from https://eregon.me/blog/2024/10/27/benchmarking-ruby-parsers.html

paracycle:
> Oh for sure, don't remove that method. I just mean in this function. @paracycle would you rather do that here or follow-up?

I'll dona follow-up since this can be shipped as is and is already more performant.

PR #3571, opened 2025-05-28
Fix Java class ParsingOptions.Scope constructors visibility

andrykonchin:
Make ParsingOptions.Scope's constructors public so they can be used in other packages.


eregon (approved):

eregon:
I'll merge this since it's a trivial fix (let me know if that's not OK).

Issue #3582, opened 2025-06-10
Parser Translation: tokenization error


viralpraxis:
```ruby
# frozen-string-literal: true

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'prism', github: 'ruby/prism'
  gem 'parser'
end

require "prism"

data = DATA.read

p "eval: #{eval(data)}"
Prism::Translation::Parser34.new.tokenize(Parser::Source::Buffer.new('(str)', source: data))

__END__

%Q(\«)
```

Output:

```shell
"eval: «"
/home/viralpraxis/.asdf/installs/ruby/3.4.4/lib/ruby/gems/3.4.0/bundler/gems/prism-3bd6837ed7b2/lib/prism/translation/parser/lexer.rb:770:in 'String#ord': invalid byte sequence in UTF-8 (ArgumentError)
	from /home/viralpraxis/.asdf/installs/ruby/3.4.4/lib/ruby/gems/3.4.0/bundler/gems/prism-3bd6837ed7b2/lib/prism/translation/parser/lexer.rb:770:in 'Prism::Translation::Parser::Lexer#escape_read'
	from /home/viralpraxis/.asdf/installs/ruby/3.4.4/lib/ruby/gems/3.4.0/bundler/gems/prism-3bd6837ed7b2/lib/prism/translation/parser/lexer.rb:698:in 'Prism::Translation::Parser::Lexer#unescape_string'
	from /home/viralpraxis/.asdf/installs/ruby/3.4.4/lib/ruby/gems/3.4.0/bundler/gems/prism-3bd6837ed7b2/lib/prism/translation/parser/lexer.rb:426:in 'Prism::Translation::Parser::Lexer#to_a'
	from /home/viralpraxis/.asdf/installs/ruby/3.4.4/lib/ruby/gems/3.4.0/bundler/gems/prism-3bd6837ed7b2/lib/prism/translation/parser.rb:319:in 'Prism::Translation::Parser#build_tokens'
	from /home/viralpraxis/.asdf/installs/ruby/3.4.4/lib/ruby/gems/3.4.0/bundler/gems/prism-3bd6837ed7b2/lib/prism/translation/parser.rb:140:in 'Prism::Translation::Parser#tokenize'
	from /tmp/prism-bug.rb:17:in '<main>'

```

PR #3583, opened 2025-06-11
Fix parser translator during string escaping with invalid utf-8

Earlopain:
`byte.ord` can raise when it is not valid in its encoding. Instead, prefer `scan_byte` over `get_byte` since that already returns the byte as an integer, sidestepping conversion issues.

Fixes https://github.com/ruby/prism/issues/3582


Earlopain:
Right, no `scan_byte`, it's much to recent. Just make the character binary then, seems good enough.

kddnewton (commented):
Let's use scan_byte and add a polyfill for when it's necessary

Earlopain:
Right, makes sense. I changed it.

kddnewton (approved):

PR #3585, opened 2025-06-13
Reject `true && not true`

mame:
A command-call-like `not true` must be rejected after `&&` and `||`.

https://bugs.ruby-lang.org/issues/21337


test/prism/errors/command_calls_31.txt
```diff
@@ -0,0 +1,12 @@
...
+            ^~~~ unexpected 'true', expecting end-of-input
+
+true || not true
+            ^~~~ expected a `(` after `not`
+            ^~~~ unexpected 'true', expecting end-of-input
+
+true && not (true)
+            ^ expected a `(` after `not`
+            ^ unexpected '(', expecting end-of-input
+
```
mame, line 12: I wanted to add a test to confirm that

```
true && not
true
```

be rejected, but I am not sure how to write the test.

test/prism/errors/command_calls_31.txt
```diff
@@ -0,0 +1,12 @@
...
+            ^~~~ expected a `(` after `not`
+            ^~~~ unexpected 'true', expecting end-of-input
+
+true || not true
+            ^~~~ expected a `(` after `not`
+            ^~~~ unexpected 'true', expecting end-of-input
+
+true && not (true)
+            ^ expected a `(` after `not`
+            ^ unexpected '(', expecting end-of-input
```
mame, line 11: This error message is a bit weird, but I am not sure how to fix it.

kddnewton:
This looks correct!! I can fix the error messages next Monday

mame:
@kddnewton Just a gentle follow-up on the PR I sent. I believe this fix is somewhat time-sensitive. Since the bug can lead to invalid Ruby code being written, it would be great to get it resolved before it spreads further. Ideally, it could also be backported to Ruby 3.4.5.

Please let me know if there's anything I can do to help move it forward. Thanks so much!

mame:
@kddnewton Second follow-up on the PR. Would appreciate it if you could take a look when you have time. cc/ @tenderlove @eileencodes

ydah:
How about following changes? Perhaps the remaining concerns (https://github.com/ruby/prism/pull/3585#discussion_r2144153832 and https://github.com/ruby/prism/pull/3585#discussion_r2144153500) have been addressed.
- https://github.com/ydah/prism/commit/f87bb971f26886d69bc61c72d7596e8eb3bebd8f

test/prism/errors/command_calls_31.txt
ydah: Perhaps we can write something like this in test/prism/errors/command_calls_31.txt:

```
true && not
true
^~~~ expected a `(` after `not`
^~~~ unexpected 'true', expecting end-of-input
```

ydah:
I've added the above commits. If they don't match your intentions, please feel free to revert them without any concern.

ydah:
@kddnewton Please feel free to let me know if there is anything blocking the merge of this PR so I can help you.

mame:
@kddnewton @tenderlove @eileencodes Hey, we really need to get this merged soon. I’d like to have it backported to the upcoming Ruby 3.4 series.

Let me repeat: this is a very serious bug. It allows invalid code to be parsed, and once people start writing such code, fixing it becomes much harder later on.

Also, I’m increasingly concerned that bug reports for Prism are being left unaddressed. If maintenance doesn't improve, I think we'll seriously need to switching the default parser back to parse.y.

cc/ @paracycle

mame:
@kddnewton @tenderlove @eileencodes @paracycle Ruby 3.4.5 has already been released. That's very unfortunate.

This reinforces my growing concern that Prism is effectively unmaintained. If this isn't resolved in time for Ruby 3.4.6, or if any other serious signs of it being unmaintained appear, I'll propose removing Prism from the Ruby interpreter.

eregon:
> Also, I’m increasingly concerned that bug reports for Prism are being left unaddressed.

I think that's an exaggeration based on this specific case, e.g., 22 open 1007 closed at https://github.com/ruby/prism/issues is what I would call very well maintained.
I guess what happened is the 3 Prism maintainers you @-mentioned were not available recently, maybe they were on leave/at RailsConf/etc?
Of course they know best so it would be great if they can reply here.
I'll note only @kddnewton was aware of this PR since a while, others only since https://github.com/ruby/prism/pull/3585#issuecomment-3044074649, 9 days ago.

I think it would be good to clarify in which situations it's OK for CRuby committers to merge changes to Prism themselves.
After all Prism is now the default parser in CRuby so I think CRuby committers need to be able to modify it when needed (e.g. based on CRuby tickets, for adding new syntax, etc)

Earlopain:
Well, I myself have two PRs that attempt to fix some sort of bug with prism:
* https://github.com/ruby/prism/pull/3564
* https://github.com/ruby/ruby/pull/13108

They have been open without any feedback for a few months now unfortunatly. I'm not a C dev and also not experienced with parsers in general but wanted to help out a bit. Previously I focused on the translation layer and the time to merge there has been phenomenal.

@kddnewton github says he is on paternity leave, so I do not expect immediate action from him in any way. The PRs from me also do not address anything serious, so personally those are not a big deal to me. It was just nice to do something a bit different for a change.

What I do want to say in regards to the maintenance though is in regards to this:

> If this isn't resolved in time for Ruby 3.4.6, or if any other serious signs of it being unmaintained appear, I'll propose removing Prism from the Ruby interpreter.

I think it would be incredibly unfortunate to lose prism. It has seen widespread adoption in the ecosystem and other parsers that used to be around have stopped development, like the parser gem. If there is no adequate replacement, this leaves many projects in an awkward spot. I can't tell about alternative ruby implementations like truffleruby/jruby/natalie, but I presume they don't carry around their old parser implementation anymore now that they use prism too. Projects like rails/ruby-lsp/solargraph/rubocop and others also adapted prism so I would say prism has been a huge success thus far.

I would like to see anything to keep prism going, and more direct involvement from other cruby devs would certainly help with that.

kddnewton:
Sorry friends, still getting the hang of 3 kids while on leave here. I will get this reviewed now.

kddnewton:
Please don't yank prism because I went on paternity leave. It is still maintained, I promise.

mame:
First of all, thank you for merging it!

I'm aware that @kddnewton is currently on paternity leave and is no longer maintaining Prism full-time.
I had understood that @tenderlove and @eileencodes were responsible for maintaining Prism at Shopify (https://bugs.ruby-lang.org/groups/53036), so I was expecting that they might be the ones to take care of this.

It is a fact that several Prism-related bugs, including a serious one like heap-buffer-overflow, have been left unaddressed for some time, and I am growing increasingly concerned about this trend.

https://bugs.ruby-lang.org/issues/21461
https://bugs.ruby-lang.org/issues/21326
https://bugs.ruby-lang.org/issues/21187
https://bugs.ruby-lang.org/issues/21256
https://bugs.ruby-lang.org/issues/21259
https://bugs.ruby-lang.org/issues/21188
https://bugs.ruby-lang.org/issues/21168
https://bugs.ruby-lang.org/issues/21139
https://bugs.ruby-lang.org/issues/20776

Unfortunately, there are several components in Ruby that are not actively maintained, but the parser, arguably one of the most critical parts of a language implementation, should not be one of them.

One of Prism's original goal was to provide a more maintainable alternative to parse.y.
However, maintainability is not only the cleanliness of the code, but also the fact that there are maintainers who is passionate and competent.
It seems that there are simply too few active maintainers who are both deeply familiar with Prism's internals and willing to devote the time and energy to maintaining it.

The code itself of parse.y is never clean, but it has been actively maintained for decades (largely thanks to @nobu).
More recently, we have also seen new committers like @ydah stepping in to work on parse.y.
This makes me feel that the maintenance situation is being reversed.

PR #3593, opened 2025-07-01
Fix crash when using arithmetic expressions in pattern matching

st0012:
Fixes https://bugs.ruby-lang.org/issues/21460

When arithmetic expressions like `-1**2` are used in pattern matching contexts, Ruby crashes with "Unexpected node type in pattern matching expression: PM_CALL_NODE". This happens because the Prism parser creates PM_CALL_NODE for arithmetic operations, but Ruby's pattern matching compiler doesn't handle call nodes.

This fix adds validation to reject PM_CALL_NODE in pattern contexts with a proper syntax error.


Earlopain:
Looks like this also fixes https://github.com/ruby/prism/issues/3381 (basically the same issue)

eileencodes (approved):

st0012:
@eileencodes Thanks!

PR #3599, opened 2025-07-17
[docs] RuboCop now uses prism as a first-class (translation) parser

Earlopain:
Since rubocop 1.75 it just does the right thing, so no need to mention much of the old configuration anymore. See https://github.com/rubocop/rubocop/pull/14035

I'm also open to removing this section entirely, I don't think it provides much value anymore.


vinistock:
I agree that this section may not make much sense now that no configuration is needed and RuboCop just uses Prism automatically. My vote goes for removing it entirely.

Earlopain:
:+1: I removed the section completly

vinistock (approved):

rafaelfranca (approved):

kddnewton:
Thanks!

Issue #3473, opened 2025-02-23
Should parse `x = def f = p 1`


qnighy:
Found during investigation of [ruby#21153](https://bugs.ruby-lang.org/issues/21153), Prism cannot parse this:

```
% ./miniruby --parser=parse.y -e "x = def f = p 1"
% ./miniruby --parser=prism -e "x = def f = p 1"
-e: -e:1: syntax error found (SyntaxError)
> 1 | x = def f = p 1
    |               ^ unexpected integer, expecting end-of-input
```

PR #3601, opened 2025-07-17
Allow command calls in endless method bodies within assignments

st0012:
(This is another attempt to solve https://github.com/ruby/prism/issues/3473 as #3598 caused `ruby/ruby` tests to fail and got reverted)

Previously, endless method definitions in assignment contexts like `x = def f = p 1` would fail to parse because command calls (method calls without parentheses) were only accepted when the surrounding binding power was less than `PM_BINDING_POWER_COMPOSITION`.

This fix specifically checks for assignment context and allows command calls in those cases while maintaining the existing behavior for other contexts. This ensures that:

- `x = def f = p 1` parses correctly (previously failed)
- `private def f = puts "Hello"` still produces the expected error

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -19505,7 +19505,15 @@ parse_expression_prefix(pm_parser_t *parser, pm_binding_power_t binding_power, b
                 pm_do_loop_stack_push(parser, false);
                 statements = (pm_node_t *) pm_statements_node_create(parser);
 
-                pm_node_t *statement = parse_expression(parser, PM_BINDING_POWER_DEFINED + 1, binding_power < PM_BINDING_POWER_COMPOSITION, false, PM_ERR_DEF_ENDLESS, (uint16_t) (depth + 1));
+                // In endless method bodies, we need to handle command calls carefully.
+                // We want to allow command calls in assignment context but maintain
+                // the same binding power to avoid changing how operators are parsed.
+                // Note that we're intentionally NOT allowing code like `private def foo = puts "Hello"`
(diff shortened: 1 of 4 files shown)


src/prism.c
```diff
@@ -19501,7 +19501,13 @@ parse_expression_prefix(pm_parser_t *parser, pm_binding_power_t binding_power, b
                 pm_do_loop_stack_push(parser, false);
                 statements = (pm_node_t *) pm_statements_node_create(parser);
 
-                pm_node_t *statement = parse_expression(parser, PM_BINDING_POWER_DEFINED + 1, binding_power < PM_BINDING_POWER_COMPOSITION, false, PM_ERR_DEF_ENDLESS, (uint16_t) (depth + 1));
+                // In endless method bodies, we need to handle command calls carefully.
```
Earlopain, line 19508: I do want to mention that the failed test mentions it being a technical limitation of parse.y: https://github.com/ruby/ruby/blob/86320a53002a3adaf35ad7434c70e86747a8b345/test/ruby/test_syntax.rb#L1797 (also https://bugs.ruby-lang.org/issues/17398#note-5). Perhaps add a comment that this is to accomodate the overly specific ruby/ruby test? It looks like this was supposed to be valid syntax but it was too difficult to implement.

tenderlove: +1 @st0012 do you mind adding a comment?  🙇

st0012: Updated 👍

PR #3602, opened 2025-07-17
Add an `feof` callback to Prism stream parsing

tenderlove:
Checking the return value of `fgets` isn't good enough to determine whether the input stream has reached EOF or not.  `fgets` could have run in to an error and returned NULL.  Additionally, we can't just check the buffer contents for `\n` because `fgets` could have read the maximum buffer size (which may not have ended with a `\n`).

I want to use this callback to implement a fix for:

  https://bugs.ruby-lang.org/issues/21188

### DIFF
--- a/ext/prism/extension.c
+++ b/ext/prism/extension.c
@@ -994,6 +994,14 @@ profile_file(int argc, VALUE *argv, VALUE self) {
     return Qnil;
 }
 
+static int
+parse_stream_eof(void *stream) {
+    if (rb_funcall((VALUE) stream, rb_intern("eof?"), 0)) {
+        return 1;
+    }
+    return 0;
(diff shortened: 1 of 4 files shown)


kddnewton (approved):
Looks right, though I think the FFI build is having issues.

tenderlove:
> Looks right, though I think the FFI build is having issues.

Ya, I'm trying to figure out the FFI issues :|

kddnewton:
Looks like we just need documentation on the new callback (https://github.com/ruby/prism/actions/runs/16603543403/job/46969381276?pr=3602#step:8:153). memcheck is an expected failure, and the test-all failure is telling us we need to merge this into ruby/ruby first.

PR #3604, opened 2025-07-21
Make `it = it` assign `nil` to match parse.y behavior [Bug #21139]

S-H-GAMELINKS:
Currently Prism returns `42` for code like this:
```ruby
42.tap { it = it; p it } # => 42
```

But parse.y returns `nil`:
```ruby
42.tap { it = it; p it } # => nil
```

In parse.y, it on the right-hand side is parsed as a local variable. In Prism, it was parsed as the implicit block parameter it, which caused this inconsistent behavior.

This change makes the right-hand side it to be parsed as a local variable, aligning with parse.y's behavior.

Bug ticket: https://bugs.ruby-lang.org/issues/21139

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -16456,13 +16456,6 @@ parse_variable(pm_parser_t *parser) {
 
             return node;
         } else if ((parser->version != PM_OPTIONS_VERSION_CRUBY_3_3) && pm_token_is_it(parser->previous.start, parser->previous.end)) {
-            if (match1(parser, PM_TOKEN_EQUAL)) {
-                pm_constant_id_t name_id = pm_parser_local_add_location(parser, parser->previous.start, parser->previous.end, 0);
-                pm_node_t *node = (pm_node_t *) pm_local_variable_read_node_create_constant_id(parser, &parser->previous, name_id, 0, false);
-
-                return node;
-            }
-
             pm_node_t *node = (pm_node_t *) pm_it_local_variable_read_node_create(parser, &parser->previous);
             pm_node_list_append(&current_scope->implicit_parameters, node);
 
@@ -21178,6 +21171,13 @@ parse_expression_infix(pm_parser_t *parser, pm_node_t *node, pm_binding_power_t 
                 }
                 PRISM_FALLTHROUGH
                 case PM_CASE_WRITABLE: {
+                    // When we have `it = value`, we need to add `it` as a local
+                    // variable before parsing the value, in case the value
+                    // references the variable.
+                    if (PM_NODE_TYPE_P(node, PM_IT_LOCAL_VARIABLE_READ_NODE)) {
+                        pm_parser_local_add_location(parser, node->location.start, node->location.end, 0);
+                    }
+
                     parser_lex(parser);
                     pm_node_t *value = parse_assignment_values(parser, previous_binding_power, PM_NODE_TYPE_P(node, PM_MULTI_TARGET_NODE) ? PM_BINDING_POWER_MULTI_ASSIGNMENT + 1 : binding_power, accepts_command_call, PM_ERR_EXPECT_EXPRESSION_AFTER_EQUAL, (uint16_t) (depth + 1));
 


test/prism/ruby/parser_test.rb
```diff
@@ -179,6 +179,25 @@ def test_it_block_parameter_syntax
       assert_equal(it_block_parameter_sexp, actual_ast.to_sexp)
     end
 
+    def test_it_assignment_syntax
```
Earlopain: This is for the parser gem translation and is not necessary to add as a test. The fixture you added should generate a snapshot instead that you should commit to test the prism behaviour.

test/prism/fixtures/it_assignment.txt
```diff
@@ -0,0 +1 @@
+42.tap { it = it; p it }
```
Earlopain, line 1: Maybe also test `42.tap { it; it = it; p it }`? This should still generate a it read for the first one.

src/prism.c
```diff
@@ -16456,6 +16456,13 @@ parse_variable(pm_parser_t *parser) {
 
             return node;
         } else if ((parser->version != PM_OPTIONS_VERSION_CRUBY_3_3) && pm_token_is_it(parser->previous.start, parser->previous.end)) {
+            if (match1(parser, PM_TOKEN_EQUAL)) {
+                pm_constant_id_t name_id = pm_parser_local_add_location(parser, parser->previous.start, parser->previous.end, 0);
+                pm_node_t *node = (pm_node_t *) pm_local_variable_read_node_create_constant_id(parser, &parser->previous, name_id, 0, false);
+
+                return node;
```
vinistock: Do we not need to append to the list of implicit parameters still? For example, in this scenario:

```ruby
42.tap do
  p it
  it = it
  p it
end
```

The block still needs to have an implicit `it` parameter as the first print is reading from it. And this matches parse.y.

vinistock (commented):
While I understand the discrepancy, I wonder if parse.y's behaviour could be counter intuitive for developers.  In this example:

```ruby
42.tap { it = it; p it }
```

The proposed parse.y behaviour would treat `it = it` as if it were `it = nil` (reading from a non-intiailized local variable, to a that same local variable). But we think it's more intuitive if it prints 42, because `it = it` is creating a new local variable from the result of reading the `it` parameter (before it gets "shadowed" by a local variable of the same name), which is 42. If developers want the the `nil` behaviour, they can explicitly write `it = nil` instead.

Maybe someone with more experience on the parser can pitch in, but I'd argue that Prism's current behaviour is more intuitive than having `it` end up as `nil`.

tenderlove:
> Maybe someone with more experience on the parser can pitch in, but I'd argue that Prism's current behaviour is more intuitive than having it end up as nil.

@vinistock do you mind commenting on the original Redmine ticket?  It seems like matz wants the behavior specified in this PR.

src/prism.c
S-H-GAMELINKS: Thank you for your feedback.

I applied this change to Ruby and tested the behavior you mentioned, and the result is:

```ruby
42.tap do
  p it # => 42
  it = it
  p it # => nil
end
```

The implicit parameter works correctly for the first it read, and the assignment properly creates a local variable it. So, no additional changes are needed for the implicit parameters list.

test/prism/fixtures/it_assignment.txt
S-H-GAMELINKS: Thank you for the comment.

You're right, that test case would be good to have as well.
I've added the test case in the following commit: https://github.com/ruby/prism/pull/3604/commits/659d769621a461a889772aa37a5d9dbf0f8743fe

test/prism/ruby/parser_test.rb
S-H-GAMELINKS: Thank you for pointing this out.
It appears I misunderstood and added the test incorrectly. I've removed this test case.

Earlopain: If I run `bundle exec rake` on your branch, I am getting two files that need to be commited:

```sh
Started
|Created snapshot at /home/user/code/ruby-prism/snapshots/it_assignment.txt.
-Created snapshot at /home/user/code/ruby-prism/snapshots/it_read_and_assignment.txt
```

Those basically contain the test I commented on here, just with the native prism ast format. This is automatically tested against and when the output changes, a test would fail.

S-H-GAMELINKS: Thank you for the comment.
I forgot to add the snapshot, and I've pushed it now.

kddnewton (changes requested):
I would prefer that we handle this in the place where we handle the `=` operator. That's where we usually handle these kinds of transformations. (The very top of the `parse_expression_infix` function.) You'll see we have to handle call nodes in a similar way.

S-H-GAMELINKS:
Thank you for your comment.
I've moved the logic to convert `it` to a local variable to the beginning of parse_expression_infix as you suggested.
All tests including memcheck pass locally, but CI shows memcheck failures. I'm investigating what might be different in the CI environment.

Earlopain:
memcheck runs with ruby-dev from 2025-07-25 and also recently started failing on main. I wouldn't worry about it too much unless you want to report it upstream, very unlikely to be related to your PR.

tenderlove (approved):
This looks good to me.

PR #3605, opened 2025-07-21
Do not use `0` to indicate the latest ruby version to parse

Earlopain:
This makes it hard to do version checks against this value. The current version checks work because there are so few possible values, and they are exclusively `!=/`==` checks.

As an example, https://github.com/ruby/prism/pull/3337 introduces new syntax for ruby 3.5 and uses `PM_OPTIONS_VERSION_LATEST` as its version guard. Because what is considered the latest changes every year, it must later be changed to `parser->version == parser->version == PM_OPTIONS_VERSION_CRUBY_3_5 || parser->version == PM_OPTIONS_VERSION_LATEST`, with one extra version each year.

With this change, the PR can instead write `parser->version >= PM_OPTIONS_VERSION_CRUBY_3_5` which is self-explanatory and works for future versions.

### DIFF
--- a/java/org/prism/ParsingOptions.java
+++ b/java/org/prism/ParsingOptions.java
@@ -12,9 +12,10 @@ public abstract class ParsingOptions {
      * See pm_options_version_t in include/prism/options.h.
      */
     public enum SyntaxVersion {
-        LATEST(0),
+        LATEST(0), // Handled in pm_parser_init
         V3_3(1),
-        V3_4(2);
+        V3_4(2),
+        V3_5(3);
 
         private final int value;
 
--- a/include/prism/options.h
+++ b/include/prism/options.h
@@ -82,14 +82,20 @@ typedef void (*pm_options_shebang_callback_t)(struct pm_options *options, const 
  * parse in the same way as a specific version of CRuby would have.
  */
 typedef enum {
-    /** The current version of prism. */
-    PM_OPTIONS_VERSION_LATEST = 0,
+    /** If an explicit version is not provided, the current version of prism will be used. */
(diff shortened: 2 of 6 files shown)


java/org/prism/ParsingOptions.java
```diff
@@ -12,9 +12,10 @@ public abstract class ParsingOptions {
      * See pm_options_version_t in include/prism/options.h.
      */
     public enum SyntaxVersion {
-        LATEST(0),
         V3_3(1),
-        V3_4(2);
+        V3_4(2),
+        V3_5(3),
+        LATEST(V3_5.value);
```
eregon: Could be:
```suggestion
        LATEST(0);
```
So that there is no risk to forget updating it.
And then 0 would be translated when "parsing" the options (already done in `pm_parser_init` but not sure if that gets called when using the serialization).
Same for `javascript/src/parsePrism.js` below

eregon (approved):
Looks great to me, I have noticed this as well as being rather brittle but didn't get around fixing it, thanks!

java/org/prism/ParsingOptions.java
Earlopain: I agree that it is nicer that way. It works because `pm_options_read` (which does the deserialization is always followed by `pm_parser_init`. I only tested ffi though but they should all go through the same thing in the end.

eregon:
This looks ready to merge for me, except maybe the PR title could be clearer like:
> Use comparison operators for parse versions checks instead of ==/!=
or
> Make parse version values monotonic to clarify version checks

@tenderlove or @Morriar Could you do a second review and merge?

PR #3606, opened 2025-07-21
Clear flags on interpolated strings

tenderlove:
When inner strings aren't frozen, we need to clear the flags on interpolated string nodes so that we don't emit wrong instructions.

The compiler is currently incorrectly emitting frozen strings because the parser is erroneously declaring interpolated strings as "frozen". We need to fix this behavior in the parser so we can fix the compiler in CRuby.  This patch is a partial fix for [this bug](https://bugs.ruby-lang.org/issues/21187)

### DIFF
--- a/snapshots/dos_endings.txt
+++ b/snapshots/dos_endings.txt
@@ -17,7 +17,7 @@
         │   │   ├── flags: ∅
         │   │   └── arguments: (length: 1)
         │   │       └── @ InterpolatedStringNode (location: (1,5)-(2,12))
-        │   │           ├── flags: static_literal
+        │   │           ├── flags: ∅
         │   │           ├── opening_loc: ∅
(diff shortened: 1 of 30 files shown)


test/prism/result/string_test.rb
```diff
@@ -28,5 +28,10 @@ def test_xstring_node_unescaped_frozen
       node = Prism.parse_statement("`foo`")
       assert_predicate node.unescaped, :frozen?
     end
+
+    def test_string_literal_concatenation
+      node = Prism.parse_statement("`foo`")
```
st0012: Should we have concatenation here? Otherwise it looks exact the same as the previous test?

tenderlove: I am not sure why I added this test. I think it's just a copy / paste error 😆 

I'll rm this, thank you!

PR #3608, opened 2025-07-22
Fix jruby CI

Earlopain:
The current version constraint is too specific and doesn't allow 10.0.1.0 which was recently released

### DIFF
--- a/gemfiles/jruby/Gemfile
+++ b/gemfiles/jruby/Gemfile
@@ -2,7 +2,7 @@
 
 source "https://rubygems.org"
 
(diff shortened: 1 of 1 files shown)


Earlopain:
Well, the tests run now but they look proper broken. Much stuff about missing constants. No idea what to do about that. Let me see if going back to 10.0.0.0 solves it

Earlopain:
I pinned to 10.0.0.0 for now and opened an upstream issue at https://github.com/jruby/jruby/issues/8923

Earlopain:
I tracked it down to autoloading with `private_constant` which prism does in a few places, reproduction in jruby issue.

PR #3609, opened 2025-07-22
Add more Prism related tests to cruby-bindings workflow

st0012:
This should prevent Prism from breaking CRuby's CI.

### DIFF
--- a/.github/workflows/cruby-bindings.yml
+++ b/.github/workflows/cruby-bindings.yml
@@ -43,5 +43,5 @@ jobs:
           make -j2
         working-directory: ruby/ruby
       - name: make test-all
-        run: make -j2 -s test-all TESTS="prism --no-retry"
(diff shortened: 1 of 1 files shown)


tenderlove:
I'm not 100% sure if we want to do this.  In some cases we have a circular dependency where we need to update Prism and CRuby at the same time.  Do we merge to CRuby first, _then_ send a PR to Prism?  I'm not sure about the order of operations in that case.

st0012:
> Do we merge to CRuby first, then send a PR to Prism? I'm not sure about the order of operations in that case.

From my (limited) experience working on default gems, yes. There will be a PR including both library change AND CRuby test changes opened on `ruby/ruby` and gets merged there.
And the relevant changes will be ported back to the library with some local tests.

I think the point of running prism compilation test on CI is to identify the need to do so before we merge a PR.

tenderlove:
@st0012 makes sense to me!

tenderlove (approved):

kddnewton:
This is fine, thanks @st0012

PR #3610, opened 2025-07-22
Improve global variable alias error messaging

allcre:
### Summary
This PR improves the error message when attempting to alias a global variable with a non-global variable. The previous generic error message was misleading and has been replaced with a more specific and descriptive one.

### The Problem
The old error message was confusing because it didn't explain the actual constraint. Consider this example:

```ruby
alias $a b
         ^ invalid argument being passed to `alias`; expected a bare word, symbol, constant, or global variable
```

This error message is misleading because:
1. `b` **is** a bare word, which the error says is expected
2. The message doesn't explain the actual rule: when aliasing global variables, both names must be global variables

### The Solution
The new error message clearly states the actual requirement:

```ruby
alias $a b
         ^ invalid argument being passed to `alias`; when aliasing global variables, both the old and new names must be global variables
```

### Changes
- Added new error constant `PM_ERR_ALIAS_ARGUMENT_GLOBAL_VARIABLE` to provide clearer error messaging
- Updated the parser to use the new error when aliasing global variables with invalid arguments
- Updated test case `aliasing_global_variable_with_non_global_variable.txt` to verify the new error message


templates/src/diagnostic.c.erb
```diff
@@ -86,6 +86,7 @@ static const pm_diagnostic_data_t diagnostic_messages[PM_DIAGNOSTIC_ID_MAX] = {
 
     // Errors that should raise syntax errors
     [PM_ERR_ALIAS_ARGUMENT]                     = { "invalid argument being passed to `alias`; expected a bare word, symbol, constant, or global variable", PM_ERROR_LEVEL_SYNTAX },
+    [PM_ERR_ALIAS_ARGUMENT_GLOBAL_VARIABLE]     = { "invalid argument being passed to `alias`; when aliasing global variables, both the old and new names must be global variables", PM_ERROR_LEVEL_SYNTAX },
```
st0012, line 89: I think the message can be more specific. Since this error only happens when the first (new) name is already a global variable, we can probably say 

```
invalid argument being passed to `alias`; expected the old name to be a global variable as well
```

eregon: `old name` might be unclear if it's the first or second argument (`alias` is notoriously confusing about this), so I'd suggest to use `the second name` instead of `the old name`, or the message in the current state of this PR (which sounds clear to me).

st0012: That's fair. My point is mostly that the message can be more concise. `the second name` indeed works better.

kddnewton:
I agree that I don't prefer the wording, but these are validated against CRuby's test suite. So if you'd like to see a change in the error message, I suggest opening an issue on the ruby issue tracker. You can consider this more of a client fulfilling an API.

PR #3611, opened 2025-07-23
CI: Migrate Travis CI ppc64le/s390x cases to the GitHub Actions.

junaruga:
/cc @hsbt

Add a job for the ppc64le/s390x cases using GitHub Actions ppc64le/s390x service.
https://github.com/IBM/actionspz

The job runs conditionally only in the original (non-fork) repository. Because we cannot run the job in the forked repositories. The ppc64le/s390x runners only work on the registered repositories.

I had to add the new job `build-ibm` because the job has to run only in non-fork repository, this repository `ruby/prism`.

While Travis ran the Ubuntu 22.04 jammy ppc64le and s390x, and Ruby CI's Ubuntu ppc64le and s390x servers are also Ubuntu 22.04, this CI will add the following Ubuntu 24.04. Because there is a merit to have both Ubuntu 24.04 and 22.04 cases to test in various environment.

* ubuntu-24.04-ppc64le
* ubuntu-24.04-s390x

Below are the cases in other `ruby/*` repositoreis.

https://github.com/ruby/zlib/blob/4cac3f7f3134004c4395090ead6eba3fea05f822/.github/workflows/test.yml#L52-L53

https://github.com/ruby/psych/blob/a576b777ab3b02a1eb6ee1f39a0df5c8c341105b/.github/workflows/test.yml#L71-L72

But if you want Ubuntu 22.04 ppc64le/s390x cases in the CI, please let me know. There are following images available too.

* ubuntu-22.04-ppc64le
* ubuntu-22.04-s390x


Earlopain:
> I had to add the new job build-ibm because the job has to run only in non-fork repository, this repository ruby/prism.

FYI, a matrix like this would work I believe:
```yml
      matrix:
        os:
          - ubuntu-latest
          - ubuntu-24.04-arm
          - macos-latest
          # - windows-latest <-- failing with fiddle error, temporarily disabled
          - ubuntu-24.04-ppc64le
          - ubuntu-24.04-s390x
        upstream:
          - ${{ github.repository == 'ruby/prism' }}
        # These jobs can't run in forked repositories
        exclude:
          - os: ubuntu-24.04-ppc64le
            upstream: false
          - os: ubuntu-24.04-s390x
            upstream: false
```

Then you could reuse the existing workflow. But it looks like there is some extra setup that is needed for these images so the way you have it now looks fine to me.

junaruga:
> > I had to add the new job build-ibm because the job has to run only in non-fork repository, this repository ruby/prism.
> 
> FYI, a matrix like this would work I believe:
> 
> ```yaml
>       matrix:
>         os:
>           - ubuntu-latest
>           - ubuntu-24.04-arm
>           - macos-latest
>           # - windows-latest <-- failing with fiddle error, temporarily disabled
>           - ubuntu-24.04-ppc64le
>           - ubuntu-24.04-s390x
>         upstream:
>           - ${{ github.repository == 'ruby/prism' }}
>         # These jobs can't run in forked repositories
>         exclude:
>           - os: ubuntu-24.04-ppc64le
>             upstream: false
>           - os: ubuntu-24.04-s390x
>             upstream: false
> ```
> 
> Then you could reuse the existing workflow. But it looks like there is some extra setup that is needed for these images so the way you have it now looks fine to me.

Oh thanks for the tip!

Yes, I noticed that the `ruby/setup-ruby` didn't work on ppc64le/s390x. So, I picked up the code from the ruby/zlib and ruby/psych.

kddnewton:
Thank you!

PR #3616, opened 2025-07-30
RubyParser translation for stabby lambdas with `it`

presidentbeef:
Similar to #3409

However, I'm a little confused and I may be missing something.

The tests pass (as you can see) and the expected output in Prism's test is:
```ruby
s(:iter, s(:lambda), s(:args), (:call, nil, :it))
```

However, when I run RubyParser myself, I get

```ruby
3.4.2 :002 > RubyParser.new.parse('-> { it }')
 => s(:iter, s(:lambda), 0, s(:call, nil, :it)) 
```

I honestly can't explain the discrepancy. :confused:

### DIFF
--- a/lib/prism/translation/ruby_parser.rb
+++ b/lib/prism/translation/ruby_parser.rb
@@ -1151,7 +1151,7 @@ module Prism
         def visit_lambda_node(node)
           parameters =
             case node.parameters
-            when nil, NumberedParametersNode
+            when nil, ItParametersNode, NumberedParametersNode
               s(node, :args)
             else
               visit(node.parameters)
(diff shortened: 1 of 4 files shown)


Earlopain:
> I honestly can't explain the discrepancy

This one causes it, can't tell why though https://github.com/ruby/prism/blob/3e8066e5fee872a243a3bcdc79894de511031a12/test/prism/ruby/ruby_parser_test.rb#L18-L24

Overall it would still be parsed as 3.3 since there is no released version of `ruby_parser` to compare against. Though that monkeypatch seems to change some things.

presidentbeef:
AHHH thank you @Earlopain!

Prism should not do this, because there is code in RubyParser where it is comparing a Sexp with a line/file to a fresh Sexp with no line/file and changing its behavior as a result ([example](https://github.com/seattlerb/ruby_parser/blob/969574474e201cf7c71e8ffa8c6ed7b43da90553/lib/ruby_parser3.yy#L1925)).

Updated this PR to fix.

kddnewton (approved):
Nice!

PR #3632, opened 2025-09-02
[Bug #17398] Allow `private def hello = puts "Hello"`

Earlopain:
This was a limitation of parse.y that prism intentionally replicated.

https://bugs.ruby-lang.org/issues/17398#note-12


Earlopain:
CI will be red until the syntax tests in ruby/ruby are adapted via https://github.com/yui-knk/ruby/tree/bug_17398

Earlopain:
@kddnewton will you be available again in time for the Ruby 3.5 release? I am wondering if I should simply wait for your review or try to get it from someone else. Same for the change in https://bugs.ruby-lang.org/issues/20925

kddnewton:
I just got back today actually, so I will be able to start reviewing things again with more regularity.

Earlopain:
Ah, that's nice to hear. Welcome back

Earlopain:
How about it? The next ruby 3.4 release is planned for early next week and @xrxr has requested a new release: https://github.com/ruby/ruby/pull/14358#issuecomment-3229656013

Additionally, @nurse is growing ever more concerned about the maintenance status of prism: https://bugs.ruby-lang.org/issues/21540#note-10. The same is also true for @mame: https://github.com/ruby/prism/pull/3585#issuecomment-3079973089. Most of the bugs listed by mame were since then adressed and reviewed from a variety of shopify people but at the moment it looks like that effort was more of a one-off thing.

I try to help out but if no one reviews then I am rather stuck. In ruby/ruby I have a PR open for more than a month now without any feedback, not from you which is ok but also not from the other two that are listed as prism maintainers which is less ideal.

kddnewton:
Superceded by https://github.com/ruby/prism/pull/3639

PR #3634, opened 2025-09-08
Use pm_arguments_end for function call

mame:
Previously, the location of CallNode was incorrect when it accepts a block parameter:

```
ruby -rprism -e 'pp Prism.parse("foo(&blk)").value.statements.body[0]']
@ CallNode (location: (1,0)-(1,8))  # <=== It should be (1,0)-(1,9)
├── flags: ∅
├── receiver: ∅
├── call_operator_loc: ∅
├── name: :foo
├── message_loc: (1,0)-(1,3) = "foo"
├── opening_loc: (1,3)-(1,4) = "("
├── arguments: ∅
├── closing_loc: (1,8)-(1,9) = ")"
*snip*

$ ruby -rprism -e 'pp Prism.parse("foo(&blk)").value.statements.body[0].slice'
"foo(&blk"
```

Note that the slice lacks the closing parenthesis.

### DIFF
--- a/snapshots/method_calls.txt
+++ b/snapshots/method_calls.txt
@@ -278,7 +278,7 @@
         │   ├── arguments: ∅
         │   ├── closing_loc: ∅
         │   └── block: ∅
-        ├── @ CallNode (location: (25,0)-(25,8))
+        ├── @ CallNode (location: (25,0)-(25,9))
         │   ├── flags: newline, ignore_visibility
         │   ├── receiver: ∅
         │   ├── call_operator_loc: ∅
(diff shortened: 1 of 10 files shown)


mame:
By the way, the AST design of `BlockArgumentNode` feels odd. Since the `&blk` is within the arguments as `foo(&blk)`, it would be more natural for it to be a member of `ArgumentsNode#arguments` rather than a special position of `CallNode#block`.

kddnewton:
> By the way, the AST design of `BlockArgumentNode` feels odd. Since the `&blk` is within the arguments as `foo(&blk)`, it would be more natural for it to be a member of `ArgumentsNode#arguments` rather than a special position of `CallNode#block`.

Originally that was the design, but we got feedback early on that it was much easier to reason about call nodes when all of the blocks were in the same place.

eregon:
Regarding BlockArgumentNode, see discussion in https://github.com/ruby/prism/issues/1310 and linked issues/PRs.

PR #3639, opened 2025-09-12
[Bug #17398] Allow `private def hello = puts "Hello"`

kddnewton:
This was a limitation of parse.y that prism intentionally replicated.

Moving this directly onto ruby/prism so that I can sync it into ruby/ruby first.


kddnewton:
Waiting on https://github.com/ruby/ruby/pull/14530

Earlopain:
Thanks for tacking care of everything. I hope I wasn't coming off as being too demanding...

kddnewton:
Not at all!

PR #3647, opened 2025-09-14
Fix warn polyfill when no uplevel is provided

Earlopain:
An unspecified uplevel is not the same as an uplevel of 1:

```
$ irb
irb(main):001> warn("foo")
foo
=> nil
irb(main):002> warn("foo", uplevel: 1)
/home/user/.rbenv/versions/2.7.8/lib/ruby/gems/2.7.0/gems/irb-1.14.0/lib/irb/workspace.rb:121: warning: foo
=> nil
```

cc @koic


Earlopain:
JRuby can't handle this is seems...

kddnewton (changes requested):
Can you either keep the Integer check or just make it .nil? Instead of the case? Either way is fine with me

PR #3652, opened 2025-09-17
Reject argument command call taking a block with more trailing arguments

Earlopain:
https://bugs.ruby-lang.org/issues/21168#note-5

The added code samples align with `parse.y`, except for `foo(bar baz do end)` which `parse.y` currently rejects but shouldn't.

### DIFF
--- a/snapshots/command_method_call_2.txt
+++ b/snapshots/command_method_call_2.txt
@@ -1,0 +1,96 @@
+@ ProgramNode (location: (1,0)-(3,17))
+├── flags: ∅
+├── locals: []
+└── statements:
+    @ StatementsNode (location: (1,0)-(3,17))
+    ├── flags: ∅
+    └── body: (length: 2)
+        ├── @ CallNode (location: (1,0)-(1,19))
+        │   ├── flags: newline, ignore_visibility
+        │   ├── receiver: ∅
+        │   ├── call_operator_loc: ∅
+        │   ├── name: :foo
+        │   ├── message_loc: (1,0)-(1,3) = "foo"
+        │   ├── opening_loc: (1,3)-(1,4) = "("
+        │   ├── arguments:
+        │   │   @ ArgumentsNode (location: (1,4)-(1,18))
+        │   │   ├── flags: ∅
+        │   │   └── arguments: (length: 1)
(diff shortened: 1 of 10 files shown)


kddnewton:
I'm okay to merge this as-is, but I'm wondering if it indicates that we're passing `accepts_command_call` incorrectly somewhere or we're not checking it somewhere. Because this behavior should be controlled by that flag.

Earlopain:
Let me check again later when I got some time. I didn't yet find the place where `accepts_command_call` would be relevant while making this PR

Earlopain:
Actually, would `accepts_command_call` be relevant? It doesn't reject commands calls but any argument after a command call (that takes a block).

kddnewton:
I think so, because https://github.com/ruby/prism/blob/90a098dd1ee33cb6a089e07bc66e58a4d0ab0e72/src/prism.c#L15584 rejects any arguments without parentheses, which is what we're trying to reject here.

Earlopain:
(I just rebased and added `foo(bar 1 do end,)` as a syntax error test)

That `accepts_command_call` just decides if `bar 1 do end` should be allowed. `parse_arguments_list` gets called twice:
```
foo(bar 1 do end, 2)
    ^~~~~~~~~~~~~~~ # accepts_command_call is irrelevant, parens are present
        ^~~~~~~~    # accepts_command_call is true. Having it be false would reject the first argument to `foo`, which should be allowed.
```

Though it's not unreasonable to think that I am missing something or that the fix is better placed at a different location

kddnewton:
Right, but I'm talking about the `bar` method call, which would have been called from here: https://github.com/ruby/prism/blob/90a098dd1ee33cb6a089e07bc66e58a4d0ab0e72/src/prism.c#L18647.

Earlopain:
Yeah, but that command call is allowed. Only stuff following it should be rejected, so I don't currently see how it would be relevant.

It's ok to reject this if you want to take this yourself. Looking at the proper solution to learn from later is good for me too, I have stared at this for a while longer now and am not really getting any further.

kddnewton:
Ahh I'm sorry, I misread the ticket. I think this is definitely correct.

kddnewton (approved):

PR #3658, opened 2025-09-19
Fix dangling pointers on Windows as well

nobu:
Share the empty source string in `pm_string_mapped_init` and `pm_string_file_init`.


kddnewton (approved):
Thanks!

PR #3661, opened 2025-09-29
Add `unparser` into README library examples

viralpraxis:
[Unparser](https://github.com/mbj/unparser) now uses Prism (if available) via Parser translator.


kddnewton:
Thanks!

Issue #3558, opened 2025-05-05
`NoMatchingPatternError` on hash + parenthesis omission


avlazarov:
Hey folks,

Been getting an odd `NoMatchingPatternError` on Ruby 3.4.3 and Prism 1.4.0 (same on latest `main` at 2cae3102e9dfa7b5e4c53a6d6c3c5554f6108387)

```
# script.rb
module A
  extend self

  def print(message:)
    puts message
  end
end

case ENV['COND']
in 'FOO'
  message = 'foo'

  A.print message:
in 'BAR'
  message = 'bar'

  A.print message:
end
```

```
→ RUBYOPT="--parser prism" COND="BAR" ruby script.rb
script.rb:13:in '<main>': "BAR": "FOO" === "BAR" does not return true (NoMatchingPatternError)
```

Additionally, matching both conditions

```
→ RUBYOPT="--parser prism" COND="FOO" ruby script.rb
foo
bar
```

No issues with `parse.y`

```
→ RUBYOPT="--parser parse.y" COND="BAR" ruby script.rb
bar
→ RUBYOPT="--parser parse.y" COND="FOO" ruby script.rb
foo
```

kddnewton:
Thanks for the report! I'll check it out

tompng:
I think the cause of this issue is:
```ruby
Prism.parse_success? "print message: in 'BAR'"  # => false
Prism.parse_success? "A.print message: in 'BAR'"  # => true, should be false
Prism.parse_success? "A.print message:\nin 'BAR'"  # => true, should be false
```
and it is not fixed yet.

kddnewton:
You're absolutely right, I'll keep going. Thanks @tompng!

PR #3666, opened 2025-09-30
Follow fix for method call with `in` bug

eileencodes:
This PR is a followup to #3558 and fixes the following cases, which should throw a syntax error but were not:

```ruby
A.print message: in 'BAR'

A.print message:
in 'BAR'
```

To fix this we check for methods calls with arguments and no parens that then call `in`.

Note I had claude write this because I wasn't sure where the problem was but then edited it to remove extra unnecessary code the robot wrote.

cc/ @tenderlove

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -22304,14 +22304,23 @@ parse_expression(pm_parser_t *parser, pm_binding_power_t binding_power, bool acc
                     return node;
                 }
                 break;
-            case PM_CALL_NODE:
+            case PM_CALL_NODE: {
                 // These expressions are also statements, by virtue of the
                 // right-hand side of the expression (i.e., the last argument to
                 // the call node) being an implicit array.
                 if (PM_NODE_FLAG_P(node, PM_CALL_NODE_FLAGS_IMPLICIT_ARRAY) && pm_binding_powers[parser->current.type].left > PM_BINDING_POWER_MODIFIER) {
                     return node;
                 }
+
+                // Method calls with arguments (without parentheses)
+                // cannot be followed by the `in` operator.
+                pm_call_node_t *cast = (pm_call_node_t *) node;
+                if (cast->arguments != NULL && cast->opening_loc.start == NULL && parser->current.type == PM_TOKEN_KEYWORD_IN) {
+                    return node;
+                }
+
                 break;
+            }
             default:
                 break;
         }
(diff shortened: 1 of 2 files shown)


src/prism.c
```diff
@@ -22304,14 +22304,23 @@ parse_expression(pm_parser_t *parser, pm_binding_power_t binding_power, bool acc
...
                 if (PM_NODE_FLAG_P(node, PM_CALL_NODE_FLAGS_IMPLICIT_ARRAY) && pm_binding_powers[parser->current.type].left > PM_BINDING_POWER_MODIFIER) {
                     return node;
                 }
+
+                // Method calls with arguments (without parentheses)
+                // cannot be followed by the `in` operator.
+                pm_call_node_t *cast = (pm_call_node_t *) node;
+                if (cast->arguments != NULL && cast->opening_loc.start == NULL && parser->current.type == PM_TOKEN_KEYWORD_IN) {
+                    return node;
+                }
```
tenderlove, line 22357: I need @kddnewton's opinion here, but I feel like testing for a specific token isn't the right fix and that we should be using binding power instead (but this is just a gut feeling).  Is `in` the only keyword that has a problem?

tompng: > Is `in` the only keyword that has a problem?

Looks like `in`, `=>`, `.`, `&.` and all binary operators except `+` `-` `**` `&` `*` are wrongly accepted after `A.print message:` but not after `a.print message:`

tompng: And ternary operator `A.print message: ? 1 : 2`

Earlopain: I agree with @tenderlove that this is not the correct fix. The code that is supposed to handle this is here: https://github.com/ruby/prism/blob/ce4abe118c490340621cc7e41c4b81ace9494511/src/prism.c#L22253-L22261

But `parse_expression_prefix` returns `PM_CONSTANT_READ_NODE`. So I feel like the fix would be to modify `parse_expression_prefix` so that it parses further instead of just one node deep.  Or this check is also needed at some other place.

But I did not manage to actually fix it myself last I looked into this.

eileencodes:
I made some changes and don't think this is done yet. I added some new tests so we have more coverage while figuring this out. The `=>`, `.` and `&.` were passing without changes but I suspect the issue is where the error is indicated?

@tompng do you have a script you're running that we can use to figure out which others are still not working correctly? Thanks!

tompng:
For this case, I don't have a script. I just listed up by glazing at all symbol keys in my keyboard and tried `Prism.parse_success?` manually.

kddnewton:
Superseded by #3948

PR #3669, opened 2025-10-03
Reject `p(p a, &block => value)` and similar

Earlopain:
They were being parsed as `p((p a, &block) => value)`. When we get to this point, we must not just have parsed a command call, always consuming the `=>` is not correct.

Closes [[Bug #21622]](https://bugs.ruby-lang.org/issues/21622)

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -14232,6 +14232,25 @@ parse_assocs(pm_parser_t *parser, pm_static_literals_t *literals, pm_node_t *nod
     return contains_keyword_splat;
 }
 
+static inline bool
+argument_allowed_for_bare_hash(pm_parser_t *parser, pm_node_t *argument) {
+    if (pm_symbol_node_label_p(argument)) {
+        return true;
+    }
+
+    switch (PM_NODE_TYPE(argument)) {
+        case PM_CALL_NODE: {
+            pm_call_node_t *cast = (pm_call_node_t *) argument;
+            if (cast->opening_loc.start == NULL && cast->arguments != NULL) {
+                return false;
+            }
+            break;
+        }
+        default: break;
+    }
+    return accept1(parser, PM_TOKEN_EQUAL_GREATER);
+}
+
 /**
  * Append an argument to a list of arguments.
  */
@@ -14389,7 +14408,7 @@ parse_arguments(pm_parser_t *parser, pm_arguments_t *arguments, bool accepts_for
                 bool contains_keywords = false;
                 bool contains_keyword_splat = false;
 
-                if (pm_symbol_node_label_p(argument) || accept1(parser, PM_TOKEN_EQUAL_GREATER)) {
+                if (argument_allowed_for_bare_hash(parser, argument)){
                     if (parsed_bare_hash) {
                         pm_parser_err_previous(parser, PM_ERR_ARGUMENT_BARE_HASH);
                     }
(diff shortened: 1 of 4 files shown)


src/prism.c
```diff
@@ -14378,7 +14397,7 @@ parse_arguments(pm_parser_t *parser, pm_arguments_t *arguments, bool accepts_for
                 bool contains_keywords = false;
                 bool contains_keyword_splat = false;
 
-                if (pm_symbol_node_label_p(argument) || accept1(parser, PM_TOKEN_EQUAL_GREATER)) {
+                if (argument_allowed_for_bare_hash(parser, argument)){
```
Earlopain, line 14411: There is very similar code here https://github.com/ruby/prism/blob/ce4abe118c490340621cc7e41c4b81ace9494511/src/prism.c#L18217

I tried a few things to find code samples which force me to do the same change there as well but I did not find such code. Left it alone for now

test/prism/errors/command_calls_35.txt
```diff
@@ -0,0 +1,41 @@
...
+              ^~ unexpected '=>', ignoring it
+                  ^ unexpected ',', expecting end-of-input
+                  ^ unexpected ',', ignoring it
+                    ^ unexpected '&', ignoring it
+
+p p p 1 => 2 => 3 => 4
+             ^~ unexpected '=>', expecting end-of-input
+             ^~ unexpected '=>', ignoring it
+
+p[p a, x: b => value]
```
Earlopain, line 37: This is code which I assumed would still be accepted with this change (see previous comment). But it is correctly rejected.

test/prism/fixtures/command_method_call.txt
```diff
@@ -39,3 +39,9 @@ def foo = bar 1
 !foo 1 or !bar 2
 
 not !foo 1
+
+foo(bar baz, key => value)
```
Earlopain, line 43: Probably already tested somewhere :shrug:

Earlopain:
There is this https://github.com/ruby/prism/blob/c89ca2af12ba20b4fd2c5ff43ebe25da1d81d8db/.github/workflows/cruby-bindings.yml#L46

But I guess you mean something a bit more comprehensive?

tenderlove:
> But I guess you mean something a bit more comprehensive?

Ya, I'd like it if the whole suite was run. But it may make CI take too long.  We've had cases where those two tests passed just fine, but merging broke upstream CI.

Earlopain:
Yeah, I thought so. I ran test-all locally with this and apart from some `LoadError: cannot load such file -- .../ruby/lib/prism/node`  (I seem to have messed up my repo somehow) it doesn't fail anything else for me.

Earlopain:
This needs to be reverted: https://github.com/ruby/ruby/actions/runs/19615315949/job/56167057220

```
/home/runner/work/ruby/ruby/build/install/lib/ruby/4.0.0+0/bundled_gems.rb:60:in 'Kernel.require': --> /home/runner/work/ruby/ruby/build/install/lib/ruby/gems/4.0.0+0/gems/shipit-engine-0.40.1/app/models/shipit/task.rb

unexpected '=>', expecting end-of-input
unexpected '=>', ignoring it

    3  module Shipit
    4    class Task < Record
  100      state_machine :status, initial: :pending do
  157        event :aborting do
> 158          transition all - %i[aborted] => :aborting
  159        end
  180      end
  497    end
  498  end
```

Seems I missed something, will revisit this later

PR #3673, opened 2025-10-06
Add a workflow to sync commits to ruby/ruby

k0kubun:
https://bugs.ruby-lang.org/issues/21628

### DIFF
--- a/.github/workflows/sync-ruby.yml
+++ b/.github/workflows/sync-ruby.yml
@@ -1,0 +1,33 @@
+name: Sync ruby
+on:
+  push:
+    branches: [master]
+jobs:
+  sync:
+    name: Sync ruby
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'ruby' }}
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Create GitHub App token
+        id: app-token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: 2060836
+          private-key: ${{ secrets.RUBY_SYNC_DEFAULT_GEMS_PRIVATE_KEY }}
+          owner: ruby
+          repositories: ruby
+
+      - name: Sync to ruby/ruby
+        uses: convictional/trigger-workflow-and-wait@v1.6.5
+        with:
+          owner: ruby
+          repo: ruby
+          workflow_file_name: sync_default_gems.yml
+          github_token: ${{ steps.app-token.outputs.token }}
+          ref: master
+          client_payload: |
+            {"gem":"${{ github.event.repository.name }}","before":"${{ github.event.before }}","after":"${{ github.event.after }}"}
+          propagate_failure: true
+          wait_interval: 10


.github/workflows/sync-ruby.yml
```diff
@@ -0,0 +1,33 @@
+name: Sync ruby
+on:
+  push:
+    branches: [master]
```
kddnewton, line 4: On mobile so I can't fix this now, but I'm pretty sure this should be [main]

.github/workflows/sync-ruby.yml
```diff
@@ -0,0 +1,33 @@
...
+          repositories: ruby
+
+      - name: Sync to ruby/ruby
+        uses: convictional/trigger-workflow-and-wait@v1.6.5
+        with:
+          owner: ruby
+          repo: ruby
+          workflow_file_name: sync_default_gems.yml
+          github_token: ${{ steps.app-token.outputs.token }}
+          ref: master
```
kddnewton, line 29: Maybe this too?

.github/workflows/sync-ruby.yml
k0kubun: yeah sorry, I've fixed that one https://github.com/ruby/prism/commit/c89ca2af12ba20b4fd2c5ff43ebe25da1d81d8db

kddnewton: Ahh sorry should have checked the commits

.github/workflows/sync-ruby.yml
k0kubun: This refers to ruby/ruby's master, so this is correct.

kddnewton: Great, thanks for doing this!

k0kubun:
FYI, ruby/prism was out-of-sync with ruby/ruby as of filing this PR, but I fixed it at https://github.com/ruby/ruby/pull/14751. So if future sync fails, it's not because it had out-of-sync files. When this workflow fails, we should fix every failure of it.

For example, https://github.com/ruby/prism/pull/3667 could not be sync-ed by `tool/sync_default_gems.rb {before}..{after}` (failure log: https://github.com/ruby/ruby/actions/runs/18298406027/job/52101460864).

I think ruby/prism has too many file names that are inconsistent with ruby/ruby. `git cherry-pick` can recognize file renames, but unless somebody pushes files in the original paths and then rename them to desired paths in ruby/ruby, this will not happen. And such explicit push-and-rename does not happen automatically, so I suspect files that are added after the initial merge (w/ renames in git history) and have an inconsistent file path in ruby/ruby tend to fail the sync today.

So I suggest either:

1. Make sure ruby/prism uses the same file paths as ruby/ruby, or
2. Let `tool/sync_default_gems.rb prism before..after` break up some cherry-picks so that they will be pushed to the original paths in the first commit and then renamed to the desired paths in ruby/ruby.
    * which I assume will help us fix the sync failures, but I'm not sure for 100%.
    * Also you probably need to do something about existing files that are already pushed to ruby/ruby without a rename.

k0kubun:
>  think ruby/prism has too many file names that are inconsistent with ruby/ruby

Actually, I'm not sure whether they're really inconsistent or just deleted on the ruby/ruby side. But either way it still fails a lot with:

```
Remove added files: snapshots
rm 'snapshots/character_literal.txt'
rm 'snapshots/character_literal.txt'
rm 'snapshots/dos_endings.txt'
rm 'snapshots/dos_endings.txt'
rm 'snapshots/dstring.txt'
rm 'snapshots/dstring.txt'
...
```

and this needs to be fixed.

Earlopain:
Ruby does not want these snapshots https://github.com/ruby/ruby/pull/11624

And I messed up that PR, next time I will just let it be reverse-synced instead of filling two PRs with potentially different content and add prism specific tests later

k0kubun:
Thanks for the context.

> Ruby does not want these snapshots

I'm personally fine with either way, but I just want `tool/sync_default_gems.rb prism {before}..{after}` to not fail under any Prism changes.

If you decide to keep removing ruby/prism's `snapshots/` from ruby/ruby (the current situation), `tool/sync_default_gems.rb`'s `sync_default_gems_with_commits` (not `sync_default_gems`, which works fine already) should be modified to make sure changing files under ruby/prism's `snapshots/` don't fail it. You might be thinking it failed because you filed two PRs, but it should skip the commit and exit normally in that case. So it's still something to be fixed.

If you decide to re-add ruby/prism's `snapshots/` to ruby/ruby, you probably want to put them in `test/prism/snapshots/` for ruby/prism too to avoid sync failures.

Earlopain:
@kddnewton would have even more context. Easiest would probably be to just put them back into ruby/ruby 🤷. There is not a lot of movement in these snapshots anymore these days, if that was the reason to exclude them. Overall I am missing a lot of knowledge to act on what you wrote.

kddnewton:
I had them under `test/prism/snapshots`, but that ended up failing _more_. This failed less. Here's the situation: we want them in ruby/prism, but @hsbt indicated that no generated files can be checked into ruby/ruby. I'm not sure of the best way to remove them. Maybe @nobu has more context on the tool/sync_default_gems.rb script? I'll bet we could filter out these files so they aren't added in the first place? (At least I hope?)

k0kubun:
Remember how every repository has `.github` for GitHub Actions that is not copied to ruby/ruby and still succeeds to execute `tool/sync_default_gems.rb`. It's handled in `ignored_file_pattern_for` in `tool/sync_default_gems.rb`, and I think you can add "prism" under the `# Gem-specific patterns` comment in that method and skip `snapshots/`.

PR #3674, opened 2025-10-08
[Bug #21623] Consistently reject endless method definitions without argument parens

Earlopain:
For these special cases, there exists no optional argument type. Since a endless method is started with `=`, there was ambiguity here. We have to simply reject these in all cases.

This adds a new error for the following reason:
* `def foo arg = nil` is interpreted as a normal method call with optional `arg` without matching `end`
* `def foo *arg = nil; end` is interpreted as a endless method call that has body `nil` with extraneous `end`

`def foo *arg = nil` is somewhere inbetween and I don't know how to otherwise indicate the error.
Now the second case above also shows the newly added error message.

Fixes [[Bug #21623]](https://bugs.ruby-lang.org/issues/21623)


tompng:
I'm not sure if it's related to this fix, perhaps a separate issue, but `def foo a, (b) = 1` is wrongly accepted.

PR #3679, opened 2025-10-14
Add support for `Prism.parse(foo, version: "current")`

Earlopain:
The docs currently say to use `Prism.parse(foo, version: RUBY_VERSION)` for this. By specifying "current" instead, we can have prism raise a more specifc error. It doesn't add much code so I think being more explicit is worth it. It's similar to `Prism::Translation::ParserCurrent` but instead of warning it raises.

For a smooth experience, it requires a prism release right around when a new ruby version is released. But that is also already true for using `RUBY_VERSION`.

Note: Does not use `ruby_version` from `ruby/version.h` because writing a test for that is not really possible. I instead pull it from `RUBY_VERSION` constant at runtime so it can be stubbed.

Ref https://bugs.ruby-lang.org/issues/21618

### DIFF
--- a/lib/prism.rb
+++ b/lib/prism.rb
@@ -37,6 +37,27 @@ module Prism
   private_constant :LexCompat
   private_constant :LexRipper
 
+  # Raised when requested to parse as the currently running Ruby version but Prism has no support for it.
+  class CurrentVersionError < ArgumentError
+    # Initialize a new exception for the given ruby version string.
+    def initialize(version)
+      message = +"invalid version: Requested to parse as `version: 'current'`; "
+      gem_version =
+        begin
+          Gem::Version.new(version)
+        rescue ArgumentError
+        end
+
+      if gem_version && gem_version < Gem::Version.new("3.3.0")
+        message << " #{version} is below the minimum supported syntax."
+      else
+        message << " #{version} is unknown. Please update the `prism` gem."
+      end
+
+      super(message)
+    end
+  end
+
   # :call-seq:
   #   Prism::lex_compat(source, **options) -> LexCompat::Result
   #
--- a/lib/prism/ffi.rb
+++ b/lib/prism/ffi.rb
@@ -423,7 +423,9 @@ module Prism
 
     # Return the value that should be dumped for the version option.
     def dump_options_version(version)
-      case version
+      current = version == "current"
+
+      case current ? RUBY_VERSION : version
       when nil, "latest"
         0 # Handled in pm_parser_init
       when /\A3\.3(\.\d+)?\z/
@@ -433,7 +435,11 @@ module Prism
       when /\A3\.5(\.\d+)?\z/
         3
       else
-        raise ArgumentError, "invalid version: #{version}"
+        if current
+          raise CurrentVersionError, RUBY_VERSION
+        else
+          raise ArgumentError, "invalid version: #{version}"
+        end
       end
     end
 
--- a/test/prism/api/parse_test.rb
+++ b/test/prism/api/parse_test.rb
@@ -140,6 +140,25 @@ module Prism
       end
     end
 
+    def test_version_current
+      if RUBY_VERSION >= "3.3"
+        assert Prism.parse_success?("1 + 1", version: "current")
+      else
+        assert_raise(CurrentVersionError) { Prism.parse_success?("1 + 1", version: "current") }
+      end
+
+      version = RUBY_VERSION.split(".").tap { |segments| segments[0] = segments[0].succ }.join(".")
+      stub_ruby_version(version) do
+        error = assert_raise(CurrentVersionError) { Prism.parse("1 + 1", version: "current") }
+        assert_includes error.message, "unknown"
+      end
+
+      stub_ruby_version("2.7.0") do
+        error = assert_raise(CurrentVersionError) { Prism.parse("1 + 1", version: "current") }
+        assert_includes error.message, "minimum"
+      end
+    end
+
     def test_scopes
       assert_kind_of Prism::CallNode, Prism.parse_statement("foo")
       assert_kind_of Prism::LocalVariableReadNode, Prism.parse_statement("foo", scopes: [[:foo]])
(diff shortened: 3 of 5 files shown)


Earlopain:
Unsure about the ffi backend. My understanding is that alternate ruby versions will use it, basically the interface to the version of prism shipped with their runtime itself. So it should also be supported there. Is that right?

lib/prism/ffi.rb
```diff
@@ -423,7 +423,15 @@ def dump_options_command_line(options)
 
     # Return the value that should be dumped for the version option.
     def dump_options_version(version)
-      case version
+      current = version == "current"
+      if current
+        case RUBY_VERSION
+        when /\A2\.7(\.\d+)?\z/, /\A3\.0(\.\d+)?\z/, /\A3\.1(\.\d+)?\z/, /\A3\.2(\.\d+)?\z/
```
eregon:
```suggestion
        when /\A2\.7(\.\d+)?\z/, /\A3\.[012](\.\d+)?\z/
          # Requested to parse as a version prism will never support
```

test/prism/api/parse_test.rb
```diff
@@ -140,6 +140,22 @@ def test_version
...
     end
 
+    def test_version_current
+      if RUBY_VERSION >= "3.3"
+        assert Prism.parse_success?("1 + 1", version: "current")
+      end
+
+      stub_ruby_version("3.6.0") do
+        error = assert_raise(ArgumentError) { Prism.parse("1 + 1", version: "current") }
+        assert_match "Prism does not know about Ruby 3.6.0 syntax", error.message
```
eregon: It's my preference but I would recommend to use `assert_equal` and have the full message as expectation here, that way it's also representative of what the user will see.
(same just below of course)

eregon:
> Unsure about the ffi backend. My understanding is that alternate ruby versions will use it, basically the interface to the version of prism shipped with their runtime itself. 

Not quite, you can think of it as Prism gem having 2 backends: a C extension, only used on CRuby, and the FFI backend used on JRuby and TruffleRuby.
Whether it's the prism default gem or later-installed prism gem doesn't really make a difference.

> So it should also be supported there. Is that right?

Yes.

eregon:
I wonder if `Prism.parse(foo, version: :current)` could be easily supported?
That seems nice to me because `:current` is a "symbolic value" to mean current version.
It's a detail though, so if not trivial probably not worth it.

test/prism/api/parse_test.rb
Earlopain: Yeah, sure.

Earlopain:
Thanks for the the review!

> Not quite, you can think of it as Prism gem having 2 backends: a C extension, only used on CRuby, and the FFI backend used on JRuby and TruffleRuby.
>  Whether it's the prism default gem or later-installed prism gem doesn't really make a difference.

Ah, right. I thought that at least on truffleruby it would use the c extension when part of the gemfile (i.e. explicitly installed)

> I wonder if Prism.parse(foo, version: :current) could be easily supported?

Probably, just convert it to a string somewhere. There is already `Prism.parse(foo, version: "latest")` though, so I would not do it in this PR. I'm not so interested in it, feel free to open an issue for that if you want it.

Earlopain:
I guess you're saying to move the C extension code for version handling into ruby as well, and then share that between ffi? Something like `Prism.resolve_ruby_version_or_raise!` or something. Is there a place for this already? I didn't really find something that would fit.

kddnewton:
Not quite. What I mean is that instead of:

```c
if (RSTRING_LEN(value) == 7 && strncmp(version, "current", 7) == 0) {
    if (strncmp || strncmp || strncmp || strncmp) { rb_raise }
    if (!pm_options_version_set(options, current_version, 3)) { rb_raise(ArgumentError, "...") }
}
```

and:

```ruby
if current
  case RUBY_VERSION
  when ...
    raise ArgumentError, "..."
  end
end
```

and in the tests:

```ruby
error = assert_raise(ArgumentError) { Prism.parse(...) }
assert_equal error.message, "..."
```

you somewhat simplify to:

```ruby
# prism.rb
class CurrentVersionError < ArgumentError
  def initialize(version)
    super("Prism does not support #{version}. Please specify a supported version instead of 'current'.")
  end
end
```

and:

```c
if (RSTRING_LEN(value) == 7 && strncmp(version, "current", 7) == 0) {
    if (!pm_options_version_set(options, current_version, 3)) { rb_raise(rb_cCurrentVersionError, "...") }
}
```

and:

```ruby
if current && RUBY_VERSION !~ /supported-versions-pattern/
  raise CurrentVersionError, "..."
end
```

and in the tests:

```ruby
error = assert_raise(CurrentVersionError) { Prism.parse(...) }
```

Earlopain:
Ah, gotcha. That is nicer. Thanks for the explanation

lib/prism.rb
```diff
@@ -37,6 +37,27 @@ module Prism
...
   private_constant :LexRipper
 
+  # Raised when requested to parse as the currently running Ruby version but Prism has no support for it.
+  class CurrentVersionError < ArgumentError
+    # Initialize a new exception for the given ruby version string.
+    def initialize(version)
+      message = +"invalid version: Requested to parse as `version: 'current'`; "
+      gem_version =
+        begin
+          Gem::Version.new(version)
```
Earlopain, line 47: ruby/ruby CI is failing here because `Gem` is not loaded: https://github.com/ruby/ruby/actions/runs/18561653353/job/52911907413

kddnewton: Ugh of course. I'll fix it. Shouldn't have cowboy'd.

Earlopain: I seem to have messed something up as well https://github.com/ruby/ruby/actions/runs/18562695205/job/52915425235
Can't look into that right now though

(This one doesn't crash: https://github.com/ruby/ruby/actions/runs/18562695142/job/52915424818)

kddnewton: It's weird. I think RUBY_VERSION is just not consistent inside hte CRuby test suite. I reduced a lot of the tests and just ended up with some bare minimum stuff, and it appears to have resolved.

PR #3680, opened 2025-10-15
Trusted publishing workflows

tenderlove:
I'm working on adding trusted publishing workflows so we can automate gem / crate / npm package releases.  Once we get this working correctly, we can simply push a tag to GitHub, and all of the packages will be released automatically.

The plan is to add one workflow per publisher.

- [x] RubyGems workflow
- [x] crates.io workflow
- [x] npmjs workflow

I set up the RubyGems workflow based on [other gems that we publish](https://github.com/ruby/psych/blob/master/.github/workflows/push_gem.yml).

I'm less confident about the other workflows, but I used [this documentation for crates.io](https://crates.io/docs/trusted-publishing) as well as [this documentation for npm](https://docs.npmjs.com/trusted-publishers).

I based the crates workflow and the npm workflow off of their respective CI workflows, but I'd really appreciate reviews from other folks.  Once we have something to release, we can test these workflows. 😄

### DIFF
--- a/.github/workflows/publish-crate.yml
+++ b/.github/workflows/publish-crate.yml
@@ -19,10 +19,8 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: ${{ matrix.ruby }}
-
-      - name: Install dependencies
-        run: bundle install --jobs 4 --retry 3
+          ruby-version: "3.4"
+          bundler-cache: true
 
       - uses: rust-lang/crates-io-auth-action@v1
         id: auth
--- a/.github/workflows/publish-gem.yml
+++ b/.github/workflows/publish-gem.yml
@@ -33,9 +33,7 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: "3.4"
-
-      - name: Install dependencies
-        run: bundle install --jobs 4 --retry 3
+          bundler-cache: true
 
       - name: Publish to RubyGems
         uses: rubygems/release-gem@v1
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -19,16 +19,14 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: ${{ matrix.ruby }}
+          ruby-version: "3.4"
+          bundler-cache: true
 
       - name: Set up node
         uses: actions/setup-node@v4
         with:
           node-version: '20'
           registry-url: 'https://registry.npmjs.org'
-
-      - name: Install dependencies
-        run: bundle install --jobs 4 --retry 3
 
       - name: rake templates
         run: bundle exec rake templates


.github/workflows/publish-crate.yml
```diff
@@ -0,0 +1,49 @@
...
+jobs:
+  publish:
+    if: github.repository == 'ruby/prism'
+    runs-on: ubuntu-latest
+    environment: release  # Optional: for enhanced security
+    permissions:
+      id-token: write     # Required for OIDC token exchange
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
```
kddnewton: I've only seen these git commit hashes used by Shopify, outside of that I've only seen version numbers, which I would prefer. So this would be:

```suggestion
        uses: step-security/harden-runner@v2
```

.github/workflows/publish-crate.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@eaecf785f6a34567a6d97f686bbb7bccc1ac1e5c # v1.237.0
+        with:
+          ruby-version: ${{ matrix.ruby }}
+
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          targets: wasm32-wasip1
```
kddnewton: I don't think we need this, as this was just to build/test the WASM build for rust.

.github/workflows/publish-crate.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Install dependencies
+        run: bundle install --jobs 4 --retry 3
+
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
```
kddnewton: I think this should be rust/target, since it builds in a subdirectory

.github/workflows/publish-crate.yml
```diff
@@ -0,0 +1,49 @@
...
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-${{ hashFiles('Cargo.toml') }}
```
kddnewton: I'm not 100% but iI think this should be `hashFiles('rust/Cargo.toml')`

.github/workflows/publish-gem.yml
```diff
@@ -0,0 +1,52 @@
...
+      name: rubygems.org
+      url: https://rubygems.org/gems/prism
+
+    permissions:
+      contents: write
+      id-token: write
+
+    strategy:
+      matrix:
+        ruby: ["ruby"]
```
kddnewton: I don't think we need a matrix here, we can just use a single Ruby version

.github/workflows/publish-gem.yml
```diff
@@ -0,0 +1,52 @@
...
+        uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@ff7abcd0c3c05ccf6adc123a8cd1fd4fb30fb493 # v4.2.2
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@eaecf785f6a34567a6d97f686bbb7bccc1ac1e5c # v1.237.0
+        with:
+          ruby-version: ${{ matrix.ruby }}
```
kddnewton:
```suggestion
          ruby-version: "3.4"
```

.github/workflows/publish-gem.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Install dependencies
+        run: bundle install --jobs 4 --retry 3
+
+      - name: Publish to RubyGems
+        uses: rubygems/release-gem@a25424ba2ba8b387abc8ef40807c2c85b96cbe32 # v1.1.1
+
+      - name: Create GitHub release
+        run: |
+          tag_name="$(git describe --tags --abbrev=0)"
+          gh release create "${tag_name}" --verify-tag --generate-notes
```
kddnewton: Let's skip this step. I don't want it to automatically generate notes. We can add a step later that will pull from the changelog, but for now let's just drop this.

.github/workflows/publish-crate.yml
tenderlove: I actually got these hashes from [here](https://github.com/ruby/psych/blob/a05262f0ebcf065fdaf9760ecce87eb3eeed59a8/.github/workflows/push_gem.yml#L29-L37). I'd rather use the `v2` thing, but I assumed since the Psych one was tested, it would work.

.github/workflows/publish-crate.yml
Earlopain: Is caching even needed here? Rust takes long to compile, yes, but does that even matter here?

.github/workflows/publish-crate.yml
Earlopain: I also don't like hash pinning (for public repos), release workflows are the only exception where I admit defeat and have to say I can kind of see the point. Still, unless someone actually reviews the changes before blindly merging dependabot PRs, it doesn't really do much in my eyes.

Either way, it should be consistent about it. There are a few other actions that just pin the major version, that kind of defeats the point.

.github/workflows/publish-crate.yml
```diff
@@ -0,0 +1,49 @@
...
+          ruby-version: ${{ matrix.ruby }}
+
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          targets: wasm32-wasip1
+
+      - name: Install dependencies
+        run: bundle install --jobs 4 --retry 3
```
Earlopain: I think you can just move this to setup-ruby and have it be automatically cached with `bundler-cache: true`. Same for the other two workflows.

.github/workflows/publish-gem.yml
Earlopain: Yes please. I like automated releases, but I don't particularly like reading release notes with 90% dependabot merges.

Earlopain:
Oh, for the gem you definitly need some setup on rubygems.org for it to work if you didn't do that already: https://guides.rubygems.org/trusted-publishing/adding-a-publisher/

.github/workflows/publish-crate.yml
tenderlove: I admit that I'm totally out of my depth here (I don't know enough to make an informed opinion).  If using the versions works, I'd rather do that so I'll change this PR.

tenderlove:
> Oh, for the gem you definitly need some setup on rubygems.org for it to work if you didn't do that already

Ya, I've already set this up with crates.io, npmjs, and RubyGems.org.  Just need to get the workflow files in shape 😄

.github/workflows/publish-crate.yml
Earlopain: It is for security. With `@v1` it is obvious that you will get new versions as they are released but even if you do `@v1.2.3` that is still true. These things are not immutable (github is working on it, not sure where that is at right now), so using the commit hash ensures that you always get exactly the same, even if someone compromised the action and made malicious releases.

https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

tenderlove: These should all be updated with `@vBLAH` now.  I tested [with this gem of mine](https://github.com/tenderlove/bitz/blob/main/.github/workflows/release.yml) and it seemed to work fine.

.github/workflows/publish-crate.yml
tenderlove: Removed

.github/workflows/publish-crate.yml
tenderlove: I got this from [here](https://github.com/ruby/prism/blob/4fb07108cb0022b329eb9d3ea44cf439a5b74464/.github/workflows/rust-bindings.yml#L43). Should it also change? Or is this somehow different?

.github/workflows/publish-crate.yml
tenderlove: I'm going to remove the caching.  I just cargoculted (lol) it from the Rust extension build, but I don't think it makes sense.

.github/workflows/publish-gem.yml
tenderlove: should be fixed

.github/workflows/publish-gem.yml
tenderlove: Sounds good. I removed it.

.github/workflows/publish-crate.yml
tenderlove: I think I've fixed this correctly 😅

tenderlove:
I think I've addressed most of the feedback. Please take another look! 🙇

.github/workflows/publish-crate.yml
tenderlove: I guess since I removed caching, this doesn't matter. Though I still wonder about the other workflow pointing at the right location.

kddnewton (approved):

Earlopain:
🤞

PR #3681, opened 2025-10-15
`-Wsign-conversion`

Earlopain:
selenium-webdriver pinned prism to 1.4.0 because of the following warning (which their compiler treats as an error):

```
src/prism.c: In function ‘context_terminator’:
src/prism.c:8651:62: error: conversion to ‘unsigned int’ from ‘int’ may change
the sign of the result [-Werror=sign-conversion]
8651 |     return token->type < 32 && (context_terminators[context] & (1 <<
token->type));
```

I have seen this warning myself during building, but now I can't reproduce anymore. Perhaps CI will tell me more.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,8 +11,8 @@ FUZZ_OUTPUT_DIR = $(CURDIR)/fuzz/output
 SOEXT ?= $(shell ruby -e 'puts RbConfig::CONFIG["SOEXT"]')
 
 CPPFLAGS := -Iinclude $(CPPFLAGS)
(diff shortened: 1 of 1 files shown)


froydnj:
You almost certainly need to cast the 1 to unsigned prior to the shift.

Earlopain:
Oh, I see that `-Wconversion` is already enabled (which I think includes `-Wsign-conversion`) :shrug:. I can't remember what I did to get this warning myself to confirm anything

kddnewton:
I think this is now solved by #3682

Earlopain:
Yeah, probably.

PR #3682, opened 2025-10-15
explicitly cast shifted constant to unsigned to avoid undefined behavior

froydnj:
cc @Earlopain 

I think this resolves the problem that #3681 was attempting to find -- shifting a signed integer in such a way that the sign bit changes is undefined behavior.  So if `token->type` is 31, `1 << 31` is actually undefined.

Forcing the constant to be unsigned should fix this, and is good for consistency because the table we're masking against is full of unsigned integers anyway.  I went ahead and changed the table initialization out of an abundance of caution, since we might have one of those token values be 31 someday; happy to undo that one if you like.


kddnewton (approved):

Earlopain:
Thanks @froydnj! I don't have a super obvious way to test this, when the next version releases I can know for sure.

froydnj:
Is it easy to figure out what compiler and version is being used in selenium-webdriver CI @Earlopain ?  If you knew that (and ideally the compiler options they are using aren't too hard to figure out either), you could run smaller test cases through https://gcc.godbolt.org .

Earlopain:
It's some bazel setup which I don't want to mess with too much. I gave a quick look through the [build logs](https://gypsum.cluster.engflow.com/invocations/default/fb6735d1-7661-447d-9ee6-a13dbd60ed8c?treeFilterStatus=failed#console) but nothing stood out to me. It can't be anything too exotic since I could swear I've seen this warning myself in the past.

Earlopain:
Looks like this was exactly what was needed, their build goes fine now

eregon:
FWIW I got the same error with gcc 8.5.0 when compiling in AlmaLinux release 8.10 (`FROM almalinux:8`, similar to RHEL 8), with prism 1.5.1.

PR #3684, opened 2025-10-16
Test against everything from CRuby

Earlopain:
I believe this is a good idea, let's see if this works.


Earlopain:
I was hoping this triggers failures from https://github.com/ruby/prism/pull/3679 but it seems it doesn't.

Regardless, this should be ok to do. @tenderlove since you were asking: it doesn't take that long, only 7m30s vs 5m30s before.

PR #3685, opened 2025-10-17
Make error and snapshot tests multi-version aware

Earlopain:
This one has been on my mind for a while now.

Currently, there are only tests against the latest syntax version.

This changes the snapshot structure as follows:
* Snapshots at their current location are tested against all syntax versions
* Snapshots inside a version folder like "3.3" are tested against all versions starting from that version
* Snapshots inside a version folder like "3.3-4.2" are tested against all versions in the given range.

This makes sure that as new syntax is added, older versions still work as expected. I also added a few tests for now valid syntax that should be invalid in older versions (and the other way around as well)

These tests run really fast. So even though it does 3x the work for these, I am still able to run the whole test suite in just 11 seconds.


Earlopain:
This did find an inconsequential bug where `class << A; return; end` was not accepted on Ruby 3.3, https://github.com/ruby/ruby/pull/11234 changed it without a version check. It's whatever for 3.3 syntax, but basically this is why I want to do this.

PR #3686, opened 2025-10-20
Reenable windows head

Earlopain:
Disabled in https://github.com/ruby/prism/commit/2a2a1ebe008861445d137647d53c4cde87cdd6b6
I'd hope this is fixed by now


kddnewton:
@Earlopain there are a couple of others that are disabled, do you want to enable those?

Earlopain:
Ah, I missed the fiddle ones. Yeah, let me try that in a bit

PR #3687, opened 2025-10-20
Reenable windows CI that were disabled because of fiddle

Earlopain:
Followup to https://github.com/ruby/prism/commit/cd91afe4fe424a3e423992c3e77558121c669b60
Hopefully those are resolved as well.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,7 +65,7 @@ jobs:
           - ubuntu-latest
           - ubuntu-24.04-arm
           - macos-latest
-          # - windows-latest <-- failing with fiddle error, temporarily disabled
+          - windows-latest
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v5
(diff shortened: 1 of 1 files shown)


Earlopain:
This is still a problem:
> D:/rubyinstaller-head-x64/lib/ruby/3.5.0+4/rubygems/specification.rb:1419:in
'block in Gem::Specification#activate_dependencies': Could not find 'fiddle' (~>
1.0) among 61 total gem(s) (Gem::MissingSpecError)
> Checked in 'GEM_PATH=D:/a/prism/prism/vendor/bundle/ruby/3.5.0+4' at:
D:/rubyinstaller-head-x64/lib/ruby/gems/3.5.0+4/specifications/win32-registry-0.1.0.gemspec,
execute `gem env` for more information

So I guess ruby itself requires fiddle to get some value from the registry.
https://github.com/ruby/ruby/blob/master/ext/win32/win32-registry.gemspec has fiddle, but that is just gone now. Seems like a problem, I will see if this should be reported upstream.

Earlopain:
Best I can do is https://bugs.ruby-lang.org/issues/21645. It's hard since it's windows only but I think there is definitly something wrong in ruby itself.

Earlopain:
Tried adding fiddle to the gemfile but that wouldn't even work (apart from currently missing headers) since that happens too late apparently.

kddnewton:
No worries, I think they're taking a look at it upstream

kddnewton (approved):

Earlopain:
(The issue is stil there in some form but this was good to merge, yeah)

PR #3690, opened 2025-10-22
Allow compiling from source on windows:

Edouard-chin:
👋 

I'm trying to compile Prism from source on a Windows machine but it unforutnately fails when trying to render the template due to the shebang that doesn't exist on Windows 
https://github.com/ruby/prism/blob/2e577c0e95345fa3866dcb626138dcc77635791f/templates/template.rb#L1

I modified the `system` call to call ruby with its full path directly instead.

### DIFF
--- a/ext/prism/extconf.rb
+++ b/ext/prism/extconf.rb
@@ -37,7 +37,7 @@ end
 def generate_templates
   Dir.chdir(File.expand_path("../..", __dir__)) do
     if !File.exist?("include/prism/ast.h") && Dir.exist?(".git")
-      system("templates/template.rb", exception: true)
+      system(RbConfig.ruby, "templates/template.rb", exception: true)
     end
   end
 end


ext/prism/extconf.rb
```diff
@@ -37,7 +37,7 @@
 def generate_templates
   Dir.chdir(File.expand_path("../..", __dir__)) do
     if !File.exist?("include/prism/ast.h") && Dir.exist?(".git")
-      system("templates/template.rb", exception: true)
+      system(Gem.ruby, "templates/template.rb", exception: true)
```
kddnewton: I _think_ this is okay, but I just want to make sure there's not an environment when `Gem` isn't available. @tenderlove do you know if that could possibly be the case?

Since we know we already have `RbConfig` because of the explicit require, maybe safer to just implement it ourselves:

```suggestion
      ruby = File.join(RbConfig::CONFIG["bindir"], "#{RbConfig::CONFIG["ruby_install_name"]}#{RbConfig::CONFIG["EXEEXT"]}")
      system(ruby, "templates/template.rb", exception: true)
```

tenderlove: Using `Gem` should be fine.  The only time it won't be available is if someone does `--disable-gems`, but since we're using RubyGems to install the gem, then I think it's OK.

Edouard-chin:
Thanks for the review folks. Let me know if you'd like to go with the custom suggested solution or just `Gem.ruby` which is simpler

ext/prism/extconf.rb
eregon: There is `RbConfig.ruby`, that seems safer to use and is just as expressive.

eregon: The user could have `RUBYOPT=--disable-gems` in env, then `Gem.ruby` here would likely break (probably many other things too, but that's out of scope), hence `RbConfig.ruby` is better.

eregon:
@Edouard-chin Can you change to `RbConfig.ruby`? Then I think it's good to go.

PR #3691, opened 2025-10-22
Test against parse.y

Earlopain:
https://github.com/ruby/prism/commit/17a6a19bbae5c8b438a94816ed67c3852547d859 broke ruby/ruby CI because some tests are only run against parse.y
This should help catch stuff like that

### DIFF
--- a/.github/workflows/cruby-bindings.yml
+++ b/.github/workflows/cruby-bindings.yml
@@ -12,6 +12,14 @@ on:
 
 jobs:
   test-all:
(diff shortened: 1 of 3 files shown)


Earlopain:
Here's a failed workflow run without the test changes: https://github.com/ruby/prism/actions/runs/18722960567/job/53400112398

@kddnewton this should be enough to get ruby/ruby green again.

kddnewton:
Thank you!

PR #3697, opened 2025-10-30
Reduce schedule of dependabot updates to be monthly

Earlopain:
I don't think we need them that frequently. For ruby, it's almost always just sorbet, which follows no sensible versioning, makes multiple releases per day and provides no changelog

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
   - package-ecosystem: 'github-actions'
     directory: '/'
     schedule:
-      interval: 'weekly'
+      interval: 'monthly'
     groups:
       action-deps:
         patterns:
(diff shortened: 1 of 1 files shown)


kddnewton:
I totally get the sentiment, but I like finding the issues early. For example, Sorbet + tapioca just broke in the last week, and it's not hard to narrow it down to the right commit since we do merge so frequently. So I'd like to keep it as is for now.

Earlopain:
Can't say I really agree. Tapioca broke because the required ruby version is set to 2.7 and so dependabot hasn't updated tapioca in a while now already (0.17 is Ruby 3.2 only and dependabot resolves based on the constraint in the gemspec, there are also resolve issues because `spoom` pins 4.0.0.dev of `rbs`)

eregon:
My 2 cents on this is I think this mostly just reveals upstream issues but probably nothing to fix in Prism (e.g. the types are fine), so there is no much point to test this so often in Prism, hence the result is I think mostly spamming people following the Prism repo/PRs with lots of rather meaningless updates is not great (FWIW I have tried to filter them out in GMail but failed, it's non-trivial).

Earlopain:
It's about the notifications for me as well, yeah. I do wish github would provide something for this but doesn't look like they care very much

PR #3699, opened 2025-11-01
Replace `serde_yaml` with `serde_json`

Earlopain:
`serde_yaml` was deprecated more than a year ago. See https://github.com/dtolnay/serde-yaml

Didn't find a nice replacement that is also typed. So why not just convert it to json?


kddnewton (approved):
Makes sense

PR #3700, opened 2025-11-01
Reject `def f a, (b) = 1`

Earlopain:
Fixes [[#Bug 21660]](https://bugs.ruby-lang.org/issues/21660), followup to https://github.com/ruby/prism/pull/3674

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -14622,18 +14622,6 @@ update_parameter_state(pm_parser_t *parser, pm_token_t *token, pm_parameters_ord
 }
 
 /**
- * Ensures that after parsing a parameter, the next token is not `=`.
- * Some parameters like `def(* = 1)` cannot become optional. When no parens
- * are present like in `def * = 1`, this creates ambiguity with endless method definitions.
- */
-static inline void
-refute_optional_parameter(pm_parser_t *parser) {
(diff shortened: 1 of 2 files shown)


tompng:
Instead of rejecting at every possible token/node that appears before EQUAL of syntax-invalid endless def, How about simply check at `accept1(parse, PM_TOKEN_EQUAL)` like this?

```c
case PM_TOKEN_KEYWORD_DEF: {
  ...
  bool accept_endless_def;
  switch (parser->current.type) {
    case PM_TOKEN_PARENTHESIS_LEFT: { // `def f(a)`, `def f(*a)`, etc
      accept_endless_def = true;
      ...
    }
    case PM_CASE_PARAMETER: { // `def f a`, `def f *a`, `def f a, (b)`, etc
      accept_endless_def = false;
      ...
    }
    default: { // def f 
      accept_endless_def = true;
      ...
    }
  }

  if (accept1(parser, PM_TOKEN_EQUAL)) {
    // `def f(PARAMETERS) =` and `def f =` can be endless method definition
    // Raise error for `def f [PARAMETERS] =`
  }
}
```

Earlopain:
No clue why I made it so complicated, your suggestion is much nicer. I applied it

PR #3707, opened 2025-11-06
Reject `END { break }` for Ruby 3.5

Earlopain:
For [[Bug #20409]](https://bugs.ruby-lang.org/issues/20409)


src/prism.c
```diff
@@ -15790,10 +15789,21 @@ parse_block_exit(pm_parser_t *parser, pm_node_t *node) {
             case PM_CONTEXT_DEF_RESCUE:
             case PM_CONTEXT_MAIN:
             case PM_CONTEXT_PREEXE:
+            case PM_CONTEXT_POSTEXE:
             case PM_CONTEXT_SCLASS:
             case PM_CONTEXT_SCLASS_ELSE:
             case PM_CONTEXT_SCLASS_ENSURE:
             case PM_CONTEXT_SCLASS_RESCUE:
+                // https://bugs.ruby-lang.org/issues/20409
+                if (context_node->context == PM_CONTEXT_POSTEXE) {
```
kddnewton: Slight preference here for this to be a fallthrough instead of checking again the type here, as in:

```
case PM_CONTEXT_POSTEXE:
  if (...) { return; }
  if (...) { return; }
PRISM_FALLTHROUGH
case ...:
  assert(...);
```

kddnewton (approved):

src/prism.c
Earlopain: Works for me, I changed it

kddnewton:
Do we know if this is mergable yet?

Earlopain:
I think this was OKed in a dev meeting. I will check and clean this up

Earlopain:
https://github.com/ruby/dev-meeting-log/blob/f76ba7e3f69feaff70672f0077976d7f63313601/2025/DevMeeting-2025-11-13.md#bug-20409-missing-reporting-some-invalid-breaks-kddnewton

I think this is accepted, `END { break }` at least? Can't tell for sure. For reference, I tested again how it currently behaves:

```
END
return - runtime error
break - runtime error
next - ok

BEGIN
return - ok
break - syntax error
next - syntax error
```

Currently I have return in end still accepted but it also leads to a runtime error. I guess this can be merged as is after rebase, what do you think?

kddnewton:
Yeah I think this can be merged as is and we can roll with it.

kddnewton:
Actually sorry I read through the meeting notes and I think this is still not resolved.

kddnewton:
@Earlopain matz said this is good to go for 4.1, can you rebase?

kddnewton (approved):

PR #3709, opened 2025-11-07
Ruby 4.0

Earlopain:
See https://github.com/ruby/ruby/commit/6d81969b475262aba251e99b518181bdf7c5a523
    
It leaves the old variant around. RuboCop for examples accesses `Prism::Translation::Parser35` to test against ruby-head. For now I left these simply as an alias


include/prism/options.h
```diff
@@ -91,11 +91,14 @@ typedef enum {
     /** The vendored version of prism in CRuby 3.4.x. */
     PM_OPTIONS_VERSION_CRUBY_3_4 = 2,
 
-    /** The vendored version of prism in CRuby 3.5.x. */
+    /** The vendored version of prism in CRuby 4.0.x. */
     PM_OPTIONS_VERSION_CRUBY_3_5 = 3,
 
+    /** The vendored version of prism in CRuby 4.0.x. */
+    PM_OPTIONS_VERSION_CRUBY_4_0 = 3,
```
Earlopain, line 98: No need to use a different number here than for "3.5". They are the same after all.

java/org/prism/ParsingOptions.java
```diff
@@ -15,7 +15,8 @@ public enum SyntaxVersion {
         LATEST(0), // Handled in pm_parser_init
         V3_3(1),
         V3_4(2),
-        V3_5(3);
+        V3_5(3),
```
Earlopain, line 18: I have a use-case for keeping the 3.5 parser translator. While I can't say the same for the other places, I'd keep them all around for now.

In a year or two I'd simply remove them.

PR #3715, opened 2025-11-14
Move gemfiles for 3.5 to 4.0

nobu:


Earlopain:
We have to wait for https://github.com/ruby/setup-ruby/issues/824 to be resolved. Currently windows is on 4.0 while the others are stuck with some outdated version of 3.5

If you know someone that is able to fix/merge the PR for that, that would be great (; I want to test against a relatively recent dev build again

kddnewton:
This is fine, I just relaxed the restriction to be >= 3.5.0.dev. We can update it again later when things are working.

PR #3720, opened 2025-11-14
Add docs for super nodes

amomchilov:


config.yml
```diff
@@ -2631,6 +2631,14 @@ nodes:
 
           super
           ^^^^^
+
+      A ForwardingSuperNode can still have a block.
+      All other arguments are forwarded as normal, except the original block is replaced with the new block.
```
amomchilov: This is pretty surprising, worth documenting explicitly.

```ruby
#!/opt/rubies/3.4.2/bin/ruby

class Parent
  def m(arg)
    p [arg, yield]
  end
end

class Child < Parent
  def m(arg)
    super { "different block" }
  end
end

# prints ["original arg", "different block"]
Child.new.m("original arg") { "original block" }
```

config.yml
```diff
@@ -4506,6 +4514,7 @@ nodes:
       - name: arguments
         type: node?
         kind: ArgumentsNode
+        comment: "Can be `nil` when there are empty parentheses, like `super()`."
```
amomchilov: Only? Any other case and it would be `ForwardingSuperNode` instead, right?

```suggestion
        comment: "Can be `nil` only when there are empty parentheses, like `super()`."
```

amomchilov:
Test failure looks unrelated:

```
    1) Failure:
  TestPatternMatching#test_alternative_pattern [/home/runner/work/prism/prism/ruby/ruby/test/ruby/test_pattern_matching.rb:207]:
  pid 116284 exit 0.
  
  1. [1/3] Assertion for "stdout"
     | <[]> expected but was
     | <["Syntax OK"]>.
  
  2. [2/3] Assertion for "stderr"
     | Expected /alternative pattern after variable capture/ to match "".
  
  3. [3/3] Assertion for "success?"
     | Expected #<Process::Status: pid 116284 exit 0> to not be success?.
```

config.yml
kddnewton: Can you move this comment to the comment on the `block` field itself instead?

config.yml
kddnewton: That's correct

kddnewton (changes requested):

config.yml
amomchilov: I moved the "All other arguments are forwarded as normal" part, which is what I think you were referring to.

I'd like to keep the "can still have a block." part here, so that it:

1. Is near where we say "without parentheses or arguments"
2. Show the location of the whole node when it has a block
3. Has a link to `SuperNode`, which readers would probably also be interested in.

I rejigged it a bit, what do you think?

PR #3724, opened 2025-11-17
Remove `bin/prism encoding`

Earlopain:
I don't think there will be many more encodings added to ruby at this time


kddnewton:
I was keeping this around for when they update the Unicode version and we need to verify that it's correct. (Or if we find out there's a bug in one of our encodings for some reason.) I don't think it's hurting to keep it around.

Earlopain:
They did update the unicode version very recently in https://github.com/ruby/ruby/pull/15170

PR #3728, opened 2025-11-18
[Bug #21669] Void value expressions

Earlopain:
For [[Bug #21669]](https://bugs.ruby-lang.org/issues/21669)

### DIFF
--- a/snapshots/3.3-4.0/void_value.txt
+++ b/snapshots/3.3-4.0/void_value.txt
@@ -1,0 +1,247 @@
+@ ProgramNode (location: (1,0)-(29,3))
+├── flags: ∅
+├── locals: [:x]
+└── statements:
+    @ StatementsNode (location: (1,0)-(29,3))
+    ├── flags: ∅
+    └── body: (length: 5)
+        ├── @ LocalVariableWriteNode (location: (1,0)-(7,3))
+        │   ├── flags: newline
+        │   ├── name: :x
+        │   ├── depth: 0
+        │   ├── name_loc: (1,0)-(1,1) = "x"
+        │   ├── value:
+        │   │   @ BeginNode (location: (1,4)-(7,3))
+        │   │   ├── flags: ∅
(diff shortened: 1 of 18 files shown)


kddnewton (approved):

kddnewton:
Where are we with this? Did nobu's upstream changes get merged?

Earlopain:
(PR ref https://github.com/ruby/ruby/pull/15498)

Not merged, no. He fixed a case I forgot to implement (https://github.com/ruby/ruby/pull/15498/commits/74ea62b99ebbaf68343224ba56e3e21892a4cd0c) but in some edge case prism doesn't emit unused variable warnings anymore.

I haven't looked into it though. Should I just continue in this PR here, or?

kddnewton:
Yeah I think just continue with this PR then, he'll merge whatever he needs to after.

Earlopain:
I think I got it, was a pretty simple fix. I will confirm later and then open a PR in ruby/ruby with these changes + nobus ones since it requires test changes in ruby/ruby. If something more needs to be done he can just push to my PR.

I rebased so it's messy but these are the only lines I touched https://github.com/ruby/prism/pull/3728/changes#diff-18342eb08afd7802e20869c1594534ce6eb1470aef86345472f703ac57b0bdfeR1191-R1207 (plus tests and version check changes)

Earlopain:
The problem was using `pm_check_value_expression` in that loop, it marks variable writes as read so that they don't emit warnings. The solution is to just not use that function, we can just check for void nodes directly.

Earlopain:
https://github.com/ruby/ruby/pull/15999. From the prism side this should be good/identical to this pr and I just copied the parse.y changes from nobu.

Earlopain:
I dropped the parse.y changes from the ruby pr. Nobu rebased against the changes I made so his PR should be good now. To be honest, a bit frustrating to do so without saying anything. So either my PR can be merged and nobu rebases or his PR gets merged and I just close mine.

Earlopain:
https://github.com/ruby/ruby/pull/15999 got merged (thanks!)

PR #3732, opened 2025-11-19
Prevent an infinite loop parsing a capture name

### LINKED ISSUE #3729: Infinite loop/memory exhaustion parsing a rexexp named capture with a malformed unicode escape
Example input foo.txt:
```foo.txt
/(?<\u{21!3}>foo)/ =~ "foo"
```
causes 
```
./bin/parse foo.txt
```
to get stuck in `pm_named_capture_escape_unicode` using CPU and allocating memory on each loop.

Similar results for entering the code into irb etc.

I've patched locally with
```diff
diff --git a/src/prism.c b/src/prism.c
--- a/src/prism.c
+++ b/src/prism.c
@@ -21163,6 +21163,9 @@ pm_named_capture_escape_unicode(pm_parser_t *parser, pm_buffer_t *unescaped, con
         }

         size_t length = pm_strspn_hexadecimal_digit(cursor, end - cursor);
+        if (length == 0) {
+            break;
+        }
         uint32_t value = escape_unicode(parser, cursor, length);
```
and the immediate problem is fixed. Stopped short of a pull request because there may be a diagnostic error that's required in this case?

stevenjohnstone:
Fixes #3729.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -20395,6 +20395,9 @@ pm_named_capture_escape_unicode(pm_parser_t *parser, pm_buffer_t *unescaped, con
         }
 
         size_t length = pm_strspn_hexadecimal_digit(cursor, end - cursor);
+        if (length == 0) {
+            break;
+        }
         uint32_t value = escape_unicode(parser, cursor, length);
 
         (void) pm_buffer_append_unicode_codepoint(unescaped, value);


Earlopain:
Thanks, can you add a test for this? In `test/prism/fixtures` create `regex_capture.txt` or similar that contains `/(?<\u{21!3}>foo)/ =~ "foo"` and `/(?<\u{21!3}>foo)/`. When you run tests with `UPDATE_SNAPSHOTS=1 bundle exec rake`, a snapshot file gets created that you should commit.

(I guess some other parsers we test against may correctly reject this code. If that is the case, simply excluding this new file for them is fine)

Issue #2667, opened 2024-04-02
Regexp fixed encoding should raise SyntaxError if source encoding doesn't match a fixed one


andrykonchin:
I just noticed this behaviour of CRuby and it seems Parser doesn't support it right now:

```ruby
 p /Ȃ/e.encoding
# regexp encoding option 'e' differs from source encoding 'UTF-8' (SyntaxError)
```

thdaraujo:
I'd love to take a stab at this! 🙏

kddnewton:
> I'd love to take a stab at this! 🙏 

Please do!

PR #3735, opened 2025-11-20
Add tests to `regexp_encoding_option_mismatch`

thdaraujo:
(related to #2667)

The error can be seen in CRuby with `parse.y` enabled:
```bash
ruby --parser=parse.y -e "p /Ȃ/e.encoding"

#=> -e:1: regexp encoding option 'e' differs from source encoding 'UTF-8'
ruby: compile error (SyntaxError)
```

It could also be seen on Ruby 3.4.7 with Prism enabled:
```bash
ruby -e "p /Ȃ/e.encoding"
-e: -e:1: syntax error found (SyntaxError)
#=> > 1 | p /Ȃ/e.encoding
        |       ^ regexp encoding option 'e' differs from source encoding 'UTF-8'
```

And by parsing directly, you get the error too:
```
bin/parse -e "p /Ȃ/e.encoding"

Errors:
[#<Prism::ParseError @type=:regexp_encoding_option_mismatch @message="regexp encoding option 'e' differs from source encoding 'UTF-8'" @location=#<Prism::Location @start_offset=7 @length=1 start_line=1> @level=:syntax>]
```

Looking at the [CRuby](https://github.com/ruby/ruby/blob/4c7525082c353bfa717cb17f9ae75d940cb69d20/parse.y#L15305-L15337) and [Prism](https://github.com/ruby/prism/blob/f2426824aafaac59cafa0342e1e232d63384afd1/src/prism.c#L7255-L7256) implementations, both are doing the same checks.

But the CRuby code is only enabled for Ripper (with the `#ifndef RIPPER`).

So maybe the problem is that Ruby + Prism  is raising the Syntax error, but it should only be raised when using Ripper, and not during normal execution? But that would contradict the original issue description, which expects the syntax error to be raised.

I might be missing some context here about what the right behaviour should be, but I went ahead and added some tests to make sure the current behavior is the expected one. Hope this makes sense! 

Thanks!

### DIFF
--- a/test/prism/errors_test.rb
+++ b/test/prism/errors_test.rb
@@ -64,6 +64,24 @@ module Prism
       assert_equal :"", Prism.parse_statement("+.@foo,+=foo").write_name
     end
 
+    def test_regexp_encoding_option_mismatch_error
+      # UTF-8 char with ASCII-8BIT modifier
+      result = Prism.parse('/Ȃ/n')
+      assert_includes result.errors.map(&:type), :regexp_encoding_option_mismatch
+
+      # UTF-8 char with EUC-JP modifier
+      result = Prism.parse('/Ȃ/e')
+      assert_includes result.errors.map(&:type), :regexp_encoding_option_mismatch
+
+      # UTF-8 char with Windows-31J modifier
+      result = Prism.parse('/Ȃ/s')
+      assert_includes result.errors.map(&:type), :regexp_encoding_option_mismatch
+
+      # UTF-8 char with UTF-8 modifier
+      result = Prism.parse('/Ȃ/u')
+      assert_empty result.errors
+    end
+
     private
 
     def assert_errors(filepath, version)


test/prism/errors_test.rb
```diff
@@ -64,6 +64,24 @@ def test_invalid_message_name
...
+
+      # UTF-8 char with Windows-31J modifier
+      result = Prism.parse('/Ȃ/s')
+      assert_includes result.errors.map(&:type), :regexp_encoding_option_mismatch
+
+      # UTF-8 char with UTF-8 modifier
+      result = Prism.parse('/Ȃ/u')
+      assert_empty result.errors
+    end
+
```
thdaraujo, line 84: should this test be added to `prism/encoding/regular_expression_encoding_test.rb` instead?

Earlopain:
Looks like this may have already been implemented in https://github.com/ruby/prism/pull/2486. It's a bit weird since the issue was openend a month after that PR got merged.

Ripper shouldn't matter here, what's important is that `ruby -c test.rb` and `ruby --parser=parse.y -c test.rb` report the same thing (I know very little about how the original parser is actually implemented). prism and parse.y seem to agree, so all seems good?

thdaraujo:
@Earlopain yeah that's what I was thinking too. Maybe we can close the issue.

Do we care about adding these tests, or should I just close this PR?

Thanks!

test/prism/errors_test.rb
kddnewton: It's fine here.

kddnewton:
Hmm... it would appear it's as you say, and the original can be closed

PR #3743, opened 2025-11-24
Reject `p(p a, &block => value)` and similar - try 2

Earlopain:
Redo of https://github.com/ruby/prism/pull/3669 with more tests

Not ready yet, I want to see about adding more checks in CI first


Earlopain:
I used the new command I added and compared against all of rubygems. It didn't find any difference so I think this is ready now. With the previous version there were 107 gems that went syntax invalid (including `shipit-engine` caught by ruby CI)

PR #3747, opened 2025-11-25
Add `bin/compare`

Earlopain:
Allows to compare the output of two different prism versions.

It compiles prism twice and forks to allow for different versions in the same script. It then passes over minimal data to see if anything has changed.

Most bugfixes should impact little to no real code and the test suite is already very extensive. Running this can give you even more confidence by comparing against real-world-rails or similar.

There are some performance gains to be had here. Basically it is already parallelized because of `fork` but it can be even better. For simplicity (and because I don't usually write such code) I leave that as an exercise for the future. The message on change can also be improved, but right now I'm not certain that's worth the effort. Just check it manually via the already existing tools.

Would have shown that https://github.com/ruby/prism/pull/3669 isn't quite right.

### DIFF
--- a/bin/compare
+++ b/bin/compare
@@ -1,0 +1,108 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# Usage: bin/compare main feature-branch .
+
+$:.unshift(File.expand_path("../lib", __dir__))
+
+require "json"
+require "socket"
+
+def create_prism(ref)
+  parent_socket, child_socket = UNIXSocket.pair
+
+  system("git checkout #{ref}", exception: true)
+  system("bundle exec rake compile", exception: true)
+
+  pid = fork do
+    parent_socket.close
+    require "prism"
+
+    child_socket.puts("Compiling done for #{ref}")
+
+    while (data = child_socket.gets(chomp: true))
+      command, path = data.split("\x00", 2)
+      case command
+      when "dump"
+        begin
+          child_socket.puts(Prism.dump_file(path).hash)
+        rescue Errno::EISDIR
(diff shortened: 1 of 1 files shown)


bin/compare
```diff
@@ -0,0 +1,74 @@
...
+# frozen_string_literal: true
+
+# Usage: bin/compare main feature-branch .
+
+$:.unshift(File.expand_path("../lib", __dir__))
+
+require "json"
+require 'socket'
+
+SEP = "123sock-sep321"
```
Earlopain: Seems a bit weird. Is this how you usually do this?

Earlopain: Nevermind, that was from when I was passing around the whole file content. I can just use the default newline now.

Earlopain:
I added a custom visitor so that this runs faster. It gets me from ~570 down to ~98 seconds for real-world-rails (~138k source files).

I looked to optimize the inspect visitor but doubt it would get close to a 6x improvement. That said, maybe there is some other approach that I missed? It just needs to produce the same output given the same ast and doesn't need to care much about collisions since it's more like a parity check. I tried using digest to avoid appending to the string each time but that turned out to be slower.

Earlopain:
Damn, wish I thought of that. It's not 100% equivalent (`x("foo", "bar")` and `x("bar", "foo")` for example produce the same dump output) but for the purpose of this check that doesn't really matter in my eyes. I'm more interested in if nodes trade places, which this still can do no problem.

It now checks real-world-rails in just 12 seconds for me (ignoring compilation) which is pretty impressive.

kddnewton (approved):

PR #3758, opened 2025-11-26
Optimize `Prism::Source#find_line`

Earlopain:
This is more concise and ruby does a better job performance-wise.

This used to be `bsearch_index` already but https://github.com/ruby/prism/commit/6d8358c08395438d5924777c1fc3001a5ebf0aa3 changed it. https://github.com/ruby/prism/pull/1733#discussion_r1373702087 said:
> Yeah the edge case was that the value matched an element exactly

But surely there would be a test to show this behaviour?

Gets called as part of pretty-printing nodes.
Further reduces the time for `SnapshotsTest` by ~16% for me.

### DIFF
--- a/lib/prism/parse_result.rb
+++ b/lib/prism/parse_result.rb
@@ -155,21 +155,8 @@ module Prism
     # Binary search through the offsets to find the line number for the given
     # byte offset.
     def find_line(byte_offset)
-      left = 0
-      right = offsets.length - 1
-
-      while left <= right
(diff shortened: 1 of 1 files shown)


kddnewton:
If I recall correctly, it was getting the wrong line number when the bounds of the node hit the newline exactly. I _think_ maybe heredocs were the issue? I'm surprised to see none of the snapshots change though. I'm going to optimistically merge this and hope things are fine assuming this is good.

PR #3761, opened 2025-11-28
Fix invalid Ruby code example in ClassNode comment

ryumasai:
Added a semicolon to make it valid Ruby code.
```diff
- class Foo end
+ class Foo; end
```


eregon (approved):

eregon:
Thanks!

kddnewton:
That syntax was valid! The point was to avoid semicolons if we could in the file, so reverting.

PR #3769, opened 2025-11-29
Update unicode tables to match that of CRuby

kddnewton:
Per https://github.com/ruby/prism/pull/3724#issuecomment-3542614900, thanks @Earlopain 

The unicode version has been updated upstream, which means new codepoints mapped to alpha/alnum/isupper flags. We need to update our tables to match.

I'm purposefully not adding a version check here, since that is such a large amount of code. It's possible that we could include different tables depending on a macro (like UNICODE_VERSION) or something to that effect, but it's such a minimal impact on the running of the actual parser that I don't think it's necessary.


Earlopain:
Ah, I get it now.

For reference, the `parser` gem doesn't bother with that and just uses the regex from the runtime ruby (`/[[:upper:]]/` and similar). So basically the same as this and I don't think anyone ever complained.

PR #3772, opened 2025-12-01
Introduce using slices instead of locations

### LINKED ISSUE #1566: consider using 32-bit offsets for start and end in `yp_location_t`
YARP provides precise location information for any number of things in its AST, which is great!  But each of those locations on a 64-bit platform costs 16 bytes, which means every YARP node (`yp_node_t`) on those platforms is at least 24 bytes, which is rather large.

For some concrete numbers, nearly half (~47%) of the memory size as measured by `YARP::Debug.memsize` on a subset of Stripe's codebase is consumed by locations.  `YARP::Debug.memsize` doesn't measure this as-is; the count(s) of locations were obtained from this branch (and note whatever the branch counts needs to be multiplied by 16 to obtain the size):

https://github.com/ruby/yarp/compare/main...froydnj:froydnj-location-counting

so I might have done something wrong.

Using 32-bit offsets (well, probably 31-bit offsets so there's still some way to mark locations as "invalid" or "unavailable") would immediately shrink the memory required by ~25%; I haven't put together a proof-of-concept branch with that change, but my belief is that there would be a performance gain as well, though probably not of the same magnitude.  One can also imagine more complicated schemes where nodes hold a 32-bit key into some side table, which would shrink things still more, but one step at a time.

This change would limit the kinds of Ruby source files that YARP could consume.  But other Ruby-consuming tools/implementations (Sorbet and--I think--TruffleRuby cc @eregon ) already maintain 32-bit offsets internally.  I believe Clang and Go also only store 32-bit offsets for their locations.  If Clang and Go can get away with it, I think YARP can too.

kddnewton:
In the C API, we want to use slices instead of locations in the AST. In this case a "slice" is effectively the same thing as the location, expect it is represented using a 32-bit offset and a 32-bit length. This will cut down on half of the space of all of the locations in the AST.

I am introducing this as a new field type to ease the migration path, so that this can be merged on its own and then fields can be moved one at a time.

Note that from the Ruby/Java/JavaScript side, this is effectively an invisible change. This only impacts the C/Rust side.

Fixes #1566

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -653,10 +653,6 @@ tokens:
     comment: "a separator between words in a list"
   - name: __END__
     comment: "marker for the point in the file at which the parser should stop"
-  - name: MISSING
-    comment: "a token that was expected but not found"
-  - name: NOT_PROVIDED
-    comment: "a token that was not present but it is okay"
 flags:
   - name: ArgumentsNodeFlags
     values:
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -61,7 +61,7 @@ The available values for `type` are:
 * `constant` - A field that is an integer that represents an index in the constant pool. This is a `pm_constant_id_t` in C.
 * `constant[]` - A field that is an array of constants. This is a `pm_constant_id_list_t` in C.
 * `location` - A field that is a location. This is a `pm_location_t` in C.
-* `location?` - A field that is a location that is optionally present. This is a `pm_location_t` in C, but if the value is not present then the `start` and `end` fields will be `NULL`.
+* `location?` - A field that is a location that is optionally present. This is a `pm_location_t` in C, but if the value is not present then the `length` field will be `0`.
 * `uint8` - A field that is an 8-bit unsigned integer. This is a `uint8_t` in C.
 * `uint32` - A field that is a 32-bit unsigned integer. This is a `uint32_t` in C.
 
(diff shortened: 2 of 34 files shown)


Earlopain:
> I am introducing this as a new field type to ease the migration path, so that this can be merged on its own and then fields can be moved one at a time.

Can you explain this a bit? To me it seems most convenient to have it done relatively quickly so that consumers don't have to bother with both variants.

Eventually this needs changes in ruby/ruby to compile which will make backporting prism versions a bit annoying, so it would be nice if its not spread across many commits in ruby/ruby I think

froydnj:
(For posterity, linking this to #1566 .)

kddnewton:
> Can you explain this a bit? To me it seems most convenient to have it done relatively quickly so that consumers don't have to bother with both variants.

Yeah this is a good point. Let's just keep it around for the lifetime of this PR while things get migrated, and then we can squash it all before merge.

kddnewton:
Just to put this explicitly: not merging this until after Christmas/Ruby 4.0 is released.

Earlopain:
Continues in https://github.com/ruby/ruby/pull/15979 for the CRuby changes

eregon:
Very nice to see this done, I recall trying and basically failing, it was lots of non-trivial work.

PR #3774, opened 2025-12-02
Clean up test excludes

Earlopain:
Mostly not having to list version-specific excludes when testing against ripper/parse.y

I also removed excludes that have since been implemented by parse.y


Earlopain:
The first commit fixes an issue where the ripper translator parses code with the latest version instead of the currently running version. Found while cleaning up the excludes.

Also openend https://bugs.ruby-lang.org/issues/21756. The rest are as far as I can tell something that should be changed in prism.

PR #3777, opened 2025-12-03
Set "trailing" on the comment node

tenderlove:
Comment nodes can know if they are trailing comments by looking at the previous token.  If we set this on the comment object then we don't need to look at source slices when attaching comments


kddnewton:
(Closing per our convo)

PR #3778, opened 2025-12-03
Percent array heredoc continuation

Earlopain:
See https://bugs.ruby-lang.org/issues/21756. Ripper fails to parse this but prism actually also doesn't handle it correctly.

When heredocs are used, even in lowercase percent arays there can be
multiple `STRING_CONTENT` tokens. We need to concat them.

Luckily we don't need to handle as many cases as in uppercase arrays where interpolation is allowed. Mostly inspired by how uppercase arrays are handled.

`%q` strings seem to work ok already.

Second commit is a pretty small fix I noticed while I was poking around. Probably a copy-paste error.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -19343,7 +19343,7 @@ parse_expression_prefix(pm_parser_t *parser, pm_binding_power_t binding_power, b
                     pm_array_node_elements_append(array, current);
                     current = NULL;
                 } else {
-                    expect1(parser, PM_TOKEN_STRING_CONTENT, PM_ERR_LIST_W_LOWER_ELEMENT);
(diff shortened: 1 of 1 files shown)


src/prism.c
```diff
@@ -19489,23 +19523,42 @@ parse_expression_prefix(pm_parser_t *parser, pm_binding_power_t binding_power, b
             parser_lex(parser);
             pm_token_t opening = parser->previous;
             pm_array_node_t *array = pm_array_node_create(parser, &opening);
-
-            // skip all leading whitespaces
-            accept1(parser, PM_TOKEN_WORDS_SEP);
```
Earlopain, line 19494: This was not needed, the loop already consumsed the separator. Removed because it was already not present in %i handling

kddnewton (approved):
Looks right to me!

PR #3780, opened 2025-12-04
Fix `%Q` with newline delimiter and heredoc interpolation

Earlopain:
The lexer did not jump to the `heredoc_end`, causing the heredoc end delimiter to be parsed twice.

Normally the heredocs get flushed when a newline is encountered. But because the newline is part of the string delimiter, that codepath is not taken.

Fixes [[Bug #21758]](https://bugs.ruby-lang.org/issues/21758)

Actually also depends on https://github.com/ruby/prism/issues/3771. While the issue was only about syntax-invalid code, this would also hit the assertion. Quite fortunate (for me) that this was not a problem anymore (;


test/prism/fixtures/heredoc_percent_q_newline_delimiter.txt
```diff
@@ -0,0 +1,22 @@
...
+#{<<B}
+B
+
+<<A; %
+A
+#{<<B}
+B
+
+# \r\n
+%Q
```
Earlopain, line 20: Seems to just work without special handling, see snapshot for confirmation

Earlopain:
parse.y failure seems unrelated. I saw the same while I was syncing my fork. Might be fixed by https://github.com/ruby/ruby/pull/15393, that is not in ruby-dev yet.

Earlopain:
Gee, I wonder why 😅. Does seem like a bit of a waste of time to fix but I got to poke around in an area which I didn't really pay much attention to before. Your [heredoc documentation](https://github.com/ruby/prism/blob/main/docs/heredocs.md) was pretty helpful

PR #3800, opened 2025-12-12
Define RubyParser::SyntaxError so we don't have to require ruby_parser anymore

zenspider:
I'd rather there be no require whatesoever as the SyntaxError class
there is basic af... but this guard should do the trick.

### DIFF
--- a/lib/prism/translation/ruby_parser.rb
+++ b/lib/prism/translation/ruby_parser.rb
@@ -2,10 +2,15 @@
 # :markup: markdown
 
 begin
-  require "ruby_parser"
+  require "sexp"
 rescue LoadError
-  warn(%q{Error: Unable to load ruby_parser. Add `gem "ruby_parser"` to your Gemfile.})
+  warn(%q{Error: Unable to load sexp. Add `gem "sexp"` to your Gemfile.})
   exit(1)
+end
+
+class RubyParser # :nodoc:
+  class SyntaxError < RuntimeError # :nodoc:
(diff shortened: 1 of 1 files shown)


zenspider:
I have a failure in rust bindings. That's green on main. Is is flaky or what?

lib/prism/translation/ruby_parser.rb
```diff
@@ -1,11 +1,10 @@
 # frozen_string_literal: true
 # :markup: markdown
 
-begin
-  require "ruby_parser"
-rescue LoadError
-  warn(%q{Error: Unable to load ruby_parser. Add `gem "ruby_parser"` to your Gemfile.})
-  exit(1)
+require "sexp"
```
Earlopain: Let's tweak the begin/rescue instead of removing it. I think I am ok to drop the hard requirement on `ruby_parser` since it only depends on that one exception but the error message for missing `sexp_processor` should still be present.

lib/prism/translation/ruby_parser.rb
```diff
@@ -1,11 +1,10 @@
...
+
+class RubyParser # :nodoc:
+  SyntaxError = Class.new RuntimeError # :nodoc:
```
Earlopain: This avoids a redeclared constant warning if ruby_parser ends up being loaded first anyways.

```suggestion
  class SyntaxError < RuntimeError; end # :nodoc:
```

PR #3808, opened 2025-12-14
 Add Prism.node_for(Method|UnboundMethod|Proc) to get a Prism node for a callable

eregon:
* See https://bugs.ruby-lang.org/issues/21005
  and https://bugs.ruby-lang.org/issues/20999

### DIFF
--- a/lib/prism/parse_result.rb
+++ b/lib/prism/parse_result.rb
@@ -152,15 +152,6 @@ module Prism
       code_units_offset(byte_offset, encoding) - code_units_offset(line_start(byte_offset), encoding)
     end
 
-    # Returns the byte offset for a given line number and column number
-    def line_and_character_column_to_byte_offset(line, column)
-      line_start = line_to_byte_offset(line)
-      line_end = offsets[line + 1 - @start_line] || source.bytesize
-      byteslice = @source.byteslice(line_start, line_end) or raise ArgumentError, "line #{line} is out of range"
-      byte_column = (byteslice[0...column] or raise).bytesize
-      line_start + byte_column
-    end
-
     # Freeze this object and the objects it contains.
     def deep_freeze
       source.freeze
@@ -289,12 +280,6 @@ module Prism
     # essentially the same as `Prism::Source#column`.
     def code_units_column(byte_offset, encoding)
       byte_offset - line_start(byte_offset)
-    end
-
-    # Specialized version of `line_and_character_column_to_byte_offset`
-    # which does not need to access the source String
-    def line_and_character_column_to_byte_offset(line, column)
-      line_to_byte_offset(line) + column
     end
   end
 
--- a/sig/prism/parse_result.rbs
+++ b/sig/prism/parse_result.rbs
@@ -25,7 +25,6 @@ module Prism
     def code_units_offset: (Integer byte_offset, Encoding encoding) -> Integer
     def code_units_cache: (Encoding encoding) -> _CodeUnitsCache
     def code_units_column: (Integer byte_offset, Encoding encoding) -> Integer
-    def line_and_character_column_to_byte_offset: (Integer line, Integer column) -> Integer
     def deep_freeze: () -> void
 
     def self.for: (String source) -> Source
@@ -42,7 +41,6 @@ module Prism
     def code_units_offset: (Integer byte_offset, Encoding encoding) -> Integer
     def code_units_cache: (Encoding encoding) -> _CodeUnitsCache
     def code_units_column: (Integer byte_offset, Encoding encoding) -> Integer
-    def line_and_character_column_to_byte_offset: (Integer line, Integer column) -> Integer
   end
 
   class Location


tompng:
Looks like Proc generated by `for in` syntax is missing

lib/prism/parse_result.rb
```diff
@@ -895,4 +909,46 @@ def initialize(locals, forwarding)
   def self.scope(locals: [], forwarding: [])
     Scope.new(locals, forwarding)
   end
+
+  # Given a Method, UnboundMethod or Proc, use its #source_location to parse the file and return a Prism node representing it.
+  # The returned node will either be a DefNode, LambdaNode or CallNode.
+  # Raises ArgumentError if it cannot be found for any reason.
+  # Only works on Ruby 4+ as it needs #source_location to contain column and end line information.
+  def self.node_for(callable)
```
kddnewton, line 912: Could we have some tests for this?

For Ruby < 4, we could define singleton methods for this, as in:

```ruby
callable = method(:whatever)
if RUBY_VERSION < "4.0"
  def callable.source_location = [whatever]
end
```

that way we can get tests on this for all versions.

lib/prism/parse_result.rb
```diff
@@ -895,4 +909,46 @@ def initialize(locals, forwarding)
...
+
+    parse_result = Prism.parse_file(file)
+    unless parse_result.success?
+      raise ArgumentError, "#{file} has syntax errors: #{parse_result.errors_format}"
+    end
+    root = parse_result.value
+    start_offset = parse_result.source.line_and_column_to_byte_offset(start_line, start_column)
+    end_offset = parse_result.source.line_and_column_to_byte_offset(end_line, end_column)
+
+    found = root.breadth_first_search do |node|
```
kddnewton: We have a method for this that I think we should use here `Node#tunnel`, which we can then iterate through backward. It's just more efficient than breadth_first_search.

lib/prism/parse_result.rb
eregon: Ah I wrote tests but forgot to `git add` them, now they are pushed.

lib/prism/parse_result.rb
eregon: Right, I was wondering if we also have a depth-first search which might be faster.
I'll try `tunnel` and see if it's faster. It seems to allocate a `Location` for each visited node though, while this avoids it. Maybe tunnel could use `line_and_column_to_byte_offset` to avoid allocations and have simpler comparisons.

lib/prism/parse_result.rb
```diff
@@ -895,4 +909,46 @@ def initialize(locals, forwarding)
...
+    source_location = callable.source_location
+    raise ArgumentError, "#source_location is nil for #{callable}" if source_location.nil?
+    raise ArgumentError, '#source_location does not contain column and end_line, this method only works on Ruby 4+' if source_location.size != 5
+    file, start_line, start_column, end_line, end_column = source_location
+
+    unless File.exist?(file)
+      raise ArgumentError, "#source_location[0] is #{file} but this file does not exist"
+    end
+
+    parse_result = Prism.parse_file(file)
```
Earlopain: If we have a method object, it's probably safe to assume that we got that with the ruby version we are currently running as

As a result the added tests will only be runnable on ruby >= 3.3 (since prism doesn't understand previous versions)

```suggestion
    parse_result = Prism.parse_file(file, version: "current")
```

lib/prism/parse_result.rb
eregon: > For Ruby < 4, we could define singleton methods for this, as in:

Done, in a separate commit.
It makes the test a bit more messy but indeed let us run it on all versions.

lib/prism/parse_result.rb
eregon: Good point, fixed. More reasons this should be the default IMO it's quite error-prone to forget this.

eregon: This means the test can only run on Ruby 3.3+ though (since Prism cannot parse Ruby 3.2 and before), but I think that's fine since anyway the method only really works on Ruby 4+.

eregon:
The 2.7 failure is quite annoying: https://github.com/ruby/prism/actions/runs/20211767571/job/58018650394?pr=3808
It can't parse `def inline_method = 42`. We can't wrap that in `eval` otherwise `Prism.node_for` won't work on it.
I suppose we could use a separate file, but it feels messy.
I tried to repro locally but the main Gemfile doesn't even work in 2.7, so it seems difficult to repro & test locally.
I wish we could just drop 2.7, it has this partial kwargs separation which makes it pretty messy in general.

Earlopain:
> I wish we could just drop 2.7, it has this partial kwargs separation which makes it pretty messy in general.

This is something we should probably talk about sometime in the future. It was added for the convenience of rubocop as far as I'm aware and I appreciate that this happened. But I do not want to keep these long EOL versions supported forever.

For now I would just put this in a separate file like you said. You shoudl be able to use the version-specific gemfile with `BUNDLE_GEMFILE=gemfiles/2.7/Gemfile`, that works for me at least.

eregon:
> Looks like Proc generated by `for in` syntax is missing

@tompng Do you mean:
```
$ bin/parse -e 'for e in foo; body; end'
@ ProgramNode (location: (1,0)-(1,23))
├── flags: ∅
├── locals: [:e]
└── statements:
    @ StatementsNode (location: (1,0)-(1,23))
    ├── flags: ∅
    └── body: (length: 1)
        └── @ ForNode (location: (1,0)-(1,23))
            ├── flags: newline
            ├── index:
            │   @ LocalVariableTargetNode (location: (1,4)-(1,5))
            │   ├── flags: ∅
            │   ├── name: :e
            │   └── depth: 0
            ├── collection:
            │   @ CallNode (location: (1,9)-(1,12))
            │   ├── flags: variable_call, ignore_visibility
            │   ├── receiver: ∅
            │   ├── call_operator_loc: ∅
            │   ├── name: :foo
            │   ├── message_loc: (1,9)-(1,12) = "foo"
            │   ├── opening_loc: ∅
            │   ├── arguments: ∅
            │   ├── closing_loc: ∅
            │   ├── equal_loc: ∅
            │   └── block: ∅
            ├── statements:
            │   @ StatementsNode (location: (1,14)-(1,18))
            │   ├── flags: ∅
            │   └── body: (length: 1)
            │       └── @ CallNode (location: (1,14)-(1,18))
            │           ├── flags: newline, variable_call, ignore_visibility
            │           ├── receiver: ∅
            │           ├── call_operator_loc: ∅
            │           ├── name: :body
            │           ├── message_loc: (1,14)-(1,18) = "body"
            │           ├── opening_loc: ∅
            │           ├── arguments: ∅
            │           ├── closing_loc: ∅
            │           ├── equal_loc: ∅
            │           └── block: ∅
            ├── for_keyword_loc: (1,0)-(1,3) = "for"
            ├── in_keyword_loc: (1,6)-(1,8) = "in"
            ├── do_keyword_loc: ∅
            └── end_keyword_loc: (1,20)-(1,23) = "end"
```
does not have a `BlockNode`?
I suppose we could return the `ForNode#statements` or probably better the `ForNode` itself.

eregon:
> or probably better the `ForNode` itself.

Implemented.
@tompng Thank you for the reminder, I didn't think of the block given to `for`.

lib/prism/parse_result.rb
eregon:
```
$ ruby -Ilib -rprism -rbenchmark/ips -e 'm = Prism::YieldNode.instance_method(:===); Benchmark.ips { it.report { Prism.node_for(m) } }'
Without tunnel:
35.855 (± 0.0%) i/s   (27.89 ms/i) -    180.000 in   5.020462s
36.105 (± 0.0%) i/s   (27.70 ms/i) -    183.000 in   5.068871s
With tunnel:
37.073 (± 0.0%) i/s   (26.97 ms/i) -    186.000 in   5.017522s
37.179 (± 0.0%) i/s   (26.90 ms/i) -    186.000 in   5.002992s
```
It a tiny bit faster, and using tunnel is very easy, so done.

lib/prism/parse_result.rb
```diff
@@ -143,6 +143,14 @@ def code_units_column(byte_offset, encoding)
       code_units_offset(byte_offset, encoding) - code_units_offset(line_start(byte_offset), encoding)
     end
 
+    # Returns the byte offset for a given line number and column number
+    def line_and_character_column_to_byte_offset(line, column)
+      line_start = offsets[line - 1]
```
eregon: Maybe we should have a method for this e.g. `line_offset(line)`.
And that probably should take `@start_line` into account.

lib/prism/parse_result.rb
```diff
@@ -143,6 +143,14 @@ def code_units_column(byte_offset, encoding)
...
+      line_end = offsets[line]
+      byte_column = (@source.byteslice(line_start, line_end) or raise)[0...column]&.bytesize #: Integer
```
eregon: The `raise` here might be a bit hacky, since e.g. if the file shrinked we might end up with `nil` for `byteslice`.
Should we raise a better error, or return `nil` in such a case?
I think a proper error (e.g. ArgumentError with some telling message) seems better and less confusing.

`[0...column]` is fine and never returns nil because that always returns a String since it starts at 0: `""[0...42] => ""`.

eregon:
Looking at things like `node.compact_child_nodes.each` maybe we should have `node.each_child_node`? It would avoid the Array allocation and since there is no point to yield `nil` there is no need to name it `node.each_compact_child_node`.
Definitely something for another PR, but @kddnewton WDYT?

lib/prism/parse_result.rb
eregon: Added

lib/prism/parse_result.rb
eregon: Fixed.

eregon:
I found that {Method,UnboundMethod,Proc}#source_location returns columns in bytes and not in characters.
I think that's a bug and filed https://bugs.ruby-lang.org/issues/21783 but nevertheless it's not a problem for this PR.

EDIT: It also means `line_and_character_column_to_byte_offset` isn't used in this PR anymore but I think it's a good addition anyway, and it will become used again when/if that CRuby bug is fixed.

lib/prism/parse_result.rb
```diff
@@ -895,4 +919,52 @@ def initialize(locals, forwarding)
...
+    start_offset = parse_result.source.line_to_byte_offset(start_line) + start_column
+    end_offset = parse_result.source.line_to_byte_offset(end_line) + end_column
+    start_byte_column = start_offset - parse_result.source.line_start(start_offset)
+
+    found = root.tunnel(start_line, start_byte_column).reverse.find do |node|
+      case node
+      when DefNode
+        node.start_offset == start_offset && node.end_offset == end_offset
+      when LambdaNode
+        # Proc#source_location returns start_column 2 for `-> { ... }` (just after the `->`)
```
eregon: Filed https://bugs.ruby-lang.org/issues/21784 for this

lib/prism/parse_result.rb
```diff
@@ -895,4 +919,52 @@ def initialize(locals, forwarding)
...
+    unless parse_result.success?
+      raise ArgumentError, "#{file} has syntax errors: #{parse_result.errors_format}"
+    end
+    root = parse_result.value
+    # CRuby currently returns the source_location columns in bytes and not characters
+    start_offset = parse_result.source.line_to_byte_offset(start_line) + start_column
+    end_offset = parse_result.source.line_to_byte_offset(end_line) + end_column
+    start_byte_column = start_offset - parse_result.source.line_start(start_offset)
+
+    found = root.tunnel(start_line, start_byte_column).reverse.find do |node|
```
eregon: [rfind](https://bugs.ruby-lang.org/issues/21678) would be nice here but of course it's only RUby 4+.
I guess we could do a polyfill but for this case I doubt it matters, most of the time is not spent here but in creating the Ruby objects for the AST, there are typically only a few nodes returned by `tunnel`.

eregon:
Re the test-all failure, I'll fix it, makes it clear test-unit `assert_raise` is quite problematic: https://github.com/test-unit/test-unit/issues/347

lib/prism/parse_result.rb
```diff
@@ -895,4 +919,52 @@ def initialize(locals, forwarding)
...
+      raise ArgumentError, "#source_location[0] is #{file} but this file does not exist"
+    end
+
+    parse_result = Prism.parse_file(file, version: "current")
+    unless parse_result.success?
+      raise ArgumentError, "#{file} has syntax errors: #{parse_result.errors_format}"
+    end
+    root = parse_result.value
+    # CRuby currently returns the source_location columns in bytes and not characters
+    start_offset = parse_result.source.line_to_byte_offset(start_line) + start_column
```
kddnewton, line 931: I'm not understanding what you're doing here exactly. We already have line and column, and it looks like you're calculating offset, and then using that to recompute the column? The column is already in bytes, and that's what `tunnel` expects, so you should just be able to use that directly.

eregon: I compute `start_offset` & `end_offset` so we can have fast comparisons below (2 comparisons vs 4) and without creating Location objects for most cases.

Right, since CRuby currently return the column in byte I can remove `start_byte_column`, will push a commit to do that.

eregon: Done, I amended the last commit to simplify this part.

eregon: BTW, it might be easier to review commit by commit, that would show the purpose of `start_offset` better.

kddnewton: Okay last thing is I don't think we need the `line_and_character_column_to_byte_offset` API now.

eregon: Yeah, it's what I mentioned in https://github.com/ruby/prism/pull/3808#issuecomment-3654952126.
I'll remove it in its own commit, so if we need it again it'll be just a `git revert` away.

eregon: Pushed

Earlopain:
For reference, this won't be available (at least in Ruby 4.0) unfortunately https://bugs.ruby-lang.org/issues/21783#note-8

kddnewton:
@eregon we've marked this as blocked until and unless the source location thing gets resolved. Another option is requesting an official node_id API. Is that something you would consider implementing in TruffleRuby?

eregon:
https://github.com/ruby/ruby/pull/15580 is merged so this is unblocked now :)
Though there is no ruby-dev-builder build with it yet, but that should be [done soon](https://github.com/ruby/ruby-dev-builder/actions/runs/20605178890).

eregon:
The CI was green before updating to latest CRuby 4.1: https://github.com/ruby/prism/actions/runs/20605030946/attempts/2?pr=3808
And now it's red due to issues unrelated to this PR:
https://github.com/ruby/prism/actions/runs/20605030946/job/59206136960?pr=3808
```
sexp_processor-4.17.4 requires ruby version < 4, >= 2.6, which is incompatible
with the current version, 4.1.0.dev
```

Upgrading to sexp_processor-4.17.5 should fix that.

eregon:
Green now

kddnewton:
The code looks largely fine, but it's not clear to me that this feature is actually sticking around, based on the discussions on the issue tracker. Let's wait until that's resolved.

eregon:
Which discussion are you thinking to?
I guess https://bugs.ruby-lang.org/issues/6012#note-40 ?
Matz explicitly approved adding the functionality back in https://github.com/ruby/ruby/pull/15580#issuecomment-3691058342

eregon:
FWIW I do plan to comment on https://bugs.ruby-lang.org/issues/21795 (sounds great to me).

PR #3822, opened 2025-12-20
[rust] implement `first`/`last` on vector-like things

froydnj:
Using prism in `rubyfmt` has shown up a number of places where we have to do `v.iter().next().unwrap()` or similar.  Using `v.iter().last().unwrap()` is also pretty sad, because we're asking the compiler to generate a loop for something that should be a simple memory load.

`std::vec::Vec` implements these little helper functions, so we might as well implement them on vector-like things here, too.

Note that the `NodeList` functions can't be `const` because `Node::new` can panic, which means that `Node::new` cannot be `const`.

cc @reese

### DIFF
--- a/rust/ruby-prism/src/lib.rs
+++ b/rust/ruby-prism/src/lib.rs
@@ -139,6 +139,11 @@ pub struct NodeList<'pr> {
 }
 
 impl<'pr> NodeList<'pr> {
+    unsafe fn at(&self, index: usize) -> Node<'pr> {
+        let node: *mut pm_node_t = *(self.pointer.as_ref().nodes.add(index));
+        Node::new(self.parser, node)
+    }
+
     /// Returns an iterator over the nodes.
     #[must_use]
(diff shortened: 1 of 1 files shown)


froydnj:
> A question for you: do you think we should share this functionality between NodeList and ConstantList with a trait? It seems like we're going to keep treating them as vector-like things, and we could have a more generic vector with an element size and it would all work in theory?

I'm a bit divided here: having a trait for just two things feels like a little Too Much Rust.  OTOH, it would have been nice to write `at()` and then get `first()` and `last()` for basically free.

The other thing that gives me pause is that I'm not sure how many more opportunities for code sharing there would be: AIUI, these structs can't implement a lot of `Vec`-like functionality (e.g. implementing `Index`) because we're handing out newly materialized objects, rather than direct references to the elements.  If you look through the list of functions on `Vec` or `slice`, that doesn't leave you a whole lot of options.

I think it's something to keep an eye on.

PR #3824, opened 2025-12-22
Reduce size of the WASM file (2M->480K)

kddnewton:


enebo:
@kddnewton Awesome.  All of these settings I believe also are the same to allow the parser implementation work compiled with Chicory (from what I recall).

PR #3832, opened 2025-12-31
Use a stable version for cargo lint

Earlopain:
When a new rust version releases, CI shouldn't fail because of new rules

### DIFF
--- a/.github/workflows/rust-bindings.yml
+++ b/.github/workflows/rust-bindings.yml
@@ -51,8 +51,6 @@ jobs:
 
   lint:
     name: cargo:lint
-    strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/rust-bindings.yml
```diff
@@ -51,8 +51,6 @@ jobs:
 
   lint:
     name: cargo:lint
-    strategy:
```
Earlopain, line 54: Editor told me this is invalid, and indeed is declares no matrix

kddnewton (approved):

kddnewton:
On mobile, can you rebase?

Issue #3838, opened 2026-01-07
Ripper translation layer relies on upstream Ripper and is not fully compatible


eregon:
https://github.com/ruby/prism/blob/main/docs/ripper_translation.md mentions
```ruby
require "prism/translation/ripper/shim"
```

However that does not work:
```
$ ruby -v -rprism -rprism/translation/ripper/shim -e 'p Ripper.lex("1 + 2")'
ruby 4.0.0 (2025-12-25 revision 553f1675f3) +PRISM [x86_64-linux]
/home/eregon/.rubies/ruby-4.0.0/lib/ruby/4.0.0/prism/translation/ripper/shim.rb:5: warning: already initialized constant Ripper
/home/eregon/.rubies/ruby-4.0.0/lib/ruby/4.0.0/prism/parse_result.rb:776:in 'Prism::LexResult#initialize': stack level too deep (SystemStackError)
	from /home/eregon/.rubies/ruby-4.0.0/lib/ruby/4.0.0/prism/lex_compat.rb:631:in 'Prism.lex'
	from /home/eregon/.rubies/ruby-4.0.0/lib/ruby/4.0.0/prism/lex_compat.rb:631:in 'Prism::LexCompat#result'
	from /home/eregon/.rubies/ruby-4.0.0/lib/ruby/4.0.0/prism.rb:69:in 'Prism.lex_compat'
	from /home/eregon/.rubies/ruby-4.0.0/lib/ruby/4.0.0/prism/translation/ripper.rb:74:in 'Prism::Translation::Ripper.lex'
	from /home/eregon/.rubies/ruby-4.0.0/lib/ruby/4.0.0/prism/lex_compat.rb:639:in 'Prism::LexCompat#result'
	from /home/eregon/.rubies/ruby-4.0.0/lib/ruby/4.0.0/prism.rb:69:in 'Prism.lex_compat'
	from /home/eregon/.rubies/ruby-4.0.0/lib/ruby/4.0.0/prism/translation/ripper.rb:74:in 'Prism::Translation::Ripper.lex'
	from /home/eregon/.rubies/ruby-4.0.0/lib/ruby/4.0.0/prism/lex_compat.rb:639:in 'Prism::LexCompat#result'
	 ... 7850 levels...
	from /home/eregon/.rubies/ruby-4.0.0/lib/ruby/4.0.0/prism/lex_compat.rb:639:in 'Prism::LexCompat#result'
[...]

PR #3839, opened 2026-01-08
Decouple ripper translator from ripper library

Earlopain:
Ripper exposes `Ripper::Lexer:State` in its output, which is a bit of a problem. To make this work, I basically copy-pasted the implementation.

I'm unsure if that is acceptable and added a test to make sure that these values never go out of sync.
I don't imagine them changing often, prism maps them 1:1 for its own usage.

This also fixed the shim by accident. `Ripper.lex` went to `Translation::Ripper.lex` when it should have been the original. Removing the need for the original resolves that issue.

See https://github.com/ruby/prism/issues/3838

cc @eregon I was actually checking this out after you made https://bugs.ruby-lang.org/issues/21827.

### DIFF
--- a/lib/prism/lex_compat.rb
+++ b/lib/prism/lex_compat.rb
@@ -2,7 +2,6 @@
 # :markup: markdown
 
 require "delegate"
-require "ripper"
 
 module Prism
   # This class is responsible for lexing the source using prism and then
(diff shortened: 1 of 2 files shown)


eregon:
Amazing, thank you, I'll try it soon.

PR #3841, opened 2026-01-09
Correctly expose ripper state

Earlopain:
It is for example used by `irb`, `rdoc`, `syntax_suggest`. It's undocumented but they need it to work. I already wrote this code, this mostly moves things to the right place.

I want to work on making irb not crash when using the ripper shim.


kddnewton:
Looks like some valid test failures

Earlopain:
Oh, this is trying to use the real ripper state again. I will namespace this properly in a bit.

Issue #3828, opened 2025-12-29
Identify unexpected nodes in error recovery scenarios


thomasmarshall:
In error recovery scenarios (for example for invalid Ruby code) Prism can return unexpected node types in fields they wouldn't be for valid Ruby. For example:

```rb
module def foo end
```

Prism parses this as a `ModuleNode` with a `DefNode` in the `constant_path` field. The allowed/expected nodes are listed in `config.yml` [here](https://github.com/ruby/prism/blob/801d98222a5e78765b5dae48d9ee88f8c2933784/config.yml#L3698-L3703) and we use `on error:` to signal that in error recovery scenarios there might be something other than `ConstantReadNode` or `ConstantPathNode`. In this case, we only have `on error: MissingNode` but it can actually be `DefNode` or something else (perhaps this is an oversight and all possible error nodes should be listed here?).

Clients using Prism may want to deal with invalid Ruby syntax in all kinds of different ways, but I suspect the most common approach would be to simply ignore unexpected nodes and maybe stop traversing that branch of the AST. In any case, clients cannot expect only valid node types in fields, as shown above.

One way for clients to deal with unexpected node types would be to guard based on each of the expected node types, for example:

```cpp
if (!PM_NODE_TYPE(node, PM_CONSTANT_READ_NODE) && !PM_NODE_TYPE(node, PM_CONSTANT_PATH_NODE)) {
  // handle error case
}
```

In some cases this might be quite verbose, with many valid node types.

Another approach (assuming the `on error` indicators in `config.yml` are exhaustive) would be to check for known error cases:

```cpp
if (PM_NODE_TYPE(node, PM_MISSING_NODE) || PM_NODE_TYPE(node, PM_DEF_NODE)) {
  // handle error case
}
```

Again, this could potentially be quite verbose and I'm not sure if `config.yml` is currently expected to exhaustively list all possible `on error` cases.

## Proposal

Prism could introduce an `UnexpectedNode` type (similar to `MissingNode`) that would be returned in error recovery scenarios, wrapping the original node type.

```cpp
if (PM_NODE_TYPE(node, PM_MISSING_NODE) || PM_NODE_TYPE(node, PM_UNEXPECTED_NODE)) {
  // handle error case
}
```

Or introduce a helper covering both cases:

```cpp
if (PM_ERROR_NODE(node)) {
  // handle error case
}
```

The client could more easily just ignore error recovery cases, or they could do something appropriate with `node->child` in the case of `UnexpectedNode`.

[...]

PR #3843, opened 2026-01-11
Validate expected nodes

thomasmarshall:
This PR aims to address #3828 by validating expected node types and wrapping unexpected node types in an `ErrorRecoveryNode`. This should make it easier for consumer programs to handle invalid Ruby code. For example:

```rb
module def foo end
```

This is parsed as a `ModuleNode` where `constant_path` is a `DefNode`. After this change, `constant_path` will be an `ErrorRecoveryNode` with its `child` set to the `DefNode`. This way clients can just check `PM_NODE_TYPE(PM_ERROR_RECOVERY_NODE)` to understand if the expected node is either missing (formerly `MissingNode`, now `ErrorRecoveryNode` with empty `child`) or an unexpected node.

As per the suggestion from @kddnewton, I folded the missing nodes and unexpected nodes into a single `ErrorRecoveryNode` type. I also added tests for currently known error recovery scenarios and switched to validating expected node types at the callsites rather than in the constructors.

From an implementation perspective, I think it _could_ be nicer to have comprehensive validations in the constructors because it doesn't require understanding any of the parsing logic (I think my changes are correct, but likely harder to review than validations in constructors). They could also potentially be codegen'd from `config.yml` somehow but I suppose we'd be doing unnecessary work in cases where unexpected nodes are not possible.

I split the changes into a series of small commits that should make it more straightforward to understand what is changing for each node type:

- The first commit captures the existing error recovery scenarios in a Ruby test.
- Then we rename `MissingNode` to `ErrorRecoveryNode`.
- Then we setup the function validation macro.
- Then there is a separate commit for each node type with an `on error` case in one of its fields in `config.yml`.
- Finally, the last commit removes `on error` from `config.yml` as per this suggestion:

> I would go ahead and get rid of the `on error` stuff in `config.yml`

However, I'm not 100% sure this is what was meant, I might have misunderstood.

This is a somewhat substantial change, so I understand if it's a little more difficult to review or feedback on. I'm very happy to make any changes to the approach and implementation, please just let me know!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,8 +51,8 @@ jobs:
       with:
         ruby-version: "3.4"
         bundler-cache: true
-    - name: Check Sorbet
-      run: bundle exec rake typecheck:tapioca typecheck:sorbet
+    # - name: Check Sorbet
+    #   run: bundle exec rake typecheck:tapioca typecheck:sorbet
     - name: Check Steep
       run: bundle exec rake typecheck:steep
     - name: Check field kinds
--- a/config.yml
+++ b/config.yml
@@ -814,8 +814,6 @@ nodes:
           - GlobalVariableReadNode
           - BackReferenceReadNode
           - NumberedReferenceReadNode
-          - on error: SymbolNode # alias $a b
-          - on error: MissingNode # alias $a 42
         comment: |
           Represents the old name of the global variable that can be used before aliasing.
 
@@ -856,8 +854,6 @@ nodes:
         kind:
           - SymbolNode
           - InterpolatedSymbolNode
-          - on error: GlobalVariableReadNode # alias a $b
-          - on error: MissingNode # alias a 42
         comment: |
           Represents the old name of the method that will be aliased.
 
@@ -1871,7 +1867,6 @@ nodes:
         kind:
           - ConstantReadNode
           - ConstantPathNode
-          - on error: CallNode # class 0.X end
       - name: inheritance_operator_loc
         type: location?
         comment: |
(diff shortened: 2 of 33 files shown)


eregon:
> I also added tests for currently known error recovery scenarios and switched to validating expected node types at the callsites rather than in the constructors.

FYI the Ruby code already has logic to check field kinds in the Ruby nodes constructors, search for `check_field_kind` in the codebase (context: https://github.com/ruby/prism/pull/3022).
Are the added checks here redundant perhaps, or do they cover more somehow?

eregon:
I should add: this PR LGTM from a quick look at the diff

thomasmarshall:
> Are the added checks here redundant perhaps, or do they cover more somehow?

The validations I added are on the C side. They check if the node is an expected type and then wrap it in an `ErrorRecoveryNode` if not. I believe that's different to `check_field_kind` which is only on the Ruby side, so in that sense they are not redundant.

kddnewton:
@Earlopain Yeah I know it's a breaking change, but I was hoping for 2.0 we could make this kind of breaking change. We could certainly make an alias, but I'm wondering if instead in the release notes we could write a Ruby snippet saying "if you can't afford the effort to migrate, here's a snippet you can throw wherever that will alias things for you". What do you think?

Earlopain:
Ah, sorry. I misunderstood your 2.0 issue to contain stuff you want to do for 1.x still (or in other words before 2.0). If it is indented for a major then I have no complains.

thomasmarshall:
Sorry, I didn't have time to address the requested changes yet @kddnewton but it looks like you already picked this up. Let me know if there's anything I can do here, otherwise thank you very much!

kddnewton:
Blocked by https://github.com/ruby/ruby/pull/16489

PR #3844, opened 2026-01-11
Use main gemfile for truffleruby/jruby - Fix CI

Earlopain:
Truffleruby 33.0.0 has been released.

Each time truffleruby/jruby release a new X.Y, the versions in the gemfiles need to be updated in order to not fail CI.
The `engine_version` can be removed but at that point might as well just use the main gemfile.

During development, the main gemfile should support these two already anyways. I moved `parser`/`ruby_parser` out of the platform check. The engine-specific gemfiles installed them but not the main one.


eregon:
Very nice!
I wonder if we also need all the others for each CRuby version in https://github.com/ruby/prism/tree/main/gemfiles
Would be great to have a single Gemfile, but I suppose that's problematic somehow given their existence?
Probably the resolution would be overly pessimistic/old gem versions if everything needs to run on 2.7?

Earlopain:
Yeah, the problem is that the main gemfile doesn't work for all ruby versions. Not sure which versions works but I do know that it's not every single one. It wouldn't be a problem if the lockfile wasn't commited but since it is there has to me at least more than one gemfile. They don't make much trouble so it shouldn't really matter.

PR #3849, opened 2026-01-13
Make irb work with the ripper shim

Earlopain:
This is everything that `irb` uses. It works in their test-suite, but there are 20 failures when using the shim that I haven't looked into at all.

`parse` is not used by `irb`. `scan` is, and it's basically `parse` but also including errors. `irb` doesn't seem to care about the errors, so I didn't implement that.


eregon (approved):
LGTM

PR #3853, opened 2026-01-16
Make the ripper shim work with rdoc

Earlopain:
The filter class is a 1:1 copy of ruby.

rdoc has 32 test failures. It seems to expect `on_sp` in some cases to render code as written.

@eregon I believe that is mostly everything that you had implemented in your truffleruby PR. It's missing `Ripper.tokenize` which is used once in irb tests and `Ripper.dedent_string` which I don't find any usage in the ruby org.

I'll add tokenize later since it seems to be documented and pretty trivial.

### DIFF
--- a/lib/prism/translation/ripper.rb
+++ b/lib/prism/translation/ripper.rb
@@ -424,6 +424,7 @@ module Prism
         end
       end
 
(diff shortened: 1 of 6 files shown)


eregon (approved):

eregon:
Thanks!

PR #3854, opened 2026-01-16
Add `Ripper.tokenize` to translation layer

Earlopain:
It's public API and trivial to implement.

### DIFF
--- a/lib/prism/translation/ripper.rb
+++ b/lib/prism/translation/ripper.rb
@@ -78,6 +78,19 @@ module Prism
         end
       end
 
+      # Tokenizes the Ruby program and returns an array of strings.
+      # The +filename+ and +lineno+ arguments are mostly ignored, since the
+      # return value is just the tokenized input.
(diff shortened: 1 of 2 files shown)


Earlopain:
@kddnewton would it be possible to merge my own PRs without review? For some like this where I'm sure it's OK I would like to merge after a day or two if no one approved yet. I guess someone will approve this soon but I'd like to have the option.

kddnewton:
Yeah feel free. I'm on vacation this week which is why I've been slow to respond. I will review when I get back after they're merged and if there's an issue I can take care of it after they're fact.

Earlopain:
Thing is, I can't. I must pass all the merge requirements and can't self-approve. But you can handle this later after you enjoyed your vacation, I don't want to distract you from your time off.

kddnewton:
No problem I just made you an admin so you should be able to now.

Earlopain:
Yup, thanks 👍

PR #3855, opened 2026-01-17
Remove some special `lex_compat` token type classes

Earlopain:
In https://github.com/ruby/prism/pull/3849, I wondered why for the lex method I added some files as excludes for the first time. Turns out, the result contains different token types that lie about equality for `LexTest`.

That seems useful only for testing, so I removed it. It's better to explicitly list the excludes and handle specially certain cases when asserting (like ignoring state for `on_heredoc_end`). Since `RipperLex` already contains full tests for the lex method (added in the PR linked above), I removed the ones in `LexTest`. They make more sense in the ripper class to me.

Also, some of the special token types were not necessary anymore, like `EndContentToken`. Probably was fixed at some point.

### DIFF
--- a/lib/prism/lex_compat.rb
+++ b/lib/prism/lex_compat.rb
@@ -233,50 +233,6 @@ module Prism
       end
     end
 
-    # Ident tokens for the most part are exactly the same, except sometimes we
-    # know an ident is a local when ripper doesn't (when they are introduced
-    # through named captures in regular expressions). In that case we don't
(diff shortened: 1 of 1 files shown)


Earlopain:
Right, the CI lext step also needs this. I'll come back to this later

Earlopain:
Seems like most of these are fixed in either prism or ruby itself now. Just ignoring the state is still needed

PR #3856, opened 2026-01-17
Remove duplicate lex tests

Earlopain:
`RipperTest` already does this (added in https://github.com/ruby/prism/pull/3849)
Since it doesn't use the token classes, it also lists out all the excludes instead of just claiming some are passing.

Also see https://github.com/ruby/prism/pull/3855

### DIFF
--- a/test/prism/lex_test.rb
+++ b/test/prism/lex_test.rb
@@ -6,43 +6,6 @@ require_relative "test_helper"
 
 module Prism
   class LexTest < TestCase
-    except = [
-      # https://bugs.ruby-lang.org/issues/21756
-      "spanning_heredoc.txt",
-      # Prism emits a single string in some cases when ripper splits them up
-      "whitequark/dedenting_heredoc.txt",
-      "heredocs_with_fake_newlines.txt",
-      # Prism emits BEG for `on_regexp_end`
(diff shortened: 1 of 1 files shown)


test/prism/lex_test.rb
```diff
@@ -82,18 +45,5 @@ def test_parse_lex_file
...
-
-    private
-
-    def assert_lex(fixture)
-      source = fixture.read
-
-      result = Prism.lex_compat(source, version: "current")
-      assert_equal [], result.errors
-
-      Prism.lex_ripper(source).zip(result.value).each do |(ripper, prism)|
```
eregon, line 94: Should we still test `Prism.lex_compat` & `Prism.lex_ripper`?
They don't seem tested by #3849 from a quick look, but maybe I missed something.

Earlopain: `lex_compat` is what drives `Translation::Ripper` so that should be good. I'm not sure what the use-case for `lex_ripper` is, seems useful only for for testing against the real one in prism itself.

Both are also pretty extensively used in the various lex CI steps. I added a basic test for them back here.

Earlopain:
Merging so the logic for comparing doesn't have to be implemented in two places.

PR #3858, opened 2026-01-18
Add Prism::Source#line_to_byte_offset, optimize Prism::Node#tunnel and fix docs of BlockNode

eregon:
Extracted from https://github.com/ruby/prism/pull/3808

I need `Prism::Source#line_to_byte_offset` for Ripper, and it's needed to optimize `Prism::Node#tunnel`.

PR #3859, opened 2026-01-18
Add Ripper :on_sp events for Prism.lex_compat and Prism::Translation::Ripper

eregon:
Ref: https://github.com/ruby/prism/issues/3838
On top of https://github.com/ruby/prism/pull/3858 should be merged first.


eregon:
Failures that were added to excludes:
```
==============================================================================================================================
Failure: test_dos_endings.txt_lex(Prism::RipperTest)
test/prism/ruby/ripper_test.rb:148:in 'block in Prism::RipperTest#assert_ripper_lex'
<internal:numeric>:257:in 'Integer#times'
test/prism/ruby/ripper_test.rb:127:in 'Prism::RipperTest#assert_ripper_lex'
test/prism/ruby/ripper_test.rb:84:in 'block (2 levels) in <class:RipperTest>'
<[[1, 9], :on_sp, "\\\r\n", nil]> expected but was
<[[1, 9], :on_sp, "\\\r\n" + "     ", nil]>

diff:
? [[1, 9], :on_sp, "\\\r\n" + "     ", nil]
==============================================================================================================================
F
==============================================================================================================================
Failure: test_seattlerb/str_lit_concat_bad_encodings.txt_lex(Prism::RipperTest)
test/prism/ruby/ripper_test.rb:148:in 'block in Prism::RipperTest#assert_ripper_lex'
<internal:numeric>:257:in 'Integer#times'
test/prism/ruby/ripper_test.rb:127:in 'Prism::RipperTest#assert_ripper_lex'
test/prism/ruby/ripper_test.rb:84:in 'block (2 levels) in <class:RipperTest>'
<[[1, 62], :on_sp, " ", nil]> expected but was
<[[1, 62], :on_sp, " \\\n" + "        ", nil]>

diff:
? [[1, 62], :on_sp, " \\\n" + "        ", nil]
==============================================================================================================================
F
==============================================================================================================================
Failure: test_seattlerb/utf8_bom.txt_lex(Prism::RipperTest)
test/prism/ruby/ripper_test.rb:148:in 'block in Prism::RipperTest#assert_ripper_lex'
<internal:numeric>:257:in 'Integer#times'
test/prism/ruby/ripper_test.rb:127:in 'Prism::RipperTest#assert_ripper_lex'
test/prism/ruby/ripper_test.rb:84:in 'block (2 levels) in <class:RipperTest>'
<[[2, 0], :on_ident, "p", CMDARG]> expected but was
<[[1, 23], :on_sp, "-w\n", BEG]>

diff:
? [[   2, 0], :on_ide     nt, "p", CMDARG]
?   1,  3         sp, "-w\         BE     
?   +++ ???         ??? -----   ?????     
==============================================================================================================================
F
==============================================================================================================================
Failure: test_unparser/corpus/semantic/dstr.txt_lex(Prism::RipperTest)
test/prism/ruby/ripper_test.rb:148:in 'block in Prism::RipperTest#assert_ripper_lex'
<internal:numeric>:257:in 'Integer#times'
test/prism/ruby/ripper_test.rb:127:in 'Prism::RipperTest#assert_ripper_lex'
test/prism/ruby/ripper_test.rb:84:in 'block (2 levels) in <class:RipperTest>'
<[[119, 3], :on_sp, " ", nil]> expected but was
<[[119, 3], :on_sp, " \\\n", nil]>

diff:
? [[119, 3], :on_sp, "     ", nil]
?                      \\\n       
?                     ?       
==============================================================================================================================
Finished in 1.197953795 seconds.
------------------------------------------------------------------------------------------------------------------------------
1892 tests, 36583 assertions, 4 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
99.7886% passed
------------------------------------------------------------------------------------------------------------------------------
```

eregon:
`:on_sp` events also have state on Ripper and it's not always the last token state, example:
```
$ ruby -rripper -e 'pp Ripper.lex("def m(a) nil end")'  
[[[1, 0], :on_kw, "def", FNAME],
 [[1, 3], :on_sp, " ", FNAME],
 [[1, 4], :on_ident, "m", ENDFN],
 [[1, 5], :on_lparen, "(", BEG|LABEL],
 [[1, 6], :on_ident, "a", ARG],
 [[1, 7], :on_rparen, ")", ENDFN],
 [[1, 8], :on_sp, " ", BEG], <=
 [[1, 9], :on_kw, "nil", END],
 [[1, 12], :on_sp, " ", END],
 [[1, 13], :on_kw, "end", END]]
```
And more complex cases like `whitequark/space_args_block.txt`:
```
EXPECTED:
[[[1, 0], :on_kw, "undef", FNAME|FITEM],
 [[1, 5], :on_sp, " ", FNAME|FITEM], # OK
 [[1, 6], :on_ident, "foo", END],
 [[1, 9], :on_comma, ",", BEG|LABEL],
 [[1, 10], :on_sp, " ", FNAME|FITEM], # what?
 [[1, 11], :on_symbeg, ":", FNAME],
 [[1, 12], :on_ident, "bar", ENDFN],
 [[1, 15], :on_comma, ",", BEG|LABEL],
 [[1, 16], :on_sp, " ", FNAME|FITEM], # what?
 [[1, 17], :on_symbeg, ":\"", FNAME],
 [[1, 19], :on_tstring_content, "foo", FNAME],
 [[1, 22], :on_embexpr_beg, "\#{", FNAME],
 [[1, 24], :on_int, "1", END],
 [[1, 25], :on_embexpr_end, "}", END],
 [[1, 26], :on_tstring_end, "\"", END],
 [[1, 27], :on_nl, "\n", BEG]]
```

Would there be any way to get the correct state with Prism?

Earlopain:
I would not bother with comparing state for this. I'm not particularly motivated to understand what is happening in detail here.

I doubt anyone is relying on this information being correct.

eregon:
Yeah it's what I'm doing in test/prism/ruby/ripper_test.rb, I'm ignoring the state of `:on_sp` events for now.
Maybe we should always give them some "0/initial" state or so? Or even `nil` or one less Array element?

Slightly related, is `Translation::Ripper::Lexer::State.new(Translation::Ripper::EXPR_BEG)` the correct way to create Ripper states? It seems wasteful to create new instances every time when there are only a few possible states and they are frozen.

Earlopain:
> Maybe we should always give them some "0/initial" state or so? Or even nil or one less Array element?

Just putting in the state from the previous token seems fine. When it lines up, great. If not, oh well. Doesn't really matter in what way it's wrong. `nil` doesn't sound great for compatibility.

> Slightly related, is Translation::Ripper::Lexer::State.new(Translation::Ripper::EXPR_BEG) the correct way to create Ripper states?

I believe that is what ripper itself does. I haven't benchmarked this in detail yet but there are definitly some hotspots. Much time is spend sorting tokens by location for example.

If changing this makes a noticable difference, I wouldn't be opposed to it. Here's what I use to quickly check:

```rb
require "ripper"
require "prism"
require "benchmark/ips"

Benchmark.ips do |x|

  x.compare!
end
```

----

I was also a bit concerned about the performance of this implementation but it is not so bad. Compared to ripper it goes from 2.3x times slower to 2.6x times slower which seems fine.

lib/prism/lex_compat.rb
```diff
@@ -857,7 +858,39 @@ def result
...
+      new_tokens = []
+
+      last_token_state = Translation::Ripper::Lexer::State.new(Translation::Ripper::EXPR_BEG)
+      last_token_end = 0
+
+      tokens.each do |token|
+        line, column = token.location
+        start_offset = @source.line_to_byte_offset(line) + column
+        if start_offset > last_token_end
+          new_tokens << Token.new([
```
Earlopain: Should get you green, or closer to it. Maybe you need to add some excludes in `lex.rake`

```suggestion
          new_tokens << IgnoreStateToken.new([
```

lib/prism/lex_compat.rb
```diff
@@ -624,12 +624,13 @@ def result
...
 
-      result = Prism.lex(source, **options)
+      result = Prism.lex(code, **options)
+      @source = result.source
       result_value = result.value
       previous_state = nil #: State?
       last_heredoc_end = nil #: Integer?
 
-      bom = source.byteslice(0..2) == "\xEF\xBB\xBF"
+      bom = code.byteslice(0..2) == "\xEF\xBB\xBF"
```
Earlopain: Using both `code` and `source` is a bit confusing to me. If you update the signaure of `Source#slice` to just forward the arguments with `...` this can become
```suggestion
      bom = @source.slice(0..2) == "\xEF\xBB\xBF"
```

Or tweak the call to pass start/length.

lib/prism/lex_compat.rb
```diff
@@ -763,7 +764,7 @@ def result
                   end_offset += 3
                 end
 
-                tokens << Token.new([[lineno, 0], :on_nl, source.byteslice(start_offset...end_offset), lex_state])
+                tokens << Token.new([[lineno, 0], :on_nl, code.byteslice(start_offset...end_offset), lex_state])
```
Earlopain: Same here

lib/prism/lex_compat.rb
```diff
@@ -857,7 +858,39 @@ def result
...
+      tokens = add_on_sp_tokens(tokens)
+
+      Result.new(tokens, result.comments, result.magic_comments, result.data_loc, result.errors, result.warnings, @source)
+    end
+
+    def add_on_sp_tokens(tokens)
+      new_tokens = []
+
+      last_token_state = Translation::Ripper::Lexer::State.new(Translation::Ripper::EXPR_BEG)
+      last_token_end = 0
```
Earlopain: Seems more accurate

```suggestion
      prev_token_state = Translation::Ripper::Lexer::State.new(Translation::Ripper::EXPR_BEG)
      prev_token_end = 0
```

lib/prism/lex_compat.rb
eregon: Nice!

lib/prism/lex_compat.rb
eregon: Right. I wish we didn't have Source#source, because then it's intrinsically confusing :sweat_smile: 
I like `code` or `snippet` to just be a String and `source` to be a `Source` object.
I applied your suggestion.

eregon:
- [x] should rebase on top of https://github.com/ruby/prism/pull/3862 and use the cache

lib/prism/lex_compat.rb
```diff
@@ -857,7 +858,39 @@ def result
...
+            :on_sp,
+            @source.slice(prev_token_end, start_offset - prev_token_end),
+            prev_token_state
+          ])
+        end
+        new_tokens << token
+
+        prev_token_state = token.state
+        prev_token_end = start_offset + token.value.bytesize
+      end
```
Earlopain, line 931: A somewhat ugly code to handle on_sp for line continuations:

```suggestion
        if start_offset > prev_token_end
          sp_value = @source.slice(prev_token_end, start_offset - prev_token_end)
          sp_line = @source.line(prev_token_end)
          sp_column = @source.column(prev_token_end)
          continuation_index = sp_value.byteindex("\\")
          # ripper emits up to three on_sp tokens when line continuations are used
          if continuation_index
            next_whitespace_index = continuation_index + 1
            next_whitespace_index += 1 if sp_value.byteslice(next_whitespace_index) == "\r"
            first_whitespace = sp_value[0...continuation_index]
            continuation = sp_value[continuation_index..next_whitespace_index]
            second_whitespace = sp_value[next_whitespace_index+1..]
            new_tokens << IgnoreStateToken.new([
              [sp_line, sp_column],
              :on_sp,
              first_whitespace,
              prev_token_state
            ]) if first_whitespace != ""
            new_tokens << IgnoreStateToken.new([
              [sp_line, sp_column + continuation_index],
              :on_sp,
              continuation,
              prev_token_state
            ])
            new_tokens << IgnoreStateToken.new([
              [sp_line + 1, 0],
              :on_sp,
              second_whitespace,
              prev_token_state
            ]) if second_whitespace != ""
          else
            new_tokens << IgnoreStateToken.new([
              [sp_line, @source.column(prev_token_end)],
              :on_sp,
              sp_value,
              prev_token_state
            ])
          end
        end
```

You can probably write this a bit more readable but this is what I have for today. With this applied, I have only a single failure for `bundle exec rake lex:ruby`  instead of the current 197 failures.

eregon: I was just looking at that, I got a repro on `top-100-gems/actioncable-7.0.4.3/lib/action_cable/connection/client_socket.rb` line 95 which looks like:
```ruby
          raise ArgumentError, "Failed to execute 'close' on WebSocket: " \
                               "The code must be either 1000, or between 3000 and 4999. " \
                               "#{code} is neither."
```
```diff
- [[95, 73], :on_sp, " ", END],
- [[95, 74], :on_sp, "\\\n", END],
- [[96, 0], :on_sp, "                               ", END],
+ [[95, 73], :on_sp, " \\\n" + "                               ", END],
```

eregon:
Taking notes, next failure is
```ruby
# top-100-gems/graphql-2.0.21/lib/generators/graphql/install/mutation_root_generator.rb
  end
end #<trailing space here
```
```diff
- [[34, 0], :on_kw, "end", END],
- [[34, 3], :on_sp, " ", END]]
+ [[34, 0], :on_kw, "end", END]]
```
UPDATE: fixed now

eregon:
The BOM (byte order mark) edge cases are madness but I think I got them correct now :sweat_smile:

rbi/prism/parse_result.rbi
```diff
@@ -25,8 +25,7 @@ class Prism::Source
   sig { returns(T::Array[String]) }
   def lines; end
 
-  sig { params(byte_offset: Integer, length: Integer).returns(String) }
-  def slice(byte_offset, length); end
+  def slice(...); end
```
Earlopain: I really don't know if this is ok. It typechecks but that's about it about my knowledge. I hate fighting with these tools. I'm just going to revert this and update the two places that pass a range.

test/prism/fixtures/bom_leading_space.txt
```diff
@@ -0,0 +1 @@
+﻿ p (42)
```
Earlopain, line 1: Commited the two snapshots as well. Basically they get created when when running `bundle exec rake`

rakelib/typecheck.rake
```diff
@@ -51,6 +51,7 @@ namespace :typecheck do
       --ignore=rakelib/
       --ignore=Rakefile
       --ignore=top-100-gems/
+      #{Dir.glob("*.rb").map { |f| "--ignore=/#{f}" }.join("\n")}
```
Earlopain, line 54: This is nice, thanks

Earlopain (approved):
I'm somewhat surprised that this seems to handle all the cases. Nice work.

Earlopain:
This causes test failures in `syntax_suggest` because it worked around missing `on_sp` in an incompatible way. I'll handle it https://github.com/ruby/ruby/pull/15914

rbi/prism/parse_result.rbi
eregon: I read some Sorbet docs and asked ChatGPT and basically Sorbet seems to not really support `...`.
I tried just leaving `sig { returns(String) }` but it was not happy about it.
Maybe overloads like this would work:
```ruby
sig { overload.params(a: Integer, b: Integer).void }
sig { overload.params(r: T::Range[Integer]).void }
def slice(*args); end
```
But it feels complicated and duplicating the type of byteslice, hence just `def slice(...); end` felt simpler.
It'd be nice if we could say "this has the exact same type as String#byteslice".

eregon:
@Earlopain Oh, sorry about that, I didn't expect that would break.
Thanks a lot for making PRs to fix it!

PR #3860, opened 2026-01-18
[rust] add support for accesing the list of line offsets

froydnj:
As per subject.  `rubyfmt` spends a non-trivial amount of time figuring out where the newlines are in the source string, and since the parser already has this information, it seems silly to have to recompute it.

I'm glad I wrote tests, though, because the implementation seems surprising to me -- I don't think I ever would have expected the list to contain offsets for newlines that aren't actually present in the source string.  I guess it's too late to change the contract here?  (I considered making this code return `list[1..]`, but I think the consistency between it and the C implementation is more important; happy to reconsider or to get the underlying library changed.)

cc @reese

### DIFF
--- a/rust/ruby-prism/src/lib.rs
+++ b/rust/ruby-prism/src/lib.rs
@@ -66,6 +66,27 @@ mod tests {
     }
 
     #[test]
+    fn line_offsets_test() {
+        let source = "";
+        let result = parse(source.as_ref());
+
+        let expected: [u32; 1] = [0];
+        assert_eq!(expected, result.line_offsets());
+
+        let source = "1 + 1";
+        let result = parse(source.as_ref());
+
+        let expected: [u32; 1] = [0];
+        assert_eq!(expected, result.line_offsets());
+
+        let source = "begin\n1 + 1\n2 + 2\nend";
+        let result = parse(source.as_ref());
+
+        let expected: [u32; 4] = [0, 6, 12, 18];
+        assert_eq!(expected, result.line_offsets());
+    }
+
+    #[test]
     fn magic_comments_test() {
         use crate::MagicComment;
 
--- a/rust/ruby-prism/src/parse_result/mod.rs
+++ b/rust/ruby-prism/src/parse_result/mod.rs
@@ -119,6 +119,15 @@ impl<'pr> ParseResult<'pr> {
         &self.source[start..end]
     }
 
+    /// Returns a slice containing the offsets of the start of each line in the source string
+    /// that was parsed.
+    #[must_use]
+    pub fn line_offsets(&self) -> &'pr [u32] {
+        unsafe {
+            let list = &(*self.parser.as_ptr()).line_offsets;
+            std::slice::from_raw_parts(list.offsets, list.size)
+        }
+    }
     /// Returns an iterator that can be used to iterate over the errors in the
     /// parse result.
     #[must_use]


froydnj:
Oh, I see -- the doc comment on `pm_newline_list_t` is a little ambiguous, or maybe I just read too fast:

https://github.com/ruby/prism/blob/6881fd55ac988dbc4e8e665a7d5c6ab3c5e0bd88/include/prism/util/pm_newline_list.h#L24-L40

The indices in the list are actually the indices for the beginning of each line, not of the actual newlines in the source string.  OK, that makes the tests make a little more sense, but in that case, I think the function name should be something more like `line_offsets` and the documentation should be changed.

froydnj:
Moving this to draft; I think if we're going to provide line offsets, we might as well go ahead and just expose an iterator over the lines instead, since that's what people are going to be doing anyway.

eregon:
The list of line offsets is useful for quite a few other things, which probably will be useful in Rust too: https://github.com/ruby/prism/blob/38c64e5c5ecd066c886eba3c0e400f7e70f3ea2d/lib/prism/parse_result.rb#L79-L169
Agreed the name `pm_newline_list_t` is confusing :sweat_smile: 
Would need to fix that without breaking compatibility though.

froydnj:
Going to submit this for review as-is; we can add `.lines()` at a later point if it seems like a common thing.

rust/ruby-prism/src/lib.rs
```diff
@@ -600,6 +600,16 @@ impl<'pr> ParseResult<'pr> {
         &self.source[start..end]
     }
 
+    /// Returns a slice containing the offsets of the start of each line in the source string
+    /// that was parsed.
+    #[must_use]
+    pub fn line_offsets(&self) -> &'pr [usize] {
```
kddnewton: I'm happy to expose this, but I'd prefer to keep the naming consistent. (Right now, newline_list is newline_list everywhere except in rust where it's line_offsets.)

froydnj: I can appreciate consistency, but I think "newline_list" is a bad name regardless, and a function named `*_list` that actually returns a slice seems very un-Rusty.  Sacrifices do have to be made in cross-language projects, though...

Would you take a pre-work PR to change the naming here -- either for the fields, the underlying data structure, or all of the above -- prior to landing this?  This seems like a good time to rename things as #3772 would likely require significant changes to clients anyway in the next release, and trying to do all the breaking changes all at once seems like a reasonable thing to do.

kddnewton: Sure, yeah that's fine. Let's rename it across the board. We won't be able to merge unless we do the corresponding work in CRuby, but I'm okay to do that if you can take the prism side of things.

froydnj: Cool, I can do that, thank you!

eregon: > Right now, newline_list is newline_list everywhere except in rust where it's line_offsets.

FWIW I checked and it's called `ParseResult#offsets` in Ruby:
https://github.com/ruby/prism/blob/54a5dd3285530ed82c41ec0d3e95bc61746afdd3/lib/prism/parse_result.rb#L42-L43

And it's already called `lineOffsets` in Java:
https://github.com/ruby/prism/blob/54a5dd3285530ed82c41ec0d3e95bc61746afdd3/templates/java/org/prism/Nodes.java.erb#L52

froydnj:
Rebased on top of #3925 and other Rust-based changes.

kddnewton (approved):

PR #3862, opened 2026-01-19
Optimize ripper translator

Earlopain:
Ref https://github.com/ruby/prism/pull/3859#issuecomment-3767532413. I was interested in how big the impact is and turns out it is a lot.

Creating state classes is pretty expensive. Since they are not modifiable, we can reuse them instead.

Benchmark script:
```rb
require "ripper"
require "prism"
require "benchmark/ips"

Benchmark.ips do |x|
  x.config(time: 10)

  x.compare!
end
```

Before:

```
               prism     1.000 i/100ms
              ripper     1.000 i/100ms
               prism      0.293 (± 0.0%) i/s     (3.42 s/i) -      3.000 in  10.248348s
              ripper      0.633 (± 0.0%) i/s     (1.58 s/i) -      7.000 in  11.055687s

Comparison:
              ripper:        0.6 i/s
               prism:        0.3 i/s - 2.16x  slower
```

After

```
               prism     1.000 i/100ms
              ripper     1.000 i/100ms
               prism      0.486 (± 0.0%) i/s     (2.06 s/i) -      5.000 in  10.280413s
              ripper      0.635 (± 0.0%) i/s     (1.58 s/i) -      7.000 in  11.027169s

Comparison:
              ripper:        0.6 i/s
               prism:        0.5 i/s - 1.31x  slower
```

### DIFF
--- a/lib/prism/translation/ripper/lexer.rb
+++ b/lib/prism/translation/ripper/lexer.rb
@@ -38,6 +38,13 @@ module Prism
           def allbits?(i) to_int.allbits?(i) end
           def anybits?(i) to_int.anybits?(i) end
           def nobits?(i) to_int.nobits?(i) end
+
+          # Instances are frozen and there are only a handful of them so we cache them here.
+          STATES = Hash.new { |h,k| h[k] = State.new(k) }
+
+          def self.cached(i)
+            STATES[i]
+          end
         end
 
         class Elem
@@ -47,7 +54,7 @@ module Prism
             @pos = pos
             @event = event
             @tok = tok
-            @state = State.new(state)
+            @state = State.cached(state)
             @message = message
           end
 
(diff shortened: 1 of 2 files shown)


lib/prism/translation/ripper/lexer.rb
```diff
@@ -38,6 +38,17 @@ def |(i) self.class.new(to_int | i) end
           def allbits?(i) to_int.allbits?(i) end
           def anybits?(i) to_int.anybits?(i) end
           def nobits?(i) to_int.nobits?(i) end
+
+          # Instances are frozen and there are only a handful of them so we cache them here.
+          # This list here is derived by running the full prism test suite and checking which
+          # state was encountered.
+          STATES = [
+            1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1025, 1026, 2064, 3088, 4224,
```
eregon: It feels like this list is likely to eventually get out of sync, how about:
```ruby
STATES = Hash.new { |h,k| h[k] = State.new(k) }
```
and then just `STATES[i]` below?

Earlopain: That seems fine, yeah.

eregon (approved):

PR #3865, opened 2026-01-20
Fix `on_*` return value of ripper translator

Earlopain:
You're supposed to return the first argument.
```rb
# Before
[[:stmts_new], [:rescue_mod, nil, nil], [:stmts_add, nil, nil], [:program, nil]]
# After
[[:stmts_new], [:rescue_mod, "1", "2"], [:stmts_add, nil, "1"], [:program, nil]]
```

The correct result would be:
`[[:rescue_mod, "1", "2"], [:stmts_new], [:stmts_add, nil, "1"], [:program, nil]]`

But the order depends on the prism AST so it seems very difficult to match.

Ref: https://github.com/ruby/prism/issues/3838

### DIFF
--- a/lib/prism/translation/ripper.rb
+++ b/lib/prism/translation/ripper.rb
@@ -3446,12 +3446,12 @@ module Prism
 
       # :stopdoc:
       def _dispatch_0; end
-      def _dispatch_1(_); end
-      def _dispatch_2(_, _); end
-      def _dispatch_3(_, _, _); end
-      def _dispatch_4(_, _, _, _); end
-      def _dispatch_5(_, _, _, _, _); end
-      def _dispatch_7(_, _, _, _, _, _, _); end
+      def _dispatch_1(arg); arg end
+      def _dispatch_2(arg, _); arg end
+      def _dispatch_3(arg, _, _); arg end
+      def _dispatch_4(arg, _, _, _); arg end
+      def _dispatch_5(arg, _, _, _, _); arg end
+      def _dispatch_7(arg, _, _, _, _, _, _); arg end
       # :startdoc:
 
       #
--- a/test/prism/ruby/ripper_test.rb
+++ b/test/prism/ruby/ripper_test.rb
@@ -86,6 +86,40 @@ module Prism
       define_method("#{fixture.test_name}_lex") { assert_ripper_lex(fixture.read) }
     end
 
+    module Events
+      attr_reader :events
+
+      def initialize(...)
+        super
+        @events = []
+      end
+
+      Prism::Translation::Ripper::PARSER_EVENTS.each do |event|
+        define_method(:"on_#{event}") do |*args|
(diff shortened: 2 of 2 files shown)


Earlopain:
Somewhat unsure about this fix but it seems to work very well.

test/prism/ruby/ripper_test.rb
```diff
@@ -86,6 +86,40 @@ class RipperTest < TestCase
...
 
+    module Events
+      attr_reader :events
+
+      def initialize(...)
+        super
+        @events = []
+      end
+
+      (Prism::Translation::Ripper::PARSER_EVENTS).each do |event|
```
eregon:
```suggestion
      Prism::Translation::Ripper::PARSER_EVENTS.each do |event|
```

test/prism/ruby/ripper_test.rb
```diff
@@ -86,6 +86,40 @@ class RipperTest < TestCase
...
+    end
+
+    def test_events
+      source = "1 rescue 2"
+      ripper = RipperEvents.new(source)
+      prism = PrismEvents.new(source)
+      ripper.parse
+      prism.parse
+      # This makes sure that the content is the same. Ordering is not correct for now.
+      assert_equal(ripper.events.sort_by(&:itself), prism.events.sort_by(&:itself))
```
eregon: Isn't `.sort_by(&:itself)` the same as `.sort`?

lib/prism/translation/ripper.rb
```diff
@@ -3446,12 +3446,12 @@ def bounds(location)
...
-      def _dispatch_3(_, _, _); end
-      def _dispatch_4(_, _, _, _); end
-      def _dispatch_5(_, _, _, _, _); end
-      def _dispatch_7(_, _, _, _, _, _, _); end
+      def _dispatch_1(arg); arg end
+      def _dispatch_2(arg, _); arg end
+      def _dispatch_3(arg, _, _); arg end
+      def _dispatch_4(arg, _, _, _); arg end
+      def _dispatch_5(arg, _, _, _, _); arg end
+      def _dispatch_7(arg, _, _, _, _, _, _); arg end
```
eregon, line 3454: Right, it's the same as https://github.com/ruby/ruby/blob/d225bb8b464e4e03d2eb6c09ef15adf727af9e2b/ext/ripper/lib/ripper/core.rb#L33-L42

eregon: I wonder why we didn't/don't just copy that part from upstream.
Maybe the unused variables cause warnings?

Earlopain: Ah, nice. I was mostly just guessing but this makes it pretty clear.

test/prism/ruby/ripper_test.rb
Earlopain: Yeah

PR #3866, opened 2026-01-20
Add `license` identifier to Rust crates

vinistock:
So that tools like [cargo-about](https://crates.io/crates/cargo-about) can automatically bundle the license for attribution.


Earlopain:
Should also update the license on crates.io from non-standard 👍

PR #3867, opened 2026-01-20
Check using Prism nodes if a command call has any arguments in Ripper translator

eregon:
* We don't know what `on_*` events might return so we cannot assume it's an Array.
* See https://github.com/ruby/prism/issues/3838#issuecomment-3774702396

### DIFF
--- a/lib/prism/translation/ripper.rb
+++ b/lib/prism/translation/ripper.rb
@@ -832,7 +832,7 @@ module Prism
       # foo(bar)
       #     ^^^
       def visit_arguments_node(node)
-        arguments, _ = visit_call_node_arguments(node, nil, false)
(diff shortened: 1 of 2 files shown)


eregon:
Thinking more about it, it seems far better to just check `node.arguments&.arguments&.any?`, that's much safer and does the correct decision based on the AST.
I'll update to that.

eregon:
Done, this feels like a proper fix.

eregon:
Added a test as well and confirmed it fails before the fix (with 132 failures).

eregon:
@Earlopain could you review this?
With this merged https://github.com/truffleruby/truffleruby/pull/4102 will have no more failures in MRI tests than before.

Issue #3869, opened 2026-01-25
Release with the improved Prism::Translation::Ripper


eregon:
We're using `Prism::Translation::Ripper` as the implementation for `::Ripper` in TruffleRuby.
This seems to work fine, but currently I'm importing a commit from ruby/prism `main` branch.

It would be better to import a Prism release, so that e.g. `gem info prism` would report the right version.
Currently with that commit from `main` it [would report itself as 1.8.0](https://github.com/truffleruby/truffleruby/blob/49d39f3e1518e9ff901ba4159bcd1e4de10c814f/lib/gems/specifications/default/prism-1.8.0.gemspec) but that's not correct since it has many extra commits.

This is not urgent and I'm more than happy to include more fixes/optimizations from @Earlopain if he plans any (do you?).

@kddnewton On a slightly related note I wonder if you'd be interested to avoid so many copies of the version number across the codebase.
I recall https://github.com/ruby/prism/pull/1295 but that didn't work when imported in CRuby.
I could do a safer variant that just share the version for files generated by `template.rb` if you're interested.
It's probably not a big deal though looking at https://github.com/ruby/prism/blob/main/docs/releasing.md#preparation.

Related: https://github.com/ruby/prism/issues/3838

Earlopain:
> This is not urgent and I'm more than happy to include more fixes/optimizations

No more fixes, it seems good enough. I was looking for more things to optimize but it was all without much impact so not really worth a PR for the extra complexity (https://github.com/ruby/prism/compare/main...Earlopain:prism:more-ripper-optimizations)

That, and the majority of time is spend here: https://github.com/ruby/prism/blob/61e1207e96ddb3236f7c88e9c7ed7e9c74a0ae5a/lib/prism/lex_compat.rb#L607-L608

It binary searches through the offsets incredibly often but the tokens are basically already sorted (ignoring heredocs). But I don't really have the motivation to try, overall it is also good enough for me.

eregon:
Took a quick look at your 2 commits, [the first one](https://github.com/Earlopain/prism/commit/fbf56f4fb293f603a3766870cc4edb490a78a4d1) is I think worth it and a small/easy change. I'd just use `nonzero?` since it's made for that, like `(line_a - line_b).nonzero? || column_a - column_b`.
Could probably also be `(line_a <=> line_b).nonzero? || column_a <=> column_b` to use the "sorting operator".
Could you open a PR with that?

[...]

PR #3870, opened 2026-01-25
Optimize ripper translator token sorting

Earlopain:
With the benchmark from https://github.com/ruby/prism/commit/2ea81398cc6f776dca686125847dbf9b0af58c16 Prism is:
* 1.33x slower before
* 1.07x slower after

### DIFF
--- a/lib/prism/lex_compat.rb
+++ b/lib/prism/lex_compat.rb
@@ -802,8 +802,12 @@ module Prism
       # Drop the EOF token from the list
       tokens = tokens[0...-1]
 
-      # We sort by location to compare against Ripper's output
-      tokens.sort_by!(&:location)
+      # We sort by location because Ripper.lex sorts.
+      # Manually implemented instead of `sort_by!(&:location)` for performance.
+      tokens.sort_by! do |token|
+        line, column = token.location
+        source.line_to_byte_offset(line) + column
+      end
 
       # Add :on_sp tokens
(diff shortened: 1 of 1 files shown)


lib/prism/lex_compat.rb
```diff
@@ -802,8 +802,12 @@ def result
       # Drop the EOF token from the list
       tokens = tokens[0...-1]
 
-      # We sort by location to compare against Ripper's output
-      tokens.sort_by!(&:location)
+      # We sort by location to compare against Ripper's output. Manually implemented
```
eregon: > to compare against Ripper's output

Just for testing or because Ripper's output is always sorted?

Earlopain: I wrote it in https://github.com/ruby/prism/issues/3869#issuecomment-3797174584, let me copy:

Most of ripper returns the tokens sorted. `Ripper.lex` does for example. There is `Ripper::Lexer.new("<<~FOO\nFOO").parse` which returns them unsorted but it is undocumented. The other `Lexer` methods are implemented on top of it and they always sort.  That is the only API that does this as far as I remember.

eregon: Thanks, saw this PR first.
Maybe we could clarify like:
```suggestion
      # We sort by location because Ripper.lex sorts. Manually implemented
```

Earlopain: Yeah, sure

PR #3871, opened 2026-01-26
Remove `Prism.lex_ripper`

Earlopain:
Since `on_sp` is emitted, it doesn't do a whole lot anymore.

This leaves one incompatibility for code like `"x#$%"`

Ripper confuses this for bare interpolation with a global, but `$%` is not a valid global name. Still,
it emits two string tokens in such a case. It doesn't make sense for prism to work around this bug,
so the affected files are added as excludes.


kddnewton:
We need splitting for escaped newlines in heredocs anyway.

Earlopain:
> I think we can remove this API entirely instead of deprecating, I can't imagine anyone is using it.

I did a codesearch and didn't find anyone using it. Let's remove 👍 

> I think I'd like us to support the splitting though

Do you mean for `"x#$%"`? I created an upstream issue for this at https://bugs.ruby-lang.org/issues/21849. Since it's a single string, I would expect ripper to emit a single token for it like prism does. I don't think it makses sense to replicate the buggy behavior. And since it's only for our tests, excludes seem fine.

If you insist I try to implement it but the value of doing that seems very low to me. Doing it correctly is a bit complex since it depends on where the string appears. Quoted heredocs for example don't allow interpolation so ripper does the correct thing there and that doesn't just apply do dash heredocs.

kddnewton:
Nah that's okay I don't insist :)

kddnewton:
But let's do remove lex_ripper

PR #3872, opened 2026-01-26
Fix `on_words_sep` for ripper translator with newlines

Earlopain:
Ripper emits a token each per line.

### DIFF
--- a/lib/prism/lex_compat.rb
+++ b/lib/prism/lex_compat.rb
@@ -657,6 +657,15 @@ module Prism
             IgnoreStateToken.new([[lineno, column], event, value, lex_state])
           when :on_embexpr_end
             IgnoreStateToken.new([[lineno, column], event, value, lex_state])
+          when :on_words_sep
+            # Ripper emits one token each per line.
+            lines = value.lines
+            lines[0...-1].each do |whitespace|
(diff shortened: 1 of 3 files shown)


test/prism/ruby/ripper_test.rb
```diff
@@ -62,18 +62,10 @@ class RipperTest < TestCase
     ]
 
     omitted_lex = [
-      "comments.txt",
```
Earlopain, line 65: Not every exclude was fixed through this commit. I just also removed all the others that are now passing.

lib/prism/lex_compat.rb
```diff
@@ -657,6 +657,16 @@ def result
             IgnoreStateToken.new([[lineno, column], event, value, lex_state])
           when :on_embexpr_end
             IgnoreStateToken.new([[lineno, column], event, value, lex_state])
+          when :on_words_sep
+            # Ripper emits one token each per line. This splits the
+            # string at newlines and emits one token for each match.
+            parts = value.scan(/(?=\s)[^\S\r\n]*(?:\r\n|\n)?/)
```
eregon: This looks a bit complicated. Would `String#lines` work?

Earlopain: That's so much better, wish I thought of that

PR #3873, opened 2026-01-26
Use each_line to avoid allocating array

kddnewton:
Though very unlikely, it could potentially allocate a large array of whitespace.


Earlopain:
:+1:

PR #3876, opened 2026-01-27
Add platforms to main gemfile

Earlopain:
I don't want to compile nokogiri each time.

I though this makes trouble with ruby-dev because platform-specific nokogiri gems don't allow ruby-dev. But it seems to correctly fall back to the ruby platform (compile) in that case.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -13,6 +13,14 @@ GEM
     erb (6.0.1)
     erb (6.0.1-java)
     ffi (1.17.3)
+    ffi (1.17.3-aarch64-linux-gnu)
+    ffi (1.17.3-aarch64-linux-musl)
+    ffi (1.17.3-arm-linux-gnu)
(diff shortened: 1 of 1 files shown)


Earlopain:
Nevermind, I didn't notice that bundler just removed the platform-specific entries when running with ruby-dev. So https://github.com/ruby/rubygems/issues/9034 is still an issue. CI sets `bundle config --local deployment true` which makes it fail.

eregon:
> CI sets `bundle config --local deployment true` which makes it fail.

Maybe setup-ruby could avoid setting that if the user explicitly specifies `BUNDLE_FROZEN=false`. But not sure it's a good idea as then it means the CI no longer check the lockfile is up-to-date, and potentially could cause issues with caching as well.

Earlopain:
Nah, I think it's fine. Either the lockfile should be dropped entirely or bundler should fix this issue.

PR #3877, opened 2026-01-27
Rename line_to_byte_offset -> byte_offset

kddnewton:
Also, include the column in here. Hopefully we can do some additional optimizations later.

### DIFF
--- a/lib/prism/parse_result.rb
+++ b/lib/prism/parse_result.rb
@@ -76,13 +76,13 @@ module Prism
       source.byteslice(byte_offset, length) or raise
     end
 
-    # Converts the line number to a byte offset corresponding to the start of that line
-    def line_to_byte_offset(line)
-      l = line - @start_line
-      if l < 0 || l >= offsets.size
-        raise ArgumentError, "line #{line} is out of range"
-      end
-      offsets[l]
+    # Converts the line number and column in bytes to a byte offset.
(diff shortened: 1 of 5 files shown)


eregon:
> Hopefully we can do some additional optimizations later.

What are you thinking to? That method looks already pretty much optimal, it's basically just an `Array#[]`.

lib/prism/parse_result.rb
eregon, line 106:
```suggestion
    # Return the column in bytes for the given byte offset.
```

kddnewton:
> > Hopefully we can do some additional optimizations later.
> 
> What are you thinking to? That method looks already pretty much optimal, it's basically just an `Array#[]`.

Not sure. Since we have both operands, we could potentially cache based on both. I'm not sure if it would be worth it, but I wanted to have both.

lib/prism/parse_result.rb
eregon: In general I think we should avoid the term `column number`, at least I would interpret that as starting at 1 like an editor column. So I'd suggest to use `column in bytes/characters/code units`.
Happy to make a PR for that if you agree.

kddnewton: Sure yeah, happy to see that change.

eregon: https://github.com/ruby/prism/pull/3880

PR #3878, opened 2026-01-27
Add Prism::Node#find_all

kddnewton:


templates/lib/prism/node.rb.erb
```diff
@@ -215,6 +215,26 @@ module Prism
 
       nil
     end
+    alias find breadth_first_search
+
+    # Returns all of the nodes that match the given block when visited in a
+    # breadth-first search. This is useful for finding all nodes that match a
```
eregon, line 221: For finding all nodes matching something, maybe a depth-first search would be faster?
I'm unsure if the order matters much in such cases, but maybe?

I'm thinking aliasing to `find_all` might be suboptimal if there is a DFS variant at some point.

templates/lib/prism/node.rb.erb
```diff
@@ -215,6 +215,26 @@ module Prism
 
       nil
     end
+    alias find breadth_first_search
```
eregon, line 218: Same thing here, `find` might be better/faster for/with DFS, especially for the example shown with `node_id`.

kddnewton: Yeah, I thought about that too. But I figured we can cross that bridge when we get to it if we really need to add a DFS variant.

PR #3880, opened 2026-01-27
Use the terminology "column in bytes/characters/code units"

eregon:
* Consistent and clear.
* Avoids the confusion that "column number" might be understood as a column in an editor starting at 1 (they all start at 0).

From https://github.com/ruby/prism/pull/3877#discussion_r2733633671

### DIFF
--- a/include/prism/util/pm_newline_list.h
+++ b/include/prism/util/pm_newline_list.h
@@ -46,7 +46,7 @@ typedef struct {
     /** The line number. */
     int32_t line;
 
-    /** The column number. */
+    /** The column in bytes. */
     uint32_t column;
 } pm_line_column_t;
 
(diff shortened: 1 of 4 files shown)


include/prism/util/pm_newline_list.h
```diff
@@ -40,13 +40,13 @@ typedef struct {
 } pm_newline_list_t;
 
 /**
- * A line and column in a string.
+ * A line and column in a struct.
```
kddnewton: This is meant to say that this struct represents a line and column in a string. So it was correct initially.

eregon: OK, I'll change it back. It sounded a bit overly specific to me, people might use line and column for other things than just for printing to a string.

eregon: done

PR #3883, opened 2026-01-28
Remove tokens from lex compat

kddnewton:
Instead of having custom classes, use arrays and track which tokens we should ignore the state for in the test.


eregon:
From https://github.com/ruby/prism/pull/3868#discussion_r2736204662 I guess?

One downside is then probably more of the upstream Ripper tests may fail, due to not ignoring the state during comparisons (for tests outside the prism repo). I think it's probably fine though.

PR #3886, opened 2026-01-29
Support `version: "nearest"`.

### LINKED ISSUE #3885: Robust version of `Prism.parse(code, version: 'current')`
In IRB, it's better to parse ruby code with current ruby version for usability, but doesn't strictly require current version.
It basically work except for few edge cases: Error tokens (valid in latest) won't be rendered RED. (example: `"if a\n&&b\nend"`)

Parsing with `Prism.parse(code, version: RUBY_VERSION >= '3.3' ? '3.3' : 'current')` is not enough.
Lower bound(`3.3`) may change in the future, RUBY_VERSION may exceed upper bound(`3.4` (prism-1.6) or `4.1` (prism >= 1.7)) that may block updating ruby version in a rails app.

It would be helpful if there is an easy way to use current version robustly (like `Prism.parse(code, version: 'nearest' or 'closest')`), or a way to get supported version range.

kddnewton:
This clamps to supported versions based on the current Ruby version.

Fixes #3885

### DIFF
--- a/ext/prism/extension.c
+++ b/ext/prism/extension.c
@@ -201,9 +201,24 @@ build_options_i(VALUE key, VALUE value, VALUE argument) {
             const char *version = check_string(value);
 
             if (RSTRING_LEN(value) == 7 && strncmp(version, "current", 7) == 0) {
-                const char *current_version = RSTRING_PTR(rb_const_get(rb_cObject, rb_intern("RUBY_VERSION")));
-                if (!pm_options_version_set(options, current_version, 3)) {
-                    rb_exc_raise(rb_exc_new_cstr(rb_cPrismCurrentVersionError, current_version));
+                const char *ruby_version = RSTRING_PTR(rb_const_get(rb_cObject, rb_intern("RUBY_VERSION")));
+                if (!pm_options_version_set(options, ruby_version, 3)) {
+                    rb_exc_raise(rb_exc_new_cstr(rb_cPrismCurrentVersionError, ruby_version));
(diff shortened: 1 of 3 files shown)


ext/prism/extension.c
```diff
@@ -201,9 +201,24 @@ build_options_i(VALUE key, VALUE value, VALUE argument) {
             const char *version = check_string(value);
 
             if (RSTRING_LEN(value) == 7 && strncmp(version, "current", 7) == 0) {
-                const char *current_version = RSTRING_PTR(rb_const_get(rb_cObject, rb_intern("RUBY_VERSION")));
-                if (!pm_options_version_set(options, current_version, 3)) {
-                    rb_exc_raise(rb_exc_new_cstr(rb_cPrismCurrentVersionError, current_version));
+                const char *ruby_version = RSTRING_PTR(rb_const_get(rb_cObject, rb_intern("RUBY_VERSION")));
```
eregon, line 204: Pre-existing issue but I think this pattern looks unsafe and could cause segfaults in the worst case.
The problem is getting the constant value (a Ruby String) but not keeping a reference to it on stack, then the GC might move the constant and if the `char*` is embedded it will move too and we'd read something else in the Ruby heap.

I think the fix is just keeping a variable with the `VALUE` and also use it in the rb_exc* below.
I don't have time to make a PR now though.

eregon: https://github.com/ruby/prism/pull/3892

PR #3887, opened 2026-01-29
[rust] add `node_ext` module with `full_name` methods for constant nodes

sei40kr:
## Summary

- Port Ruby's `node_ext.rb` functionality to Rust bindings
- Add `FullName` trait with `full_name` and `full_name_parts` methods, implemented for `ConstantReadNode`, `ConstantWriteNode`, `ConstantTargetNode`, `ConstantPathNode`, and `ConstantPathTargetNode`
- Add `ConstantPathError` enum for error handling

## Test Coverage

Rust bindings include 11 unit tests compared to Ruby's 8 tests. Additional coverage includes:

- `ConstantWriteNode` (not tested in Ruby)
- `ConstantTargetNode` (not tested in Ruby)
- `MissingNodes` error case (not tested in Ruby)

## Public API Changes (cargo-public-api)

```
Added items to the public API
=============================
+pub enum ruby_prism::ConstantPathError
+pub ruby_prism::ConstantPathError::DynamicParts
+pub ruby_prism::ConstantPathError::MissingNodes
+impl core::clone::Clone for ruby_prism::ConstantPathError
+impl core::cmp::Eq for ruby_prism::ConstantPathError
+impl core::cmp::PartialEq for ruby_prism::ConstantPathError
+impl core::error::Error for ruby_prism::ConstantPathError
+impl core::fmt::Debug for ruby_prism::ConstantPathError
+impl core::fmt::Display for ruby_prism::ConstantPathError
+pub trait ruby_prism::FullName<'pr>
+pub fn ruby_prism::FullName::full_name(&self) -> Result<Vec<u8>, ConstantPathError>
+pub fn ruby_prism::FullName::full_name_parts(&self) -> Result<Vec<&'pr [u8]>, ConstantPathError>
+impl<'pr> ruby_prism::FullName<'pr> for ruby_prism::ConstantReadNode<'pr>
+impl<'pr> ruby_prism::FullName<'pr> for ruby_prism::ConstantWriteNode<'pr>
+impl<'pr> ruby_prism::FullName<'pr> for ruby_prism::ConstantTargetNode<'pr>
+impl<'pr> ruby_prism::FullName<'pr> for ruby_prism::ConstantPathNode<'pr>
+impl<'pr> ruby_prism::FullName<'pr> for ruby_prism::ConstantPathTargetNode<'pr>
```

## Test plan

- [x] `bundle exec rake cargo:test` passes
- [x] `bundle exec rake cargo:lint` passes
- [x] Doctests pass

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/rust/ruby-prism/src/node_ext.rs
+++ b/rust/ruby-prism/src/node_ext.rs
@@ -1,0 +1,250 @@
+//! Node extension methods for the prism parser.
+//!
+//! This module provides convenience methods on AST nodes that aren't generated
+//! from the config, mirroring Ruby's `node_ext.rb`.
+
+use std::fmt;
+
+use crate::{ConstantPathNode, ConstantPathTargetNode, ConstantReadNode, ConstantTargetNode, ConstantWriteNode, Node};
+
+/// Errors for constant path name computation.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ConstantPathError {
+    /// The constant path contains dynamic parts (e.g., `var::Bar::Baz`).
+    DynamicParts,
+    /// The constant path contains missing nodes (e.g., `Foo::`).
+    MissingNodes,
+}
+
+impl fmt::Display for ConstantPathError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::DynamicParts => {
+                write!(f, "Constant path contains dynamic parts. Cannot compute full name")
+            },
+            Self::MissingNodes => {
+                write!(f, "Constant path contains missing nodes. Cannot compute full name")
+            },
+        }
+    }
+}
+
+impl std::error::Error for ConstantPathError {}
+
+/// Trait for nodes that can compute their full constant name.
+///
+/// Implemented by constant-related nodes (`ConstantReadNode`,
+/// `ConstantWriteNode`, `ConstantTargetNode`, `ConstantPathNode`, and
+/// `ConstantPathTargetNode`).
+pub trait FullName<'pr> {
+    /// Returns the list of parts for the full name of this constant.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ConstantPathError`] if the path contains dynamic parts or
+    /// missing nodes.
+    fn full_name_parts(&self) -> Result<Vec<&'pr [u8]>, ConstantPathError>;
+
+    /// Returns the full name of this constant.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ConstantPathError`] if the path contains dynamic parts or
+    /// missing nodes.
+    fn full_name(&self) -> Result<Vec<u8>, ConstantPathError> {
+        let parts = self.full_name_parts()?;
+        let mut result = Vec::new();
+        for (index, part) in parts.iter().enumerate() {
+            if index > 0 {
+                result.extend_from_slice(b"::");
+            }
+            result.extend_from_slice(part);
+        }
+        Ok(result)
+    }
+}
+
+/// Computes `full_name_parts` for a `Node` by dispatching to the appropriate
+/// `FullName` implementation.
+#[allow(clippy::option_if_let_else)]
+fn full_name_parts_for_node<'pr>(node: &Node<'pr>) -> Result<Vec<&'pr [u8]>, ConstantPathError> {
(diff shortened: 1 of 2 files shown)


rust/ruby-prism/src/node_ext.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// ```
+    /// # use ruby_prism::parse;
+    /// let result = parse(b"Foo");
+    /// let stmt = result.node().as_program_node().unwrap()
+    ///     .statements().body().iter().next().unwrap();
+    /// let constant = stmt.as_constant_read_node().unwrap();
+    /// assert_eq!(constant.full_name_parts(), vec!["Foo"]);
+    /// ```
+    #[must_use]
+    pub fn full_name_parts(&self) -> Vec<Cow<'pr, str>> {
```
kddnewton: Why copy on write here?

rust/ruby-prism/src/node_ext.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// # use ruby_prism::parse;
+    /// let result = parse(b"Foo");
+    /// let stmt = result.node().as_program_node().unwrap()
+    ///     .statements().body().iter().next().unwrap();
+    /// let constant = stmt.as_constant_read_node().unwrap();
+    /// assert_eq!(constant.full_name(), "Foo");
+    /// ```
+    #[must_use]
+    pub fn full_name(&self) -> Cow<'pr, str> {
+        String::from_utf8_lossy(self.name().as_slice())
```
kddnewton: I don't feel great about utf8 here, it's not guaranteed by the parser (there are ~90 encodings supported).

sei40kr:
Thanks for the review!

> Why copy on write here?

The `Cow<'pr, str>` was not an intentional API choice — it was just the return type of `String::from_utf8_lossy()` leaking through. When the bytes are valid UTF-8 it returns `Cow::Borrowed(&str)`, and when they're not it returns `Cow::Owned(String)` with replacement characters.

> I don't feel great about utf8 here, it's not guaranteed by the parser (there are ~90 encodings supported).

Agreed. I've updated the PR to return raw bytes (`&'pr [u8]` / `Vec<u8>`) instead, removing all `from_utf8_lossy` calls. This is consistent with the existing `ConstantId::as_slice()` API and avoids silently corrupting non-UTF-8 data.

rust/ruby-prism/src/node_ext.rs
```diff
@@ -0,0 +1,444 @@
...
+    /// # Errors
+    ///
+    /// Returns [`ConstantPathError::DynamicParts`] if the path contains
+    /// dynamic parts, or [`ConstantPathError::MissingNodes`] if the path
+    /// contains missing nodes.
+    pub fn full_name_parts(&self) -> Result<Vec<&'pr [u8]>, ConstantPathError> {
+        let mut parts = Vec::new();
+        let mut current: Option<Node<'pr>> = Some(self.as_node());
+
+        while let Some(ref node) = current {
```
kddnewton: Can we make full_name and full_name_parts into a trait? It is an interface, and it should make the ConstantPathNode implementation simpler.

rust/ruby-prism/src/node_ext.rs
sei40kr: The `Cow<'pr, str>` was a byproduct of using `String::from_utf8_lossy()`, which returns `Cow<'_, str>`. It was not an intentional design choice. Addressed in ec991ae by switching to `&'pr [u8]`, which removed the need for Cow entirely.

rust/ruby-prism/src/node_ext.rs
sei40kr: Good point. Fixed in ec991ae by switching from `String::from_utf8_lossy` / `Cow<str>` to raw `&'pr [u8]`, so we no longer assume any specific encoding.

rust/ruby-prism/src/node_ext.rs
sei40kr: Done in 33b4e6f. Extracted a `FullName<'pr>` trait with `full_name_parts` and a default `full_name` implementation. This simplified the `ConstantPathNode` impl significantly — replaced the iterative while-loop + `reverse()` + `is_stovetop()` pattern with recursive delegation via a `full_name_parts_for_node` helper.

PR #3890, opened 2026-01-29
[rust] add `Location` line/column and chop methods

sei40kr:
## Summary

- Add `start_line()` method to `Location` (Rust equivalent of Ruby's `Location#start_line`)
- Add `end_line()` method to `Location` (Rust equivalent of Ruby's `Location#end_line`)
- Add `start_column()` method to `Location` (Rust equivalent of Ruby's `Location#start_column`)
- Add `end_column()` method to `Location` (Rust equivalent of Ruby's `Location#end_column`)
- Add `chop()` method to `Location` (Rust equivalent of Ruby's `Location#chop`)

Line/column computation is implemented in Rust using binary search on the line offsets array, matching the approach used by the Ruby and Java bindings.

## Public API Changes

```
Removed items from the public API
=================================
(none)

Changed items in the public API
===============================
(none)

Added items to the public API
=============================
+pub const fn ruby_prism::Location<'pr>::chop(&self) -> Self
+pub fn ruby_prism::Location<'pr>::end_column(&self) -> u32
+pub fn ruby_prism::Location<'pr>::end_line(&self) -> i32
+pub fn ruby_prism::Location<'pr>::start_column(&self) -> u32
+pub fn ruby_prism::Location<'pr>::start_line(&self) -> i32
```

## Test plan

- [x] `bundle exec rake cargo:test`
- [x] `bundle exec rake cargo:lint`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/include/prism/util/pm_line_offset_list.h
+++ b/include/prism/util/pm_line_offset_list.h
@@ -96,7 +96,7 @@ int32_t pm_line_offset_list_line(const pm_line_offset_list_t *list, uint32_t cur
  * @param start_line The line to start counting from.
  * @return The line and column of the given offset.
  */
-pm_line_column_t pm_line_offset_list_line_column(const pm_line_offset_list_t *list, uint32_t cursor, int32_t start_line);
+PRISM_EXPORTED_FUNCTION pm_line_column_t pm_line_offset_list_line_column(const pm_line_offset_list_t *list, uint32_t cursor, int32_t start_line);
 
 /**
  * Free the internal memory allocated for the list.
--- a/rust/ruby-prism-sys/build/main.rs
+++ b/rust/ruby-prism-sys/build/main.rs
@@ -9,6 +9,7 @@ fn main() {
 
     let ruby_build_path = prism_lib_path();
     let ruby_include_path = prism_include_path();
+    emit_rerun_hints(&ruby_include_path);
 
     // Tell cargo/rustc that we want to link against `libprism.a`.
     println!("cargo:rustc-link-lib=static=prism");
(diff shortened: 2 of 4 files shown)


kddnewton:
Let's wait on #3860 for this.

rust/ruby-prism-sys/src/lib.rs
```diff
@@ -35,3 +35,25 @@ mod bindings {
...
+    /// parser.
+    pub line: i32,
+    /// The 0-indexed column number in bytes.
+    pub column: u32,
+}
+
+extern "C" {
+    /// Return the line and column number for the given byte offset relative to
+    /// the beginning of the source.
+    pub fn pm_newline_list_line_column(
```
kddnewton: This doesn't exist anymore, because it was renamed. Please rebase.

rust/ruby-prism-sys/src/lib.rs
```diff
@@ -35,3 +35,25 @@ mod bindings {
...
+/// Line and column information for a given byte offset relative to the
+/// beginning of the source.
+#[repr(C)]
+#[derive(Debug, Clone, Copy)]
+pub struct pm_line_column_t {
+    /// The 1-indexed line number relative to the start line configured on the
+    /// parser.
+    pub line: i32,
+    /// The 0-indexed column number in bytes.
+    pub column: u32,
```
froydnj: The commit message for this explains that these are troublesome for the WASM target...but why are they troublesome, and can that be fixed instead?  They really shouldn't be treated any differently than other structures, and it would be better to not maintain these declarations by hand.  I trust bindgen more than manual declarations.

kddnewton: Agreed, I don't want us manually maintaining parallel structures.

kddnewton (changes requested):

sei40kr:
Updated based on review feedback:

- **Removed manual C struct declarations** as suggested by @froydnj and @kddnewton. Instead of maintaining parallel `pm_line_column_t` / `pm_line_offset_list_line_column` declarations in Rust, the line/column computation is now implemented directly in Rust using binary search on the `line_offsets` array. This matches the approach used by the Ruby and Java bindings.
- **Rebased onto main** (the `pm_newline_list` → `pm_line_offset_list` rename is no longer relevant since we don't call the C function).
- **Preserved `emit_rerun_hints`** from @kddnewton's rebase.

PR #3891, opened 2026-01-29
Handle String-like and IO-like objects for Ripper.sexp

eregon:
* RSpec relies on this in https://github.com/rspec/rspec/blob/rspec-support-v3.13.6/rspec-support/lib/rspec/support/source.rb#L57 which is given an RSpec::Support::EncodedString.
* CI failure caused by this on truffleruby: https://github.com/sporkmonger/addressable/actions/runs/21457707372/job/61802608154#step:7:14

### DIFF
--- a/lib/prism/translation/ripper.rb
+++ b/lib/prism/translation/ripper.rb
@@ -480,7 +480,17 @@ module Prism
 
       # Create a new Translation::Ripper object with the given source.
       def initialize(source, filename = "(ripper)", lineno = 1)
-        @source = source
+        if source.is_a?(IO)
+          @source = source.read
+        elsif source.respond_to?(:gets)
+          @source = +""
+          while line = source.gets
+            @source << line
+          end
+        else
+          @source = source.to_str
+        end
+
         @filename = filename
         @lineno = lineno
         @column = 0
--- a/test/prism/ruby/ripper_test.rb
+++ b/test/prism/ruby/ripper_test.rb
@@ -145,6 +145,36 @@ module Prism
       assert_equal(Ripper.tokenize(source), Translation::Ripper.tokenize(source))
     end
 
+    def test_sexp_coercion
+      string_like = Object.new
+      def string_like.to_str
+        "a"
+      end
+      assert_equal Ripper.sexp(string_like), Translation::Ripper.sexp(string_like)
+
+      File.open(__FILE__) do |file1|
+        File.open(__FILE__) do |file2|
+          assert_equal Ripper.sexp(file1), Translation::Ripper.sexp(file2)
+        end
+      end
+
+      File.open(__FILE__) do |file1|
+        File.open(__FILE__) do |file2|
+          object1_with_gets = Object.new
+          object1_with_gets.define_singleton_method(:gets) do
+            file1.gets
+          end
+
+          object2_with_gets = Object.new
+          object2_with_gets.define_singleton_method(:gets) do
+            file2.gets
+          end
+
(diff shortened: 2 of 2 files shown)


test/prism/ruby/ripper_test.rb
```diff
@@ -145,6 +145,12 @@ def test_tokenize
       assert_equal(Ripper.tokenize(source), Translation::Ripper.tokenize(source))
     end
 
+    def test_sexp_coercion
+      string_like = Object.new
+      def string_like.to_str; "a"; end
+      assert_equal [:program, [[:vcall, [:@ident, "a", [1, 0]]]]], Translation::Ripper.sexp(string_like)
```
Earlopain: Doesn't really matter what the result is, we just want the same behaviour as native

```suggestion
      assert_equal Ripper.sexp(string_like), Translation::Ripper.sexp(string_like)
```

eregon:
> It looks like the rules are a bit more complex: https://github.com/ruby/ruby/blob/acc414572c213f2690eee56bac3d2355a24e5930/ext/ripper/ripper_init.c.tmpl#L357-L376
> 
> Do you want to implement those also? File should be easy at least.

I see. If you want to do it feel free to push to this branch, otherwise I can get to it tomorrow or so.

Earlopain:
Tomorrow is fine

eregon:
Done, now it also handles IO object (= T_FILE) and `#gets` objects, same logic as upstream.

lib/prism/translation/ripper.rb
```diff
@@ -480,7 +480,17 @@ def self.lex_state_name(state)
 
       # Create a new Translation::Ripper object with the given source.
       def initialize(source, filename = "(ripper)", lineno = 1)
-        @source = source
+        if source.is_a?(IO)
+          @source = source.read
```
Earlopain, line 484: This could be backed by `Prism.parse_stream` but since we need the whole source anyways I think this is fine. BTW, wouldn't work for the `gets` case since prism also expects `eof?` to be present

lib/prism/translation/ripper.rb
```diff
@@ -480,7 +480,17 @@ def self.lex_state_name(state)
...
+        elsif source.respond_to?(:gets)
+          @source = +""
+          while line = source.gets
+            @source << line
+          end
+        else
+          @source = source.to_str
+        end
```
Earlopain, line 492: Maybe like this to skip `respond_to` when it's already a string? It's better for cruby but a bit more complex. I remeber something like this being done in the json gem.

```suggestion
        case source
        when String then @source = source
        when IO     then @source = source.read
        else
          if source.respond_to?(:gets)
            @source = +""
            while line = source.gets
              @source << line
            end
          else
            @source = source.to_str
          end
        end
```

Earlopain, line 492: Maybe like this to skip respond_to when it's already a string? It's better for cruby but a bit more complex. I remeber something like this being done in the json gem.

```suggestion
        case source
        when String then @source = source
        when IO     then @source = source.read
        else
          if source.respond_to?(:gets)
            @source = +""
            while line = source.gets
              @source << line
            end
          else
            @source = source.to_str
          end
        end
```

eregon: I think the `is_a?` and `respond_to?` are reasonably fast on CRuby, so I would keep it simple as it is.
Not a big fan of `case/when` as it adds extra indentation.

kddnewton (approved):

kddnewton:
Just following up on this since I saw the mention of RSpec: https://github.com/rspec/rspec/pull/303.

PR #3892, opened 2026-01-29
RB_GC_GUARD() the VALUE for RSTRING_PTR() to ensure the char* does not move

eregon:
* See https://github.com/ruby/prism/pull/3886#discussion_r2741579984.


eregon:
@Earlopain Nice, I was thinking indeed there is probably a more direct way to access the value.

This is using https://github.com/ruby/ruby/blob/3237be163c313af0b6626b209e55bdb0b33c9f0f/include/ruby/version.h#L107-L113 for reference.

PR #3894, opened 2026-02-01
Add xfree_sized and xrealloc_sized

byroot:
### Context

Ruby internals have had a `ruby_sized_xfree` function for a very longtime. Until recently it was just used to feed the freed size into the GC statistics, to allow it to be smarter as of when to trigger.

However C23 does introduce `free_sized`: https://en.cppreference.com/w/c/memory/free_sized.html, which when used allow allocator to either deallocate faster or be more secure by aborting when the size doesn't match.

It is however important to note that passing an incorrect size is undefined behavior, so the allocator could just as well leak or crash.

Very recently `glibc 2.43` added support for `free_sized`: https://sourceware.org/pipermail/libc-announce/2026/000052.html, and other common allocators like `jemalloc` have supported it for years.
However if the allocator doesn't support it, it's easily shimed with a simple macro that discard the size.

The last few days I've changed a majority of `xfree` calls inside MRI to use `ruby_sized_xfree`, and added some debug checks inside it to ensure the passed size is correct, and that surfaced a few minor bugs (mostly strings losing a few bytes of capacity when switching between encodings). (e.g. https://github.com/ruby/ruby/pull/16010)

Now most of Ruby's own codebase have been converted, and most of the remaining unsized frees are within `prism` and similarly vendored codebases. I haven't yet hooked MRI to the allocator, for now it's still only used for GC statistics.

### `xrealloc_sized`

I'm not sure why, but C23 standard doesn't defined `realloc_sized`, but similarly Ruby had `ruby_sized_xrealloc` for a while, and I think it does make a lot of sense for that purpose given realloc is essentially `malloc + free`.

### Proposal

I'd like to also convert Prism to use sized frees as well, so that when compiled inside Ruby it can use `ruby_sized_xfree`, hence play better with the GC and potentially benefit from C23 `free_sized` performance benefits in the future.

In this PR I implemented debug checks for the allocator sizes, and converted a few call sites to the new API, as well as added a couple convenience macros that make using `free_sized` much easier.

A side benefit of these debug checks is that if you accidentally use raw `free` or raw `malloc`, it will blow up.

If there is interest from the maintainers, I can convert the rest of the prism codebase.

### DIFF
--- a/include/prism/defines.h
+++ b/include/prism/defines.h
@@ -161,10 +161,12 @@
  * ```
  * #ifndef PRISM_XALLOCATOR_H
  * #define PRISM_XALLOCATOR_H
- * #define xmalloc      my_malloc
- * #define xrealloc     my_realloc
- * #define xcalloc      my_calloc
- * #define xfree        my_free
+ * #define xmalloc          my_malloc
+ * #define xrealloc         my_realloc
+ * #define xcalloc          my_calloc
+ * #define xfree            my_free
+ * #define xrealloc_sized   my_realloc_sized // (optional)
+ * #define xfree_sized      my_free_sized    // (optional)
  * #endif
  * ```
  */
@@ -202,6 +204,28 @@
          */
         #define xfree free
     #endif
+#endif
+
+#ifndef xfree_sized
+/**
+ * The free_sized function that should be used. This can be overridden with the
+ * PRISM_XALLOCATOR define.
+ * If not defined, defaults to calling xfree.
+ */
+    #define xfree_sized(p, s) xfree(((void)(s), (p)))
+#endif
+
+#ifndef xrealloc_sized
+/**
+ * The xrealloc_sized function that should be used. This can be overridden with the
+ * PRISM_XALLOCATOR define.
+ * If not defined, defaults to calling xrealloc.
+ */
+    #define xrealloc_sized(p, ns, os) xrealloc((p), ((void)(os), (ns)))
+#endif
+
+#ifdef PRISM_BUILD_DEBUG
+    #include "prism/debug_allocator.h"
 #endif
 
 /**
--- a/Doxyfile
+++ b/Doxyfile
@@ -24,6 +24,7 @@ OUTPUT_DIRECTORY       = doc
 JAVADOC_AUTOBRIEF      = YES
 OPTIMIZE_OUTPUT_FOR_C  = YES
 INPUT                  = src src/util include include/prism include/prism/util
+EXCLUDE                = include/prism/debug_allocator.h
 HTML_OUTPUT            = c
 SORT_MEMBER_DOCS       = NO
 GENERATE_LATEX         = NO
--- a/ext/prism/extension.c
+++ b/ext/prism/extension.c
@@ -413,7 +413,7 @@ dump(int argc, VALUE *argv, VALUE self) {
     if (options.freeze) rb_obj_freeze(value);
 
 #ifdef PRISM_BUILD_DEBUG
-    xfree(dup);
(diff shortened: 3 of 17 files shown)


include/prism/defines.h
```diff
@@ -204,6 +206,47 @@
     #endif
 #endif
 
+#ifndef xfree_sized
```
kddnewton, line 209: I think these should be in the else for PRISM_XALLOCATOR, in case folks want to define that, but maybe I'm not following it correctly?

byroot: If they are defined in `PRISM_XALLOCATOR` they won't be overritten here.

The idea is that the `sized` variants are shimed perfectly by just calling the normal function without the size

include/prism/defines.h
```diff
@@ -204,6 +206,47 @@
...
+#endif
+
+/**
+ * Convenience macro to call xfree_sized with a pointer to a fixed size object.
+ * Example:
+ *      pm_call_node_t *target = xmalloc(sizeof(pm_call_node_t));
+ *      ...
+ *      PM_FREE_SIZED(target);
+ */
+#define PM_FREE_SIZED(p) xfree_sized((p), (sizeof(*(p))))
```
kddnewton: I'm uncomfortable with this macro, because it means we're assuming every allocation points to one object, and that object matches exactly sizeof the struct. So for example this wouldn't work if you passed in an array, or an object with extra memory off the end (like our constant pools), or even if you passed in void*. I'd be more comfortable not defining this macro, and instead using the slow path for the case where you don't explicitly pass a size.

include/prism/defines.h
```diff
@@ -204,6 +206,47 @@
...
+
+/**
+ * Convenience macro to call xfree_sized with a pointer to an array of fixed size objects.
+ * Example:
+ *      size_t capacity = 12;
+ *      pm_call_node_t *targets = xcalloc(capacity, sizeof(pm_call_node_t));
+ *      ...
+ *      PM_FREE_SIZED_N(target, capacity);
+ */
+#define PM_FREE_SIZED_N(p, n) xfree_sized((p), (sizeof(*(p))) * n)
```
kddnewton: Same here, I'm uncomfortable using sizeof for the reasons stated above.

include/prism/defines.h
byroot: Well, you'd supposed to use it in cases it works, and debug mode ensure you didn't get it wrong.

But I'm fine with removing it. I'll do that tomorrow.

byroot:
Alright, I removed the two macros, and converted almost all `xfree` and `xrealloc` call sites.

There is 5 `xfree` callsites left, for 4 of them it's freeing raw strings for which the buffer size wasn't tracked.

The last one is `pm_integer_free`, in theory the size if tracked in `integer->length` but apparently it's not always correct, and I haven't figured out why (but would be valuable to)

include/prism/defines.h
```diff
@@ -204,6 +206,28 @@
...
+ * The xrealloc_sized function that should be used. This can be overridden with the
+ * PRISM_XALLOCATOR define.
+ * If not defined, defaults to calling xrealloc.
+ */
+    #define xrealloc_sized(p, ns, os) xrealloc((p), ((void)(os), (ns)))
+#endif
+
+#ifdef PRISM_BUILD_DEBUG
+    #include "prism/debug_allocator.h"
+#endif
```
byroot, line 229: Right now `PRISM_BUILD_DEBUG` is opt-in, and only set on CI for valgrind. But perhaps it should be set by default when running tests, as it would be better for contributors to not have to know about it.

PR #3896, opened 2026-02-02
Add escaping for doxygen

Earlopain:
Many characters have special meaning and break formatting

Before:
grafik

After:
grafik

@byroot can you check if this fixes https://github.com/ruby/prism/pull/3895 for you? It doesn't break when I run it locally even though I use the same version as you.

### DIFF
--- a/templates/template.rb
+++ b/templates/template.rb
@@ -49,6 +49,14 @@ module Prism
       end
     end
 
+    # This module contains methods for escaping characters in Doxygen comments.
+    module Doxygen
+      # Similar to /verbatim ... /endverbatim but doesn't wrap the result in a code block.
+      def self.verbatim(value)
+        value.gsub(/[\.*%!`#<>_+-]/, '\\\\\0')
+      end
+    end
+
     # A comment attached to a field or node.
     class ConfigComment
       attr_reader :value
--- a/templates/include/prism/ast.h.erb
+++ b/templates/include/prism/ast.h.erb
@@ -22,7 +22,7 @@
  */
 typedef enum pm_token_type {
 <%- tokens.each do |token| -%>
-    /** <%= token.comment %> */
+    /** <%= Prism::Template::Doxygen.verbatim(token.comment) %> */
     PM_TOKEN_<%= token.name %><%= " = #{token.value}" if token.value %>,
 
 <%- end -%>


byroot:
Yep, that works for me.

templates/template.rb
```diff
@@ -49,6 +49,13 @@ def self.escape(value)
       end
     end
 
+    # This module contains methods for escaping characters in Doxygen comments.
+    module Doxygen
+      def self.escape(value)
+        value.gsub(/[\.*%!`#<>_+-]/, '\\\\\0')
```
byroot, line 56: I'm not certain this is correct though, as e.g. mainy comments do contain markdown.

The one comment I had issue with had a single backtick, but many other have proper mardown codeblocks.

byroot: e.g. I think your PR breaks: https://github.com/ruby/prism/blob/609cd5b74f5f3dd985440100f8587861e7ef9e7f/config.yml#L707

Earlopain: Ah, right. I tried `/verbatim` but that wraps the result in a codeblock. I think I can just rename this method and it should be ok.

Earlopain: (It's only applied to that one specific place where token comments are shown)

byroot: I still don't follow. These comment do contain doxygen code (markdown really but WTV) so it shouldn't be fully escaped.

IMO it's not that we're missing some automated escaping, it's that the YAML file do contain invalid markdown/doxygen, the fix can only be manual.

Earlopain: It's only for these comments https://github.com/ruby/prism/blob/62c7300c1faaac481bef098a7648e05a5c6fbb28/config.yml#L327-L655 (https://ruby.github.io/prism/c/ast_8h.html#adb9ccf5e858ff029b17095d6298cf42b)

Makes sense for these I think, there is no markdown (yet). You would have to escape many other places manually otherwise. I tried that but doxygen doesn't really do what I want it to.

I'm also fine with your PR. I never looked at the generated docs before so it doesn't really matter that much to me.

byroot: Ah I see. I guess that currently work since none of these comments do use Doxygen formatting, but would be a problem in the future if one wanted to.

You have more context than me so up to you (or to Kevin).

kddnewton (approved):
Yeah this is fine.

Issue #3837, opened 2026-01-07
parse difference for heredoc with multiple consecutive line continuations


froydnj:
I think this is like #2842, but with more line continuations:

```ruby
description = <<~EOS
  This line\
  \
  Another line.

  Extra rationale.
EOS
```

`bin/prism ripper` gives:

```
Ripper:
[:program,
 [:stmts_add,
  [:stmts_new],
  [:assign,
   [:var_field, [:@ident, "description", [1, 0]]],
   [:string_literal,
    [:string_add,
     [:string_add,
      [:string_add, [:string_add, [:string_add, [:string_content], [:@tstring_content, "This line\\\n", [2, 2]]], [:@tstring_content, "  \\\n", [3, 0]]], [:@tstring_content, "Another line.\n", [4, 2]]],
      [:@tstring_content, "\n", [5, 0]]],
     [:@tstring_content, "Extra rationale.\n", [6, 2]]]]]]]
Prism:
[:program,
 [:stmts_add,
  [:stmts_new],
  [:assign,
   [:var_field, [:@ident, "description", [1, 0]]],
   [:string_literal,
    [:string_add,
     [:string_add, [:string_add, [:string_add, [:string_content], [:@tstring_content, "  This line\\\n", [2, 0]]], [:@tstring_content, "Another line.\n", [4, 2]]], [:@tstring_content, "\n", [5, 0]]],
     [:@tstring_content, "Extra rationale.\n", [6, 2]]]]]]]
Output is not identical
```

The second line continuation is nowhere to be found in the prism output.  I'm not sure how deeply you care about this, given that the ultimate output is unchanged, but it was a little surprising that the parse tree didn't really correspond to the source for this bit.

Found while debugging [fables-tables/rubyfmt#772](https://github.com/fables-tales/rubyfmt/issues/772). (cc @reese)

kddnewton:
I'm not sure if I should consider this a bug in our ripper translation layer or our AST. I could see a case for inserting an empty string node that contains the content. I'll take a look.

froydnj:
I could have been clearer in my description -- the second line continuation doesn't even show up in `bin/prism parse` output, so I think this is definitely an AST issue.

kddnewton:
Yeah, I mean it's an issue insofar as you're trying to recreate the original source, not for use cases like compilers since it doesn't impact the actual value of the string.

Until this is fixed, you can probably use the fact that when subsequent string parts are not continuous you know there's an empty string that was dropped from the output.

[...]

PR #3897, opened 2026-02-02
Preserve line-continuation only in dedent heredocs

Earlopain:
Closes https://github.com/ruby/prism/issues/3837

While these lines are whitespace only from a runtime perspective, the line continuation is significant for AST consumers.

Sort of a followup to https://github.com/ruby/prism/commit/faab217d9382186133862ce4d5ba9a9e090a9550

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -15822,6 +15822,19 @@ parse_heredoc_dedent_string(pm_string_t *string, size_t common_whitespace) {
 }
 
 /**
+ * If we end up trimming all of the whitespace from a node and it isn't
+ * part of a line continuation, then we'll drop it from the list entirely.
+ */
+static inline bool
+heredoc_dedent_discard_string_node(pm_parser_t *parser, pm_string_node_t *string_node) {
+    if (string_node->unescaped.length == 0) {
+        const uint8_t *cursor = parser->start + PM_LOCATION_START(&string_node->content_loc);
+        return pm_memchr(cursor, '\\', string_node->content_loc.length, parser->encoding_changed, parser->encoding) == NULL;
+    }
+    return false;
+}
+
+/**
  * Take a heredoc node that is indented by a ~ and trim the leading whitespace.
  */
 static void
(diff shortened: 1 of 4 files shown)


kddnewton:
I'm assuming this has no impact on generated code right?

Earlopain:
Yeah, from what I tested this gives the exact same instructions.

src/prism.c
```diff
@@ -15821,6 +15821,19 @@ parse_heredoc_dedent_string(pm_string_t *string, size_t common_whitespace) {
...
 
+/**
+ * If we end up trimming all of the whitespace from a node and it isn't
+ * part of a line continuation, then we'll drop it from the list entirely.
+ */
+static inline bool
+heredoc_dedent_discard_string_node(pm_parser_t *parser, pm_string_node_t *string_node) {
+    if (string_node->unescaped.length == 0) {
+        const uint8_t *cursor = parser->start + PM_LOCATION_START(&string_node->content_loc);
+        return pm_memchr(cursor, '\\', string_node->content_loc.length, parser->encoding_changed, parser->encoding) == NULL;
```
kddnewton, line 15832: I know you've checked this thoroughly, but I'm just uncomfortable with this because it seems like it _could_ include other kinds of strings, but by virtue of the dedent algorithm we're cool.

Just to be super clear, are we saying that we want to include the string node if it matches `[\s\t]*\\\r?\n`, but otherwise not?

If that's the case, I think I'd prefer that we explicitly check that pattern here, without relying on `pm_memchr`.

Earlopain: > it seems like it could include other kinds of strings, but by virtue of the dedent algorithm we're cool.

Hm, I wouldn't say that is the case, we only take this branch when the unescaped string is empty. That's only possible when the string indeed is dedenting whitespace only, anything else would result in it having some value.

Otherwise, your regexp is pretty much what I am checking for, with the assumption that just checking `\` is good enough for the above reason. Otherwise, this is the implementation I came up with:

```c
static inline bool
heredoc_dedent_discard_string_node(pm_parser_t *parser, pm_string_node_t *string_node) {
    if (string_node->unescaped.length == 0) {
        const uint8_t *start = parser->start + PM_LOCATION_START(&string_node->content_loc);
        const uint8_t *end = parser->start + PM_LOCATION_END(&string_node->content_loc);

        while (start < end && pm_char_is_inline_whitespace(*start)) start++;
        if (start < end && *start == '\\') {
            start++;
            if (start < end && *start == '\r') {
                start++;
            }
            if (start < end && *start == '\n' && start + 1 == end) {
                // The string is whitespace only and ends with a line continuation.
                return false;
            }
        }
        // The string may be empty, or all whitespace was removed via dedenting.
        return true;
    }
    // The string has content. It may be whitespace not removed via dedenting or any other characters.
    return false;
}
```

Do you prefer this still?

Earlopain: What do you think @kddnewton? Should I go with this or keep as is? I don' think what I have in the comment is very clear

PR #3898, opened 2026-02-03
Small CI maintenance

Earlopain:
* JRuby should be fixed now
* Windows builds should be available


Earlopain:
Ah, it's too early. setup-ruby hasn't been released yet.

Also, https://github.com/IBM/actionspz/issues/75 ...

PR #3902, opened 2026-02-03
Large update to WASM Prism backend with Chicory

headius:
This PR includes a large set of changes to update the WASM backend using latest versions of Chicory and help from @andreaTP to clean it up and get it ready for a public release.

The changes here should not affect the existing native-library Java wrapper, but we will need to agree upon a package name and Maven groupID that represent a domain we have some control over (org.prism does not qualify since prism.org is completely unrelated to this project).

I would like to get Maven artifacts published for the native and WASM wrappers before JRuby 10.1 is released in the next few weeks.


kddnewton (commented):
I'm generally fine with all of these changes. Would ruby-lang.org work?

headius:
@kddnewton Something based on that domain would be fine, but no hyphens in Java package names. Too bad `ruby.org` is being squatted.

I would propose the following:

* Maven Group ID and Artifact ID: "org.ruby-lang" and "prism"
* Package name: "org.ruby_lang.prism"

We'd need to register the group ID with the Maven Central folks (https://central.sonatype.org/register/namespace/#for-dns) and make sure anyone who wants to be able to publish Maven artifacts has access.

For the moment I've got the WASM bits all under the following:

* Group and Artifact: "org.jruby", "chicory-prism" (could be changed to a ruby-lang group and "prism-wasm" or similar).
* Package name: org.jruby.parser.prism for general bits and org.jruby.parser.prism.wasm for the wasm logic specifically.

headius:
@kddnewton @enebo I believe this could be merged now and additional work done as smaller one-off PRs.

This PR includes:

* All updates for Chicory and the WASM build to create a fully functional AOT WASM parser library.
* Move all Java code out of "org.prism" and into appropriate packages. General Prism wrappers are moved to "org.ruby_lang.prism" and the JRuby-specific WASM library wrapper is under "org.jruby.parser.prism.wasm".
* Prepare the WASM wrapper library for release under Maven coordinates "org.jruby:chicory-prism" (name may change after discussion)

I have not:

* Set up a separate Maven build and artifact for the base Prism wrapper library (the native version and code under `java/`).
* Registered a group name for publishing the base wrapper library (org.ruby-lang group ID will need to be registered with Maven Central)

This PR is now complete and sufficient to get an artifact for the WASM wrapper released to Maven Central for JRuby's use.

java-wasm/pom.xml
```diff
@@ -81,7 +86,7 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.14.1</version>
         <configuration>
-          <release>11</release>
+          <release>21</release>
```
eregon, line 89: Is it OK to bump files under `java/` to 21 too in the Rakefile:
https://github.com/ruby/prism/blob/55948fe69992ef5c6615d9fb395384fd7a00d171/Rakefile#L48-L49

I recall we had to downgrade to Java 8 (for JRuby 9 IIRC) in https://github.com/truffleruby/truffleruby/commit/9acddea1dfba3a41034abfc33dfacfd98fcea07b, is that still necessary?

headius: JRuby 9.x is EOL and there may not be many remaining users that are also using the prism gem, but let's keep it Java 8 compatible for now. It doesn't hurt anything. 

The wasm piece is unlikely to ever be used by 9.4, so I choose JRuby 10's minimum have level there.

eregon: > It doesn't hurt anything.

Well it prevents using any feature from Java 9 to 21.
I don't recall what it was but there were some changes needed to make it compatible with Java 8.
But yes, it's not a big deal for now.

kddnewton:
Hey @headius — just want to check if this is good to go? It's still marked as draft so I wasn't sure if you had other stuff to put in.

headius:
@kddnewton Sorry about that! Yes, this can be merged. There will be additional clean-up work as we decide how to publish the base artifact (currently the only Maven artifact that's publishable is the JRuby-specific WASM library).

kddnewton:
@headius sorry bug could you rebase? Once that's done happy to merge

headius:
@kddnewton will do!

headius:
@kddnewton It is done!

headius:
Future rebases should be trivial so this can be merged whenever you are satisfied. I have re-tested the wasm library with JRuby and it appears to be functioning correctly. Once this gets merged I will be pushing a 0.0.1 version of the wasm artifact and setting up some JRuby CI to test it more thoroughly (our support for Prism as a primary parser has lagged somewhat in the last year).

The other part of this plan is to release a wasm version of the gem, so that JRuby users can install and package applications without requiring any external native library. I'll start putting that PR together after this one is finalized.

headius:
@kddnewton Thank you!

PR #3909, opened 2026-02-06
Remove pack parsing

### LINKED ISSUE #3907: Status of `pack`/`unpack` parsing
[Prism contains code to parse pack/unpack directives](https://github.com/ruby/prism/blob/55948fe69992ef5c6615d9fb395384fd7a00d171/include/prism/pack.h). These haven't changed in a while in ruby. But for 4.1 https://bugs.ruby-lang.org/issues/21785 and https://bugs.ruby-lang.org/issues/21796 may happen which adds two new directives.

Is anyone actually using it? ruby doesn't as far as I can tell and still uses its own implementation. Maybe some other implementation does?

Basically I am asking if this is still a goal of prism.

Earlopain:
It was never adopted by ruby or other implementations. Ruby 4.1 may add new directives, so now is a good time to remove it.

It was actually never properly documented (just shows `Prism::Pack` with no methods) so removing seems safe to do even now. I did a github code search and didn't find anyone using the ruby or c methods.

I'll merge this if no changes are requested.

Closes #3907


Earlopain:
https://github.com/ruby/ruby/pull/16085 to fix ruby CI, should be merged first. Looks like I won't do it after all. I asked in the bugtracker https://bugs.ruby-lang.org/issues/21839#note-7, let's see what happens with that.

eregon:
There is a small typo in the commit message: `Ruyb 4.1 `

kddnewton:
Yeah this is good. We just need to merge it into CRuby first to make sure references to the headers and such go away.

kddnewton:
@Earlopain can you rebase?

Issue #3911, opened 2026-02-08
Prism.lex_compat creates wrong on_sp token when used with heredoc and unclosed embexpr


tompng:
```ruby
Prism.lex_compat("<<A+B\n\#{C").value
# => [[[1, 0], :on_heredoc_beg, "<<A", BEG], [[1, 3], :on_sp, "+B\n\#{C", BEG]]
```

Found in IRB's truffleruby ci failure. Actual ruby code:
https://github.com/ruby/irb/blob/342f57b2d0a39c519a53e6d5a3938f6235119c9c/test/irb/test_ruby_lex.rb#L17-L21
```ruby
①+<<A-②
#{③*<<B/④
#{⑤&<<C|⑥
```
This failing test is not important in IRB anymore and I think it will be removed soon (in https://github.com/ruby/irb/pull/1160).

Earlopain:
Generally I would say that syntax invalid code not matching up is somewhat expected but the prism output doesn't look all that correct:

```rb
Prism.lex("<<A+B\n\#{C").value
=> 
[[HEREDOC_START(1,0)-(1,3)("<<A"), 1],
 [EMBEXPR_BEGIN(2,0)-(2,2)("\#{"), 1],
 [CONSTANT(2,2)-(2,3)("C"), 32],
 [EOF(2,3)-(2,3)(""), 32],
 [EOF(2,3)-(2,3)(""), 32]]
```

There's clearly some stuff missing and EOF appears twice. Seems like it isn't jumping back to the right place after looking for the heredoc end.

Earlopain:
I reverted the fix since it breaks some test in irb. I will take a closer look about what's going on with that.

tompng:
This is a result of `Prism.parse("<<D\n\#{").value` change before and after the revert, which seems the cause of irb test failure

a964703ba9c972388248a67f33a042208182cc74
```
@ ProgramNode (location: (1,0)-(1,3))
├── flags: ∅
├── locals: []
└── statements:
    @ StatementsNode (location: (1,0)-(1,3))
    ├── flags: ∅
    └── body: (length: 1)
        └── @ InterpolatedStringNode (location: (1,0)-(1,3))
            ├── flags: newline
            ├── opening_loc: (1,0)-(1,3) = "<<D"
            ├── parts: (length: 1)
            │   └── @ EmbeddedStatementsNode (location: (2,0)-(2,2))
            │       ├── flags: ∅
[...]

PR #3916, opened 2026-02-10
Remove unneeded exit conditions for `parse_stream`

Earlopain:
As I understand it, the code operates line-by-line. When finding `__END__`, it has to decide if it is the end marker or just appears as part of of a multi-line literal (see tests).

Basically, if we found `__END__` on a line by itself and there are syntax errors, it's impossible to be the end marker.

This is in preparation for https://github.com/ruby/prism/issues/3911. That issue is caused by `lex_modes` not being properly reset after reaching EOF during heredoc body parsing. It is still in `PM_LEX_EMBEXPR` mode, which causes it to not reset the parser to the heredoc identifier end.

Tests are just formatted so it's better to see what they are doing.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -22283,28 +22283,6 @@ pm_parse_stream_read(pm_buffer_t *buffer, void *stream, pm_parse_stream_fgets_t 
 }
 
 /**
- * Determine if there was an unterminated heredoc at the end of the input, which
- * would mean the stream isn't finished and we should keep reading.
- *
- * For the other lex modes we can check if the lex mode has been closed, but for
- * heredocs when we hit EOF we close the lex mode and then go back to parse the
- * rest of the line after the heredoc declaration so that we get more of the
- * syntax tree.
- */
-static bool
-pm_parse_stream_unterminated_heredoc_p(pm_parser_t *parser) {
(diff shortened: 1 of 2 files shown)


Earlopain:
I tested with the examples in https://github.com/ruby/prism/issues/1247 and it seems ok.

There is a difference for the following code:
```rb
1 +

__END__

foo
```

```sh
# Before (source stopped directly after __END__)
$ cat ../test.rb | ruby
-: -:3: syntax errors found (SyntaxError)
  1 | 1 +
  2 | 
> 3 | __END__
    | ^~~~~~~ unexpected end-of-input, assuming it is closing the parent top level context
    | ^~~~~~~ unexpected end-of-input; expected an expression after the operator

# After (source is now the entire code)
$ cat ../test.rb | ./miniruby
-: -:3: syntax errors found (SyntaxError)
  1 | 1 +
  2 | 
> 3 | __END__
    | ^~~~~~~ unexpected end-of-input, assuming it is closing the parent top level context
    | ^~~~~~~ unexpected end-of-input; expected an expression after the operator
  4 | 
  5 | foo

```

This however now matches the output when reading the code from file. So it seems fine.

PR #3918, opened 2026-02-11
Fix lexer skipping tokens when heredoc body is unterminated

### LINKED ISSUE #3911: Prism.lex_compat creates wrong on_sp token when used with heredoc and unclosed embexpr

Earlopain:
Fixes #3911

When we hit EOF and still have lex modes left, it means some content was unterminated.
Heredocs specifically have logic that needs to happen when the body finished lexing.
If we don't reset the mode back to how it was before, it will not continue lexing at the correct place.

Also skips the heredoc_end tokens that prism emits in the ripper translator. They are for bookkeeping only and ripper doesn't do so.

### DIFF
--- a/lib/prism/lex_compat.rb
+++ b/lib/prism/lex_compat.rb
@@ -768,21 +768,24 @@ module Prism
         source.byte_offset(line, column)
       end
 
-      # Add :on_sp tokens
-      tokens = insert_on_sp(tokens, source, result.data_loc, bom, eof_token)
+      tokens = post_process_tokens(tokens, source, result.data_loc, bom, eof_token)
 
       Result.new(tokens, result.comments, result.magic_comments, result.data_loc, result.errors, result.warnings, source)
     end
 
     private
 
-    def insert_on_sp(tokens, source, data_loc, bom, eof_token)
+    def post_process_tokens(tokens, source, data_loc, bom, eof_token)
       new_tokens = []
 
       prev_token_state = Translation::Ripper::Lexer::State[Translation::Ripper::EXPR_BEG]
       prev_token_end = bom ? 3 : 0
 
       tokens.each do |token|
+        # Skip missing heredoc ends.
+        next if token[1] == :on_heredoc_end && token[2] == ""
+
+        # Add :on_sp tokens.
         line, column = token[0]
         start_offset = source.byte_offset(line, column)
 
--- a/test/prism/ruby/ripper_test.rb
+++ b/test/prism/ruby/ripper_test.rb
@@ -81,6 +81,16 @@ module Prism
       define_method("#{fixture.test_name}_lex") { assert_ripper_lex(fixture.read) }
     end
 
+    def test_lex_ignored_missing_heredoc_end
+      ["", "-", "~"].each do |type|
+        source = "<<#{type}FOO\n"
+        assert_ripper_lex(source)
+
+        source = "<<#{type}'FOO'\n"
+        assert_ripper_lex(source)
+      end
+    end
+
     module Events
       attr_reader :events
 


test/prism/lex_test.rb
```diff
@@ -48,11 +48,94 @@ def test_parse_lex_file
...
+    def test_lex_interpolation_unterminated_with_content
+      # No trailing newline.
+      # FIXME: Emits EOL twice.
+      assert_lexed(<<~'RUBY'.strip, <<~'LEXED')
+        "#{C
+      RUBY
+        STRING_BEGIN(1,0)-(1,1)("\"")
+        EMBEXPR_BEGIN(1,1)-(1,3)("\#{")
+        CONSTANT(1,3)-(1,4)("C")
+        EOF(1,4)-(1,4)("")
```
Earlopain: This containing EOF twice is preexisting and not caused by this change

test/prism/lex_test.rb
```diff
@@ -48,11 +48,94 @@ def test_parse_lex_file
     end
 
     if RUBY_VERSION >= "3.3"
-      def test_lex_compare
-        prism = Prism.lex_compat(File.read(__FILE__), version: "current").value
-        ripper = Ripper.lex(File.read(__FILE__))
+      def test_lex_compat
+        source = "foo bar"
```
Earlopain, line 52: I added heredocs below where ripper doesn't do the state correctly. So I just have it parse some other source

src/prism.c
```diff
@@ -9856,6 +9862,14 @@ parser_lex(pm_parser_t *parser) {
             // We'll check if we're at the end of the file. If we are, then we
             // need to return the EOF token.
             if (parser->current.end >= parser->end) {
+                // We may be missing closing tokens. We should pop modes one by one
+                // to do the appropriate cleanup like moving next_start for heredocs.
+                // Only when no mode is remaining will we actually emit the EOF token.
+                if (parser->lex_modes.current->mode != PM_LEX_DEFAULT) {
+                    lex_mode_pop(parser);
+                    goto switch_lex_mode;
```
Earlopain: Unsure about this but it was the easiest way to accomplish it. There are already gotos in the method (`lex_next_token`) so maybe it is fine?

test/prism/lex_test.rb
```diff
@@ -48,11 +48,94 @@ def test_parse_lex_file
...
+        HEREDOC_END(3,0)-(3,0)("")
+        PLUS(1,3)-(1,4)("+")
+        CONSTANT(1,4)-(1,5)("B")
+        NEWLINE(1,5)-(2,0)("\n")
+        EOF(2,3)-(3,0)("\n")
+      LEXED
+    end
+
+    def assert_lexed(code, expected)
+      actual = Prism.lex(code).value.map { |token| token[0].pretty_inspect }
```
Earlopain: I don't think there will be many tests of this sort, for now writing them like this seems fine. I added some errors tests but not everything will be visible there (like double EOF). The error test highlights `+` after the heredoc identifer which it didn't do before so if you think that is enough I can remove this

src/prism.c
kddnewton: Could this just be return parser_lex?

test/prism/lex_test.rb
```diff
@@ -48,11 +48,94 @@ def test_parse_lex_file
...
+        ripper = Ripper.lex(source)
         assert_equal(ripper, prism)
       end
     end
+
+    def test_lex_interpolation_unterminated
+      assert_lexed(<<~'RUBY'.strip, <<~'LEXED')
+        "#{
+      RUBY
+        STRING_BEGIN(1,0)-(1,1)("\"")
```
kddnewton: I really don't like these kinds of snapshot tests because they're usually very difficult to update, and it's not clear what exactly is being tested. Could you instead lex the content, get the tokens, and make assertions against the specific behavior you're looking for?

src/prism.c
Earlopain: Ah, yeah

test/prism/lex_test.rb
Earlopain: I simplified it a bit. Token type was really the only thing I was looking for. Is this good or still too much?

PR #3922, opened 2026-02-12
Test some bundled gems

Earlopain:
They make extensive use of prism.


Earlopain:
Eh, don't really want to look into how to make this work right now

PR #3924, opened 2026-02-13
Fix lexing for unterminated strings/heredocs etc.

Earlopain:
When we hit EOF and still have lex modes left, it means some content was unterminated. Heredocs specifically have logic that needs to happen when the body finished lexing. If we don't reset the mode back to how it was before, it will not continue lexing at the correct place.

Followup to https://github.com/ruby/prism/pull/3918. We can't call into `parser_lex` since it resets token locations. So I went back to `goto`.

Closes https://github.com/ruby/prism/issues/3911

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -9783,6 +9783,12 @@ parser_lex(pm_parser_t *parser) {
     unsigned int semantic_token_seen = parser->semantic_token_seen;
     parser->semantic_token_seen = true;
 
+    // We'll jump to this label when we are about to encounter an EOF.
+    // If we still have lex_modes on the stack, we pop them so that cleanup
+    // can happen. For example, we should still continue parsing after a heredoc
+    // identifier, even if the heredoc body was syntax invalid.
+    switch_lex_modes:
+
     switch (parser->lex_modes.current->mode) {
         case PM_LEX_DEFAULT:
         case PM_LEX_EMBEXPR:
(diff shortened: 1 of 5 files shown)


src/prism.c
```diff
@@ -15433,7 +15447,7 @@ parse_string_part(pm_parser_t *parser, uint16_t depth) {
             pm_token_t opening = parser->previous;
             pm_statements_node_t *statements = NULL;
 
-            if (!match1(parser, PM_TOKEN_EMBEXPR_END)) {
+            if (!match3(parser, PM_TOKEN_EMBEXPR_END, PM_TOKEN_HEREDOC_END, PM_TOKEN_EOF)) {
```
Earlopain, line 15450: `#{` currently gives the interpolation a statement with `MissingNode`. That's not correct, the missing end token leads to syntax error. It also messes up the locations when it finds the synthetic heredoc end token. I don't think there are any other tokens to consider here, should be all hopefully.

```rb
# Before
Prism.parse("\"\#{").value.statements.body[0]
=> 
@ InterpolatedStringNode (location: (1,0)-(1,3))
├── flags: newline
├── opening_loc: (1,0)-(1,1) = "\""
├── parts: (length: 1)
│   └── @ EmbeddedStatementsNode (location: (1,1)-(1,3))
│       ├── flags: ∅
│       ├── opening_loc: (1,1)-(1,3) = "\#{"
│       ├── statements:
│       │   @ StatementsNode (location: (1,1)-(1,3))
│       │   ├── flags: ∅
│       │   └── body: (length: 1)
│       │       └── @ MissingNode (location: (1,1)-(1,3))
│       │           └── flags: ∅
│       └── closing_loc: (1,3)-(1,3) = ""
└── closing_loc: ∅

# After
Prism.parse("\"\#{").value.statements.body[0]
=> 
@ InterpolatedStringNode (location: (1,0)-(1,3))
├── flags: newline
├── opening_loc: (1,0)-(1,1) = "\""
├── parts: (length: 1)
│   └── @ EmbeddedStatementsNode (location: (1,1)-(1,3))
│       ├── flags: ∅
│       ├── opening_loc: (1,1)-(1,3) = "\#{"
│       ├── statements: ∅
│       └── closing_loc: (1,3)-(1,3) = ""
└── closing_loc: ∅
```

PR #3927, opened 2026-02-16
Fix C extension docs in ruby/ruby

Earlopain:
Because ruby/ruby doesn't use our rake task it also uses the default markup (rdoc), unless specified otherwise. https://github.com/ruby/prism/pull/3572 already added it explicitly to ruby files.

For C however, it is not so easy because of a rdoc bug: https://github.com/ruby/rdoc/issues/1597 Adding the comment would remove all documentation since it gets parsed as ruby.

So, specify it for each method individually. Remove it from the template since there is only one C file relevant for rdoc. Fixes this:

image

which is hopefully the last issue. I diffed the docs in ruby/prism and this creates no difference here.


kddnewton:
Honestly instead of doing this I'd be more keen to just use actual rdoc syntax

Earlopain:
Rdoc will switch the default to markdown in the future https://github.com/ruby/rdoc/?tab=readme-ov-file#markup-formats

No clue when that actually happens but in the future we can just remove all these comments (and in ruby files too)

PR #3929, opened 2026-02-16
Parser support for no block parameter syntax

kddnewton:
@koic does this work?

### DIFF
--- a/bin/prism
+++ b/bin/prism
@@ -270,17 +270,17 @@ module Prism
 
     # bin/prism parser [source]
     def parser(argv)
-      require "parser/ruby34"
+      require "parser/current"
       source, filepath = read_source(argv)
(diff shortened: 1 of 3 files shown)


Earlopain:
I'm a bit cautious but I think it should be fine. Last time for `itblock` it caused issues for people because they had older `parser` versions. But that only really was a problem because they had already written code with `it` in it. This syntax only became valid the with new ruby version (which is long not released yet). Previously discussed in https://github.com/ruby/prism/pull/3504

I opened https://github.com/whitequark/parser/pull/1087 but now don't think we need to wait on that.

koic:
Parser 3.3.10.2 has been released with https://github.com/whitequark/parser/pull/1087 included.

Issue #3931, opened 2026-02-17
Extension not fully rebuild when `config.yml` changes


Earlopain:
Ref https://github.com/ruby/prism/pull/3926#issuecomment-3907350089

I can reproduce it with the following:

```sh
$ bundle exec rake clean
$ git checkout 7c16f777bd4708a513097d9e1f6c91639f9c5a6f
bundle exec rake compile

$ git checkout 27c24fdc0d8c9d14f819f83050d573315e8474ee # changes to config.yml
$ bundle exec rake # crashes/segfaults
```

I don't really know much about this but noticed that some object files don't get rebuild, which they should probably. 

```sh
$ find tmp -name "*.o" | xargs ls -la
10:53 tmp/x86_64-linux/prism/4.0.0/api_node.o
10:53 tmp/x86_64-linux/prism/4.0.0/diagnostic.o
10:50 tmp/x86_64-linux/prism/4.0.0/encoding.o
10:50 tmp/x86_64-linux/prism/4.0.0/extension.o
10:53 tmp/x86_64-linux/prism/4.0.0/node.o
10:50 tmp/x86_64-linux/prism/4.0.0/options.o
10:50 tmp/x86_64-linux/prism/4.0.0/pm_buffer.o
10:50 tmp/x86_64-linux/prism/4.0.0/pm_char.o
10:50 tmp/x86_64-linux/prism/4.0.0/pm_constant_pool.o
10:50 tmp/x86_64-linux/prism/4.0.0/pm_integer.o
10:50 tmp/x86_64-linux/prism/4.0.0/pm_list.o
10:50 tmp/x86_64-linux/prism/4.0.0/pm_memchr.o
10:50 tmp/x86_64-linux/prism/4.0.0/pm_newline_list.o
10:50 tmp/x86_64-linux/prism/4.0.0/pm_string.o
10:50 tmp/x86_64-linux/prism/4.0.0/pm_strncasecmp.o
10:50 tmp/x86_64-linux/prism/4.0.0/pm_strpbrk.o
10:53 tmp/x86_64-linux/prism/4.0.0/prettyprint.o
10:53 tmp/x86_64-linux/prism/4.0.0/prism.o
10:50 tmp/x86_64-linux/prism/4.0.0/regexp.o
10:53 tmp/x86_64-linux/prism/4.0.0/serialize.o
10:50 tmp/x86_64-linux/prism/4.0.0/static_literals.o
[...]

PR #3932, opened 2026-02-17
Recompile when config.yml changes

eregon:
* Fixes https://github.com/ruby/prism/issues/3931.
* The first line is the default one, we want to keep that default dependency too.
* Some docs about this: https://docs.ruby-lang.org/en/4.0/MakeMakefile.html

### DIFF
--- a/ext/prism/depend
+++ b/ext/prism/depend
@@ -1,0 +1,2 @@
+$(OBJS): $(HDRS) $(ruby_headers)
+$(OBJS): $(srcdir)/../../config.yml
--- a/prism.gemspec
+++ b/prism.gemspec
@@ -42,6 +42,8 @@ Gem::Specification.new do |spec|
     "docs/serialization.md",
     "docs/testing.md",
     "ext/prism/api_node.c",
+    "ext/prism/depend",
+    "ext/prism/extconf.rb",
     "ext/prism/extension.c",
     "ext/prism/extension.h",
     "include/prism.h",


Earlopain:
Ah, right:
> If “depend” file exist in the source directory, that content will be included in the generated makefile, with formatted by [depend_rules](https://docs.ruby-lang.org/en/4.0/MakeMakefile.html#method-i-depend_rules) method.

So quite easy actually. But shouldn't this include all the headers? This basically just takes care of all the gitignored/generated ones but what if for example `defines.h` changes?

Earlopain:
I had unknown symbol errors after https://github.com/ruby/prism/pull/3964 was merged so without looking too much into it I guess my comment is somewhat relevant. At least it failed loudly this time :sweat_smile:

eregon:
> So quite easy actually. But shouldn't this include all the headers? This basically just takes care of all the gitignored/generated ones but what if for example `defines.h` changes?

Yes, I think you're right, I'll try it.

Earlopain:
I found that adding the headers to `$headers`(I think) in extconf seemed to work. But I barely found any usage of that so not sure about that.

Would be nice since they can just be globbed in ruby

eregon:
Yes, I came to the same conclusion, https://github.com/ruby/prism/pull/3967
And those headers are included in `ruby_headers` so there is no need to depend on `LOCAL_HDRS` explicitly (but we can for clarity if wanted)

PR #3933, opened 2026-02-18
rbs-inline

kddnewton:
Generate RBS through inline annotations instead of through maintaining separate RBS files.

I have wanted this for a while, but frankly it has been daunting to try to convert the whole codebase, and previously Steep would just hang. It looks like it has gotten better to the point where this is now possible. The `typecheck:steep` codebase will now first extract out inline type annotations and put them into the sig directory. You can effectively ignore all files in that directory for the purposes of review.

At present, this means basically the entire codebase is typechecked with steep, except the translation directory which depends on type signatures for parser/ripper/ruby_parser which do not exist.

### DIFF
--- a/rakelib/typecheck.rake
+++ b/rakelib/typecheck.rake
@@ -138,8 +138,32 @@ namespace :typecheck do
               method.sigs << sig
             end
           end
-      when RBS::AST::Members::Alias, RBS::AST::Members::InstanceVariable,
-          RBS::AST::Declarations::Interface, RBS::AST::Declarations::TypeAlias
+      when RBS::AST::Members::Alias
+        case [node.new_name, node.old_name]
+        when [:dispatch, :visit]
+          parent <<
+            RBI::Method.new("dispatch", visibility: @visibility, comments: compile_comments(node)) do |method|
+              method.add_param("node")
+              method.sigs <<
+                RBI::Sig.new(
+                  params: [RBI::SigParam.new("node", RBI::Type.nilable(RBI::Type.simple("Node")))],
+                  return_type: RBI::Type.untyped
+                )
+            end
+        when [:script_lines, :source_lines],
+             [:find, :breadth_first_search],
+             [:find_all, :breadth_first_search_all],
+             [:deconstruct, :child_nodes]
+          found = parent.nodes.find { |child| child.is_a?(RBI::Method) && child.name == node.old_name.name }
+          parent <<
+            found.dup.tap do |method|
+              method.name = node.new_name.name
+              method.comments = compile_comments(node)
+            end
+        else
+          raise
+        end
+      when RBS::AST::Members::InstanceVariable, RBS::AST::Declarations::Interface, RBS::AST::Declarations::TypeAlias
         # skip
       else
         raise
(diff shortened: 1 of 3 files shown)


eregon:
Nice!

Earlopain:
rdoc doesn't seem to really handle this which I find kind of sad since I just cleaned it all up. The type annotations are put into method docs and the magic comment appears like 20 times on the main page

kddnewton:
@Earlopain ahh you're right, that's pretty ugly. I'll see if I can modify RDoc in some way before merging this.

kddnewton:
Turns out there's a special syntax in RDoc where you can use `#--` to mark part of the comment as private so it won't show up in the docs. I added that to all of the annotations and now it should be the same. @Earlopain can you check if this is better for you?

Earlopain:
I didn't know about that syntax, TIL. Unfortunatly rdoc doesn't ignore the rbs_inline comments, so I applied the same thing to them. I think rdoc should just do this by itself (github.com/ruby/rdoc/issues/1614) but for now it doesn't. Without, it looks like this 
image

(`typed` sneaked in for some reason, which wasn't present before. I ignored that as well)

Earlopain:
BTW, can we generate rbi signatures with this instead of handwriting them? There's https://github.com/Shopify/spoom?tab=readme-ov-file#translate-sigs-between-rbi-and-rbs, maybe we don't need to have them in two places

gemfiles/typecheck/Gemfile
```diff
@@ -6,8 +6,10 @@ gem "minitest"
 gem "parser"
 gem "rake-compiler"
 gem "rake"
+gem "rbs-inline"
 gem "ruby_parser"
 gem "sorbet", "<= 0.6.12666" # until tapioca is bumped
 gem "steep", ">= 1.7.0.dev.1"
 gem "tapioca"
 gem "test-unit"
+gem "tsort" # until rbs is bumped
```
Earlopain: I don't know what's up with dependabot but you can just manually bump rbs and this is not necessary anymore.

rakelib/typecheck.rake
```diff
@@ -76,6 +77,13 @@ namespace :typecheck do
     end
   end
 
+  desc "Generate RBS with rbs-inline"
+  task rbs_inline: :templates do
+    with_gemfile do
+      sh "bundle", "exec", "rbs-inline", "lib", "--output", "lib"
```
Earlopain: Should this be checked in CI or be part of the release process?

kddnewton: I think it's okay because the typecheck will fail if it's not run.

kddnewton:
@Earlopain want to check now? It now generates RBI from the RBS.

kddnewton:
I agree, I don't really want to maintain the type AST translation either, but I'm hoping to upstream changes to spoom to make this obsolete. Hopefully we aren't stuck maintaining this for long.

The real benefit is that not only does it get rid of templates/manually-managed RBI, it also types _way_ more of the codebase. So there are real benefits for consumers if we do it this way.

Are you totally opposed to merging as-is? I'm okay if you are, I can try to upstream changes first and leave this as blocked, it might just be a while.

PR #3949, opened 2026-03-02
Reject infix operators on command call on writes

### LINKED ISSUE #3106: Assigning command call is allowed where rejected in parse.y

kddnewton:
Fixes #3106

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -21606,6 +21606,26 @@ pm_call_node_command_p(const pm_call_node_t *node) {
 }
 
 /**
+ * Determine if a given write node has a command call as its right-hand side. We
+ * need this because command calls as the values of writes cannot be extended by
+ * infix operators.
+ */
+static inline bool
+pm_write_node_command_p(const pm_node_t *node) {
+    pm_node_t *value;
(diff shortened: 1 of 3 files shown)


Earlopain:
This is casting too wide a net. This code used to be ok but now isn't:

```rb
# steep-1.10.0/lib/steep/interface/function.rb
kp = keyword_params | other.keyword_params or return
                                           ^~ unexpected 'or', expecting end-of-input
                                           ^~ unexpected 'or', ignoring it
```

I have a few more cases in https://gist.github.com/Earlopain/908f5744ae2c32bf4ab1d0a590b4e6c5 but I did not check them all. They all seem rather similar though (`a = prefix != nil and prefix.length > 0`, `query_options = options.keys & [:select, :where, :order, :limit] and !query_options.empty?`, etc.) so guessing it's the same cause for now.

Can you take a look?

Earlopain:
I reverted for now in https://github.com/ruby/prism/pull/3960 with some added tests for this

PR #3950, opened 2026-03-02
Require arguments to Source.for

### LINKED ISSUE #3861: Prism::Source.for(code) returns an incomplete Prism::Source object
```ruby
irb -Ilib -rprism
irb(main):001> src = Prism::Source.for("1 + 2")
=> #<Prism::ASCIISource:0x00007f59cab9f840 @offsets=[], @source="1 + 2", @start_line=1>
irb(main):002> src.line_start(3)
=> nil
```

The problem is `@offsets` is `[]` which is never valid. 
(`@start_line` defaults to 1 which is probably fine)

Another way to see it:
```ruby
irb(main):004> Prism.parse("1\n2").source
=> #<Prism::ASCIISource:0x00007f59c9affc60 @offsets=[0, 2], @source="1\n2", @start_line=1>
irb(main):005> Prism::Source.for("1\n2")
=> #<Prism::ASCIISource:0x00007f59c998cf68 @offsets=[], @source="1\n2", @start_line=1>
```

`Prism::Source.for` is notably used by `Prism::DSL#source` and might be used by code outside of Prism (I don't know).
It's also used in the Ripper translation layer, I'm fixing that usage in https://github.com/ruby/prism/pull/3859.

The method takes arguments for the offsets and start_line:
https://github.com/ruby/prism/blob/d3d3de96f5d6395f9cc5133c6fa0e16bb9b3aec2/lib/prism/parse_result.rb#L13
But because they are optional and Prism doesn't provide a way to compute the offsets (AFAIK) I don't think that's good enough.

Some solutions I can imagine:
* Remove `Prism::Source.for`, only allow to create complete Source objects through Prism.{lex,parse}*. This is what the Java code does, the Source constructor is private, and the only way to get a Source is through lexing/parsing and get it from the ParseResult.
* Have the default compute the offsets itself, probably most convenient but a bit of a performance trap (at least we should document it on the method). In the past we have seen that computing line offsets from Ruby is pretty slow.
* Have the default be `nil` and automatically compute the offsets when needed.

kddnewton:
Fixes #3861

### DIFF
--- a/lib/prism/parse_result.rb
+++ b/lib/prism/parse_result.rb
@@ -19,9 +19,17 @@ module Prism
     # be used instead of `new` and it will return either a `Source` or a
     # specialized and more performant `ASCIISource` if no multibyte characters
     # are present in the source code.
+    #
+    # Note that if you are calling this method manually, you will need to supply
+    # the start_line and offsets parameters. start_line is the line number that
+    # the source starts on, which is typically 1 but can be different if this
+    # source is a subset of a larger source or if this is an eval. offsets is an
+    # array of byte offsets for the start of each line in the source code, which
+    # can be calculated by iterating through the source code and recording the
+    # byte offset whenever a newline character is encountered.
     #--
-    #: (String source, ?Integer start_line, ?Array[Integer] offsets) -> Source
-    def self.for(source, start_line = 1, offsets = [])
+    #: (String source, Integer start_line, Array[Integer] offsets) -> Source
+    def self.for(source, start_line, offsets)
       if source.ascii_only?
         ASCIISource.new(source, start_line, offsets)
       elsif source.encoding == Encoding::BINARY
@@ -55,8 +63,8 @@ module Prism
 
     # Create a new source object with the given source code.
     #--
-    #: (String source, ?Integer start_line, ?Array[Integer] offsets) -> void
-    def initialize(source, start_line = 1, offsets = [])
+    #: (String source, Integer start_line, Array[Integer] offsets) -> void
+    def initialize(source, start_line, offsets)
       @source = source
       @start_line = start_line # set after parsing is done
       @offsets = offsets # set after parsing is done
--- a/templates/lib/prism/dsl.rb.erb
+++ b/templates/lib/prism/dsl.rb.erb
@@ -5,7 +5,7 @@ module Prism
   # The DSL module provides a set of methods that can be used to create prism
   # nodes in a more concise manner. For example, instead of writing:
   #
-  #     source = Prism::Source.for("[1]")
+  #     source = Prism::Source.for("[1]", 1, [])
   #
   #     Prism::ArrayNode.new(
   #       source,
@@ -62,7 +62,7 @@ module Prism
     #--
     #: (String string) -> Source
     def source(string)
-      Source.for(string)
+      Source.for(string, 1, [])
     end
 
     # Create a new Location object.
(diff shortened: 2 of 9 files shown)


templates/lib/prism/dsl.rb.erb
```diff
@@ -136,7 +136,7 @@ module Prism
     #--
     #: () -> Source
     def default_source
-      Source.for("")
+      Source.for("", 1, [])
```
eregon, line 139:
```suggestion
      Source.for("", 1, [0])
```
Since that's correct for this case

templates/lib/prism/dsl.rb.erb
```diff
@@ -5,7 +5,7 @@ module Prism
   # The DSL module provides a set of methods that can be used to create prism
   # nodes in a more concise manner. For example, instead of writing:
   #
-  #     source = Prism::Source.for("[1]")
+  #     source = Prism::Source.for("[1]", 1, [])
```
eregon, line 8:
```suggestion
  #     source = Prism::Source.for("[1]", 1, [0])
```
Since that's correct for this case

templates/lib/prism/dsl.rb.erb
```diff
@@ -62,7 +62,7 @@ module Prism
     #--
     #: (String string) -> Source
     def source(string)
-      Source.for(string)
+      Source.for(string, 1, [])
```
eregon, line 65: This will still cause issues when using `Source` methods as it's an incorrect value for offsets.
Maybe computing it here would be best? Or set it to `nil` so it fails more clearly and early?

templates/lib/prism/serialize.rb.erb
```diff
@@ -27,7 +27,7 @@ module Prism
     #: (String input, String serialized, bool freeze) -> ParseResult
     def self.load_parse(input, serialized, freeze)
       input = input.dup
-      source = Source.for(input)
+      source = Source.for(input, 1, [])
```
eregon, line 30: As a note, the 4 cases in this file are fine because `source.replace_offsets(offsets)` is used a few lines later.

lib/prism/parse_result.rb
```diff
@@ -19,9 +19,17 @@ class Source
...
-    #: (String source, ?Integer start_line, ?Array[Integer] offsets) -> Source
-    def self.for(source, start_line = 1, offsets = [])
+    #
+    # Note that if you are calling this method manually, you will need to supply
+    # the start_line and offsets parameters. start_line is the line number that
+    # the source starts on, which is typically 1 but can be different if this
+    # source is a subset of a larger source or if this is an eval. offsets is an
+    # array of byte offsets for the start of each line in the source code, which
+    # can be calculated by iterating through the source code and recording the
+    # byte offset whenever a newline character is encountered.
```
eregon, line 29: A correct `offsets` Array always contains `0` as the first element, even though there is no newline there.
```suggestion
    # source is a subset of a larger source or if this is an eval. offsets is an
    # array of byte offsets for the start of each line in the source code, which
    # can be calculated by iterating through the source code and recording the
    # byte offset whenever a newline character is encountered.
    # The first element is always 0 to mark the first line.
```

kddnewton:
> Thank you for this PR, I have some suggestions, could you take a look? I can make a PR with them, but it would be helpful to discuss them here first.

@eregon I'm good with all of your changes, and I agree we should compute it eagerly in the DSL.

PR #3965, opened 2026-03-05
Add a simple benchmark script to bin/prism

kddnewton:


Earlopain:
Do we still need the older `benchmark` command? It's confusing to have two so similarly named.

FWIW, I never used the benchmark command when looking at performance. It runs against so many variants which makes it cumbersome to use when actually wanting to compare/improve performance.

kddnewton:
Yeah let's get rid of it.

eregon:
It seems good to be able to compare to the other parsers though, and check whether the translation layers are faster than the original, but yeah if it's never used probably not much point to keep it here.

For https://eregon.me/blog/2024/10/27/benchmarking-ruby-parsers.html I used something that was probably based on that script but a bit more complicated with more variants and getting more date.

PR #3966, opened 2026-03-05
Correctly handle `and?` and similar on ruby 4.0

Earlopain:
It gets confused for syntax introduced in https://bugs.ruby-lang.org/issues/20925 But it actually should be a plain method call.
`!`/`?` are not valid as part of an identifier, methods however allow them as the last character.

Fixes [[Bug #21946]](https://bugs.ruby-lang.org/issues/21946)

### DIFF
--- a/test/prism/fixtures/4.0/leading_logical.txt
+++ b/test/prism/fixtures/4.0/leading_logical.txt
@@ -14,8 +14,3 @@ and 3
 or 2
 or 3
 
-1
-andfoo
-
-2
-orfoo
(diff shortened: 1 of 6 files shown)


test/prism/fixtures/4.0/leading_logical.txt
```diff
@@ -14,8 +14,3 @@ and 3
 or 2
 or 3
 
-1
-andfoo
```
Earlopain, line 18: This should be true on all versions, not just 4.0. I moved it out to the non-version-specific fixture.

kddnewton (approved):

PR #3967, opened 2026-03-05
Recompile when Prism headers change

eregon:
* See https://github.com/ruby/prism/pull/3932#issuecomment-3915768654

### DIFF
--- a/prism.gemspec
+++ b/prism.gemspec
@@ -42,6 +42,7 @@ Gem::Specification.new do |spec|
     "docs/serialization.md",
     "docs/testing.md",
     "ext/prism/api_node.c",
+    "ext/prism/extconf.rb",
     "ext/prism/extension.c",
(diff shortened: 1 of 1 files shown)


Earlopain:
I think your previous PR can be reverted, since this covers that? I can check that tomorrow if you don't want to

eregon:
Right, all *.c and *.h should be tracked, and if config.yml changes without changing those it shouldn't matter for the extension, i.e. it doesn't need to be recompiled then.
I'll push a revert of that commit.

eregon:
Done.
FWIW changing `config.yml` by e.g. adding a comment still recompiles everything most likely because the generated .h files are still regenerated (`Prism::Template.render` writes unconditionally) and get a new mtime. That's fine though (and not a common use case, most changes to config.yml would change .h/.c files).

PR #3969, opened 2026-03-06
Fix the handling of `do` on commands

kddnewton:
Introduce keyword_do_block to match parse.y behavior, as well as do a big audit of behavior to line up as closely as possible.


src/prism.c
```diff
@@ -21483,6 +21524,29 @@ pm_call_node_command_p(const pm_call_node_t *node) {
...
+        case PM_CONSTANT_WRITE_NODE: value = ((pm_constant_write_node_t *) node)->value; break;
+        case PM_GLOBAL_VARIABLE_WRITE_NODE: value = ((pm_global_variable_write_node_t *) node)->value; break;
+        case PM_INSTANCE_VARIABLE_WRITE_NODE: value = ((pm_instance_variable_write_node_t *) node)->value; break;
+        case PM_LOCAL_VARIABLE_WRITE_NODE: value = ((pm_local_variable_write_node_t *) node)->value; break;
+        default: return false;
+    }
+    if (!PM_NODE_TYPE_P(value, PM_CALL_NODE)) return false;
+
+    const pm_call_node_t *call = (const pm_call_node_t *) value;
+    return pm_call_node_command_p(call) && (call->receiver == NULL || call->call_operator_loc.length > 0);
```
Earlopain: I feel like I also tried this but I guess not. Can you put it in a different PR?

kddnewton: Ahh whoops, yeah that was different work.

PR #3972, opened 2026-03-07
Copy and embed for serialization format

### LINKED ISSUE #1537: Copy & embed any string/constant pool entry in the serialized output

kddnewton:
Fixes #1537

@eregon @enebo would you both be able to experiment with this branch and tell me what you think? If it works better for JRuby/TruffleRuby I'm happy to just merge this as the default and remove the branching logic inside the deserializers

### DIFF
--- a/rakelib/typecheck.rake
+++ b/rakelib/typecheck.rake
@@ -345,11 +345,14 @@ namespace :typecheck do
 
   desc "Typecheck with Sorbet"
   task sorbet: :templates do
+    require "fileutils"
+
     locals = {
       polyfills: Dir["lib/prism/polyfill/**/*.rb"],
       gem_rbis: Dir["sorbet/rbi/**/*.rbi"]
     }
 
+    FileUtils.mkdir_p("sorbet")
     File.write("sorbet/typed_overrides.yml", ERB.new(<<~YAML, trim_mode: "-").result_with_hash(locals))
       false:
         - ./lib/prism/lex_compat.rb
--- a/rbi/generated/prism.rbi
+++ b/rbi/generated/prism.rbi
@@ -16,7 +16,7 @@ module Prism
   # resembles the return value of Ripper.lex.
   #
   # For supported options, see Prism.parse.
-  sig { params(source: String, options: T.untyped).returns(LexCompat::Result) }
+  sig { params(source: String, options: ::T.untyped).returns(LexCompat::Result) }
   def self.lex_compat(source, **options); end
 
   # Load the serialized AST using the source as a reference into a tree.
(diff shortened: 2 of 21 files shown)


templates/src/serialize.c.erb
```diff
@@ -304,28 +287,18 @@ pm_serialize_content(pm_parser_t *parser, pm_node_t *node, pm_buffer_t *buffer)
...
-                // Since this is an owned or constant constant, we are going to
-                // write its contents into the buffer after the constant pool.
-                // So effectively in place of the source offset, we have a
-                // buffer offset. We will add a leading 1 to indicate that this
-                // is a buffer offset.
-                uint32_t content_offset = pm_sizet_to_u32(buffer->length);
-                uint32_t owned_mask = 1U << 31;
+            // Write the constant contents into the buffer after the constant
+            // pool. In place of the source offset, we store a buffer offset
+            // with the high bit set to indicate embedded content.
```
eregon: Since they would all be embedded we wouldn't need the high bit anymore (once the deserializers are updated of course), right?

eregon:
Some stats:
master https://github.com/ruby/prism/actions/runs/22784388008/job/66097760909
this branch https://github.com/ruby/prism/actions/runs/22789229164/job/66112499398?pr=3972

All fields:
```
master:
Total sizes for top 100 gems:
total serialized size: 113025865
total serialized/total source: 1.253

average: 1.350
median:  1.343
1st quartile: 0.980
3rd quartile: 1.681
min - max: 0.129 - 4.922

this branch:
Total sizes for top 100 gems:
total serialized size: 124178078
total serialized/total source: 1.377

average: 1.601
median:  1.609
1st quartile: 1.172
3rd quartile: 1.988
min - max: 0.185 - 5.556
```

Only semantic fields:
```
master:
Total sizes for top 100 gems:
total serialized size:  66147370
total serialized/total source: 0.733

average: 0.800
median:  0.779
1st quartile: 0.568
3rd quartile: 1.007
min - max: 0.076 - 3.675

this branch:
Total sizes for top 100 gems:
total serialized size:  77299583
total serialized/total source: 0.857

average: 1.051
median:  1.035
1st quartile: 0.760
3rd quartile: 1.301
min - max: 0.080 - 4.286
```
So `total serialized size` increases as expected, but the increase is far smaller than `total source size` so it seems a rather clear gain in retained memory (when doing lazy method loading).

eregon:
> @eregon @enebo would you both be able to experiment with this branch and tell me what you think? If it works better for JRuby/TruffleRuby I'm happy to just merge this as the default and remove the branching logic inside the deserializers

Conceptually it should be strictly better for retained memory, and it should make it possible to lazily deserialize methods. Right now the laziness for methods in TruffleRuby is by lazily translating the Prism AST of a `def` to TruffleRuby AST, but the entire Prism AST is still loaded eagerly (because `Loader.java` currently only supports that, and having to keep the `serialized` byte[] + `source` byte[] around for source substrings seems not great, at least currently we don't need to keep the `serialized` byte[] around).
So instead of keeping the entire Prism AST around for lazy methods we'd keep the `serialized` byte[] and deserialize on first execution of the method.

In practice if you're thinking of parse performance in TruffleRuby/JRuby then I think we should adapt deserializers as well already before measuring, e.g. to no longer have to handle the "owned string bit".

I checked that the source bytes are only used by `org.prism.Loader#loadString` and `org.prism.Loader.ConstantPool#get`, both of which should no longer need the source bytes with this PR.

eregon:
Regarding stopping at DefNode in Loader.java to have lazy deserialization, I'd be happy to look into that, though it seems better done separately from this PR.

enebo:
I think this looks good.  We have plans to do an update on JRuby side this week.  Ultimately as @eregon said the big opt for prism will be completely going lazy for methods (leaving serialized sections un-ASTd much less built into IR).  Not having to make the AST eagerly for methods we may never call is one step past what we do in legacy parsing.

eregon:
I also wanted to say for me this PR with the deserializers changes would be fine to merge, for TruffleRuby POV I think there is no need to benchmark before merging.

headius:
Worth mentioning that lazy def nodes will also help the wasm implementation narrow the gap with native.

templates/java/org/ruby_lang/prism/Loader.java.erb
```diff
@@ -61,15 +59,10 @@ public class Loader {
...
-                } else {
-                    int position = buffer.position();
-                    buffer.position(start & 0x7FFFFFFF);
-                    buffer.get(bytes, 0, length);
-                    buffer.position(position);
-                }
+                int position = buffer.position();
+                buffer.position(start);
+                buffer.get(bytes, 0, length);
+                buffer.position(position);
```
eregon, line 65: This code already existed before but it becomes more obvious now, this should use an absolute bulk get like `ByteBuffer#get(int index, byte[] dst)` and then it wouldn't need to mess with `position` at all.
Absolute bulk get is available since Java 13, @headius @enebo OK to bump to 21 (or 17) then?
I don't think JRuby 9 would/could use such a recent Prism anyway, would it?

Happy to make a PR for this.

eregon: https://github.com/ruby/prism/pull/3977

eregon:
Some quick numbers (benchmarks from https://github.com/truffleruby/truffleruby/issues/3117):
```
$ TRUFFLERUBY_METRICS_REPS=5 jt metrics time --experimental-options -e0
$ jt -q test fast --dry-run -- --experimental-options --cpusampler --cpusampler.Period=1 --metrics-profile-require=summary | grep "metrics "
```

On v1.9.0:
```
     0.111 parsing-core
     0.109 parsing-core
     0.119 parsing-core
     0.108 parsing-core

 metrics parsing                                                               ||             1510ms   8.1% ||             1510ms   8.1% || (metrics)~1:0
 metrics parsing                                                               ||             1358ms   7.6% ||             1358ms   7.6% || (metrics)~1:0
 metrics parsing                                                               ||             1484ms   7.9% ||             1484ms   7.9% || (metrics)~1:0
```
On a1567ec0d8975c513a5d9540c5500dd58dec9a26 (before this PR):
```
     0.092 parsing-core
     0.095 parsing-core
     0.099 parsing-core
     0.092 parsing-core

 metrics parsing                                                               ||             1172ms   6.8% ||             1172ms   6.8% || (metrics)~1:0
 metrics parsing                                                               ||             1308ms   7.2% ||             1308ms   7.2% || (metrics)~1:0
 metrics parsing                                                               ||             1330ms   7.5% ||             1330ms   7.5% || (metrics)~1:0
```
On c8fd89e008267acc973d58ea759d792b2b4296e5 (after this PR + #3977 ):
```
     0.098 parsing-core
     0.096 parsing-core
     0.093 parsing-core
     0.096 parsing-core

 metrics parsing                                                               ||             1314ms   7.4% ||             1314ms   7.4% || (metrics)~1:0
 metrics parsing                                                               ||             1249ms   7.1% ||             1249ms   7.1% || (metrics)~1:0
 metrics parsing                                                               ||             1308ms   7.3% ||             1308ms   7.3% || (metrics)~1:0
```

Much faster than 1.9.0 and pretty close to before this PR.

PR #3976, opened 2026-03-08
Remove generated types from checkin

kddnewton:
We don't need these checked in, just make sure we have them when the gem gets published.


Earlopain:
What about when using it as a git gem? I think that would not work (if I require types that is)

kddnewton:
Hmm... you're right. And this is becoming more trouble than it's worth. Closing for now.

PR #3977, opened 2026-03-08
Use ByteBuffer#get(int index, byte[] dst) in Loader.java

eregon:
* This is an absolute bulk get which does not modify the position, and therefore saves from getting, setting and restoring the position.

From https://github.com/ruby/prism/pull/3972#discussion_r2902032897

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -45,7 +45,7 @@ elsif RUBY_ENGINE == "jruby"
     ext.name = "prism"
     ext.ext_dir = "java"
     ext.lib_dir = "tmp"
-    ext.release = "8"
+    ext.release = "21"
     ext.gem_spec = Gem::Specification.load("prism.gemspec")
   end
 end


eregon:
The [CI passes](https://github.com/ruby/prism/actions/runs/22827672810/job/66210556339?pr=3977) even though that method doesn't exist on Java 8.
It's because it's using `-target 1.8 -source 1.8` and that's not checking if JDK methods exist.
`--release=8` (via `ext.release = "8"`) does check it and find it:
```
javac --release=8 -Xlint -d tmp/java/prism -cp /home/eregon/.rubies/jruby-10.0.4.0/lib/jruby.jar java/org/ruby_lang/prism/AbstractNodeVisitor.java java/org/ruby_lang/prism/Loader.java java/org/ruby_lang/prism/MarkNewlinesVisitor.java java/org/ruby_lang/prism/Nodes.java java/org/ruby_lang/prism/ParseResult.java java/org/ruby_lang/prism/Parser.java java/org/ruby_lang/prism/ParsingOptions.java
warning: [options] source value 8 is obsolete and will be removed in a future release
warning: [options] target value 8 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
java/org/ruby_lang/prism/Loader.java:65: error: no suitable method found for get(int,byte[])
                buffer.get(start, bytes);
                      ^
    method ByteBuffer.get() is not applicable
      (actual and formal argument lists differ in length)
    method ByteBuffer.get(int) is not applicable
      (actual and formal argument lists differ in length)
    method ByteBuffer.get(byte[],int,int) is not applicable
      (actual and formal argument lists differ in length)
    method ByteBuffer.get(byte[]) is not applicable
      (actual and formal argument lists differ in length)
1 error
3 warnings
```

I think this is a good reason to update to JDK 21+ in Prism (which is what JRuby 10 requires).
I think JRuby 9 won't update to a newer Prism from latest master, isn't it? (especially now that Prism [has release branches](https://bugs.ruby-lang.org/issues/21928)).
cc @headius @enebo

enebo:
Looks good to me

blog: http://blog.enebo.com       twitter: tom_enebo
mail: ***@***.***

PR #3998, opened 2026-03-16
Do not pass source bytes to Loader.java at all and mark newlines for lazy DefNodes

eregon:
Follow-up of https://github.com/ruby/prism/pull/3983, implementing a variant of the first option, the difference being we create a new `MarkNewlinesVisitor` object vs keeping it around.
We also need to keep the Source object around to have access to line offsets, but that's fine as the first commit completely removes the need to pass the source `byte[]` to the Loader :)
With this, we can use lazy DefNode in TruffleRuby and be fully compatible with eager loading.

From the commits:

[Do not pass source bytes to Loader.java at all](https://github.com/ruby/prism/commit/fc4d405caa2518ed33562e1c79375ddcfab69c25)
* That way we are sure it's not retained by it.
* Source#bytes seems to have no usages.
* Arrays.binarySearch() works just fine if the value is > the last array element.

[Mark newlines for lazy DefNodes](https://github.com/ruby/prism/commit/156d4fbda9b0b8eb264ed31c46196423169fc671)
* By running the MarkNewlinesVisitor on them when they are lazily loaded.
* See https://github.com/ruby/prism/pull/3983.
* Achieves the same nodes as with eager DefNode.

### DIFF
--- a/java/org/ruby_lang/prism/MarkNewlinesVisitor.java
+++ b/java/org/ruby_lang/prism/MarkNewlinesVisitor.java
@@ -6,9 +6,9 @@ final class MarkNewlinesVisitor extends AbstractNodeVisitor<Void> {
     private final Nodes.Source source;
     private boolean[] newlineMarked;
 
-    MarkNewlinesVisitor(Nodes.Source source, boolean[] newlineMarked) {
+    MarkNewlinesVisitor(Nodes.Source source) {
         this.source = source;
-        this.newlineMarked = newlineMarked;
+        this.newlineMarked = new boolean[1 + source.getLineCount()];
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


eregon:
Benchmarks in https://github.com/truffleruby/truffleruby/pull/4182#issue-4052614089, it's about 35% faster to parse the TruffleRuby core library!

eregon:
This is basically fixing https://github.com/ruby/prism/pull/3983 to make it usable, merging to unblock https://github.com/truffleruby/truffleruby/pull/4182 and a chain of PRs on top of it.

If anyone wants to access the `sourceBytes` they can easily keep it themselves as they are given to `Parser.parseAndSerialize(sourceBytes, parsingOptions)`.

PR #4005, opened 2026-03-16
[ruby-4.0] Support auto-sync for ruby-4.0 branch

k0kubun:
I want to make it easier to keep ruby/ruby `ruby_4_0` and ruby/prism `ruby-4.0` in sync. This PR allows ruby/prism `ruby-4.0` branch to be automatically synchronized to ruby/ruby `ruby_4_0` branch.

After merging this PR, this branch will effectively be part of Ruby's stable branches, so this PR also registers myself as the codeowner for the `ruby-4.0` branch.

ref: https://github.com/ruby/ruby/commit/b01ca60e88d93ff7e93e11951ff8a1ba57462463


Earlopain:
Very nice 👍

PR #4006, opened 2026-03-17
Optimizations

kddnewton:
This branch has a ton of optimizations, broadly grouped into:

* Splitting up functions that would reallocate if they need more space into fast paths and slow paths, with the fast path being static inline in the header so it gets inlined into callers.
* Adding a bloom filter to local scopes so that it's faster to reject locals that are not found.
* Having the lexer scan the value of small integers so for integers < 2**32 we only scan them once.
* Building SIMD and SWAR paths through our custom strpbrk, and then caching the lookup tables that we build for those function calls so that in repeat calls (like when lexing strings, regexps, and list literals) we do not have to recompute them.
* Lowering the threshold for switching the locals table over to a hash from 9 to 5.
* Adding SIMD and SWAR paths through scanning identifiers for ASCII.
* Making some fast paths through `parser_lex_magic_comment` because it shows up shockingly high in profiles.
* Putting a lot of the parser metadata into its own arena, as opposed to ad-hoc malloc/free pairs.
* Changing the constant pool to use a hash function based on wyhash that goes word-at-a-time as opposed to byte-at-a-time.
* Putting the contents of the constant on the constant bucket (this would seem like a net negative for memory, but because we resize more appropriately now and its backed by the parser arena, it's kind of a wash).

With all this in place, I'm seeing about 22% speed gains, using about the same amount of memory.

### DIFF
--- a/include/prism/parser.h
+++ b/include/prism/parser.h
@@ -101,19 +101,19 @@ typedef struct {
 } pm_heredoc_lex_mode_t;
 
 /**
- * When lexing Ruby source, the lexer has a small amount of state to tell which
- * kind of token it is currently lexing. For example, when we find the start of
- * a string, the first token that we return is a TOKEN_STRING_BEGIN token. After
- * that the lexer is now in the PM_LEX_STRING mode, and will return tokens that
- * are found as part of a string.
- */
-/**
  * The size of the breakpoints and strpbrk cache charset buffers. All
  * breakpoint arrays and the strpbrk cache charset must share this size so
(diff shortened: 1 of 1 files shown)


headius:
Very nice. I wonder what the impact will be on the WASM build performance. I'll try to get that stuff more finalized today and into CI.

Earlopain:
https://github.com/ruby/ruby/pull/16418 is already merged. I pushed up the unique commits from that PR and also https://github.com/ruby/ruby/pull/16423, it's back in sync now

PR #4013, opened 2026-03-18
Remove currently deprecated functionality

Earlopain:
Mostly in the node extension. The method for marking something as deprecated is currently unused but seems fine leave as is.


Earlopain:
They pin to ~> 1.0 but didn't always. I would guess mixing these older rubocop-ast with newer rubocop versions is broken in some way but it's whatever.

PR #4014, opened 2026-03-18
Bump required ruby version to 3.2

Earlopain:
3.1 and earlier are EOL today. I want to write new syntax and use new methods without having to worry about the consequences. A major version is a nice oportunity to do this.

2.7 to 3.2 is a bit of a jump but there are very few changes to prism that actually require any changes from consumers.

I don't want to support all runtime ruby versions in perpetuity. But perhaps this is a bit too much, what do you think @kddnewton?


koic:
From the perspective of the Ruby ecosystem, this raises some concerns.

A parser is a fundamental component of the Ruby language, and ideally it should remain usable across as wide a range of Ruby versions as possible. Its support policy should be treated separately from Ruby’s own EOL policy.

One important aspect of the Parser gem, which RuboCop originally depended on, is that it runs on Ruby 2.0 and above. This separation between runtime requirements and the ability to parse older and newer Ruby syntax has been a key benefit.

For example, RuboCop runs on Ruby 2.7 or later while still being able to analyze code targeting Ruby 2.0 and above. Since RuboCop now depends on Prism, the runtime version requirement of Prism has a significant impact as a dependency.

With the rise of AI-agentic coding, the role of linters and formatters has become even more important. As a widely used tool in the ecosystem, continuing runtime support for Ruby 2.7 for as long as reasonably possible remains important.

For additional context, please see the previous discussion:
https://github.com/rubocop/rubocop/issues/12600#issuecomment-1932707748

kddnewton:
Yeah, I don't think we can do this. Last time I checked it had to do with 2.7.0 being a version that was installed as the base Ruby on a lot of linux distributions. I don't see a lot we can gain from doing this (dropping a couple of polyfills that are only a couple of lines), and there's a ton we could lose (breaking the whole ecosystem).

Issue #4009, opened 2026-03-17
Eliminate templated STRING_TYPE and use a factory


headius:
Currently there are several templates that must be customized to either use `java.lang.String` or `org.jruby.RubySymbol` for various string or identifier representations. This is done via the `JAVA_STRING_TYPE` constant which is customized in template.rb to the specific types used by TruffleRuby and JRuby. Despite the name, this appears to only be used for representing identifiers, e.g. for variables, methods, and class names.

I'd like to eliminate this variability in the Java API by instead providing a factory class that can construct appropriate representations.

The interface might look something like this:

```
public interface IdentitierFactory<IDType> {
    public IDType newIdentifier();
}
```

The parser would be set up to accept such a factory on instantiation, and indentifiable nodes it produces would aggregate values of T rather than the templated `String` or `RubySymbol` that they do now. This would not change usage of the API significantly, other than requiring TruffleRuby and JRuby to both pass in a factory object for creating identifiers.

Nodes that aggregate identifiers (or aggregate nodes that aggregate identifiers) would be generified to propagate this IDType parameter.

```
public class Nodes {
    ...
    public static class LocalVariableReadNode<IDType> extends Node {
         ...
         public final IDType name;
```

This would allow a single Java API to be published that all JVM consumers of the Prism API could use, with only the backends differing (initially publishing a WASM version and later a multi-platform native backend).

As this will require some discussion I'm putting it here as an issue and will for now proceed with publishing a JRuby-specific API artifact so we can ship that in JRuby 10.1.

eregon:
What are concrete issues with the current templated approach?
I don't see a clear problem/motivation in the issue description.

My concerns with this proposal are:
* Slower performance due to extra `checkcast` on every usage of a generic type
* More verbose node definitions, making it look needlessly complicated and boilerplate-Java-like
* More verbose creation of nodes (at least TruffleRuby uses that to desugar some nodes into simpler ones)

IOW, this sounds a perfect use case for templating given we already have templating.

eregon:
What about always using `java.lang.String` as the `JAVA_STRING_TYPE`?
From your messages on Slack I thought that would be the proposal.
That's a neutral type, makes the API significantly easier to use than a factory, and should work fine under JRuby too.

I agree it doesn't make much sense to publish under `org.ruby_lang` if it's still JRuby-specific (would be better under a JRuby Maven namespace then).

headius:
[...]

PR #4016, opened 2026-03-19
Switch identifiers to byte[]

headius:
To remove the compile-time dependency on JRuby in the Java API, we need to replace RubySymbol identifiers with something implementation-independent. The simplest option is to use the original identifier bytes (byte[]).

Fixes https://github.com/ruby/prism/issues/4009.

### DIFF
--- a/templates/java/org/ruby_lang/prism/Loader.java.erb
+++ b/templates/java/org/ruby_lang/prism/Loader.java.erb
@@ -19,21 +19,9 @@ public class Loader {
 
     // Overridable methods
 
-    public Charset getEncodingCharset(String encodingName) {
-        encodingName = encodingName.toLowerCase(Locale.ROOT);
-        if (encodingName.equals("ascii-8bit")) {
-            return StandardCharsets.US_ASCII;
-        }
-        return Charset.forName(encodingName);
-    }
-
-    public <%= id_type %> bytesToName(byte[] bytes) {
-        <%- if id_type == "String" -%>
-        return new String(bytes, encodingCharset).intern();
-        <%- else -%>
-        return bytes;
-        <%- end -%>
-    }
+    <%- if id_type == "String" -%>
+    public abstract <%= id_type %> bytesToName(byte[] bytes);
+    <%- end -%>
 
     private static final class ConstantPool {
 
@@ -63,7 +51,11 @@ public class Loader {
                 byte[] bytes = new byte[length];
                 buffer.get(start, bytes);
 
+                <%- if id_type == "byte[]" -%>
+                constant = bytes;
+                <%- else %>
                 constant = loader.bytesToName(bytes);
+                <%- end %>
                 cache[index] = constant;
             }
 
@@ -74,9 +66,6 @@ public class Loader {
 
     private final ByteBuffer buffer;
     protected String encodingName;
-    <%- if id_type == "String" -%>
-    private Charset encodingCharset;
-    <%- end -%>
     private ConstantPool constantPool;
     private Nodes.Source source = null;
 
@@ -104,9 +93,6 @@ public class Loader {
         byte[] encodingNameBytes = new byte[encodingLength];
         buffer.get(encodingNameBytes);
         this.encodingName = new String(encodingNameBytes, StandardCharsets.US_ASCII);
-        <%- if id_type == "String" -%>
-        this.encodingCharset = getEncodingCharset(this.encodingName);
-        <%- end -%>
 
         source.setStartLine(loadVarSInt());
         source.setLineOffsets(loadLineOffsets());


templates/java/org/ruby_lang/prism/Loader.java.erb
```diff
@@ -31,7 +31,7 @@ public class Loader {
         <%- if string_type == "String" -%>
         return new String(bytes, encodingCharset).intern();
         <%- else -%>
-        return null; // Must be implemented by subclassing Loader
+        return bytes; // Must be implemented by subclassing Loader
```
eregon:
```suggestion
        return bytes;
```

headius:
This is still a draft and needs to be integrated into JRuby so I know if it's working properly.

headius:
This builds properly now and @enebo is looking at integrating it into our prism service for JRuby's internal parsing.

templates/java/org/ruby_lang/prism/Loader.java.erb
```diff
@@ -70,7 +74,7 @@ public class Loader {
 
     private final ByteBuffer buffer;
     protected String encodingName;
-    <%- if string_type == "String" -%>
+    <%- if id_type == "String" -%>
     private Charset encodingCharset;
```
eregon: I think `encodingCharset` could be removed completely (as well as `getEncodingCharset()`) and then `bytesToName` would e.g. `throw new AbstractMethodError();` for the `id_type == "String"` case, forcing it to be implemented in the subclass.
From https://github.com/ruby/prism/issues/4018#issuecomment-4100303712

headius:
Barring any further changes from discussions in #4018 etc, I believe this is ready to go. @enebo has switched over our jruby-prism library to this `byte[]` version and believes that it's probably ok to proceed.

templates/java/org/ruby_lang/prism/Loader.java.erb
headius: That would be reasonable to me. I'll make that change before marking for review.

eregon:
With the default of `byte[]` it sort of breaks the build-java CI job:
https://github.com/ruby/prism/actions/runs/23362184649/job/67968008137?pr=4016#step:4:107
```
      CallNode[Li]
        CallNodeFlags: 33
        receiver: null
        name: "[B@13da7ab0"
```

vs on main (https://github.com/ruby/prism/actions/runs/23356421280/job/67948275346#step:4:107):
```
      CallNode[Li]
        CallNodeFlags: 33
        receiver: null
        name: "x"
```

That CI job only checks that the deserializer works fine but doesn't check the output directly, hence it doesn't show as failed.

It also means AST printing with `byte[]` as identifier type is broken, so that's what should be fixed I think.

templates/java/org/ruby_lang/prism/Loader.java.erb
eregon, line 14: This must be `abstract` if bytesToName() is abstract, otherwise javac fails.

templates/java/org/ruby_lang/prism/Nodes.java.erb
eregon, line 379: This and below need a fix to print a `byte[]` properly

templates/java/org/ruby_lang/prism/Loader.java.erb
eregon: This is what we lose by making the Loader abstract and no longer `javac`-compiling the truffleruby case in CI (i.e. CI didn't notice this bug).
Maybe it shouldn't be abstract and be a runtime AbstractMethodError (https://github.com/ruby/prism/pull/4016#discussion_r2967553672), then at least we could keep testing that `javac`-compiling with `PRISM_JAVA_BACKEND=truffleruby` works in CI and doesn't break without noticing.

templates/java/org/ruby_lang/prism/Loader.java.erb
eregon: I had forgotten but `encodingCharset` was already conditionally added based on the `id_type`, so I see little value to remove it besides cleanup, but that cleanup causes some complications as noticed in my other comments.

IOW, maybe worth dropping that commit to keep it simple.
I'm also fine if you remove it but then the rest should be addressed (keep checking it javac-compiles in CI with `j.l.String`, make printing work again).

kddnewton (approved):

eregon:
@kddnewton Why did you merge a PR that does not even compile with `javac` as noted in https://github.com/ruby/prism/pull/4016#discussion_r2967971664 (and breaks `toString()` output)?
Or IOW why did you ignore my comments and review?

headius:
I thought I tested the TruffleRuby version but I must not have.

I will fix the listed issues in another PR.

eregon:
Reverted in #4022, @kddnewton saw my "Looks good" above and assumed it was good to merge.
@headius Could you redo the PR?

PR #4017, opened 2026-03-19
Implement Chicory WASM backend for the Ruby API

headius:
This PR will provide a new Ruby API backend (in addition to the C extension and the FFI wrapper) based on the Chicory WASM runtime for the JVM.

This version of the gem will be published for JRuby users as the default cross-platform gem. Future gems will make platform-specific binary builds available for users that desire the additional performance.

The goal here is to get a native-code free version of the prism gem released for JRuby users.


headius:
This depends upon changes in #3944 and will be rebased once that merges to main.

eregon:
I wonder if it'd make sense to support the WASM case in `lib/prism/ffi.rb` directly and e.g. define `LibRubyParser` as `WASM::PRISM` or so in that case, or to share some of the logic.
This looks like an early prototype as much of the logic is copied and probably not working yet (since it refers `LibRubyParser` and that's not set in the new file).

headius:
> This looks like an early prototype

Yes, that's why it's a draft.

headius:
Most recent commit is an attempt to move the `PRISM_SERIALIZE_ONLY_SEMANTIC_FIELDS` into a macro. I'll explain why.

The JVM WASM builds need to take two forms:

* A semantic-only version for the WASM build for JRuby's internal parser, which uses the Loader API.
* A non-semantic version for the WASM build for the JRuby gem, which uses the same logic as FFI.

These two artifacts should be sibling modules under `java/` and built together as part of the Maven build lifecycle. Ideally there would remain three steps to build this:

* Generate the templated sources
* Build the two WASM files
* Build the Maven artifacts

But because the templated sources must be generated twice, once for semantic-only and again for non-semantic, the build process ends up being much more complicated:

* Generate the semantic-only sources
* Build the semantic-only WASM
* Generate the non-semantic sources
* Build the non-semantic WASM
* Build the Maven artifacts

By moving the semantic-only flag into a C macro, a single set of sources can be generated and customized at C compile time, allowing a single make target to build both WASM forms.

Ideally the other two major C templating flags could also move into build-time macros:

* Pretty printing could be omitted at C compile time and eliminated from the resulting binary.
* Node ID could be omitted for builds that do not want it. (I had to tweak the `INCLUDE_NODE_ID` template flag because it was intertwined with the templated `PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS`

Really, the only tricky part is for the Java API that has two forms: with node ID and without node ID, and I think we can handle that a different way (perhaps by adding another flag to the serialization header?)

templates/java/api/src/main/java-templates/org/ruby_lang/prism/Loader.java.erb
```diff
@@ -326,12 +326,12 @@ public class Loader {
...
 
     <%-
-    base_params = [*("nodeId" if Prism::Template::INCLUDE_NODE_ID), "startOffset", "length"]
+    base_params = [*("nodeId" unless Prism::Template::OMIT_NODE_ID), "startOffset", "length"]
     base_params_sig = base_params.map { "int #{_1}" }.join(", ")
     -%>
     private Nodes.Node loadNode() {
         int type = buffer.get() & 0xFF;
-        <%- if Prism::Template::INCLUDE_NODE_ID -%>
+        <%- unless Prism::Template::OMIT_NODE_ID -%>
```
eregon: Could you keep the constant as `Prism::Template::INCLUDE_NODE_ID`?
That way:
* The diff is much smaller and focused on the semantic fields change
* There is no double negation (`unless ... OMIT`)

The ENV var can be either.

templates/template.rb
```diff
@@ -7,12 +7,11 @@
 
 module Prism
   module Template # :nodoc: all
-    SERIALIZE_ONLY_SEMANTICS_FIELDS = ENV.fetch("PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS", false)
     CHECK_FIELD_KIND = ENV.fetch("CHECK_FIELD_KIND", false)
 
     JAVA_BACKEND = ENV["PRISM_JAVA_BACKEND"] || "default"
     JAVA_IDENTIFIER_TYPE = JAVA_BACKEND == "truffleruby" ? "String" : "byte[]"
-    INCLUDE_NODE_ID = !SERIALIZE_ONLY_SEMANTICS_FIELDS || JAVA_BACKEND == "jruby"
+    OMIT_NODE_ID = ENV.fetch("PRISM_OMIT_NODE_ID", false)
```
eregon:
```suggestion
    INCLUDE_NODE_ID = ENV.fetch("PRISM_INCLUDE_NODE_ID", "true") != "false"
```
(mentioned on Slack)

templates/src/serialize.c.erb
```diff
@@ -91,8 +91,12 @@ pm_serialize_node(pm_parser_t *parser, pm_node_t *node, pm_buffer_t *buffer) {
             size_t length_offset = buffer->length;
             pm_buffer_append_string(buffer, "\0\0\0\0", 4); /* consume 4 bytes, updated below */
             <%- end -%>
-            <%- unless Prism::Template::SERIALIZE_ONLY_SEMANTICS_FIELDS && !node.flags -%>
+            <%- if node.flags -%>
             pm_buffer_append_varuint(buffer, (uint32_t) node->flags);
+            <%- else -%>
+            if (!PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS) {
```
eregon:
```suggestion
            #ifndef PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS
```
would be better because it then goes away during preprocessing vs later in the compiler, which could potentially make a difference e.g. in inlining due to the function being bigger.
It's also clearer this is not a runtime check, and setting it must be done with `-D`.

That would mean `PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS` must not be defined by default.

templates/template.rb
headius: Done.

templates/src/serialize.c.erb
headius: Done.

.github/workflows/java-wasm-bindings.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           bundler-cache: true
 
       - name: rake templates
-        run: PRISM_EXCLUDE_PRETTYPRINT=1 PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS=1 bundle exec rake templates
+        run: PRISM_EXCLUDE_PRETTYPRINT=1 PRISM_OMIT_NODE_ID=1 bundle exec rake templates
```
eregon:
```suggestion
        run: PRISM_EXCLUDE_PRETTYPRINT=1 PRISM_INCLUDE_NODE_ID=false bundle exec rake templates
```

eregon:
I reviewed [the changes turning the semantic-fields template flag into `#ifdef`](https://github.com/ruby/prism/pull/4017/changes/b81eba8d3deeee8b50592e21b61b07ec6a13b983..cec815b6f32f88a5b30ab60eda5826a93e6d167a), they look good to me.
I think it would be good to eventually squash them to avoid too much back-and-forth in the git history.

headius:
> I think it would be good to eventually squash them

I'll try to clean them up a bit.

.github/workflows/java-wasm-bindings.yml
headius: Done.

kddnewton:
@headius I'm going to close this for now just for cleanliness. Feel free to open again if you'd like to revisit!

headius:
@kddnewton Yeah I understand. There's nothing here that can't be picked right back up so I'll try to get it closer to completion (with some contributor help perhaps) before re-pushing.

Given the move to release pre-compiled binaries in the gem, there is renewed interest in avoiding compilation on JRuby as well. It's so close to being ready; I only had a few API endpoints to finish up.

headius:
See #4080 for plumbing added to precompile the native C extensions for CRuby. Similar logic will be used to precompile WASM (+ Redline AOT binaries) and release the JRuby gem with that build included.

PR #4021, opened 2026-03-20
Prism.find

kddnewton:
Take a method, unbound method, proc, or thread backtrace location. This is our equivalent to RubyVM::AbstractSyntaxTree.of, and could be leveraged in something like error highlight.

Note that this uses CRuby-specific APIs on CRuby, and falls back to using location-based APIs when those aren't available.

Closes #3808.


eregon:
Nice.
And when `Ruby::CodeLocation` or so is added (https://bugs.ruby-lang.org/issues/6012#note-54) then we can also find the precise node on non-CRuby if there are multiple on the same line.

One potential gotcha of using `node_id` is it could return a wrong node if the file is modified.
That could also happen with `Ruby::CodeLocation` but would be less likely to have a node matching the exact same bounds which isn't the right one.
In practice I think it's quite rare that the file is modified concurrently after the program started.

lib/prism.rb
```diff
@@ -81,6 +85,16 @@ def self.load(source, serialized, freeze = false)
     Serialize.load_parse(source, serialized, freeze)
   end
 
+  # Given a Method, UnboundMethod, Proc, or Thread::Backtrace::Location,
+  # returns the Prism node representing it. On CRuby, this uses node_id for
+  # an exact match. On other implementations, it falls back to best-effort
+  # matching by source location line number.
+  #--
+  #: (Method | UnboundMethod | Proc | Thread::Backtrace::Location callable, ?rubyvm: bool) -> Node?
+  def self.find(callable, rubyvm: !!defined?(RubyVM))
```
eregon, line 94: Is it intentional to expose a `rubyvm:` kwarg as public API here?
If it's just for testing, tests could use `NodeFind.find` directly.

PR #4022, opened 2026-03-20
Revert "Switch identifiers to byte[]"

kddnewton:
Reverts ruby/prism#4016


eregon (approved):
Thanks. Let's redo the PR to address the issues detailed in #4016.

PR #4023, opened 2026-03-20
Switch identifiers to byte[]

headius:
To remove the compile-time dependency on JRuby in the Java API, we need to replace RubySymbol identifiers with something implementation-independent. The simplest option is to use the original identifier bytes (byte[]).

Fixes https://github.com/ruby/prism/issues/4009.

### DIFF
--- a/java-wasm/README.md
+++ b/java-wasm/README.md
@@ -5,7 +5,7 @@ This dir contains the chicory-prism artifact, a version of prism compiled to WAS
 Generate the templated sources:
 
 ```
-PRISM_EXCLUDE_PRETTYPRINT=1 PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS=1 PRISM_JAVA_BACKEND=jruby bundle exec rake templates
+PRISM_EXCLUDE_PRETTYPRINT=1 PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS=1 bundle exec rake templates
 ```
 
 Compile to WASM using WASI SDK version 25:


eregon:
Could you add a check in CI in the build-java job that javac-compiling with `PRISM_JAVA_BACKEND=truffleruby` works?

templates/java/org/ruby_lang/prism/Nodes.java.erb
```diff
@@ -139,6 +139,14 @@ public abstract class Nodes {
 
         protected abstract String toString(String indent);
     }
+
+    protected static String asString(Object value) {
+        return value.toString();
+    }
+
+    protected static String asString(byte[] value) {
+        return new String(value, StandardCharsets.ISO_8859_1);
```
eregon: Since UTF-8 is the most common encoding for Ruby files should we use that?
Or maybe US-ASCII since that's the only safe thing we know here?
(we can't access the source encoding in nodes)

I checked and the javadoc of `String(byte[] bytes, Charset charset)` says if it can't map the bytes it just uses a replacement character, which seems good here:
> This method always replaces malformed-input and unmappable-character sequences with this charset's default replacement string.

headius: Decoding this as ISO-8859-1 allows us to see every byte in the identifier as an individual character regardless of original encoding. It won't be easily readable for multi-byte characters, but the alternative suggestion will completely drop any non-US-ASCII characters that aren't UTF-8.

So it's a question of whether we want the `toString` output to always display bytes as characters or drop characters for identifiers with non-UTF-8 byte values (>127).

If the node aggregated the encoding name, we could simply use the correct Charset (provided it is supported by JDK).

If we were using jcodings, we could call on that to transcode to UTF-8 and always print the correct string.

We could also just print out the `Arrays.toString` output and see the raw bytes directly, but that's probably the least useful for user-friendly `toString` output.

eregon: I think the best and safest here is US-ASCII, then we don't misinterpret anything.
For example if there is a UTF-8 String with a single byte 233, it would incorrectly be rendered as `é` if we use `ISO_8859_1`.

> If the node aggregated the encoding name, we could simply use the correct Charset (provided it is supported by JDK).

Yeah but it's too much of a footprint overhead.

This `toString()` is mostly for debugging, so it's fine if it doesn't show everything, but I think it should not be misleading by interpreting bytes it doesn't know the encoding of.

(US-ASCII is safe because all source encodings are ASCII-compatible, a .rb file in UTF-16/32 is just an immediate error)

.github/workflows/java-wasm-bindings.yml
```diff
@@ -52,3 +52,36 @@ jobs:
         with:
           name: prism.wasm
           path: java-wasm/src/test/resources/prism.wasm
+
+  build-truffleruby:
```
eregon: Re the CI job I meant adding a few lines to https://github.com/ruby/prism/blob/ec3a427ed5f650a24661ddf36c18941727348977/.github/workflows/main.yml#L177-L189
Compiling to WASM seems unnecessary heavy here, it's just about `javac` with `PRISM_JAVA_BACKEND=truffleruby`.

eregon: https://github.com/ruby/prism/actions/runs/23366495516/job/67981450562?pr=4023 seems fast enough but it seems overly complicated for what we're testing here, and the job name `build-truffleruby` is confusing since it doesn't build TruffleRuby.

I think using https://github.com/ruby/prism/blob/ec3a427ed5f650a24661ddf36c18941727348977/Rakefile#L44-L50 is simplest.

headius: Already fixed.

templates/java/org/ruby_lang/prism/Nodes.java.erb
headius: I think I found the best compromise: US-ASCII with characters outside [0x20,0x7F] escaped. I've made that change for both identifiers and other StringFields.

eregon: Nice, similar to how the Ruby BINARY encoding does `inspect`.
For me I use this output mostly while already in the Java debugger so US-ASCII  and `?` for non-ASCII would be enough, but this is even better.

headius:
@kddnewton I believe we are all good here!

PR #4039, opened 2026-03-24
Maven tweaks for usability

headius:
Based on review comments from #4031 I've made a few additional tweaks.

* Template sources are now generated into `src/main/java-templates` so they are not deleted by `mvn clean`.
* The WASM build is generated to `src/main/wasm` since it is an input to the build (and not a test resource as before).
* The build-java-truffleruby GHA job now uploads a copy of the generated sources for TruffleRuby, to make it easier to copy.


headius:
The archive of the sources generated specifically for TruffleRuby could be moved to a separate job if that would be cleaner.

eregon:
Could Loader.java and Nodes.java simply be in `java/api/src/main/java/org/ruby_lang/prism`, as siblings of e.g ParseResult.java?
They are not generated by the Maven build and it's expected to run `rake` so it should be fine, no?

.github/workflows/main.yml
```diff
@@ -199,6 +199,15 @@ jobs:
         bundler-cache: true
     - name: Run Java Loader test
       run: PRISM_SERIALIZE_ONLY_SEMANTICS_FIELDS=1 PRISM_JAVA_BACKEND=truffleruby bundle exec rake compile
+    - name: Archive sources for TruffleRuby
+      run: |
+        jar --create --file java/prism-truffleruby-sources.jar -C java/api/src/main/java `ls -r java/api/src/main/java`
+        jar --update --file java/prism-truffleruby-sources.jar -C java/api/src/main/java-templates `ls -r java/api/src/main/java-templates`
+        jar --update --file java/prism-truffleruby-sources.jar -C java/native/src/main/java `ls -r java/native/src/main/java`
```
eregon: I appreciate the thought and effort but it's better to directly copy files from my Prism checkout, that way it also works if e.g. I have some local changes in Prism I'm testing in TruffleRuby.
So this can be removed.

java/native/src/main/java/org/ruby_lang/prism/jni/Parser.java
```diff
@@ -1,4 +1,4 @@
-package org.ruby_lang.prism;
+package org.ruby_lang.prism.jni;
 
 public abstract class Parser {
```
eregon: This class, which is the only class under `java/native` doesn't really do anything currently.
Notably the corresponding C code is not here, it's in https://github.com/truffleruby/truffleruby/blob/master/src/main/c/yarp_bindings/src/yarp_bindings.c.

Maybe we should actually remove this class? Or import the C file here?
If the C file would get compiled in CI that could be nice, as it'd catch cases that it needs to be updated, e.g. with the recent changes in the C API.

OTOH this is an area where usages might want different things (e.g. how the reading is done, maybe from a file/using mmap/etc), and notably truffleruby compiles that .c file together with libprism.a.

I'd tend to say the simplest is to just remove this file, but then maybe there is an issue because `java/native` is empty?

headius: The logical change would be to put the JNI C code in here as well and use Maven to build it. But there's the larger issue of building, packaging, and releasing that JNI library.

I have looked briefly at the available Maven tooling for building and releasing native libraries but never attempted to use them:

https://www.mojohaus.org/maven-native/native-maven-plugin/

Unfortunately the source links for this plugin appear to be dead. I'm not sure what is considered "standard" for building and releasing JNI libraries at this point.

headius: Ah, the link from the mojohouse page was incorrect. The plugins are here:

https://github.com/mojohaus/maven-native

headius: Looking into other projects that publish native libraries, I have found no standard tooling to do this.

lwjgl builds and publishes an extensive set of native binaries, but they also have an enormous set of ant build files to do it. MacOS example: https://github.com/LWJGL/lwjgl3/blob/4ef1eebe4af235b2934a165e82aeefcaf8d9b893/config/macos/build.xml#L140

Of course JRuby's JFFI also publishes many native libraries, but we have always built them by hand and committed them to the repository due to the challenges of doing this automatically.

eregon: I think the simplest for now would be to remove this file then, we can always add it back later if useful. It's like 3 lines of code and not useful on its own.

It doesn't need to be solved in this PR, but if it removes an extra Maven module and/or simplifies things it seems useful to remove it earlier than later.

eregon: FWIW TruffleRuby and more specifically `mx` has its own way to build & package this stuff but it's not reusable here as I'd think we wouldn't want a dependency on mx.

.github/workflows/main.yml
headius: Ok, I'll just remove this.

java/native/src/main/java/org/ruby_lang/prism/jni/Parser.java
headius: It's come to my attention while updating progress with @enebo that TruffleRuby is the only consumer of the JNI library. For JRuby's internal parser use, we bind the parser with JNR or WASM, and for the gem we will use FFI or WASM.

So perhaps you should just own this class and we can eliminate the artifact for now. It adds no functionality beyond what the WASM library will provide, and we have no good answer to handling the native libraries yet.

headius: Resolved after @eregon agreed TR will own the JNI binding and I removed the native module.

.github/workflows/main.yml
headius: Removed from commits.

headius:
After removing the native module altogether along with the archive of sources for TruffleRuby, this boiled down to just some template-generation tweaks.

The updated plan for published artifacts (from a JRuby perspective):

* prism-parser-api: the Java API
* prism-parser-wasm: the WASM backend of the internal Prism parser build (no non-semantic contents, used by JRuby)
* prism-parser-wasm-full: (name pending) the WASM build of Prism with full non-semantic info, to be used from FFI by JRuby for now and for a full non-semantic enhancement to the API in the future.

Of these, the first two are working as published Maven modules today and JRuby 10.1 will ship with support for them. Additional Ruby artifacts built around these:

* prism gem for JRuby: defaulting to using the WASM backend (prism-parser-wasm-full)
* prism-native gem for JRuby: essentially what the FFI version of the gem contains today, opt-in for JRuby users who require native performance out of the gem.
* jruby-prism-parser gem: (name pending) a gem whos sole purpose is to build the native non-semantic prism parser and install into the current JRuby distribution for our internal parsing, for users who want native prism performance at boot with JRuby.

Of these, only the jruby-prism-parser gem exists but it has never been merged into the main prism repository. The WASM-based gem is WIP in #4017. The native gem for JRuby will be a rework of the existing FFI gem.

@kddnewton This is ready to be merged.

PR #4069, opened 2026-04-06
Add CI job to build C sources with various compiler versions and fix array-bounds error with GCC 9

eregon:
From https://github.com/ruby/prism/pull/4051#pullrequestreview-4021612093

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -174,6 +174,18 @@ jobs:
       env:
         PRISM_BUILD_MINIMAL: "1"
 
+  build-with-various-compilers:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ruby
+        bundler-cache: true
+    - name: Build with various compilers
+      run: bundle exec rake build_in_docker
+
   build-java:
     runs-on: ubuntu-latest
     env:
--- a/Rakefile
+++ b/Rakefile
@@ -79,15 +79,14 @@ end
 
 task :build_in_docker => :templates do
   # Versions from https://github.com/ruby/ruby/blob/master/.github/workflows/compilers.yml
-  compilers = (7..15).map { |v| "gcc-#{v}" }
-  compilers.each do |compiler|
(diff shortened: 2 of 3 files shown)


kddnewton (dismissed):
What's the deal with the changes to parser.h here?

eregon:
@kddnewton Could you check the commit message? It explains it.

eregon:
The new job took [10min15s](https://github.com/ruby/prism/actions/runs/24032615787/job/70084557719?pr=4069), which is faster than [memcheck](https://github.com/ruby/prism/actions/runs/24032615787/job/70084557766?pr=4069).
So maybe it's fast enough for now?

Not sure where the time is spent but I wondered about image size:
```
$ docker images
ghcr.io/ruby/ruby-ci-image                                gcc-15               771cea939adf  6 days ago      3.18 GB
docker.io/library/gcc                                     15                   9a52dbcd9b63  2 weeks ago     1.57 GB
```
Maybe we could use `gcc:15` but then we'd also need `ruby` inside, or change `Makefile` to no longer depend on ruby for `SOEXT` (or just fallback to `.so` if no `ruby`).

eregon:
> @kddnewton Could you check the commit message? It explains it.

To explain a bit more, the code does `memset(breakpoints, 0, PM_STRPBRK_CACHE_SIZE);` but that's at least theoretically invalid because PM_STRPBRK_CACHE_SIZE=16 and `breakpoints = lex_mode.as.list.breakpoints` and that's defined as `uint8_t breakpoints[11];` or `uint8_t breakpoints[7];`.
It's an out of bound access, but maybe it doesn't cause any issue in practice due to alignment and/or the `union`, nevertheless it seems far safer to use `uint8_t breakpoints[PM_STRPBRK_CACHE_SIZE];` (and fixes the GCC 9 error).

eregon:
I switched to `gcc` images and it's quite a bit faster :) [6m49s](https://github.com/ruby/prism/actions/runs/24041442282/job/70113976853?pr=4069)

eregon:
@Earlopain Could you help reviewing this?

Earlopain (approved):
Yeah, this looks fine to me. Originally this was done to fix some ASAN errors https://github.com/ruby/prism/commit/1cabef7946746d9468891c8827d10754bbeee281

PR #4077, opened 2026-04-14
Implement bracket/braces events for ripper

Earlopain:
Drops the check against order. Very often ripper emits events in a order that is not easy to mimic. It's only getting worse now that most events are implemented.

Perhaps the test can be brought back at a later time. For now, I used it while making this commit but adding all these exceptions makes not much sense. It already was a pretty long list.

Also, now that there are so many new events, I should look at optimizing `bounds`. I haven't checked performance but suspect the impact is noticable. We should do something similar to https://github.com/ruby/ruby/commit/c45f781771314a71856c9b348c640ba532f54349.

### DIFF
--- a/lib/prism/translation/ripper.rb
+++ b/lib/prism/translation/ripper.rb
@@ -22,16 +22,9 @@ module Prism
     # - on_comma
     # - on_ignored_nl
     # - on_ignored_sp
-    # - on_kw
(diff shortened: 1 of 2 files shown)


kddnewton (approved):
Ahh nice, that's much cleaner

PR #4080, opened 2026-04-16
Add native gem precompilation support

### LINKED ISSUE #4026: Support pre-compiled gems in our rubygems publishing workflow

flavorjones:
Closes #4026

Package precompiled native gems for Ruby 3.3, 3.4, and 4.0 across 10 platforms: aarch64-linux-gnu, aarch64-linux-musl, aarch64-mingw-ucrt, arm-linux-gnu, arm-linux-musl, arm64-darwin, x64-mingw-ucrt, x86_64-darwin, x86_64-linux-gnu, and x86_64-linux-musl.

- Add Ruby-version-aware extension loader in lib/prism.rb with GLIBC error message for musl/glibc mismatches
- Add cross-compilation support to Rakefile using rake-compiler-dock
- Add a reusable `build-gems.yml` workflow called by both the CI workflow (`test-gems.yml`) and the publish workflow (`publish-gem.yml`), building source and native gems across all platforms and Ruby versions
- Add scripts for building, verifying, and testing gems

Update `publish-gem.yml` to handle multi-gem publishing. Note that `rubygems/release-gem` only supports a single gem, so this uses `rubygems/configure-rubygems-credentials` and loops `gem push` over each built `.gem`. Add a `workflow_dispatch` trigger alongside the existing tag-push trigger, and create/update a GitHub release for the tag with all gems plus a `CHECKSUMS.txt` manifest.

Move testing of the "ruby" platform gem (compile-from-source) out of `main.yml` into `test-gems.yml` so we're testing the gem built by the workflow tooling.

---

This approach uses rake-compiler-dock to build a broad set of platform gems, including GNU and Musl linux variants.

I also have an experimental branch that uses https://github.com/shopify/cibuildgem, however I think for a gem like prism it may be prudent to choose the more mature rake-compiler-dock toolchain, until cibuildgem is at parity.


kddnewton:
Only minor comment (that can be left until after merge) is that I'd love these `bin` scripts to be in rake. Otherwise the approach looks reasonable.

It seems like right now the GitHub action triggers on a bunch of stuff, but presumably after you get CI passing we'll switch it to just tags right?

flavorjones:
> Only minor comment (that can be left until after merge) is that I'd love these bin scripts to be in rake. Otherwise the approach looks reasonable.

I don't think it's obvious that they should be rake tasks, but I do want to try to follow the project conventions. If that's what you want, we can try to make it work.

The shell scripts are intentionally operating outside of bundler, and test-gem-file-contents is a self-contained minitest suite. `build-gems` will be replaced with a publishing workflow.

> It seems like right now the GitHub action triggers on a bunch of stuff, but presumably after you get CI passing we'll switch it to just tags right?

I mean, it's obviously up to you when you want to run these tests, but my implementation decision to run these tests on every pull request is intentional. The new workflow isn't publishing, it's testing:

- ensure that the gem can be precompiled on all target platforms
- run the test suite against the installed precompiled gem to make sure it's functionally correct and does not generate runtime linker errors

and I think these are valuable tests for any PR that modifies the source code or the build process.

flavorjones:
Looking into the failing windows-11-arm builds with https://github.com/sparklemotion/nokogiri/pull/3619

flavorjones:
Force-pushed some changes:

- extracted a reusable workflow that's used for both **testing** native gems and **publishing** native gems
- did some security hardening (pinning actions, extracting matrix variables into env vars to prevent injection, etc.)

flavorjones:
@kddnewton This is ready for your review. LMK if you have any questions, happy to jump on slack or whatever! ❤️

flavorjones:
@kddnewton Please let me know if there's anything I can do here. This needs to be rebased because of dependabot changes to the CI workflows, but I'd prefer to hold off on putting any more work in here until it's been reviewed we think it's worth finishing up.

kddnewton:
@flavorjones so sorry I have gotten so far behind on my open source stuff. Thank you for working on this, it looks really good. Let me get a little bit of time to make sure I understand everything and I will do a proper pass.

flavorjones:
@kddnewton I'm happy to pair with you on a review, too, if that will make it easier for you. I'm at your disposal, just let me know.

kddnewton:
Yes please! I'll message you on slack

flavorjones:
@kddnewton Rebased onto `main`.

Note that I removed the version-pinning of github actions since that's a distinct hardening step which (if adopted) should be applied consistently to _all_ of the publishing workflows.

kddnewton (approved):

headius:
Next step on JRuby side is to ship a precompiled WASM build and finish up #4017. That PR was closed but there's only bits of work left to do to ship it (including tying the WASM precompile and release into the same logic as the C extension precompile and release.

PR #4083, opened 2026-04-20
Optimize ripper bounds, and other small perf tweaks

Earlopain:
Basically a port of https://github.com/ruby/ruby/commit/c45f781771314a71856c9b348c640ba532f54349 into ruby

It's quite effective at ~97% hit rate for me. Speeds it up from ~6.77x slower to only 4.07x slower.

For the lexer `on_sp` it also gives a bit of an improvement: 1.04x slower to 1.10x faster

I guess the class may be universally useful but for now I just made it nodoc. I also didn't look much into the window, it seemed good enough.

----

Other commit is a small hack to skip redudant pack/unpacking of locations. Prism by default packs locations into a 64bit integer and later extracts start offset / length to create the location. But we know we access nearly all of the locations anyways, so we can request a frozen result to compute them eagerly. It gives about a 4% speed boost.

---

Benchmark from https://github.com/ruby/prism/commit/2ea81398cc6f776dca686125847dbf9b0af58c16

### DIFF
--- a/lib/prism/translation/ripper.rb
+++ b/lib/prism/translation/ripper.rb
@@ -1007,7 +1007,7 @@ module Prism
             on_stmts_add(on_stmts_new, on_void_stmt)
           else
             body = node.statements.body
-            body.unshift(nil) if void_stmt?(location, node.statements.body[0].location, allow_newline)
+            body = [nil, *body] if void_stmt?(location, node.statements.body[0].location, allow_newline)
 
             bounds(node.statements.location)
             visit_statements_node_body(body)
@@ -1024,7 +1024,7 @@ module Prism
                 [nil]
               else
                 body = else_clause_node.statements.body
-                body.unshift(nil) if void_stmt?(else_clause_node.else_keyword_loc, else_clause_node.statements.body[0].location, allow_newline)
+                body = [nil, *body] if void_stmt?(else_clause_node.else_keyword_loc, else_clause_node.statements.body[0].location, allow_newline)
                 body
               end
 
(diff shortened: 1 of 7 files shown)


kddnewton:
I don't think I want that generally in parse result. Can you put that into the ripper translator? (Just not sure if we truly want to support that for everyone going forward.)

I see you must have determined `unshift` wasn't as fast as reassigning and splatting the array. That really surprises me, are you sure that's the case?

Earlopain:
> I don't think I want that generally in parse result. Can you put that into the ripper translator?

Yeah, I can move it. I had it nodoc but that's fine too.

> I see you must have determined unshift wasn't as fast as reassigning and splatting the array. That really surprises me, are you sure that's the case?

It's probably not. But it also only happens for what ripper considers void statements. The change is only to make https://github.com/ruby/prism/pull/4083/commits/3e8c264d9069a350214e1bccbc51f4c9780e0288 work, and overall it comes out positive.

lib/prism/translation/ripper.rb
kddnewton, line 4118: If you want to speed this up, I think you should take advantage of opt_case_dispatch by making every key that's possible hash-able. That would look something like

```suggestion
        case token[0]
        when "."
          on_period(token)
        when "`"
          on_backtick(token)
        when "_"
          on_ident(token)
        when "@"
          token[1] == "@" ? on_cvar(token) : on_ivar(token)
        when "$"
          on_gvar(token)
        else
          if allow_keywords && KEYWORDS.include?(token)
            on_kw(token)
          elsif token.match?(/^[[:upper:]]\w*$/)
            on_const(token)
          elsif token.match?(/^[[:punct:]]/)
            on_op(token)
          else
            on_ident(token)
          end
        end
```

that should help because the case will become a jump table, with the other arms falling through to the linear check.

Earlopain: I tried that but it doesn't work out to be faster (maybe because `token[0]` allocates a string?). Good idea regardless, didn't know about that trick

Out of curiosity I gave `token.getbyte(0)` a try (should be safe I think, although the code becomes less readable) but not much difference for that as well. So I think I'll just leave this be as is

kddnewton (approved):

PR #4092, opened 2026-04-27
Fix missing Ripper doc caused by shim with `:nodoc:`

ybiquitous:
This fixes the missing Ripper documentation, which currently returns 404 Not Found:

image

https://docs.ruby-lang.org/en/master/Ripper.html

We need to use `:stopdoc:`/`startdoc:` instead of `:nodoc:` to ignore the shim markup.

See also https://ruby.github.io/rdoc/doc/markup_reference/rdoc_rdoc.html#directives

> [!NOTE]
> This PR is thanks to @st0012's [advice](https://x.com/_st0012/status/2048765110966247579?s=20). 😃

### DIFF
--- a/lib/prism/translation/ripper/shim.rb
+++ b/lib/prism/translation/ripper/shim.rb
@@ -2,4 +2,6 @@
 
 # This writes the prism ripper translation into the Ripper constant so that
 # users can transparently use Ripper without any changes.
-Ripper = Prism::Translation::Ripper # :nodoc:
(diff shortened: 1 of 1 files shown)


kddnewton (approved):

Earlopain:
Nice, thanks. I openend https://github.com/ruby/rdoc/issues/1662 for this on the rdoc side since it started happening after a change they made but this is fine as well.

PR #4094, opened 2026-04-29
Check CRuby depend is up-to-date

Earlopain:
I often forget that this is a thing, and it fails CRuby CI
PRs that make such changes should first be made to ruby/ruby to avoid this


kddnewton (approved):
Nice that's helpful!

PR #4095, opened 2026-04-29
Fix CI path excludes

Earlopain:
They need wildcards, otherwise they don't match.
Previously it always ran on PRs but in https://github.com/ruby/ruby-prism/commit/c4d202dc4866d6f03cc6cbda083a8950c6e6e42e I also applied it to PRs.
That means that previously it didn't run on main only.

This fixes this.

### DIFF
--- a/.github/workflows/cpp-bindings.yml
+++ b/.github/workflows/cpp-bindings.yml
@@ -4,9 +4,9 @@ on:
   push:
     paths: &paths
       - ".github/workflows/cpp-bindings.yml"
-      - "include/"
-      - "src/"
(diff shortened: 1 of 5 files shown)


Earlopain:
https://github.com/ruby/prism/pull/4093 bumped JRuby but it fails CI (didn't actually run) so I reverted it (cc @andreaTP
 https://github.com/ruby/prism/actions/runs/25101220259/job/73550648784)

andreaTP:
Thanks for the heads up, cc. @enebo @headius I think it might be on their plate

PR #4098, opened 2026-04-29
Fix UTF-8 code units to match the number of bytes

vinistock:
I was seeing some weird results for UTF-8 in the Ruby LSP and I realized we actually made a mistake when implementing the code unit offsets. The idea is for this method to return code units (not code points!). That means:

- UTF-32: number of characters
- UTF-16: the weird code unit count with 1/2 byte characters being length 1 and 3/4 byte characters length 2
- UTF-8: **number of bytes**

[Spec reference](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#positionEncodingKind): see the comments above each position encoding kind.

We are actually not returning the number of bytes, but the string length, which is incorrect. It's easy to see the mistake in the tests: the location of an emoji (4 bytes) was 1 before (single character).

This PR makes sure that we're returning the number of bytes for UTF-8, which is the amount expected for code units.

Basically, I added a UTF-8 counter that just returns the number of bytes and tried to ensure naming consistency across the counters.

### DIFF
--- a/lib/prism/parse_result.rb
+++ b/lib/prism/parse_result.rb
@@ -189,6 +189,8 @@ module Prism
     #--
     #: (Integer byte_offset, Encoding encoding) -> Integer
     def code_units_offset(byte_offset, encoding)
+      return byte_offset if encoding == Encoding::UTF_8
+
       byteslice = (source.byteslice(0, byte_offset) or raise).encode(encoding, invalid: :replace, undef: :replace)
 
       if encoding == Encoding::UTF_16LE || encoding == Encoding::UTF_16BE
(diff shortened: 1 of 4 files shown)


kddnewton:
To get typecheck to pass you need to run `BUNDLE_GEMFILE=gemfiles/typecheck/Gemfile bundle exec rake typecheck:rbs_inline typecheck:rbi`

lib/prism/parse_result.rb
```diff
@@ -250,6 +252,14 @@ def find_line(byte_offset) # :nodoc:
   #   has not yet been implemented.
   #
   class CodeUnitsCache
+    # Counter used for UTF-8, where one code unit equals one byte.
+    class UTF8Counter
```
Earlopain:
```suggestion
    class UTF8Counter # :nodoc:
```

vinistock:
Just FYI: running the command generated other RBI/RBS changes that aren't related to this PR, so I think the files might be out of sync in main despite not causing any type checking failures.

I only included the updates for parse result.

Earlopain (approved):
👍 looks good, thanks!

PR #4102, opened 2026-05-06
Check host_os rather than RUBY_PLATFORM

headius:
RUBY_PLATFORM has been "java" on JRuby since the mid 2000s, so use "host_os" from RbConfig::CONFIG to compare the host platform.

This fixes installs on FreeBSD and OpenBSD which need to use gmake instead of make.

cc @jeremyevans who ran into this upgrading minitest while running JRuby on OpenBSD. Please confirm!

### DIFF
--- a/ext/prism/extconf.rb
+++ b/ext/prism/extconf.rb
@@ -50,7 +50,7 @@ def make(env, target)
   Dir.chdir(File.expand_path("../..", __dir__)) do
     system(
       env,
-      RbConfig::CONFIG['host_os'].match?(/openbsd|freebsd/) ? "gmake" : "make",
+      RbConfig::CONFIG['host_os'].match?(/openbsd|freebsd/i) ? "gmake" : "make",
       target,
       exception: true
     )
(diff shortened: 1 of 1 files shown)


ext/prism/extconf.rb
```diff
@@ -50,7 +50,7 @@ def make(env, target)
   Dir.chdir(File.expand_path("../..", __dir__)) do
     system(
       env,
-      RUBY_PLATFORM.match?(/openbsd|freebsd/) ? "gmake" : "make",
+      RbConfig::CONFIG['host_os'].match?(/openbsd|freebsd/) ? "gmake" : "make",
```
jeremyevans: The existing regexp was already broken as it wasn't case insensitive.  This fixes things for me:

```suggestion
      RbConfig::CONFIG['host_os'].match?(/openbsd|freebsd/i) ? "gmake" : "make",
```

headius:
@jeremyevans Thanks for the catch!

@kddnewton This can be squashed and merged once green.

kddnewton (approved):

PR #4106, opened 2026-05-08
Respect `encoding` option in `Prism.lex` and friends / Fix ripper encoding handling

Earlopain:
Two separate fixes for https://github.com/truffleruby/truffleruby/issues/4175:
* `Prism.lex(..., encoding: foo)` always created tokens in utf-8 when no magic comment is present, ignoring the passed encoding
* In Ripper, when no encoding comment is present, the encoding is taken from the passed string instead.

cc @eregon


eregon (approved):
Thank you!

kddnewton (approved):

PR #4120, opened 2026-06-01
Reject modifier conditionals in `if`/`unless` predicates

kai-matsudate:
Fixes [Bug #22089](https://bugs.ruby-lang.org/issues/22089).

Prism accepted a statement modifier (`if`/`unless`/`while`/`until`) in an `if`/`unless`/`elsif` condition, while parse.y rejects it.

```ruby
if a if b then end   # parse.y: SyntaxError, Prism: accepted
```

Those conditions were parsed with a binding-power floor of `PM_BINDING_POWER_MODIFIER`. That floor is inclusive: statement modifiers have left binding power `PM_BINDING_POWER_MODIFIER`, so they were consumed as part of the condition. `while`/`until` conditions already use `PM_BINDING_POWER_COMPOSITION` and reject this syntax.

This raises the floor to `PM_BINDING_POWER_MODIFIER + 1`, which keeps `and`/`or` and tighter operators in the condition but excludes statement modifiers, matching parse.y.

`while`/`until` use `PM_BINDING_POWER_COMPOSITION` for the same effect because there is currently no precedence between `MODIFIER` and `COMPOSITION`. I used `PM_BINDING_POWER_MODIFIER + 1` to match the existing `+ 1` floor idiom (`PM_BINDING_POWER_DEFINED + 1`, `PM_BINDING_POWER_MATCH + 1`) and state the boundary directly. I can switch this to `PM_BINDING_POWER_COMPOSITION` if keeping the paths aligned is preferred.

Added an error fixture under `test/prism/errors/`.

### DIFF
--- a/src/prism.c
+++ b/src/prism.c
@@ -15486,7 +15486,7 @@ parse_conditional(pm_parser_t *parser, pm_context_t context, size_t opening_newl
     pm_token_t keyword = parser->previous;
     pm_token_t then_keyword = { 0 };
 
-    pm_node_t *predicate = parse_predicate(parser, PM_BINDING_POWER_MODIFIER, context, &then_keyword, (uint16_t) (depth + 1));
+    pm_node_t *predicate = parse_predicate(parser, PM_BINDING_POWER_MODIFIER + 1, context, &then_keyword, (uint16_t) (depth + 1));
     pm_statements_node_t *statements = NULL;
 
     if (!match3(parser, PM_TOKEN_KEYWORD_ELSIF, PM_TOKEN_KEYWORD_ELSE, PM_TOKEN_KEYWORD_END)) {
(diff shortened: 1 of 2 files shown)


kddnewton:
Yeah this is great, I'm good with almost all the changes here. I think we should go with `PM_BINDING_POWER_COMPOSITION` though. The `+ 1` is really to indicate associativity, as opposed to the blanket binding power.

kai-matsudate:
The two failing `cargo:test` jobs look unrelated to this change.
I'm not familiar with the Rust/wasm, but it looks like a wasi-sdk toolchain issue rather than the parser.
rust-lang/rust#146843 seems related.
All other jobs pass.

kai-matsudate:
> Yeah this is great, I'm good with almost all the changes here. I think we should go with `PM_BINDING_POWER_COMPOSITION` though. The `+ 1` is really to indicate associativity, as opposed to the blanket binding power.

@kddnewton Thanks for the advice, switched to `PM_BINDING_POWER_COMPOSITION` in 399a641.

kddnewton (approved):
Thanks!

PR #4124, opened 2026-06-03
Fix WASI undefined symbol in rust examples

Earlopain:
Trying to see what to do about the recent wasi failures.

I'm fairly sure its because of https://github.com/rust-lang/rust/pull/149868 / https://blog.rust-lang.org/2026/04/04/changes-to-webassembly-targets-and-handling-undefined-symbols/


Earlopain:
Bumping WASI seems to do the trick. But I have no idea why jruby on windows fails during actions/download-artifact now. There are 0 usable logs.

Edit: Should have know it was just github being itself. A rerun of all jobs resolved it but not an individual rerun of the failing job

kddnewton (approved):

PR #4136, opened 2026-06-17
Fix `Prism.parse_comments` locations for FFI backend

Earlopain:
Offsets are not serialized but the locations require them to work correctly.

`start_line` returned 0 previously.

### DIFF
--- a/include/prism/internal/serialize.h
+++ b/include/prism/internal/serialize.h
@@ -15,6 +15,11 @@
 #ifndef PRISM_EXCLUDE_SERIALIZATION
 
 /*
+ * Serialize the given list of offsets to the given buffer.
+ */
+void pm_serialize_line_offset_list(pm_line_offset_list_t *list, pm_buffer_t *buffer);
+
+/*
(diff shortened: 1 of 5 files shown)


kddnewton (approved):

eregon:
Great fix. It turns out this impacts `rbs`: https://github.com/ruby/rbs/pull/2996#discussion_r3846269842
How can we fix there? I think prism would need a release with this fix.

Actually based on https://github.com/ruby/rbs/issues/348#issuecomment-3805668694 maybe `rbs` could use `Prism.parse` which avoids the issue.

Earlopain:
Would need a release, yeah. But as you noticed rbs shouldn't use `parse_comments` in the long term. You can sync prism in truffleruby but I think gemfiles would always overwrite this.

eregon:
> You can sync prism in truffleruby but I think gemfiles would always overwrite this.

Yes, my thinking as well (would only maybe work with `--prefer-local` which is not commonly used), though I'll do it anyway.

PR #4151, opened 2026-06-29
Remove excess error from config

Earlopain:
Seems to have been left unused since https://github.com/ruby/prism/commit/826657232e5217b5c5de6992ea639d61807419c2

Now this causes build failures in windows CI, not sure why only now:
```
src/diagnostic.c:781:46: error: array subscript 329 is above array bounds of 'const pm_diagnostic_data_t[329]' [-Werror=array-bounds=]
compiling build/shared/integer.o
  781 |     const char *message = diagnostic_messages[diag_id].message;
compiling build/shared/json.o
      |                           ~~~~~~~~~~~~~~~~~~~^~~~~~~~~
compiling build/shared/line_offset_list.o
src/diagnostic.c: In function 'pm_diagnostic_list_append':
src/diagnostic.c:97:35: note: while referencing 'diagnostic_messages'
   97 | static const pm_diagnostic_data_t diagnostic_messages[PM_DIAGNOSTIC_ID_MAX] = {
      |                                   ^~~~~~~~~~~~~~~~~~~
```


Earlopain:
Hm, I thought it would be related to this, but I guess not

Earlopain:
So, removing these two asserts:
https://github.com/ruby/prism/blob/880833daca47a9f4ed06c7524e0b196f8d01cfc2/templates/src/diagnostic.c.erb#L448
https://github.com/ruby/prism/blob/880833daca47a9f4ed06c7524e0b196f8d01cfc2/templates/src/diagnostic.c.erb#L458

Resolves the error. That doesn't make a whole lot of sense to me. Do you have an idea @kddnewton?

kddnewton:
Looking into this now

kddnewton:
Proving to GCC that it's okay seemed to work (https://github.com/ruby/prism/pull/4158)

kddnewton:
If you want to reopen with just removing the dead diagnostics that'd be fine

Earlopain:
Weird, wish I knew why that's needed but I suppose its fine. Almost seems like a compiler bug but I'm not going to setup windows to investigate that. Thanks for handling.

PR #4158, opened 2026-07-05
Assume for the GCC bounds checks

kddnewton:


Earlopain:
I get this warning now:
>  warning: ‘assume’ attribute ignored [-Wattributes]

```
$ gcc --version
gcc (GCC) 16.1.1 20260625
```

kddnewton:
Ugh that's annoying. Thanks for providing the version, I'll fix. That's weird it fixes the problem, maybe it's only needed for old GCC?

kddnewton:
@Earlopain https://github.com/ruby/prism/pull/4170

PR #4160, opened 2026-07-07
Fix a bug that `Prism.parse_stream` may write too much data to internal buffer

kou:
`pm_source_stream_read()` has 4096 bytes internal buffer to read a line:
https://github.com/ruby/prism/blob/39fb41f06e5f4ccbe1783122fa8a4470e09dfcfe/src/source.c#L368-L369

It uses `gets(4095)` to read the next chunk to the buffer: https://github.com/ruby/prism/blob/39fb41f06e5f4ccbe1783122fa8a4470e09dfcfe/ext/prism/extension.c#L1076

But `gets(4095)` may read 4095 or more data when the character at 4095 is a multi-byte character. In the case, `parse_stream_fgets()` writes 4097 or more data to the buffer.

We can avoid it by increasing internal buffer size to `4096 + 6` where `6` is the max character length in all available encodings. We can reduce it by using `stream.external_encoding` and `rb_enc_mbmaxlen()` but it may slow down. So I used constant `6` in this change.

This also changes `pm_source_stream_fgets_t`'s return type to `int` (read bytes) from `char *` (read data) to reduce needless length check by caller.

### DIFF
--- a/src/source.c
+++ b/src/source.c
@@ -365,27 +365,49 @@ bool
 pm_source_stream_read(pm_source_t *source) {
     pm_buffer_t *buffer = source->stream.buffer;
 
+    /*
+     * We should use rb_enc_mbmaxlen() of stream.external_encoding for
+     * the max character length but we don't want to do it here for
+     * performance. We can detect the max character length in all
+     * available encodings by the following command line in the Ruby
+     * source directory:
+     *
+     *   grep -Erh "max (enc|byte) length" enc |
+     *     grep -Eo '[0-9]+' |
+     *     sort |
+     *     tail -n1 # => 6
+     */
+#define MAX_ENC_LEN 6
+
 #define LINE_SIZE 4096
-    char line[LINE_SIZE];
 
-    while (memset(line, '\n', LINE_SIZE), source->stream.fgets(line, LINE_SIZE, source->stream.stream) != NULL) {
-        size_t length = LINE_SIZE;
-        while (length > 0 && line[length - 1] == '\n') length--;
+    /*
+     * io.gets(size) may read size or more bytes when the last
+     * character is a multi-byte character. "+ MAX_ENC_LEN" is for the
+     * case.
+     */
+    char line[LINE_SIZE + MAX_ENC_LEN];
 
-        if (length == LINE_SIZE) {
-            /*
-             * If we read a line that is the maximum size and it doesn't end
-             * with a newline, then we'll just append it to the buffer and
-             * continue reading.
-             */
-            length--;
-            pm_buffer_append_string(buffer, line, length);
-            continue;
-        }
+    /* This must not be '\0' to detect the `\0` written by fgets(). */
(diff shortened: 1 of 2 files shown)


ext/prism/extension.c
```diff
@@ -1071,9 +1071,21 @@ parse_stream_eof(void *stream) {
  */
 static char *
 parse_stream_fgets(char *string, int size, void *stream) {
-    RUBY_ASSERT(size > 0);
+    /* We should use rb_enc_mbmaxlen() of stream.external_encoding but
+     * we don't want to do it here for performance. We can detect the
+     * max character length in all available encodings by the
+     * following command line in the Ruby source directory:
+     *
+     *   grep -Eh "max (enc|byte) length" enc/*.c |
```
Earlopain: > ../../../../ext/prism/extension.c:1079:46: warning: ‘/*’ within comment [-Wcomment]

Earlopain (commented):
There is this condition https://github.com/ruby/prism/blob/39fb41f06e5f4ccbe1783122fa8a4470e09dfcfe/src/source.c#L375-L384
which will now basically never be reached.

kou:
Good point.

I've changed the "reduce gets size in fgets callbacks" approach to the "increase the internal buffer size in `pm_source_stream_read()`" approach. I also changed the optimization condition to "whether the last character is `\n` or not" that also works when all of the internal buffer isn't used.

src/source.c
```diff
@@ -365,28 +365,50 @@ bool
...
+     * io.gets(size) may read size or more bytes when the last
+     * character is a multi-byte character. "+ MAX_ENC_LEN" is for the
+     * case.
+     */
+    char line[LINE_SIZE + MAX_ENC_LEN];
+
+    /* This must not be '\0' to detect the `\0` written by fgets(). */
+#define SENTINEL_CHAR 'x'
+
+    while (memset(line, SENTINEL_CHAR, LINE_SIZE + MAX_ENC_LEN), source->stream.fgets(line, LINE_SIZE, source->stream.stream) != NULL) {
```
kou, line 394: We can keep using `\n` for the sentinel character but I've changed to another character for easy to understand. If you don't like it, I'll revert it.

src/source.c
```diff
@@ -365,28 +365,50 @@ bool
...
+        size_t length = LINE_SIZE + MAX_ENC_LEN;
+        while (length > 0 && line[length - 1] != '\0') length--;
```
kou, line 396: We can keep using `== SENTINEL_CHAR` not `!= '\0'` here. I think that `!= '\0'` is easier to understand but I'll revert this if you don't like it.

kou:
@Earlopain @kddnewton Could you take a look at this?

kddnewton:
I think we want to change our fgets to properly not write more than the given size instead of moving the fix into pm_source_stream_read, which I think is too far down the stack. But let me verify/check.

kddnewton:
Closed by #4172

PR #4163, opened 2026-07-10
Add location slice to deconstruct_keys

kddnewton:
It is nice to be able to slice out a location in pattern matching without having to explicitly get the location object and call slice manually.

Also, construct the response for deconstruct_keys lazily in order to avoid reifying location objects if we don't have to.


Earlopain:
:+1: I was just trying to use it like this

PR #4166, opened 2026-07-12
various fixes for RubyParser

zenspider:
See individual commits for details.

One thing this doesn't have is the ACTUAL fixes I want for parser changes, because that would cause failures in the tests against RubyParser.

Is there a plan for how to deal with drift if I'm hoping to EOL RubyParser? Add them to `known_failures`?

Also, it looks like an import hasn't been for some time, and maybe the last one was done against git and not against a release... I'm hesitant to add/change those files as I'm not sure I'm doing the process correctly:

```
$ git clean -qf -- test/prism/fixtures ; rake seattlerb:import
```

### DIFF
--- a/lib/prism/translation/ruby_parser.rb
+++ b/lib/prism/translation/ruby_parser.rb
@@ -1562,7 +1562,7 @@ module Prism
 
         # Create a new compiler with the given options.
         def copy_compiler(in_def: self.in_def, in_pattern: self.in_pattern)
-          Compiler.new(file, in_def: in_def, in_pattern: in_pattern)
+          self.class.new(file, in_def: in_def, in_pattern: in_pattern)
         end
 
         # Create a new Sexp object from the given prism node and arguments.
@@ -1624,7 +1624,6 @@ module Prism
         end
       end
 
-      private_constant :Compiler
       attr_accessor :scopes # :nodoc:
 
       def initialize scopes:nil # :nodoc:
@@ -1677,7 +1676,7 @@ module Prism
         end
 
         result.attach_comments!
-        result.value.accept(Compiler.new(filepath))
+        result.value.accept(self.class::Compiler.new(filepath))
       end
     end
   end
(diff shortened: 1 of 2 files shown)


zenspider:
Added an optional commit to update the tests to current. I can pull it to a different PR to keep this discussion scoped.

Earlopain:
Generally adding failures to excludes in `ruby_parser_test.rb` somewhere is totally fine. It would be nice not to if possible but there are so many edge-cases, and often the result can also be different without actually affecting any consumer. I guess it depends on what drift you mean, that's not clear to me yet.

If I run the import, I also get `pct_q_backslash_nl.txt`, I think you might have forgotten to commit that one?

zenspider:
@Earlopain prolly gets swallowed as I'm on APFS and it is case insensitive by default... but since that commit broke EVERYTHING I should prolly punt it for now

kddnewton:
I'm seeing a couple of changes in here:

1. Explicitly requiring timeout — ✅ makes sense
2. Allowing passing scopes into the translator — ✅ makes sense
3. Fixing the rake task regexp — ✅ makes sense
4. Allowing subclassing the compiler — ?? this one I'm curious about. Not a big deal I suppose, but generally I try to avoid inheritance as much as possible so I'm curious about the use case

zenspider:
4. is simple... it allows me to make changes to prism locally until they get upstreamed... with the compiler marked private and not duplicating the same thing when calling `copy_compiler` that makes it harder.

I _can_ get around it with a prepended module, but that starts to feel clever to me.

Oh, and for some context... In flog/flay/etc I've switched over to `NotRubyParser` which is a subclass of Prism's compat layer. I pop things in there to override until they get upstreamed.

kddnewton (approved):
Okay sounds good

kddnewton:
Actually only thing I'd like to change is to make `scopes` the same type as everywhere else in prism. That is, not automatically wrapping it in a [], since that only allows one scope. But I'll change that after merge.

PR #4167, opened 2026-07-13
Versioning support for `ruby_parser`

Earlopain:
Also add `RubyParser.{latest,for_current_ruby}`

cc @zenspider.

### DIFF
--- a/test/prism/ruby/ruby_parser_test.rb
+++ b/test/prism/ruby/ruby_parser_test.rb
@@ -76,6 +76,7 @@ module Prism
       "whitequark/pattern_matching_single_line_allowed_omission_of_parentheses.txt",
       "whitequark/pattern_matching_single_line.txt",
       "whitequark/ruby_bug_11989.txt",
+      "whitequark/ruby_bug_15789.txt",
       "whitequark/ruby_bug_18878.txt",
       "whitequark/ruby_bug_19281.txt",
       "whitequark/slash_newline_in_heredocs.txt",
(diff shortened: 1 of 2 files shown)


test/prism/ruby/ruby_parser_test.rb
```diff
@@ -77,6 +76,7 @@ class RubyParserTest < TestCase
       "whitequark/pattern_matching_single_line_allowed_omission_of_parentheses.txt",
       "whitequark/pattern_matching_single_line.txt",
       "whitequark/ruby_bug_11989.txt",
+      "whitequark/ruby_bug_15789.txt",
```
Earlopain, line 79: `ruby_parser` when no specific version is given actually tries parsers until it finds one that doesn't report a syntax error, which is why this only needs an exclude now. I guess it parses in an earlier version where numbered parameters weren't a thing yet.

zenspider:
FWIW, I don’t think this needs to be merged. The current API seems sufficient. On Jul 13, 2026, at 04:47, Earlopain ***@***.***> wrote:﻿
@Earlopain commented on this pull request.



In test/prism/ruby/ruby_parser_test.rb:
> @@ -77,6 +76,7 @@ class RubyParserTest < TestCase
       "whitequark/pattern_matching_single_line_allowed_omission_of_parentheses.txt",
       "whitequark/pattern_matching_single_line.txt",
       "whitequark/ruby_bug_11989.txt",
+      "whitequark/ruby_bug_15789.txt",

ruby_parser when no specific version is given actually tries parsers until it finds one that doesn't report a syntax error, which is why this only needs an exclude now. I guess it parses in an earlier version where numbered parameters weren't a thing yet.

kddnewton:
Per what @zenspider has said here, I'm going to close this one.

PR #4169, opened 2026-07-14
Allow custom C flags in ruby-prism-sys

### LINKED ISSUE #4165: Allow `ruby-prism-sys` to use a custom sysroot
`ruby-prism-sys` already supports `wasm32-wasip1` through the WASI SDK. I would also like to target freestanding WebAssembly (`wasm32-unknown-unknown`), where the application provides its own minimal sysroot.

This can currently be configured through `CFLAGS_<target>` and `BINDGEN_EXTRA_CLANG_ARGS_<target>`, but those variables affect every C and bindgen dependency in the same Cargo build. Vendoring `ruby-prism-sys` solely to pass additional compiler arguments also creates unnecessary maintenance work.

Would you accept a crate-scoped `RUBY_PRISM_CFLAGS` option whose arguments are applied consistently to both `cc::Build` and `bindgen::Builder`?

This would allow freestanding builds to provide options such as:

- `--sysroot=/path/to/sysroot`
- additional include paths with `-I`
- Prism configuration macros such as `-DPRISM_HAS_NO_FILESYSTEM=1`

without affecting unrelated C or bindgen dependencies.

I would be happy to submit a PR.

magic-akari:
Add `RUBY_PRISM_CFLAGS` for crate-scoped compiler arguments. Parse the value with shell quoting and apply the resulting arguments to both the vendored C build and bindgen.

Pass existing vendored and WASI arguments to bindgen explicitly through the same argument list instead of mutating `BINDGEN_EXTRA_CLANG_ARGS`. This allows freestanding targets to provide sysroots, include paths, and Prism configuration macros without affecting unrelated Cargo dependencies.

- Closes: #4165.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -222,6 +222,7 @@ version = "1.9.0"
 dependencies = [
  "bindgen",
  "cc",
+ "shlex 1.3.0",
 ]
 
 [[package]]
--- a/rust/ruby-prism-sys/Cargo.toml
+++ b/rust/ruby-prism-sys/Cargo.toml
@@ -24,6 +24,7 @@ include = ["src/", "build/", "Cargo.toml", "Cargo.lock", "README.md", "vendor"]
 [build-dependencies]
 bindgen = "0.72"
 cc = { version = "1.0", optional = true }
+shlex = "1.3"
 
 [features]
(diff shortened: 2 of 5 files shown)


kddnewton (approved):
Makes sense

PR #4172, opened 2026-07-15
Fix up buffer treatment of stream parsing

kddnewton:
A couple of things to fix, notably that we relied on the stream returning strings and relied on it not lying about how much was being written. Now we no longer trust either and we verify instead.

Supersedes #4160

### DIFF
--- a/ext/prism/extension.c
+++ b/ext/prism/extension.c
@@ -1067,13 +1067,40 @@ parse_stream_eof(void *stream) {
 }
 
 /**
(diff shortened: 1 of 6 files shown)


kou:
We need the same change in Ruby because Ruby has a custom `pm_source_stream_fgets_t` for stdin: https://github.com/ruby/ruby/pull/17914

kddnewton:
Yup you're right https://github.com/ruby/ruby/pull/17940

PR #4173, opened 2026-07-16
Add MRuby and Human-ruby  to runtimes list

JoeDupuis:
mruby has now fully transitioned to Prism (https://github.com/mruby/mruby/tree/master/mrbgems/mruby-compiler)

This PR adds the mruby VM to the runtime list, along with the [human-ruby](https://github.com/JoeDupuis/human-ruby) VM, which uses the same bytecode.

The human-ruby VM is a bit of a silly thing, but I thought it would be fun to list.


kddnewton:
I'm gonna do just mruby, thanks for the contribution!

PR #4179, opened 2026-07-21
Ship pregenerated bindings in ruby-prism-sys

### LINKED ISSUE #4175: ruby-prism-sys: cross-compilation fails out of the box; consider pregenerated bindings
`ruby-prism-sys` runs both `cc` and `bindgen` at build time. When host != target, clang gets `--target` but no target sysroot, so it parses host headers and fails.

## Reproduction (macOS arm64 host)

```
$ cargo new xrepro --lib && cd xrepro
$ cargo add ruby-prism@1.9
$ rustup target add x86_64-unknown-linux-gnu
$ cargo build --target x86_64-unknown-linux-gnu
```

```
vendor/prism-1.9.0/include/prism/defines.h:12:10: fatal error: 'ctype.h' file not found
failed to build Prism from source
```

In environments where the C toolchain is configured per target (rake-compiler-dock, building precompiled Ruby gems), the C stage succeeds and the same failure moves to bindgen:

```
/usr/include/features.h:461:12: fatal error: 'sys/cdefs.h' file not found
Unable to generate prism bindings: ClangDiagnostic(...)
```

## Suggestion

#4169 (unreleased) adds `RUBY_PRISM_CFLAGS`, and `CFLAGS_<target>` / `BINDGEN_EXTRA_CLANG_ARGS_<target>` already work — but all of these require users to discover and hardcode per-target sysroot paths.

Would you consider shipping pregenerated bindings (run bindgen at release time, like most -sys crates)? Prism's public API is largely fixed-width types, so the bindings should be target-independent. Cross-compilation would then work out of the box, and the libclang build requirement would go away.

sorafujitani:
Closes #4175.

`ruby-prism-sys` now ships the bindgen output as a committed file (`src/bindings.rs`), so building the crate no longer requires bindgen or libclang. This follows the pattern of libsqlite3-sys and zstd-sys.

## Notes

- The committed file is generated with `layout_tests(false)`. Layout tests hardcode the generating host's struct sizes, which breaks 32-bit targets (wasm32-wasip1 fails with a const-eval overflow). Without them the output is source-portable; I verified byte-identical output on macOS arm64 and Linux x86_64.
- The `buildtime_bindgen` feature restores build-time generation. It is needed when `RUBY_PRISM_CFLAGS` changes the API surface.
- `rake cargo:bindings` regenerates the committed file. CI regenerates and diffs it, runs the layout tests via `buildtime_bindgen` on the 64-bit test matrix, and checks that the committed bindings compile for wasm32-wasip1.
- One correction to the issue thread: I claimed the bindgen output had no layout tests. It does; bindgen 0.72 emits them as `const` assertions rather than `#[test]` functions. That is why the committed copy disables them.
- Scope: this removes the libclang/bindgen requirement. Cross-compiling still needs a C cross toolchain for the `cc` build of the vendored sources.


.github/workflows/rust-bindings.yml
```diff
@@ -47,10 +47,40 @@ jobs:
       - name: Run tests
         run: bundle exec rake cargo:test
 
+      - name: Run tests with buildtime_bindgen
+        run: cargo test --features buildtime_bindgen -- --nocapture
+        working-directory: rust/ruby-prism-sys
+
+      - name: Check pregenerated bindings on wasm32-wasip1
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        run: cargo check --target wasm32-wasip1 --no-default-features
```
kddnewton, line 56: I don't quite know enough about this, but I'm assuming we want to check other bindings besides just this right? Should we check x64/a64/etc?

rust/ruby-prism-sys/src/bindings.rs
Earlopain, line 1: Please mark this as generated in `.gitattributes`

.github/workflows/rust-bindings.yml
sorafujitani: 64-bit targets are already tested: the matrix builds the committed bindings on x64 (ubuntu, windows) and arm64 (macos) and runs the test suite against them. 
this step only exists because wasm32 is the only supported target with a different pointer width.

rust/ruby-prism-sys/src/bindings.rs
sorafujitani: fixed
https://github.com/ruby/prism/pull/4179/changes/d210591b0f23ade4586baed62ca9a413245a3aee

.github/workflows/rust-bindings.yml
kddnewton: Nice, thanks for the clarification

kddnewton (approved):

Issue #4123, opened 2026-06-03
JRuby bump causes CI failures


Earlopain:
Upgrading from 10.0.5.0 to 10.1.0.0 causes the following error in the wasm bindings:
(https://github.com/ruby/prism/actions/runs/26855989543/job/79198822453?pr=4121)

```
java.lang.NoSuchMethodError: 'io.roastedroot.redline.UniversalInstance$Builder io.roastedroot.redline.UniversalInstance.builder(com.dylibso.chicory.wasm.WasmModule)'
	at org.ruby_lang.prism.wasm.PrismParser.builder(PrismParser.java:92)
	at org.ruby_lang.prism.wasm.Prism.<init>(Prism.java:24)
	at org.ruby_lang.prism.wasm.Prism.<init>(Prism.java:19)
	at org.jruby.parser.prism.JRubyTest$TestingPrism.<init>(JRubyTest.java:89)
	at org.jruby.parser.prism.JRubyTest.jrubyReproducer(JRubyTest.java:99)
```

```
Error:  Errors: 
Error:    JRubyTest.jrubyReproducer:99 » NoSuchMethod 'io.roastedroot.redline.UniversalInstance$Builder io.roastedroot.redline.UniversalInstance.builder(com.dylibso.chicory.wasm.WasmModule)'
Error:    WASMTest.test1:35 » NoSuchMethod 'io.roastedroot.redline.UniversalInstance$Builder io.roastedroot.redline.UniversalInstance.builder(com.dylibso.chicory.wasm.WasmModule)'
Error:    WASMTest.test2:51 » NoSuchMethod 'io.roastedroot.redline.UniversalInstance$Builder io.roastedroot.redline.UniversalInstance.builder(com.dylibso.chicory.wasm.WasmModule)'
Error:    WASMTest.testMBCIdentifier:67 » NoSuchMethod 'io.roastedroot.redline.UniversalInstance$Builder io.roastedroot.redline.UniversalInstance.builder(com.dylibso.chicory.wasm.WasmModule)'
Error:    WASMTest.testVersion:79 » NoSuchMethod 'io.roastedroot.redline.UniversalInstance$Builder io.roastedroot.redline.UniversalInstance.builder(com.dylibso.chicory.wasm.WasmModule)'
[INFO] 
Error:  Tests run: 5, Failures: 0, Errors: 5, Skipped: 0
```

[...]

PR #4183, opened 2026-07-27
Bump jruby

Earlopain:
`jruby-complete` is being replaced by `jruby` https://github.com/jruby/jruby/pull/9512

For https://github.com/ruby/prism/issues/4123


headius:
This is a good change for consumers like Prism that just want a "complete JRuby" in classpath without depending on the "uberjar" jruby-complete artifact. Thanks @Earlopain!

cc @chadlwilson who pointed out that we could deprecate most jruby-complete uses by depending on the "jruby" artifact.

kddnewton (approved):

PR #4185, opened 2026-07-28
Better compatibility for hash string-like symbol keys in parser translator

Earlopain:
There's some special handling in regards to empty symbols and those containing newlines that should apply here. Makes it produce the correct amount of sym/dsym nodes


kddnewton (approved):

Earlopain:
Quickdraw

PR #4197, opened 2026-08-04
Abort on allocation failure in static literals

kddnewton:


Earlopain:
Any particular reason to only do this here? There are a few other places that just skip work or return NULL and will spectacularly fail if that actually happens.

kddnewton:
There was a reason, and I'm forgetting it now. I think there way technically a leak if some weird circumstance happened.

I'm conflicted on this tbh. On the one hand now that xmalloc is hooked up for most of our consumers, we don't really have to worry about it (cruby, mruby, etc.). Those will do GC on allocation failure, and otherwise abort themselves.

On the otherhand, it would be nice to have a full project story for allocation failures that flows through the whole callstack. Because right now if you use prism and you don't use the xallocator interface, there's no way to handle running out of memory.

So I'm not sure, is the answer.

Earlopain:
Oh well. And yeah, for ruby itself it's not a problem anymore.

PR #4198, opened 2026-08-04
Make `ruby_memcheck` optional

Earlopain:
It consistently fails CI when nokogiri tries to install from source. But those steps don't actually need it and the nokogiri version with this fix is still quite a long while out.

Also bumps bundler to remove a workaround.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -259,7 +259,8 @@ jobs:
   memcheck:
     runs-on: ubuntu-24.04
     env:
-      BUNDLER_VERSION: '0' # https://github.com/ruby/ruby/pull/16909
+      BUNDLE_WITH: 'memcheck'
+      BUNDLER_VERSION: '0'
     steps:
     - uses: actions/checkout@v7
     - name: Install valgrind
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -259,7 +259,8 @@ jobs:
   memcheck:
     runs-on: ubuntu-24.04
     env:
-      BUNDLER_VERSION: '0' # https://github.com/ruby/ruby/pull/16909
+      BUNDLE_WITH: 'memcheck'
+      BUNDLER_VERSION: '0'
```
Earlopain, line 263: Seems that newer bundler need this regardless when running with ruby-head, so I just removed the comment.

kddnewton:
So if I understand correctly this makes the memcheck task attempt to install with the memcheck group, and the regular gemfile no longer tries to install memcheck right? So the memcheck task itself is still required?

Earlopain:
It's still required, yes. `optional: true` tells bundler to not install it by default and then the CI step enables it back with `BUNDLE_WITH=memcheck` so it actually does something. If it were not installed in CI it fails with the dummy task on `LoadError`.

kddnewton (approved):

PR #4199, opened 2026-08-05
Split some tokens

kddnewton:
Some tokens can represent multiple things depending on context, like `(`, `{`, and `\``. These tokens are represented in parse.y as different tokens, and therefore have different grammar rules applied to them. This PR mirrors that logic in order to make comparing against the upstream grammar simpler.

It has the added benefit of making it so that we no longer have to rely on any internal state in our translation engines (except for in lex compat).

@Earlopain do you think we could drop the lex state from the public API after this?

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -584,6 +584,8 @@ tokens:
     comment: "a numbered reference to a capture group in the previous regular expression match"
   - name: PARENTHESIS_LEFT
     comment: "("
+  - name: PARENTHESIS_LEFT_GROUPING
+    comment: "( scanned at the beginning of an expression"
   - name: PARENTHESIS_LEFT_PARENTHESES
     comment: "( for a parentheses node"
   - name: PERCENT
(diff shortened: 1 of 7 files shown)


Earlopain:
I would welcome to remove the state from the lex result. But doing so in a compatible way might be a bit bothersome. When I rewrote `syntax_suggest` to not use ripper anymore in https://github.com/ruby/syntax_suggest/pull/251 I made an effort to not use it and that mostly worked out ok. For things that were too bothersome (like detecting endless methods just from tokens) I took the AST to help. But I'd want to check what else currently uses the state and why (if any)

BTW, this broke some irb colorizer tests. That's fine and easy to fix though ~~but it now also accepts `def f(a,) end` when it previously didn't. Can you take a look?~~

kddnewton:
I see you crossed out ```def f(a,) end when it previously didn't. Can you take a look?```. Was that incorrect?

Earlopain:
That's the new trailing comma syntax. I tested against older irb locally and forgot that this changed. Oops

PR #4202, opened 2026-08-06
Remove rubyvm: keyword argument from Prism.find

eregon:
Remove the `rubyvm:` keyword argument from `Prism.find`. The internal dispatch already checks `defined?(::RubyVM)`, so exposing this as a public parameter seems unnecessary. Removing it makes the API cleaner and makes it straightforward to test each `Find` implementation directly using the concrete classes.

~Additionally, unified `LineMethodFind`, `LineLambdaFind`, and `LineProcFind` into a single `LineCallableFind` class, since they share the same structure (parse file, match by start line).~

### DIFF
--- a/lib/prism/node_find.rb
+++ b/lib/prism/node_find.rb
@@ -14,11 +14,11 @@ module Prism
   module NodeFind # :nodoc:
     # Find the node for the given callable or backtrace location.
     #--
-    #: (Method | UnboundMethod | Proc | Thread::Backtrace::Location callable, bool rubyvm) -> Node?
-    def self.find(callable, rubyvm)
+    #: (Method | UnboundMethod | Proc | Thread::Backtrace::Location callable) -> Node?
+    def self.find(callable)
       case callable
       when Proc
-        if rubyvm
+        if defined?(::RubyVM)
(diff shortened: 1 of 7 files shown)


eregon:
(CI seems to not run because of a [GitHub Actions outage](https://www.githubstatus.com/incidents/qcvjkzcs7j74))

lib/prism/node_find.rb
```diff
@@ -14,25 +14,17 @@ module Prism
...
-        if rubyvm
+      when Proc, Method, UnboundMethod
+        if defined?(::RubyVM)
           RubyVMCallableFind.new.find(callable)
-        elsif callable.lambda?
-          LineLambdaFind.new.find(callable)
         else
-          LineProcFind.new.find(callable)
-        end
-      when Method, UnboundMethod
```
kddnewton, line 28: I would prefer these to stay as separate classes, as the point is that you only have to switch on the callable in one place.

kddnewton (dismissed):

lib/prism/node_find.rb
eregon: Done. It also makes the diff a lot smaller.

Earlopain (approved):

PR #4203, opened 2026-08-07
How about a new release?  Bump version to 1.10.0 

rwstauner:
Are we in a good state where we could have a release?
The last one was January.

I'm particularly interested in making some of the parser bug fixes available to rubocop, for example, so I think if we could just use a newer prism gem we'd get new validations for free.
This would help reduce incompatibilities in code that needs to run on both ruby 4 and ruby head.

I put together a PR in case that made anything easier, but happy to change the version number or cangelog or anything else as you see fit.

What do you think, can we release a new version, soon?


kddnewton:
Very soon, but it's going to be 2.0.0, and it will need a beta first because it has a fair number of breaking changes. It won't be this week; but it will be soon I promise!

rwstauner:
Ok great.  Thanks, Kevin!

PR #4206, opened 2026-08-17
Return a plain Token for `Prism.lex` etc.

Earlopain:
Instead of a 2-element array. Nobody should care about the state, and nobody seems to. A code search on Github shows that it only gets discarded (like via `value.map(&:first)`, or similar things.

But because our own ripper translator still needs it for compatibility, we have to keep it around in some way. I played around with something like `Prism.lex_internal` that keeps the same return value but that duplicates much logic like serialization/deserialization.

This achieves the same thing and is much smaller in scope.

### DIFF
--- a/ext/prism/extension.c
+++ b/ext/prism/extension.c
@@ -846,14 +846,7 @@ parse_lex_token(pm_parser_t *parser, pm_token_t *token, void *data) {
     parse_lex_data_t *parse_lex_data = (parse_lex_data_t *) data;
 
     VALUE value = pm_token_new(parser, token, parse_lex_data->encoding, parse_lex_data->source, parse_lex_data->freeze);
-    VALUE yields = rb_assoc_new(value, INT2FIX(pm_parser_lex_state(parser)));
-
-    if (parse_lex_data->freeze) {
-        rb_obj_freeze(value);
-        rb_obj_freeze(yields);
-    }
-
-    rb_ary_push(parse_lex_data->tokens, yields);
+    rb_ary_push(parse_lex_data->tokens, value);
 }
 
 /**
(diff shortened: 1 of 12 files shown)


ext/prism/extension.c
```diff
@@ -846,14 +846,7 @@ parse_lex_token(pm_parser_t *parser, pm_token_t *token, void *data) {
     parse_lex_data_t *parse_lex_data = (parse_lex_data_t *) data;
 
     VALUE value = pm_token_new(parser, token, parse_lex_data->encoding, parse_lex_data->source, parse_lex_data->freeze);
```
Earlopain, line 848: This already freezes the returned token if applicable, not necessary here again

lib/prism/parse_result.rb
```diff
@@ -1175,6 +1176,20 @@ def deep_freeze
...
       freeze
     end
+
+    # Backwards compatibility for Prism.lex/Prism.lex_file/Prism.parse_lex
+    # when they returned a 2-element array.
+
+    #: (Integer index) -> Token
+    def [](index) # :nodoc:
+      return self if index == 0
+      raise ArgumentError, "Invalid index #{index}"
```
kddnewton, line 1193: You may as well put in a check for index == 1 and return the state

lib/prism/lex_compat.rb
```diff
@@ -621,13 +621,13 @@ def result
...
 
-      result_value.each_with_index do |(prism_token, prism_state), index|
+      result_value.each_with_index do |prism_token, index|
         lineno = prism_token.location.start_line
         column = prism_token.location.start_column
 
         event = RIPPER.fetch(prism_token.type)
         value = prism_token.value
-        lex_state = Translation::Ripper::Lexer::State[prism_state]
+        lex_state = Translation::Ripper::Lexer::State[prism_token.instance_variable_get(:@state)]
```
kddnewton: If we're truly using it, I don't want to use ivar_get. Can you instead add a `_state` method to the token and explicitly say "this is private, don't use"?

kddnewton: Or maybe call it `ripper_state` with the same thing?

kddnewton (changes requested):

lib/prism/parse_result.rb
Earlopain: Eventually I would want to try and remove these but for now that seems fine

lib/prism/lex_compat.rb
Earlopain: `_ripper_state` it is

kddnewton (approved):

PR #4207, opened 2026-08-18
Move typecheck to main gemfile, bump versions

Earlopain:
The typecheck rake tasks has a problem where it can't find dependencies for the rbi task because it doesn't shell out.

This removes the gemfile workaround and simply makes them available from the main gemfile


Earlopain:
This is still not working out. sorbet only being precompiled and having no generic ruby platform seems to make this impossible. I'd really like to do this but for now I'm going to see about doing a different workaround for the issue I was running into with gems not being loaded properly with `BUNDLE_GEMFILE`

PR #4208, opened 2026-08-19
Reverse sync

Earlopain:


Earlopain:
I just love copilot being shoved down my throat everywhere

Earlopain:
rdoc now depends on prism so we are in the unique situation that we actually need to compile ourselves for it to properly work.

PR #4209, opened 2026-08-19
Correctly apply rdoc prerequisite

Earlopain:
I thought this didn't actually work but it seems it is necessary this way. Unclear why `enhance` worked for the docs workflow but not for pages


Earlopain:
I wish I could block this waste of resources

PR #4212, opened 2026-08-26
Remove now unneeded rdoc private annotations

Earlopain:
RDoc 8 parses rbs (including inline) and so
automatically removes the annotations for us

### DIFF
--- a/rakelib/typecheck.rake
+++ b/rakelib/typecheck.rake
@@ -30,7 +30,8 @@ namespace :typecheck do
             # break when RDoc private starts
             break
           when /\A:/
-            # skip RBS type annotations
+            # skip RBS type annotations and remove preceeding empty line
+            comments.pop if comments.last&.text&.empty?
           else
             comments << RBI::Comment.new(line)
           end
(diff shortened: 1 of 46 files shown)


rakelib/typecheck.rake
```diff
@@ -30,7 +30,8 @@ namespace :typecheck do
             # break when RDoc private starts
             break
           when /\A:/
-            # skip RBS type annotations
+            # skip RBS type annotations and remove preceeding empty line
+            comments.pop if comments.last&.text&.empty?
```
Earlopain, line 34: Hid an actual change here. It's just to keep rbi files as they are currently. Previously the line was removed because it was the private annotation (see above)

kddnewton:
I get that it works now, but do we need to do this? It's a pretty large change. Is it better looking in the docs now if we remove the --?

Earlopain:
It's only so large because of the generated rbi (don't know why rbs inline doesn't exclude its own type hints :shrug:). It looks exactly the same in the docs.

I don't want these in new code but if there're everywhere its just gonna get copy-pasted

kddnewton (approved):

kddnewton:
alrighty

PR #4213, opened 2026-08-26
Some playground improvements

Earlopain:
* Add color to empty locations, numbers
* Format constant type like ruby symbol
* Fix display of string type as `[Object Object]`
* Include slice for location fields
* Add `location: ` to node locations like in ruby inspect output

Before:
image
After:
image

### DIFF
--- a/doc/playground.js
+++ b/doc/playground.js
@@ -6,6 +6,9 @@ const editorDiv = document.getElementById("editor");
 const loading = document.getElementById("loading");
 const toast = document.getElementById("toast");
 
+const encoder = new TextEncoder();
+const decoder = new TextDecoder();
+
 // Load Prism WASM and Monaco, show error if either fails
 let instance, monaco;
 try {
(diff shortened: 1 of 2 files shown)


doc/playground.js
```diff
@@ -252,11 +252,18 @@ function offsetToLineCol(source, offset) {
...
+function formatLoc(source, loc, includeSlice) {
   if (!loc || loc.startOffset === undefined) return null;
   const start = offsetToLineCol(source, loc.startOffset);
   const end = offsetToLineCol(source, loc.startOffset + loc.length);
-  return { start, end, text: `${start.line}:${start.col}-${end.line}:${end.col}` };
+
+  let text = `${start.line}:${start.col}-${end.line}:${end.col}`;
+
+  if (includeSlice) {
+    const slice = source.slice(loc.startOffset, loc.startOffset + loc.length);
```
Earlopain: So, this is not correct. slice does by character but the offsets are in bytes and I don't know how you'd handle this in javascript. It's already the same problem with `offsetToLineCol` but here it would be a bit more noticable I think

Image

Earlopain: Should be ok with https://github.com/ruby/prism/pull/4213/commits/ffeef1e102498c35978970199ca5830d97ff46e5, I think

PR #4220, opened 2026-09-04
Make `SymbolNode#value_loc` non-optional

Earlopain:
Sort of a followup for https://github.com/ruby/prism/commit/929aec16504f799dae358757784741ec8b2b1c6a / https://github.com/ruby/prism/pull/3772

It changed the value for `:''` into nil/null, which is a bit inconvenient. For syntax-valid code I expect it to always be present.

It's also inconsistent with `StringNode`. In effect, this reverts the snapshots changes and the two changes in ruby for the ripper/parser compiler.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -4553,7 +4553,7 @@ nodes:
       - name: opening_loc
         type: location?
       - name: value_loc
-        type: location?
+        type: location
       - name: closing_loc
         type: location?
       - name: unescaped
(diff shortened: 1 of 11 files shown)


Earlopain:
We can't have locations that can legitimatly be zero-length be optional anymore, now that optional locations use zero length to indicate absence. I quickly checked the other optional locations and didn't find anything else with this ambiguity

PR #4221, opened 2026-09-05
Avoid quadratic value-expression check on and/or chains

makenowjust:
Ref https://bugs.ruby-lang.org/issues/22294

`pm_check_value_expression()` descends the left operand of every and/or node.
Because `pm_and_node_create` and `pm_or_node_create` already assert the value of their left operand when the node is built, walking a left-associative chain such as `a && a && ... && a` re-checks the whole left spine once per operator, which is quadratic in the length of the chain.

The left operand of an existing and/or node was therefore already checked, so stop at the node instead of descending.
This makes the check linear.
The only observable change is that a void value on the left branch of a chain is now reported once, at the innermost node where it is created, rather than once per enclosing operator; the errors fixture that pinned the duplicate is updated to match.

### DIFF
--- a/test/prism/errors/destroy_call_operator_write_arguments.txt
+++ b/test/prism/errors/destroy_call_operator_write_arguments.txt
@@ -1,7 +1,6 @@
 t next&&do end&=
         ^~ unexpected 'do'; expected an expression after the operator
   ^~~~ unexpected void value expression
-  ^~~~ unexpected void value expression
               ^~ unexpected '&=', expecting end-of-input
               ^~ unexpected '&=', ignoring it
   ^~~~ Invalid next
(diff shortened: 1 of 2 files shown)


test/prism/errors/destroy_call_operator_write_arguments.txt
```diff
@@ -1,6 +1,5 @@
 t next&&do end&=
         ^~ unexpected 'do'; expected an expression after the operator
-  ^~~~ unexpected void value expression
```
makenowjust, line 3: `parse.y` still reports this duplicated warning, but it is redundant; I'll open a pull request to fix it for `parse.y` too.

makenowjust: See ruby/ruby#18639.

Earlopain (approved):
Nice, thanks.

BTW, you can do `Prism.profile` to get a more accurate number for just the parse step.

PR #4223, opened 2026-09-07
 Remove all remaining skips in newline_test.rb

### LINKED ISSUE #4217: Remove `NewlineTest`?
It often breaks just because some source file changed. Latest is https://github.com/ruby/ruby/commit/f4813a34c274a764b99d6e706dd016514b66cf45 which needs to be reverse synced. It looks like in the last two years the list of excludes has just grown and nobody seems to care all that much. https://github.com/ruby/prism/commits/main/test/prism/newline_test.rb

It compares newline flags in the prism ast against those from iseq. But either something is not correct or it's not an exact science (currently the rubyvm result is already postprocessed in some way).

Everything is already mostly recorded in the ast snapshot + the visitor does a small bit of work as well based on that ast. So it only really compares against RubyVM with parse.y. I would like to stop doing that. Thoughts?

eregon:
* No skips left in `newline_test.rb`
* No workarounds left in `newline_test.rb`, the logic to mark newlines matches exactly the same lines as CRuby 3.4+ (so 3.4, 4.0, master prism, master parse.y) for the files included in `newline_test.rb`
* The corresponding Java logic is updated too
* The corresponding Java logic is now tested to mark exactly the same lines

Fixes https://github.com/ruby/prism/issues/4217

### DIFF
--- a/lib/prism/parse_result/newlines.rb
+++ b/lib/prism/parse_result/newlines.rb
@@ -65,6 +65,81 @@ module Prism
         end
       end
 
+      # Permit def nodes to mark newlines within themselves. The body of an
+      # endless method definition never emits newline events, so in that case
+      # mark every line as already seen while visiting it instead. Nested
+      # scopes (blocks, lambdas, etc.) reset the lines and emit events again.
+      #
+      #: (DefNode node) -> void
+      def visit_def_node(node)
+        old_lines = @lines
+        @lines = Array.new(old_lines.size, !node.equal_loc.nil?)
+
+        begin
+          super(node)
+        ensure
+          @lines = old_lines
+        end
+      end
+
+      # Permit class nodes to mark newlines within themselves.
+      #
+      #: (ClassNode node) -> void
+      def visit_class_node(node)
+        old_lines = @lines
+        @lines = Array.new(old_lines.size, false)
+
+        begin
+          super(node)
+        ensure
+          @lines = old_lines
+        end
+      end
+
+      # Permit module nodes to mark newlines within themselves.
+      #
+      #: (ModuleNode node) -> void
+      def visit_module_node(node)
+        old_lines = @lines
+        @lines = Array.new(old_lines.size, false)
+
+        begin
+          super(node)
+        ensure
+          @lines = old_lines
+        end
+      end
+
+      # Permit singleton class nodes to mark newlines within themselves.
+      #
+      #: (SingletonClassNode node) -> void
+      def visit_singleton_class_node(node)
+        old_lines = @lines
+        @lines = Array.new(old_lines.size, false)
+
+        begin
+          super(node)
+        ensure
+          @lines = old_lines
+        end
+      end
+
+      # Statements inside string interpolation do not emit newline events, so
+      # mark every line as already seen while visiting them. Nested scopes
+      # (blocks, lambdas, defs, etc.) reset the lines and emit events again.
+      #
+      #: (EmbeddedStatementsNode node) -> void
+      def visit_embedded_statements_node(node)
+        old_lines = @lines
+        @lines = Array.new(old_lines.size, true)
+
+        begin
+          super(node)
+        ensure
+          @lines = old_lines
+        end
+      end
+
       # Mark if nodes as newlines.
       #
       #: (IfNode node) -> void
--- a/lib/prism/parse_result/newlines.rb
+++ b/lib/prism/parse_result/newlines.rb
@@ -140,6 +140,42 @@ module Prism
         end
       end
 
+      # The predicate of a while loop is compiled at the end of the loop,
+      # after the body, so any statements it contains (from parentheses)
+      # emit their line events again even if the lines were already seen.
+      #
+      #: (WhileNode node) -> void
+      def visit_while_node(node)
+        old_lines = @lines
+        @lines = Array.new(old_lines.size, false)
+
+        begin
+          visit(node.predicate)
+        ensure
+          @lines = old_lines
+        end
+
+        visit(node.statements)
+      end
+
+      # The predicate of an until loop is compiled at the end of the loop,
+      # after the body, so any statements it contains (from parentheses)
+      # emit their line events again even if the lines were already seen.
+      #
+      #: (UntilNode node) -> void
+      def visit_until_node(node)
+        old_lines = @lines
+        @lines = Array.new(old_lines.size, false)
+
+        begin
+          visit(node.predicate)
+        ensure
+          @lines = old_lines
+        end
+
+        visit(node.statements)
+      end
+
       # Mark if nodes as newlines.
       #
       #: (IfNode node) -> void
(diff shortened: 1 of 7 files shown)


eregon:
Actually it's possible to match lines from RubyVM exactly for all the files being tested, done with the 2 `Match RubyVM newlines` commits.
I'm not entirely sure it's a good idea, still making my mind about it and welcoming other opinions on it.

eregon:
https://github.com/ruby/prism/actions/runs/34165671547/job/101876103431?pr=4223 fails, yet `bundle exec ruby -Ilib test/prism/newline_test.rb` passes locally on ruby-dev, very weird.
Also the failure is because line 95 of `ruby/parameters_signature_test.rb` is expected twice, but this makes no sense, line 95 of that file is: `object = Object.new`.
https://github.com/ruby/prism/actions/runs/34165671547/job/101876103431?pr=4223#step:8:869

eregon:
Claude found why it differs:
> `require "test/unit"` (the `test-unit` gem, 3.7.8) globally sets `RubyVM::InstructionSequence.compile_option specialized_instruction: false`. `opt_new` is a specialized instruction, so with it disabled the `opt_new` codepath is never generated, the duplicate `:line` trace point vanishes, and the test passes.

Wow, test-unit disabling `specialized_instruction`s sounds pretty bad.

eregon:
-> https://bugs.ruby-lang.org/issues/22299

test/prism/newline_test.rb
```diff
@@ -2,33 +2,15 @@
...
-      result/warnings_test.rb
-      ruby/find_fixtures.rb
-      ruby/find_test.rb
-      ruby/parser_test.rb
-      ruby/ripper_test.rb
-      ruby/ruby_parser_test.rb
-      ruby/parameters_signature_test.rb
-    ]
-
+    # Do not add skips here, every change is a regression in TracePoint :line events
```
Earlopain: Since it seems that no one actually knows how line events should behave (as they are not speced), how can you say that how they currently are is correct?

I am ok with the change overall, I guess you will update java to match, but this just replicates cruby behaviour and we shouldn't make such a statement. Not having a list of skips to easily add to should be enough

lib/prism/parse_result/newlines.rb
```diff
@@ -162,6 +285,99 @@ def newline_flag!(lines) # :nodoc:
...
+      first = elements.first
+      if first && !static_literal?
+        first.newline_flag!(lines)
+      else
+        super
+      end
+    end
+  end
+
+
```
Earlopain:
```suggestion
```

Earlopain (commented):
Overall I'm ok with this change, nice cleanup to the skips, but I'm thinking this might overfit for the few special cases the test code actually contains. Would be fine though in my eyes, we can see how it actually goes.

test/prism/newline_test.rb
eregon: I am not saying the CRuby `:line` events are correct (they are what they are and for most of them have been the same for a long time), but any change to them is an incompatibility and that's what I want to bring attention to.
So adding a skip here is not "skipping some Prism test" it's "incompatible change about TracePoint :line events, independent of Prism".
That's really for skips added from ruby/ruby though.
If we need a skip from changing ruby/prism then it's probably that `lib/prism/parse_result/newlines.rb` needs to be improved to match CRuby.
Suggested improvement to the wording welcome :)

If you think that's still too much, I can drop the comment/commit for it though.

Earlopain:
> Wow, test-unit disabling specialized_instructions sounds pretty bad.

Actually, it's `power_assert` (which `test-unit` unconditonally requires)

test/prism/newline_test.rb
eregon: > as they are not speced

Yeah, this is a bit of an unusual place to test this. I guess we could try to move this to ruby/spec, and then that would test (with quite a few more indirections) whether the Prism Java newlines marking matches the specs. It wouldn't test the Prism Ruby newlines marking, so I think we still need this test.
It's probably quite messy to write ruby/specs for this because we'd need to use a :line TracePoint and gather the lines at runtime from the actual execution.

Earlopain: How about basically what you just wrote? A bit more context doesn't hurt.
> If you are coming from ruby/ruby, a test failure here means that TracePoint `:line` events changed.
> Before adding a skip, make sure that you actually intended for such a difference to happen.

> Yeah, this is a bit of an unusual place to test this

I mean, it's fine the way it currently is. There's also tests for ripper and rubyvm (parsey specifically) for which the same is true

eregon: Updated to that

eregon:
Right, maybe we should switch to https://github.com/ruby/test-unit-ruby-core in Prism so it's the same test harness used in CRuby CI and Prism CI?

Earlopain:
I'm not so familiar with it, but my understanding is that it just adds some extra assertions. I tested it and in the end `power_assert` still changes compile options anyways.

eregon:
> I'm not so familiar with it, but my understanding is that it just adds some extra assertions. I tested it and in the end `power_assert` still changes compile options anyways.

Ah interesting I thought it doesn't use the `test-unit` gem (but some kind of shim using minitest under the good), that probably changed at some point.

So CRuby `test-all` seems to use [yet another test harness](https://github.com/ruby/ruby/blob/master/tool/lib/test/unit.rb) then (which doesn't change compile options).

eregon:
I'm looking at the [last failure](https://github.com/ruby/prism/actions/runs/34361644992/job/102499847584?pr=4223), it's a divergence between the parse.y and prism compilers since 3.4, in 3.3 they agreed.
=> It's https://bugs.ruby-lang.org/issues/22302

eregon:
It's https://bugs.ruby-lang.org/issues/22302

eregon:
This is now ready for review, summary (I also updated the PR description to the same):
* No skips left in `newline_test.rb`
* No workarounds left in `newline_test.rb`, the logic to mark newlines matches exactly the same lines as CRuby 3.4+ (so 3.4, 4.0, master prism, master parse.y) for the files included in `newline_test.rb`
* The corresponding Java logic is updated too
* The corresponding Java logic is now tested to mark exactly the same lines

eregon:
For the 2 changes to logic to match CRuby:
* `Match RubyVM newlines for parenthesized while/until predicates` 2 :line events for loops is a little bit weird, but somewhat consistent with things like methods/blocks creating a new "scope" and having a new set of lines to mark. It's also a fairly simple change.
* `Match RubyVM newlines exactly, removing all compensations in newline_test.rb` marking the value instead of the assignment makes sense for a bytecode interpreter, and is easy, we already did some of that. The part about static literals is more subtle but also makes sense for bytecode interpreters, as the elements will be executed first before building the array (unless it's a static literal and then a single :line event where the array starts).

So while I think it's debatable whether that's the best location for :line events, matching CRuby is best for compatibility and it turns out to not be too hard either.

kddnewton (approved):
Looks great, nice work @eregon. Wish this weren't so difficult to replicate, but what can you do 🤷

PR #4230, opened 2026-09-16
Set code examples in `config.yml` to be ruby explicitly

Earlopain:
Because of the `^^^` hints, it's not valid ruby and so rdoc gives it no syntax highlighting.
We can convince it otherwise and syntax-highlight syntax-invalid code.

Although it is syntax-invalid, the results look very good. I thought a bit about how to make it syntax-valid code instead so the explicit codeblocks are not necessary but didn't come up a solution.

Before:
grafik

After:
grafik

I had copilot push out the following python script to modify the config. I didn't validate it and looked at the results only, which appear correct.

<details><summary>slop.py</summary>
<p>

```py
#!/usr/bin/env python3
from pathlib import Path
import re
import sys

path = Path(sys.argv[1] if len(sys.argv) > 1 else "config.yml")
lines = path.read_text().splitlines(keepends=True)

comment_start = re.compile(r"^(\s*)comment:\s*\|\s*$")

def indent_width(line: str) -> int:
    return len(line) - len(line.lstrip(" "))

def is_blank(line: str) -> bool:
    return not line.strip()

def transform_comment_block(block: list[str], content_indent: int) -> list[str]:
    """Wrap source examples indented four spaces beyond comment prose in ```rb fences."""
    output: list[str] = []
    index = 0

    while index < len(block):
        line = block[index]

        # Source examples are conventionally indented four spaces beyond prose.
        if not is_blank(line) and indent_width(line) >= content_indent + 4:
            example: list[str] = []

            while index < len(block):
                current = block[index]

                # Keep blank lines inside an example when another indented source
                # line follows later in the same contiguous region.
                if is_blank(current):
                    lookahead = index + 1
                    while lookahead < len(block) and is_blank(block[lookahead]):
                        lookahead += 1

                    if (
                        lookahead < len(block)
                        and indent_width(block[lookahead]) >= content_indent + 4
                    ):
                        example.append(current)
                        index += 1
                        continue

                    break

                if indent_width(current) < content_indent + 4:
                    break

                example.append(current)
                index += 1

            prefix = " " * content_indent
            newline = "\n" if example and example[0].endswith("\n") else ""

            output.append(f"{prefix}```rb{newline}")
            for source_line in example:
                if is_blank(source_line):
                    output.append(source_line)
                else:
                    output.append(source_line[4:])
            output.append(f"{prefix}```{newline}")
            continue

        output.append(line)
        index += 1

    return output

result: list[str] = []
index = 0

while index < len(lines):
    match = comment_start.match(lines[index].rstrip("\n"))

    if not match:
        result.append(lines[index])
        index += 1
        continue

    result.append(lines[index])

    key_indent = len(match.group(1))
    content_indent = key_indent + 2
    index += 1

    block_start = index
    while index < len(lines):
        line = lines[index]

        # Blank lines are part of the literal block.
        if is_blank(line):
            index += 1
            continue

        # A line at the comment key's indentation or less starts the next YAML item.
        if indent_width(line) <= key_indent:
            break

        index += 1

    result.extend(transform_comment_block(lines[block_start:index], content_indent))

path.write_text("".join(result))
```

</p>
</details>


kddnewton:
These examples get templated into a bunch of other places. Could we instead template it into the Ruby files where this goes, but not do it in the general config.yml?

Earlopain:
Ah, right. Actually, rust and doxygen accept these style of code via markdown. Only javadoc doesn't (though it does in some later java version). So I changed templating to do nothing in rust anymore instead of trying to insert these programatically and skip codefences for java so stuff is as before.

How about it?

Earlopain:
I want to link to the playground from the code examples. It's convenient when the examples are behind something that can rather easily be parsed out https://github.com/Earlopain/prism/commit/41254598d9a89581ffcbd42ea64f3db663316e2e

grafik

I didn't do it this way because of this but I wanted to somehow integrate the playground into the docs for a while, and this came to mind after I made this PR

PR #4231, opened 2026-09-19
Stop marking PM_SOURCE_FILE_NODE as static

byroot:
Ref: https://bugs.ruby-lang.org/issues/17593.
Ref: https://github.com/ruby/ruby/pull/18924

Because `__FILE__` is currently considered a static literal, it's entirely evaluated during compilation.

The problem is that ISeqs can be serialized and reloaded later, but since `__FILE__` is entirely computed during compilation, a serialized ISeq is coupled with the location of the source code.

This prevent compiling the the iseq with the source in one path, and loading it later with the source in another path, which prevent bootsnap precompilation from working with some builds systems.

I'm working on changing the compiler so that `__FILE__` is no longer a static value, but a hidden methodthat returns the current ISeq path.


byroot:
NB: this is mostly a draft. Perhaps a better solution would be to turn `__FILE__` into a regular method call, just like `__dir__` (at least it would be way simpler for the compiler).

headius:
+1 we had to tweak the JRuby parse.y equivalent years ago to introduce a special AST node for __FILE__ so we could cache either AST or IR and provide a new file path when loading those caches.

I'm less in favor of making it a method, since that has a bunch of other side effects, but I'll comment on the related PR or issue.

cc @enebo

kddnewton:
Sorry I'm not entirely clear from the upstream ticket, is this good to go or is further discussion needed?

byroot:
Sorry, @kddnewton. I don't know if this is good to go, I still need to figure out how to fully achieve this, my PR on the compiler works on the happy path, but crash in other cases like pattern matching. So I'm not sure if Prism will need more changes or not yet.

But the crux of the idea is that `__FILE__` would no longer be a "static literal", as it's value could change between compilation and execution.

kddnewton:
It's no problem! I'm just going to mark this as draft for my own bookkeeping until it's good to go.

byroot:
I pushed another change, and I got a green build on https://github.com/ruby/ruby/pull/18924 now, so I think this is it.

kddnewton (changes requested):
So this is definitely correct, but we will need to gate this on version, since this is going to be change for 4.1

byroot:
I see. Happy to tackle this myself if you give me a few directions / some example.

byroot:
I see:

```c
if (parser->version < PM_OPTIONS_VERSION_CRUBY_4_1) {
```

Is that what you mean by versioning?

kddnewton (approved):
Nailed it!

